PR #89, opened 2014-01-16
Add basic Proguard rules to README

daj:
I was tempted to suggest that more clever Proguard rules might be possible, e.g. by keeping members with Wire annotations, unfortunately I don't have time to investigate further.

-keepclassmembers class ** {
    @com.squareup.wire.ProtoField public *;
    @com.squareup.wire.ProtoEnum public *;
}


swankjesse:
Sorry for dropping the ball on this. I'm closing as obsolete!

daj:
Sorry, I'm confused - why is this obsolete?  The current README still doesn't list any Proguard rules, and I don't recall seeing any change that removed the need for Proguard rules, so I think the PR still applies.  I can reopen a PR if you like?

swankjesse:
You're right! All good.

PR #110, opened 2014-03-16
If any of the class names Message then the compiled files will be messed up

UglyTroLL:
i.e. In proto file, we define something like
```
message Message {
	....
}
```

This change uses the full name([packageName].[className]) of com.squareup.wire.Message class in compiled java class instead of class name only. Since the proto could define a message names 'Message', which could mess up everything.

I'm able to fix WireCompilerErrorTest but not the WireCompilerTest, haven't figured out where do all the expected files come from.


JakeWharton:
https://github.com/square/wire/issues/247

PR #175, opened 2014-12-02
Apply Builder Pattern to WireCompiler.

mheath:
The constructors on WireCompiler are hairy and have way too many parameters. This is causing multiple problems:
* It's not clear which argument maps to which parameter when instantiating WireCompiler instances.
* CheckStyle will fail as soon as another parameter is added (the limit is 10)
* There's no clear interface for invoking the compiler programmatically (i.e. https://github.com/square/wire/blob/master/wire-maven-plugin/src/main/java/com/squareup/wire/mojo/WireGenerateSourcesMojo.java#L93)

In my opinion, the builder pattern solves these issues.

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/MissingOutputDirectoryWireCompilerException.java
+++ b/wire-compiler/src/main/java/com/squareup/wire/MissingOutputDirectoryWireCompilerException.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2013 Square Inc.
+ *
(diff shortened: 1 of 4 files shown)


ericzundel:
I concur that Builder pattern would be great. But I think I've just made lots of conflicts for this PR by moving command line processing out into a separate class CommandLineOptions...

swankjesse:
Sorry for dropping the ball on this. I'm closing as obsolete!

PR #177, opened 2014-12-18
Initial commit of a gradle plugin for Wire.

### LINKED ISSUE #176: Gradle plugin
Since this implementation of protobuf was written with Android in mind, it's fitting that we provide a first-party plugin for Gradle.

The plugin will be released to Maven central along with the other Wire artifacts as `wire-gradle-plugin`. The plugin ID would be `com.squareup.wire`. We should also look at deploying to [Gradle's new plugin registry][1] which might allow skipping the actual dependency declaration in the future.

Compiler options will be exposed through an extension object named `wire`. Examples of this would be generating code for options, extension registry name, etc. For example:
```groovy
wire {
  generateOptions true
  extensionRegistry 'com.example.protos.Extensions'
}
```

Open questions:
 * Would we only target protos which were hosted in-module (e.g., `src/main/proto`) and not ones pulled in from external sources either on the classpath or via HTTP, etc.?

 [1]: https://plugins.gradle.org/

jpd236:
This plugin adds support for generating Java code from .protos using
the Wire compiler to Android library/application projects.

For basic usage, add .proto files to src/main/proto and apply the
'com.squareup.wire' plugin. More customization is available via the
'wire' extension on any sourceSet.

Fixes #176


davidpelfree:
Hi @JakeWharton.

Is this pull request going to be merged soon? 
I am currently writing a project which is going to be depend on this Gradle plugin, so I want to be sure I am taking the merged version, not the pull-request version.

Thanks.

jpd236:
Friendly ping :) Have a project I'd like to release soon that makes use of this.

benjamin-bader:
Ping!  This will be quite useful.

JakeWharton:
I have two large hesitations on merging this as-is:

 * Its build is not integrated into the main (currently Maven) build. This means that we need to augment the build system to run both and it introduces headache into developing both side-by-side as they're not part of the same project (making refactors, testing, etc. harder than it should be).
 * It forces the use of the Android plugin. We currently ship with a Maven plugin for Wire because we have pure-Java projects using it and it is foreseeable that pure-Java Gradle projects would also want to take advantage.

As I see it, there's two paths forward:

 * Convert Wire to build with Gradle. This is a tad harder than it should be because I need to figure out how to build a Maven plugin with Gradle (it's doable since Maven plugins are just jars; the reverse is not doable). I also loathe Gradle projects for open source as they've historically been more of a pain to make releases with, although we've pretty much got that down.
 * Split this plugin out as its own project maintained in parallel by us. It would live as a separate project and simply depend on the released versions of Wire. The only downside here is the split brain. Similar to what I mention above, the refactoring, testing, etc. becomes a bit worse since the two aren't being developed in the same project. Upsides are that I think leaving out pure-Java Gradle modules is a reasonable thing to do from the get go which alleviates my other hesitation above. We also have experience with managing cross-project dependencies (Okio-->[Wire|Moshi|OkHttp]-->Retrofit).

I'm inclined to go with option 2 since it will allow releasing and iterating more quickly, as I have a few features that I would want to add. Perhaps long term we could work to bring it back "into the fold" of this main project.

Thoughts?

jpd236:
Thanks for the reply!

Re: supporting vanilla Java projects - I don't think that would be too difficult a feature to add. I'm not sure how that would impact the decision you lay out (convert Wire to a gradle build or split the plugin out as a separate project), though.

Building X plugins with Y might be possible in either direction, but it'd be hacky either way:

http://stackoverflow.com/questions/27555560/how-can-i-build-a-gradle-plugin-with-maven
http://stackoverflow.com/questions/11209382/build-maven-plugins-using-gradle

and splitting the project out would nicely work around the issue I had of not being able to easily point to the from-source version of Wire from within the gradle build.

jpd236:
The above commit addresses concern 2 about pure-Java projects. I don't think I can really address the other concern about project layout/structure and I'm happy with whatever path you prefer to go down. Let me know :)

mingfai:
hi,  I have a few Gradle Android projects that use Wire, and for every project I maintain Maven pom.xml to compile Wire protocol, and then install to Maven local, quite an annoying process. so since the Gradle Plugin issues were created, I've been watching it, but able to hold my desire to post "+1" so far. :-) 

Just want to ask if there may be a sample skeleton project or any open source project that use the latest commit? I'm happy to help to test and possibly do minor bug fix.

jpd236:
If you want to give it a try locally, run:
```
./gradlew install
```
from the wire-gradle-plugin directory, and then add:
```
buildscript {
    repositories {
        jcenter()
        mavenLocal()
    }

    dependencies {
        classpath 'com.squareup.wire:wire-gradle-plugin:+'
    }
}

apply plugin: 'com.squareup.wire'

dependencies {
    compile 'com.squareup.wire:wire-runtime:1.6.+'
}
```
to your build.gradle. Then just add protos to src/main/proto, as you'd add Java source to src/main/java.

JakeWharton:
https://github.com/square/wire-gradle-plugin

JakeWharton:
Also, big thanks!

mingfai:
@jpd236 thx. i tested that before and it works. 
 
I do encounter one issue that whenever build.gradle of a Java project (apparently not in Android project) is changed and I do a "Sync Now" operation in Intellij IDEA, the IDE set "/build" as exclusion, but as the plugin (correctly) generate the class under "/build/generated/source/proto", they got excluded and can't be seen by the IDE. 

It's completely an issue of IDEA not this plugin. And I also don't have a chance to fully test on the behavior yet. If IDEA doesn't allow skipping the auto exclusion, I may need to find a way to change the proto generate path. (which is a bad idea to me, as it is a workaround of IDEA problem)

jpd236:
I filed the above bug on your behalf. A quick search indicates that it is a broader problem and there'd have to be some hack to work around it, which isn't great.

Since the plugin is checked in now, I'd suggest filing any new issues / comments under the new project that Jake set up. Thanks for the review and for setting up the new repository, Jake!

PR #180, opened 2014-12-23
Add --quiet and --dry-run flags

ericzundel:
Pants wants to know which .java files are going to be emitted from the compiler early in its build.

This change adds the ability to squelch the diagnostic output with --quiet
It also adds the ability to bypass generating code and just report the files it would create.

This change touches a lot of stuff, maybe more than it has to.  For example, I moved 'writer' to a function parameter instead of storing it in a field variable.  Let me know if you want me to land it as smaller changes or be less intrusive.


wire-compiler/src/main/java/com/squareup/wire/logger/WireLoggerFactory.java
```diff
@@ -0,0 +1,20 @@
+package com.squareup.wire.logger;
+
+/**
+ * Created by zundel on 12/23/14.
+ */
+public class WireLoggerFactory {
```
ericzundel: I wanted to use this in static contexts, let me see what I can do to refactor.

wire-compiler/src/main/java/com/squareup/wire/OutputArtifact.java
```diff
@@ -0,0 +1,44 @@
+package com.squareup.wire;
+
+import java.io.File;
+
+/**
+ * A container class that represents an artifact output from the compiler.
+ */
+public class OutputArtifact {
```
ericzundel: I've never seen auto/value before, that's neat, thanks for mentioning it.  Looks like it could save a lot of boiler plate when writing value classes that you put into hash maps and trees.  I'm leaning toward not using it here though for the following reasons:

1) We aren't putting these in hash maps or sorted data structures.  We don't need equals/hashCode/toString at the moment
2) Some of the fields are derived. I'll have to change that up because you have to specify all the fields in the constructor.
3)  Its not being used elsewhere in the code. I'll have to explain why I chose it and its not compelling for above reasons 1 & 2.
4) I'll have to add a new external dependency.

ericzundel:
After correcting the other checkstyle errors and addressing some of the other comments, I have one checkstyle error left:

```
[INFO] Starting audit...
/Users/zundel/Src/wire/wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java:260:3: More than 10 parameters (found 11).
Audit done.
...
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.10:checkstyle (default) on project wire-compiler: An error has occurred in Checkstyle report generation. Failed during checkstyle execution: There are 1 checkstyle errors. -> [Help 1]
```

I will create a separate PR that refactors the options into a static class.

wire-compiler/src/main/java/com/squareup/wire/OutputArtifact.java
JakeWharton: Derived methods work with AutoValue. The dependency is compile-time only. There is no runtime component.

I'm also not convinced on point 1, just because we aren't doing something doesn't mean we shouldn't include correct implementations of those methods. Ultimately this is a value type so not being able to put it in a `Set` or use `equals` comparisons seems incorrect.

JakeWharton:
You can just disable that checkstyle check

checkstyle.xml
```diff
@@ -65,10 +65,6 @@
       <property name="max" value="100"/>
     </module>
     <module name="MethodLength"/>
-    <module name="ParameterNumber">
```
ericzundel: Removed as per @JakeWharton comment earlier.  It now failed on the CommandLineOptions constructor.

wire-compiler/src/main/java/com/squareup/wire/OutputArtifact.java
```diff
@@ -0,0 +1,45 @@
...
+  private final String javaPackage;
+
+  public OutputArtifact(String outputDirectory, String javaPackage, String className) {
+    this.outputDirectory = outputDirectory;
+    this.className = className;
+    this.javaPackage = javaPackage;
+  }
+
+
+  public String outputDirectory() {
```
ericzundel: @JakeWharton  did you mean for me to remove the get prefixes on the names on all of these methods?

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -132,15 +134,17 @@ public WireCompiler(String protoPath, List<String> sourceFileNames, List<String>
       throws WireException {
     this(new CommandLineOptions(protoPath, outputDirectory, sourceFileNames, roots,  registryClass,
         emitOptions,  new LinkedHashSet<String>(enumOptions), serviceWriterConstructor.getName(),
-        serviceWriterOptions));
+        serviceWriterOptions, false, false));
   }
 
   public WireCompiler(CommandLineOptions options) throws WireException {
-    this(options, new IO.FileIO());
+    this(options, new IO.FileIO(), new ConsoleWireLogger(options.quiet));
```
ericzundel: Switching to dependency injection caused an issue here.  Berfore, the IO instance had a way to pull in an instance of the logger.  But now, if i wanted to do that, I'd have to instantiate the logger first, then pass it to the constructor of IO.FileIO and to the constructor for WireCompiler.  The this() redirection doesn't give you much opportunity to slip in extra code, so I'd have to do something with a static method.  Rather than complicating this code, I just dropped the log message out of IO.FileIO because it didn't seem important.

wire-compiler/src/test/java/com/squareup/wire/WireCompilerTest.java
```diff
@@ -138,18 +139,29 @@ private void testProtoWithRegistry(String[] sources, String registryClass, Strin
...
   private void testProtoWithRoots(String[] sources, String roots, String[] outputs)
       throws Exception {
+    String[] extraArgs = {};
+    this.testProtoWithRoots(sources, roots, outputs, extraArgs);
+  }
+
+  private void testProtoWithRoots(String[] sources, String roots, String[] outputs,
+      String[] extraArgs)
+      throws Exception {
     int numFlags = 4;
```
ericzundel: I followed the pattern of other tests, but note that we don't need to invoke the compiler with an array of strings if we don't want to, we could construct a CommandLineOptions instance.  But it would be a bit awkward and require us to pass in null or empty list/set arguments.   I think a better programmatic interface would be to be able to initialize CommandLineOptions with builder pattern.

wire-compiler/src/test/java/com/squareup/wire/logger/MockWireLogger.java
```diff
@@ -0,0 +1,16 @@
+package com.squareup.wire.logger;
+
+import com.squareup.wire.OutputArtifact;
```
ericzundel: This class isn't necessary - I could have substituted StringWireLogger in the one case where I decided to use MockWireLogger.  Or we could pull in a mocking library like mockito.

ericzundel:
ping

danrice-square:
LGTM

wire-compiler/src/main/java/com/squareup/wire/IO.java
```diff
@@ -29,9 +28,10 @@
    *   <output directory>/<java package converted to slashed form>/<className>.java
    * }</pre>
    */
-  JavaWriter getJavaWriter(String outputDirectory, String javaPackage, String className)
+  JavaWriter getJavaWriter(OutputArtifact outputArtifact)
       throws IOException;
 
+
```
JakeWharton: Delete

wire-compiler/src/main/java/com/squareup/wire/logger/ConsoleWireLogger.java
```diff
@@ -0,0 +1,32 @@
+package com.squareup.wire.logger;
+
+import com.squareup.wire.OutputArtifact;
+
+
+public final class ConsoleWireLogger implements WireLogger {
```
JakeWharton: This shouldn't be public

wire-compiler/src/main/java/com/squareup/wire/logger/ConsoleWireLogger.java
```diff
@@ -0,0 +1,32 @@
+package com.squareup.wire.logger;
+
+import com.squareup.wire.OutputArtifact;
+
+
```
JakeWharton: Delete extra line

wire-compiler/src/main/java/com/squareup/wire/logger/WireLogger.java
```diff
@@ -0,0 +1,9 @@
+package com.squareup.wire.logger;
+
+import com.squareup.wire.OutputArtifact;
+
+public interface WireLogger {
```
JakeWharton: Shouldn't be public

wire-compiler/src/test/java/com/squareup/wire/WireCompilerErrorTest.java
```diff
@@ -82,10 +82,11 @@ public JavaWriter getJavaWriter(String javaOut, String javaPackage, String class
 
     CommandLineOptions options = new CommandLineOptions(".",  ".", Arrays.asList("test.proto"),
         new ArrayList<String>(), null, true, Collections.<String>emptySet(), null,
-        Collections.<String>emptyList());
+        Collections.<String>emptyList(), false, false);
+
```
JakeWharton: Delete extra line

wire-compiler/src/test/java/com/squareup/wire/logger/MockWireLogger.java
```diff
@@ -0,0 +1,16 @@
...
+
+
```
JakeWharton: Delete extra line

wire-compiler/src/test/java/com/squareup/wire/logger/StringWireLogger.java
```diff
@@ -0,0 +1,34 @@
+package com.squareup.wire.logger;
+
+import com.squareup.wire.OutputArtifact;
+
+
```
JakeWharton: Delete extra line

wire-compiler/src/main/java/com/squareup/wire/logger/WireLogger.java
danrice-square: `WireCompiler` references it, what do you suggest?

wire-compiler/src/main/java/com/squareup/wire/logger/ConsoleWireLogger.java
danrice-square: `WireCompiler` references it, what do you suggest?

wire-compiler/src/main/java/com/squareup/wire/IO.java
danrice-square: Fixed (in follow-on PR)

wire-compiler/src/main/java/com/squareup/wire/logger/ConsoleWireLogger.java
danrice-square: Fixed (in follow-on PR)

wire-compiler/src/test/java/com/squareup/wire/WireCompilerErrorTest.java
danrice-square: Fixed (in follow-on PR)

wire-compiler/src/test/java/com/squareup/wire/logger/MockWireLogger.java
danrice-square: Fixed (in follow-on PR)

wire-compiler/src/main/java/com/squareup/wire/logger/WireLogger.java
JakeWharton: Put them in the same package and make it packaged scoped.

danrice-square: Done.

PR #182, opened 2015-01-05
Break out command line processing into a separate class

ericzundel:
This is an offshoot of https://github.com/square/wire/pull/180

 @JakeWharton  said it would be OK to just remove the 10 argument checkstyle rule, but I decided to float this change because it rids us of the need to change the constructor for the compiler every time we add a new option and adds tests for the cmdline parsing logic.


wire-compiler/src/main/java/com/squareup/wire/CommandLineOptions.java
```diff
@@ -0,0 +1,137 @@
...
+  final List<String> sourceFileNames;
+  final List<String> roots;
+  final String javaOut;
+  final String registryClass;
+  final boolean emitOptions;
+  final Set<String> enumOptions;
+  final String serviceWriter;
+  final List<String> serviceWriterOptions;
+
+  CommandLineOptions(String protoPath, List<String> sourceFileNames, List<String> roots,
```
ericzundel: This is a potential place to use [@AutoValue](https://github.com/google/auto) again.   I don't think it adds a lot of value in this case either, but if you are gung ho to use this library, let me know and I will add it here and in the OutputArtifact case in PR 180.

wire-compiler/src/main/java/com/squareup/wire/CommandLineOptions.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Scanner;
+import java.util.Set;
+
+
+final class CommandLineOptions {
+  public static final String PROTO_PATH_FLAG = "--proto_path=";
```
danrice-square: Reorder to match order in javadoc below

wire-compiler/src/main/java/com/squareup/wire/CommandLineOptions.java
danrice-square: Reordering the args to match order in javadoc below

wire-compiler/src/main/java/com/squareup/wire/CommandLineOptions.java
```diff
@@ -0,0 +1,137 @@
...
+    this.roots = roots;
+    this.javaOut = javaOut;
+    this.registryClass = registryClass;
+    this.emitOptions = emitOptions;
+    this.enumOptions = enumOptions;
+    this.serviceWriter = serviceWriter;
+    this.serviceWriterOptions = serviceWriterOptions;
+  }
+
+  /**
```
danrice-square: I think this comment should still be in WireCompiler since it defines the semantics for `WireCompiler(String...)`.

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -221,32 +128,44 @@ public static void main(String... args) throws Exception {
     return null;
   }
 
+  @Deprecated
   public WireCompiler(String protoPath, List<String> sourceFileNames, List<String> roots,
```
danrice-square: Does anybody use this? Could we change the callers to use String args and delete this?

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -478,11 +397,11 @@ private void getTypes(Type parent, List<Type> types) {
   }
 
   private void emitRegistry() throws IOException {
-    int packageClassSep = registryClass.lastIndexOf(".");
-    String javaPackage = registryClass.substring(0, packageClassSep);
-    String className = registryClass.substring(packageClassSep + 1);
+    int packageClassSep = options.registryClass.lastIndexOf(".");
```
danrice-square: We could do this parsing in CommandLineOptions and provide separate getters for the package and class names.

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
ericzundel: That would be awesome. I checked the maven plugin and it ends up calling WireCompiler.main() with a note that we need a better interface to invoke the compiler.

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -221,32 +128,44 @@ public static void main(String... args) throws Exception {
     return null;
   }
 
+  @Deprecated
   public WireCompiler(String protoPath, List<String> sourceFileNames, List<String> roots,
       String outputDirectory, String registryClass, boolean emitOptions, List<String> enumOptions,
       Constructor<?> serviceWriterConstructor, List<String> serviceWriterOptions) {
-    this(protoPath, sourceFileNames, roots, outputDirectory, registryClass, emitOptions,
-        enumOptions, serviceWriterConstructor, serviceWriterOptions, new IO.FileIO());
+    this(new CommandLineOptions(protoPath, sourceFileNames, roots, outputDirectory, registryClass, emitOptions,
```
danrice-square: Line too long

wire-compiler/src/main/java/com/squareup/wire/CommandLineOptions.java
ericzundel: I was afraid it would only get out of date if we left it in WireCompiler.  I also considered moving the javadoc comment into a usage() string, but I don't think many folks actually invoke the compiler from the command line - they probably use a build tool  to invoke it.

danrice-square:
3 lines are too long according to checkstyle

danrice-square:
LGTM with nits

wire-compiler/src/test/java/com/squareup/wire/CommandLineOptionsTest.java
```diff
@@ -0,0 +1,132 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public class CommandLineOptionsTest {
+
+  @Test public void testProtoPath() throws Exception {
```
JakeWharton: You don't need test prefixes for JUnit 4+

wire-compiler/src/main/java/com/squareup/wire/CommandLineOptions.java
```diff
@@ -0,0 +1,140 @@
...
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Scanner;
+import java.util.Set;
+
+
```
JakeWharton: nit: delete extra newline

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -83,115 +74,31 @@
...
-      protoPath = System.getProperty("user.dir");
-      System.err.println(PROTO_PATH_FLAG + " flag not specified, using current dir " + protoPath);
-    }
+
+    CommandLineOptions options = new CommandLineOptions(args);
+
     WireCompiler wireCompiler =
-        new WireCompiler(protoPath, sourceFileNames, roots, javaOut, registryClass, emitOptions,
-            enumOptions, serviceWriterConstructor, serviceWriterOptions);
+        new WireCompiler(options);
```
JakeWharton: Doesn't need wrapped. Or even put into a local. Just `new WireCompiler(options).compile();` will do.

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -221,32 +128,46 @@ public static void main(String... args) throws Exception {
...
-  public WireCompiler(String protoPath, List<String> sourceFileNames, List<String> roots,
-      String outputDirectory, String registryClass, boolean emitOptions, List<String> enumOptions,
+  @Deprecated
+  public WireCompiler(String protoPath, String outputDirectory, List<String> sourceFileNames,
+      List<String> roots, String registryClass, boolean emitOptions, List<String> enumOptions,
       Constructor<?> serviceWriterConstructor, List<String> serviceWriterOptions) {
-    this(protoPath, sourceFileNames, roots, outputDirectory, registryClass, emitOptions,
-        enumOptions, serviceWriterConstructor, serviceWriterOptions, new IO.FileIO());
+    this(new CommandLineOptions(protoPath, outputDirectory, sourceFileNames, roots,  registryClass,
+        emitOptions,  new HashSet<String>(enumOptions), serviceWriterConstructor.getName(),
```
JakeWharton: nit: double space

JakeWharton: should be `LinkedHashSet`

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -221,32 +128,46 @@ public static void main(String... args) throws Exception {
     return null;
   }
 
-  public WireCompiler(String protoPath, List<String> sourceFileNames, List<String> roots,
-      String outputDirectory, String registryClass, boolean emitOptions, List<String> enumOptions,
+  @Deprecated
+  public WireCompiler(String protoPath, String outputDirectory, List<String> sourceFileNames,
```
JakeWharton: The parameters to this method were re-ordered which is a backwards-incompatible public API change and will break people's (including all our Gradle-based Android projects') code.

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -221,32 +128,46 @@ public static void main(String... args) throws Exception {
...
-  WireCompiler(String protoPath, List<String> sourceFileNames, List<String> roots,
-      String outputDirectory, String registryClass, boolean emitOptions, List<String> enumOptions,
-      Constructor<?> serviceWriterConstructor, List<String> serviceWriterOptions, IO io) {
+  WireCompiler(CommandLineOptions options, IO io) {
+    this.options = options;
+
+    String protoPath = options.protoPath;
+    if (options.javaOut == null) {
+      System.err.println("Must specify " + CommandLineOptions.JAVA_OUT_FLAG + " flag");
+      System.exit(1);
```
JakeWharton: This method assumes invocation from a `main()` which is not always true. This user-friendly validation should be performed in `main()` and it should throw exceptions here.

wire-compiler/src/test/java/com/squareup/wire/WireCompilerErrorTest.java
```diff
@@ -79,10 +79,12 @@ public JavaWriter getJavaWriter(String javaOut, String javaPackage, String class
   private Map<String, String> compile(String source) {
     StringIO io = new StringIO("test.proto", source);
 
+    CommandLineOptions options = new CommandLineOptions(".",  ".", Arrays.asList("test.proto"),
+        new ArrayList<String>(), null, true, Collections.EMPTY_SET, null,
+        Collections.EMPTY_LIST);
```
JakeWharton: `new ArrayList<String>()` --> `Collections.<String>emptyList()`
`Collections.EMPTY_LIST` --> `Collections.emptyList()`
`Collections.EMPTY_SET` --> `Collections.emptySet()`

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
ericzundel: Good idea.  I'll refactor this to throw an exception.  I moved it because I was thinking that having validation logic in main() is keeping us from having a nice way to invoke the compiler without passing strings for command line options.  The next refactoring moves logging into a logger method so we don't assume console output.

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
danrice-square: Good catch, sorry for the churn.

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
ericzundel: I missed this comment on my earlier pass.  Since parsing the class name didn't seem to have to do with command line parsing, I left this out for now.  I thought about moving all command line argument validation into the same place, but I couldn't find a great reason to move it all into CommandLineOptions.

ericzundel:
Travis-CI is passing, PTAL

ericzundel:
I am going to rebase this once https://github.com/square/wire/pull/182 lands and then re-post it for review.

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -264,20 +185,29 @@ public void compile() throws IOException {
       this.protoFile = entry.getValue();
       this.protoFileName = protoFileName(protoFile.getFileName());
       System.out.println("Compiling proto source file " + sourceFileName);
-      compileOne();
+      try {
+        compileOne();
+      } catch (IOException e) {
+        throw new WireException("Error compiling " + entry.getKey(), e);
+
```
danrice-square: Remove extra line

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
danrice-square: O.K.

wire-maven-plugin/src/main/java/com/squareup/wire/mojo/WireGenerateSourcesMojo.java
```diff
@@ -67,6 +67,7 @@
   }
 
   private void compileProtos() throws MojoExecutionException {
+    //TODO(Eric Ayers) refactor to use CommandLineOptions class?
```
danrice-square: It's not a public API, at least for now. I would omit this TODO.

wire-compiler/src/main/java/com/squareup/wire/CommandLineOptions.java
danrice-square: O.K., I see there is a reference to it.

danrice-square:
LGTM

JakeWharton:
Squashed and merged.

PR #186, opened 2015-01-09
Use an annotation value to retrieve the underlying type of a repeated Message or Enum field

danrice-square:
@rjrjr


rjrjr:
LGTM

I still feel a little bad about the DRY violation in the annotation, but the underlying code looks cleaner.

PR #192, opened 2015-03-05
Serializable messages

mattprecious:
@danrice-square @swankjesse @JakeWharton

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/MessageSerializedForm.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/MessageSerializedForm.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2015 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire;
+
+import java.io.IOException;
+import java.io.ObjectStreamException;
+import java.io.Serializable;
+import java.io.StreamCorruptedException;
+
+final class MessageSerializedForm implements Serializable {
+  private static final long serialVersionUID = 0L;
+
+  private final byte[] bytes;
+  private final Class<? extends Message> messageClass;
+
+  public MessageSerializedForm(Message message, Class<? extends Message> messageClass) {
+    this.bytes = message.toByteArray();
+    this.messageClass = messageClass;
+  }
+
+  Object readResolve() throws ObjectStreamException {
+    try {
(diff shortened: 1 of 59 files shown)


wire-runtime/src/main/java/com/squareup/wire/MessageSerializedForm.java
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+package com.squareup.wire;
+
+import java.io.IOException;
+import java.io.ObjectStreamException;
+import java.io.Serializable;
+import java.io.StreamCorruptedException;
+
+public final class MessageSerializedForm implements Serializable {
```
JakeWharton: Not public?

wire-runtime/src/main/java/com/squareup/wire/MessageSerializedForm.java
```diff
@@ -0,0 +1,39 @@
...
+  private final Class<? extends Message> messageClass;
+
+  public MessageSerializedForm(Message message, Class<? extends Message> messageClass) {
+    this.bytes = message.toByteArray();
+    this.messageClass = messageClass;
+  }
+
+  Object readResolve() throws ObjectStreamException {
+    try {
+      return new Wire().parseFrom(bytes, messageClass);
```
danrice-square: This implies that the serialized classes can't contain extensions, i.e., any extensions will end up in the UnknownField bin. Could we serialize a list of extension names as needed?

mattprecious: I don't see any clean way of doing this from my quick look around at how extensions are done. The extension classes are lost once they're added to the registry in the Wire constructor... can this be a follow-up or just omitted completely?

mattprecious:
Needed to emit `serialVersionUID` in all the messages :(

wire-runtime/src/main/java/com/squareup/wire/MessageSerializedForm.java
danrice-square: Seems like something we could follow up with, provided we allow for there to be additional fields in `MessageSerializedForm` in the future.

danrice-square:
Do we actually need to generate code for all of descriptor.proto?

wire-runtime/src/test/java/com/google/protobuf/MessageOptions.java
```diff
@@ -51,7 +52,7 @@
   /**
    * The parser stores options it doesn't recognize here. See above.
    */
-  @ProtoField(tag = 999, label = REPEATED)
+  @ProtoField(tag = 999, label = REPEATED, messageType = UninterpretedOption.class)
```
danrice-square: Thanks for fixing this.

mattprecious:
No. They're generated with the `gen_tests` script, though, so it gets annoying dealing with git changes. It would be great if they were omitted from the script.

wire-runtime/src/main/java/com/squareup/wire/MessageSerializedForm.java
mattprecious: Was there something in particular you had in mind? I think fields can be added here in the future without any incompatibility with the current structure. When an old version gets deserialized the new fields will be null, and then we can treat it like a message without extensions.

If anything more complicated arises we can branch based on the version UID.

swankjesse:
LGTM

danrice-square:
LGTM

PR #194, opened 2015-03-06
Fix for imports from external packages

mattprecious:
@danrice-square @JakeWharton


wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -890,13 +890,14 @@ private void getExternalTypes(Type parent, List<String> types) {
...
 
   // Convert 'Bar.Baz.Moo' to 'Bar' for use in imports.
-  private String getTopLevelMessageName(MessageType messageType, String fqName) {
-    String packageName = getPackageFromFullyQualifiedJavaName(fqName);
-    String unqualifiedName = fqName.substring(packageName.length() + 1);
-    int dotIndex = unqualifiedName.indexOf('.');
-    if (dotIndex != -1) {
-      String parentType = unqualifiedName.substring(0, dotIndex);
-      return fullyQualifiedJavaName(messageType, parentType);
+  private String getTopLevelMessageName(String fqName) {
```
mattprecious: @danrice-square, my version is a lot more naive so I'm not sure if I'm missing anything here... but all the tests are passing including the new one which fails without this change.

JakeWharton:
LGTM. If it's not tested, it's not supported!

danrice-square:
LGTM

PR #197, opened 2015-04-13
added the --files argument to the maven plugin

nicktate:
Added `--files` support to maven-plugin.

> Instead of supplying individual filename arguments on the command line, the --files flag may be used to specify a single file containing a list of .proto files. The file names are interpreted relative to the value given for the --proto_path flag.

### DIFF
--- a/wire-maven-plugin/src/main/java/com/squareup/wire/mojo/WireGenerateSourcesMojo.java
+++ b/wire-maven-plugin/src/main/java/com/squareup/wire/mojo/WireGenerateSourcesMojo.java
@@ -3,8 +3,13 @@ package com.squareup.wire.mojo;
 import com.google.common.base.Joiner;
 import com.google.common.collect.Lists;
 import com.squareup.wire.WireCompiler;
+
+import java.io.File;
+import java.io.FileNotFoundException;
 import java.util.Collections;
 import java.util.List;
+import java.util.Scanner;
+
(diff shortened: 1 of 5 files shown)


wire-maven-plugin/src/main/java/com/squareup/wire/mojo/WireGenerateSourcesMojo.java
```diff
@@ -87,6 +98,17 @@ private void compileProtos() throws MojoExecutionException {
     }
     Collections.addAll(args, protoFiles);
 
+    if (files != null) {
+        File filePointer = new File(files);
+        String[] fileNames;
+        try {
+            fileNames = new Scanner(filePointer, "UTF-8").useDelimiter("\\A").next().split("\n");
```
swankjesse: Use Okio.BufferedSource instead of Scanner?

wire-maven-plugin/src/test/resources/exemplar/proto.include
```diff
@@ -0,0 +1 @@
+squareup/wire/exemplar_two.proto
```
swankjesse: include 2 files to confirm split is being exercised?

swankjesse: ... testing the Maven plugin is awkward right now!

swankjesse:
Obsolete!

Issue #195, opened 2015-03-19
wire-runtime crashes when re-serializing unknown repeated fields.


asarazan:
See https://github.com/asarazan/WireBugReduced for a reduced test case.

TL;DR
1. Send an object with a repeated field to a client that doesn't know about that field
2. Have the client parse, then call toByteArray() on that object
3. Crash:

```
Caused by: java.io.IOException: Out of space: position=7, limit=7
at com.squareup.wire.WireOutput.writeRawBytes(WireOutput.java:244)
at com.squareup.wire.WireOutput.writeRawBytes(WireOutput.java:233)
at com.squareup.wire.UnknownFieldMap$LengthDelimitedFieldValue.write(UnknownFieldMap.java:170)
at com.squareup.wire.UnknownFieldMap.write(UnknownFieldMap.java:260)
at com.squareup.wire.Message.writeUnknownFieldMap(Message.java:238)
at com.squareup.wire.MessageAdapter.write(MessageAdapter.java:310)
at com.squareup.wire.MessageAdapter.writeMessage(MessageAdapter.java:504)
at com.squareup.wire.MessageAdapter.writeValueNoTag(MessageAdapter.java:491)
at com.squareup.wire.MessageAdapter.writeValue(MessageAdapter.java:465)
at com.squareup.wire.MessageAdapter.write(MessageAdapter.java:300)
at com.squareup.wire.MessageAdapter.toByteArray(MessageAdapter.java:359)
```

asarazan:
The issue appears to stem from `LengthDelimitedFieldValue` at lines 164 and 168:

```
    @Override public int getSerializedSize() {
      return WireOutput.varint32Size(value.size()) + value.size();
    }

    @Override public void write(int tag, WireOutput output) throws IOException {
      output.writeTag(tag, WireType.LENGTH_DELIMITED);
      output.writeVarint32(value.size());
      output.writeRawBytes(value.toByteArray());
    }
```

`getSerializedSize` doesn't take the tag length into account, whereas `write` does, leading to not enough space being allocated.

JakeWharton:
Nice find. Feel free to send a PR with a test if you have time. Otherwise we'll get to this soon!

asarazan:
PR sent.

PR #198, opened 2015-04-18
Unknown repeated fields

asarazan:
Fix for issue #195 : "wire-runtime crashes when re-serializing unknown repeated fields."

When writing a repeated unknown field, we were writing the tag value before each repetition, but only allocating enough space for a single tag value at the front of the section.

We are now allocating enough space for each tag write.


danrice-square:
Looks great, would you be willing to squash down to one commit then I will merge.

asarazan:
All set.

danrice-square:
Thanks for catching and fixing this!

PR #202, opened 2015-05-20
Update to the latest ProtoParser.

JakeWharton:
This adds support for 'oneof' & 'map', among a more sane syntax model.

@swankjesse @danrice-square

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/ServiceWriter.java
+++ b/wire-compiler/src/main/java/com/squareup/wire/ServiceWriter.java
@@ -2,7 +2,7 @@
 package com.squareup.wire;
 
 import com.squareup.javawriter.JavaWriter;
-import com.squareup.protoparser.Service;
+import com.squareup.protoparser.ServiceElement;
 import java.io.IOException;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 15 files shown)


JakeWharton:
Note: PP4 is not done, but should be by the end of the week.

wire-compiler/src/main/java/com/squareup/wire/ServiceWriter.java
```diff
@@ -35,5 +35,6 @@ protected ServiceWriter(JavaWriter writer, List<String> options) {
    * @param service the service being generated.
    * @param importedTypes a set of strings containing known imports.
    */
-  public abstract void emitService(Service service, Set<String> importedTypes) throws IOException;
+  public abstract void emitService(ServiceElement service, Set<String> importedTypes)
```
JakeWharton: Note: this is a public API change, but that's the consequence of exposing a dependencies types. This will have to be a Wire 2.0 if we adhere to semantic versioning.

danrice-square:
LGTM

wire-compiler/src/main/java/com/squareup/wire/ServiceWriter.java
swankjesse: +1 to doing Wire 2.0. Do we wanna bundle other changes with the API bump?

swankjesse:
Awesome.

wire-compiler/src/main/java/com/squareup/wire/ServiceWriter.java
JakeWharton: I don't think anything that could be done quickly and in spare time.

PR #204, opened 2015-05-22
Remove redundant 'private' modifier from enum constructor.

JakeWharton:
Enums can only be constructed in their constants and IntelliJ flags the redundancy as a warning.

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/MessageWriter.java
+++ b/wire-compiler/src/main/java/com/squareup/wire/MessageWriter.java
@@ -147,7 +147,7 @@ public class MessageWriter {
         parameters.add(trailingSegment(option.name));
       }
 
-      writer.beginConstructor(EnumSet.of(PRIVATE), parameters, null);
+      writer.beginConstructor(Collections.<Modifier>emptySet(), parameters, null);
       writer.emitStatement("this.value = value");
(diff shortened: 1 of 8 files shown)


wire-compiler/src/main/java/com/squareup/wire/MessageWriter.java
```diff
@@ -147,7 +147,7 @@ public void emitType(JavaWriter writer, TypeElement type, String currentType,
         parameters.add(trailingSegment(option.name));
       }
 
-      writer.beginConstructor(EnumSet.of(PRIVATE), parameters, null);
+      writer.beginConstructor(Collections.<Modifier>emptySet(), parameters, null);
```
danrice-square: Can also use `EnumSet.noneOf(Modifier.class)`.

danrice-square:
LGTM

wire-compiler/src/main/java/com/squareup/wire/MessageWriter.java
JakeWharton: I had that originally but then was curious of its behavior. It allocates a mutable set with the capacity to hold all of the enum values and makes a copy of the enum constant array. Granted, this is the JVM so it's hard to nitpick so much, but I'm a trained allocation killer from doing Android.

PR #205, opened 2015-05-22
Add support for 'oneof' fields.

danrice-square:
@JakeWharton @swankjesse


wire-compiler/src/main/java/com/squareup/wire/MessageWriter.java
```diff
@@ -454,6 +475,89 @@ private void emitMessageFields(JavaWriter writer, MessageElement messageType) th
...
+
+      for (int i = 0, count = oneOfElement.fields().size(); i < count; i++) {
+        FieldElement value = oneOfElement.fields().get(i);
+        String enumValueName = oneOfEnumValueName(sanitize(value.name()));
+        writer.emitEnumValue(enumValueName + "(" + value.tag() + ")",
+            (i == count - 1));
+      }
+
+      writer.emitEmptyLine();
+      writer.emitField("int", "value");
```
JakeWharton: + `EnumSet.of(PRIVATE, FINAL)`

wire-runtime/src/test/proto/one_of.proto
```diff
@@ -0,0 +1,29 @@
...
+option java_package = "com.squareup.wire.protos.oneof";
+
+message OneOfMessage {
+
+  // Must have a foo or a bar.
+  oneof choice {
+    // What foo.
+    int32 foo = 1;
+    // Such bar.
+    string bar = 2;
```
JakeWharton: Can you either flip the tag values or set this to 3 so that we ensure we're testing that tag values are used in the choice enum?

wire-runtime/src/main/java/com/squareup/wire/MessageAdapter.java
```diff
@@ -584,8 +608,12 @@ M read(WireInput input) throws IOException {
               storage.add(tag, value);
             } else if (extension != null) {
               setExtension((ExtendableBuilder<?>) builder, extension, value);
+            } else if (label.isOneOf()) {
+              // In order to maintain the 'oneof' invariant., call the builder setter method rather
```
JakeWharton: nit: extra period before comma

JakeWharton:
Great. LGTM.

PR #208, opened 2015-06-15
Drop the oneof choice enum.

swankjesse:
It is particularly weird in JSON encoding, since it may be
either absent or redundant.

Checking the presence of a oneof with a series of null checks
is simpler and safer.

### DIFF
--- a/wire-runtime/src/test/java/com/squareup/wire/protos/oneof/OneOfMessage.java
+++ b/wire-runtime/src/test/java/com/squareup/wire/protos/oneof/OneOfMessage.java
@@ -3,7 +3,6 @@
 package com.squareup.wire.protos.oneof;
 
 import com.squareup.wire.Message;
-import com.squareup.wire.ProtoEnum;
 import com.squareup.wire.ProtoField;
 
 import static com.squareup.wire.Message.Datatype.INT32;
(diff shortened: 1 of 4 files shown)


wire-runtime/src/test/java/com/squareup/wire/protos/oneof/OneOfMessage.java
```diff
@@ -126,7 +89,6 @@ public Builder bar(String bar) {
       this.bar = bar;
 
       this.foo = null;
```
swankjesse: Note that we still null out the competing oneof options!

JakeWharton:
Needs a rebase.

JakeWharton:
LGTM but @danrice-square should approve. You might also want to make your case for this change here so all understand where it's coming from.

swankjesse:
Rebased.

danrice-square:
If you are trying to accomplish something json-specific, the right place would be in WireGsonConverter. In general you don't want to use a raw Gson instance on a Wire Message.

Otherwise, I'd be interested in feedback from developers using oneof to see what style they prefer.

swankjesse:
I suspect we're not likely to find a bunch of `oneof` users, because it's such a rare feature.

We could ship this, and see if anyone complains. It's way easier to add the enum later than to take it out!

JakeWharton:
Can we resolve this? We need to ship a new version of Wire with this support.

swankjesse:
My preference is to merge this, and add back the enum later if there is demand. Doing the opposite is more difficult because it's a backwards-incompatible change.

danrice-square:
LGTM

PR #220, opened 2015-06-29
Prepare to switch to JavaPoet.

swankjesse:
The current Wire implementation has to work very hard to overcome
the lack of support for imports in JavaWriter. With JavaPoet, we
will just link everything together in the first step, and then
use fully qualified names when emitting code.

This adds a new model that adds new places for the linked elements
to live.


danrice-square:
LGTM

PR #221, opened 2015-06-29
Link extensions to options.

swankjesse:


wire-compiler/src/main/java/com/squareup/wire/model/ProtoTypeName.java
```diff
@@ -45,14 +45,24 @@
   public static final ProtoTypeName UINT32 = new ProtoTypeName("uint32");
   public static final ProtoTypeName UINT64 = new ProtoTypeName("uint64");
 
+  public static final ProtoTypeName FILE_OPTIONS = get("google.protobuf", "FileOptions");
```
JakeWharton: Shouldn't this class be called `WireTypeName` to fit the naming scheme?

swankjesse: Yes.

I started by trying to tease apart Java types like `com.squareup.cash.Payment` from Proto types like `squareup.cash.Payment`, and wanted to have a name to identify proto things vs. Java things.

As it turns out, that isn't really important yet, and `Wire` is proto.

swankjesse: (will fix)

wire-compiler/src/test/java/com/squareup/wire/model/LinkerTest.java
```diff
@@ -0,0 +1,58 @@
...
+ * limitations under the License.
+ */
+package com.squareup.wire.model;
+
+import java.util.LinkedHashSet;
+import java.util.Set;
+import org.junit.Test;
+
+import static java.util.Arrays.asList;
+import static org.junit.Assert.assertEquals;
```
JakeWharton: Big PR to pull in AssertJ or Truth later?

JakeWharton: I've gotten really good at IntelliJ's regex find-replace in path to do this for me...

swankjesse: Hell yes. And Guava.

wire-compiler/src/main/java/com/squareup/wire/model/ProtoTypeName.java
```diff
@@ -45,14 +45,24 @@
...
+  public static final ProtoTypeName MESSAGE_OPTIONS = get("google.protobuf", "MessageOptions");
+  public static final ProtoTypeName FIELD_OPTIONS = get("google.protobuf", "FieldOptions");
+  public static final ProtoTypeName ENUM_OPTIONS = get("google.protobuf", "EnumOptions");
+  public static final ProtoTypeName ENUM_VALUE_OPTIONS = get("google.protobuf", "EnumValueOptions");
+  public static final ProtoTypeName SERVICE_OPTIONS = get("google.protobuf", "ServiceOptions");
+  public static final ProtoTypeName METHOD_OPTIONS = get("google.protobuf", "MethodOptions");
```
JakeWharton: Package scope? Or am I not foreseeing some use for these by plugins?

wire-compiler/src/test/java/com/squareup/wire/model/LinkerTest.java
swankjesse: (though I'd prefer to land https://github.com/square/wire/pull/222 first, which branches off this)

JakeWharton:
LGTM :+1:

wire-compiler/src/main/java/com/squareup/wire/model/ProtoTypeName.java
swankjesse: Can do package scope. I started out thinking of this entire thing as an implementation detail; but I think it's going to be the thing that plugins are going to want to read.

PR #223, opened 2015-06-30
Implement trimming from the root set.

swankjesse:
This trims the in-memory model rather than limiting which types are emitted
during serialization. This approach may be more reusable when using Wire to
target other languages.

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/model/Loader.java
+++ b/wire-compiler/src/main/java/com/squareup/wire/model/Loader.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire.model;
+
+import com.squareup.protoparser.ProtoFile;
+import com.squareup.wire.IO;
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Load proto files and their transitive dependencies, parse them, and prepare for linking. The
+ * returned values are not linked and should not be used prior to linking.
+ */
+public final class Loader {
+  private final String repoPath;
+  private final IO io;
+  private final Set<String> protoFileNames = new LinkedHashSet<String>();
+  private final List<WireProtoFile> loaded = new ArrayList<WireProtoFile>();
+
+  public Loader(String repoPath, IO io) {
+    this.repoPath = repoPath;
+    this.io = io;
+  }
+
+  /** Recursively add {@code protoFile} and its dependencies. */
+  public void add(String protoFileName) throws IOException {
+    if (!protoFileNames.add(protoFileName)) {
+      return;
+    }
+
+    ProtoFile protoFile = io.parse(repoPath + File.separator + protoFileName);
+    WireProtoFile wireProtoFile = WireProtoFile.get(protoFile);
+    loaded.add(wireProtoFile);
+
+    // Recursively add dependencies.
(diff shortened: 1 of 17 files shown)


wire-compiler/src/main/java/com/squareup/wire/model/RootSet.java
```diff
@@ -0,0 +1,212 @@
...
+        continue;
+      }
+
+      WireService service = servicesIndex.get(name);
+      if (service != null) {
+        markSet.markService(service);
+        continue;
+      }
+
+      // If the root set contains a method name like 'Service#Method', only that RPC is marked.
```
JakeWharton: How granular of you!

wire-compiler/src/main/java/com/squareup/wire/model/Loader.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Load proto files and their transitive dependencies, parse them, and prepare for linking. The
+ * returned values are not linked and should not be used prior to linking.
+ */
+public final class Loader {
```
swankjesse: Currently public because I'm experimenting with use from `WireCompiler`; will be package-private or internal before anything interesting happens.

wire-compiler/src/main/java/com/squareup/wire/model/RootSet.java
```diff
@@ -0,0 +1,212 @@
...
+import java.util.Collection;
+import java.util.Deque;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/** A collection of messages, enums and services. */
+public final class RootSet {
```
swankjesse: Got any better names? This is like a working set of protos.

wire-compiler/src/main/java/com/squareup/wire/model/RootSet.java
```diff
@@ -0,0 +1,212 @@
...
+    List<WireProtoFile> retained = new ArrayList<WireProtoFile>();
+    for (WireProtoFile protoFile : protoFiles) {
+      retained.add(protoFile.retainAll(markSet.marks));
+    }
+
+    return new RootSet(retained);
+  }
+
+  /** Visits all types and services and mark them. */
+  private static class MarkSet {
```
swankjesse: Kind of like a mark + copy garbage collector.

wire-compiler/src/main/java/com/squareup/wire/model/RootSet.java
swankjesse: It's a bit begrudging. The tests have it!

JakeWharton: The tests for what?

swankjesse: https://github.com/square/wire/blob/master/wire-compiler/src/test/java/com/squareup/wire/WireCompilerTest.java#L313

wire-compiler/src/main/java/com/squareup/wire/model/RootSet.java
JakeWharton: This isn't a stateful type and it returns a new instance when doing a retain pass, so why not just use a normal `List` and do the index work as the first thing on the call to `retainRoots` which could just be a static method somewhere.

swankjesse: It's so simple. Yes.

wire-compiler/src/main/java/com/squareup/wire/model/RootSet.java
JakeWharton: Crazy. The more you know™.

wire-compiler/src/main/java/com/squareup/wire/model/WireMessage.java
```diff
@@ -115,4 +94,28 @@ void link(Linker linker) {
       option.link(ProtoTypeName.MESSAGE_OPTIONS, linker);
     }
   }
+
+  @Override WireType retainAll(Set<String> identifiers) {
```
JakeWharton: I'm pretty sure there's a tiny simplification to be had by switching these methods to return collections. You never have to worry about null checking and have logic for one vs. many. It's always just either an `isEmpty()` check or `addAll` when composing.

JakeWharton: Actually it's not too bad as is... I thought it'd be worse.

JakeWharton:
LGTMeeee

PR #226, opened 2015-07-02
Extensions on JavaPoet.

swankjesse:
Now passing 19 of 35 WireCompiler tests. Still remaining are services
and options.

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
+++ b/wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
@@ -1298,9 +1298,17 @@ public class WireCompiler {
         continue; // Don't emit anything for files not explicitly compiled.
       }
 
+      TypeWriter typeWriter = new TypeWriter(javaGenerator, options.emitOptions);
+
       for (com.squareup.wire.model.WireType type : wireProtoFile.types()) {
         ClassName javaTypeName = (ClassName) javaGenerator.typeName(type.protoTypeName());
-        TypeSpec typeSpec = new TypeWriter(javaGenerator, shouldEmitOptions()).toTypeSpec(type);
+        TypeSpec typeSpec = typeWriter.toTypeSpec(type);
+        writeJavaFile(javaTypeName, typeSpec, wireProtoFile.sourcePath());
+      }
+
+      if (options.emitOptions && !wireProtoFile.wireExtends().isEmpty()) {
+        ClassName javaTypeName = javaGenerator.extensionsClass(wireProtoFile);
+        TypeSpec typeSpec = typeWriter.extensionsType(javaTypeName, wireProtoFile);
         writeJavaFile(javaTypeName, typeSpec, wireProtoFile.sourcePath());
       }
     }
--- a/wire-compiler/src/main/java/com/squareup/wire/java/JavaGenerator.java
+++ b/wire-compiler/src/main/java/com/squareup/wire/java/JavaGenerator.java
@@ -20,6 +20,7 @@ import com.squareup.javapoet.ClassName;
 import com.squareup.javapoet.ParameterizedTypeName;
 import com.squareup.javapoet.TypeName;
 import com.squareup.wire.ExtendableMessage;
+import com.squareup.wire.Extension;
 import com.squareup.wire.Message;
 import com.squareup.wire.internal.Util;
 import com.squareup.wire.model.ProtoTypeName;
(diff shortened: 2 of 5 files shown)


wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -1298,9 +1298,17 @@ private void compileWithJavaPoet() throws WireException {
         continue; // Don't emit anything for files not explicitly compiled.
       }
 
+      TypeWriter typeWriter = new TypeWriter(javaGenerator, options.emitOptions);
```
JakeWharton: I giggle internally at the juxtaposition this class name evokes every time I see it. Especially since it's not done obnoxiously like making a calculator implementation in `Abacus.java` or something.

wire-compiler/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -45,9 +46,13 @@
   public static final ClassName BYTE_STRING = ClassName.get(ByteString.class);
   public static final ClassName STRING = ClassName.get(String.class);
   public static final ClassName LIST = ClassName.get(List.class);
+  public static final ClassName MESSAGE = (ClassName) TypeName.get(Message.class);
+  public static final ClassName EXTENDABLE_MESSAGE =
+      (ClassName) TypeName.get(ExtendableMessage.class);
```
JakeWharton: Why not use `ClassName.get` for all these?

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -340,12 +339,7 @@ private MethodSpec messageEquals(WireMessage type) {
     result.addStatement("if (other == this) return true");
     result.addStatement("if (!(other instanceof $T)) return false", javaType);
 
-    // TODO(jwilson): extensions!
-    //if (compiler.hasExtensions(messageType)) {
-    //  writer.emitStatement("if (!extensionsEqual(o)) return false");
-    //}
-
-    if (fields.size() == 1) {
+    if (fields.size() == 1 && type.extensions().isEmpty()) {
```
JakeWharton: Is this special case even worth it anymore? The byte code is equivalent, it just saves one line of source.

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -649,4 +664,76 @@ private static CodeBlock codeBlock(String format, Object... args) {
...
+
+    return builder.build();
+  }
+
+  private FieldSpec extensionField(
+      WireProtoFile wireProtoFile, TypeName extendType, WireField field) {
+    TypeName fieldType = javaGenerator.typeName(field.type());
+
+    CodeBlock.Builder initializer = CodeBlock.builder();
+    initializer.add("$[Extension\n");
```
JakeWharton: Nice to see this operator in use

wire-compiler/src/main/java/com/squareup/wire/model/WireProtoFile.java
```diff
@@ -75,6 +75,25 @@ public String sourcePath() {
     return sourcePath;
   }
 
+  /**
+   * Returns the name of this proto file, like {@code simple_message} for {@code
+   * squareup.protos.person.simple_message.proto}.
```
JakeWharton: should be slashes

wire-compiler/src/test/java/com/squareup/wire/WireCompilerTest.java
```diff
@@ -90,7 +90,9 @@ private void testProto(String[] sources, String[] outputs, String serviceWriter,
     invokeCompiler(args);
 
     List<String> filesAfter = getAllFiles(testDir);
-    assertEquals(filesAfter.toString(), outputs.length, filesAfter.size());
+    if (outputs.length != filesAfter.size()) {
```
JakeWharton: Why the conditional?

JakeWharton:
lgtm

wire-compiler/src/main/java/com/squareup/wire/java/JavaGenerator.java
swankjesse: Because I am dumb. Fixed.

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
swankjesse: Possibly not, but I want to minimize delta in the migration. Once that's done we can simplify this stuff.

wire-compiler/src/main/java/com/squareup/wire/model/WireProtoFile.java
swankjesse: Yes, definitely. Fixed.

wire-compiler/src/test/java/com/squareup/wire/WireCompilerTest.java
swankjesse: That snuck in by mistake. (I wanted to hit a breakpoint when the values were different.)

PR #227, opened 2015-07-03
Get much closer on options on JavaPoet.

swankjesse:
The code to convert options from an OptionElement to a Map is exhausting.
But it works now, and we have precise type information which is comforting.


checkstyle.xml
```diff
@@ -101,7 +101,7 @@
     <!--module name="AvoidInlineConditionals"/-->
     <module name="CovariantEquals"/>
     <module name="EmptyStatement"/>
-    <module name="EqualsAvoidNull"/>
+    <!--<module name="EqualsAvoidNull"/>-->
```
swankjesse: Checkstyle insists that this can cause null pointer exceptions, and I should flip the `.equals()`. I disagree.

```
if (lunch.equals("taco")) { ... }
```

wire-compiler/src/main/java/com/squareup/wire/model/Options.java
```diff
@@ -0,0 +1,256 @@
...
+    this.optionElements = ImmutableList.copyOf(elements);
+  }
+
+  public String packageName() {
+    return packageName;
+  }
+
+  /**
+   * Returns a map with the values for these options. Map values may be either a single entry, like
+   * {@code {default: "5"}}, or more sophisticated, with nested maps and lists.
```
JakeWharton: nit: Is Javadoc smart enough here? I think you'll probably need to use `<code>` HTML tags due to the curly braces.

JakeWharton:
> The code to convert options from an OptionElement to a Map is exhausting.

It's pretty terrible on the parsing side as well. Having done the parsing side the conversion code here was actually fairly easy to read.

PR #230, opened 2015-07-04
Message options with JavaPoet.

swankjesse:
Currently passing 23 of 35 test cases. Only service tests are failing.


JakeWharton:
:+1: keep on chuggin'

PR #232, opened 2015-07-04
Services on JavaPoet.

swankjesse:


swankjesse:
Passing 29 or 35 test cases. The last thing is emitting the extension registry.

wire-compiler/src/main/java/com/squareup/wire/java/ServiceFactory.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.squareup.javapoet.TypeSpec;
+import com.squareup.wire.model.WireService;
+import java.util.List;
+
+/**
+ * Generates a Java service class for a .proto RPC service. Implementing classes must have a no
+ * arguments constructor.
+ */
+public interface ServiceFactory {
```
swankjesse: This is the minimum interface to get us to JavaPoet, but we could probably take it even further. It'd be great to make this thing much more general-purpose, so that plugins could generate whatever they wanted.

wire-compiler/src/main/java/com/squareup/wire/java/RxJavaServiceFactory.java
```diff
@@ -0,0 +1,125 @@
...
+    typeBuilder.addField(endpointName, "endpoint", PRIVATE, FINAL);
+    typeBuilder.addType(endpointBuilder.build());
+    return typeBuilder.build();
+  }
+
+  private String upperToLowerCamel(String string) {
+    return UPPER_CAMEL.to(LOWER_CAMEL, string);
+  }
+
+  protected ClassName interfaceName(
```
swankjesse: I don't think this really makes sense as a general purpose extension point, but it gets us closer to the goal of the switch. Once that happens a lot of clean up should be possible.

wire-compiler/src/main/java/com/squareup/wire/java/ServiceFactory.java
JakeWharton: Just so I'm clear, you're just trying to get the existing tests to pass, right?

I want the whole service generator thing to be ripped out. It's far too presumptuous and limited to be of value. I don't think any apps use it, plus it's about to be obsoleted by Retrofit 2. The existing implementations here should absolutely not be shipping in the compiler when it's done.

> plugins could generate whatever they wanted

I think this is the _only_ path forward. The specialization of services is limiting and we have tons of other code we want to generate that has full access to the object graph.

swankjesse:
I pushed registry generation, which brings the JavaPoet fork to 100% of the `WireCompiler` tests passing.

wire-compiler/src/main/java/com/squareup/wire/CommandLineOptions.java
```diff
@@ -18,6 +19,7 @@
   public static final String NO_OPTIONS_FLAG = "--no_options";
   public static final String ENUM_OPTIONS_FLAG = "--enum_options=";
   public static final String SERVICE_WRITER_FLAG = "--service_writer=";
+  public static final String SERVICE_FACTORY_FLAG = "--service_factory=";
```
JakeWharton: You split two constants for the same subject. Move this up or down one.

wire-compiler/src/main/java/com/squareup/wire/java/ServiceFactory.java
swankjesse: We're on the same page.

swankjesse: (I'm trying to separate refactoring from behavior change.)

JakeWharton: :+1: :+1:

PR #233, opened 2015-07-04
Switch to JavaPoet.

swankjesse:
This large change flips a flag to turn on the new code path, and it changes
some insignificant formatting in many of the wire compiler tests.

Most of the compiler tests are changed due to things like static
imports, which aren't yet supported in JavaPoet. When those sorts
of things are supported, the code formatting will be improved.


wire-runtime/src/test/java/com/squareup/services/RxJavaService2SomeEndpoints.java
```diff
@@ -59,4 +42,18 @@ public RxJavaService2SomeEndpoints(Endpoint endpoint) {
   public Func1<LetsDataRequest, LetsDataResponse> getLetsData() {
     return letsData;
   }
+
+  public interface Endpoint {
+    /**
+     * Sends some more data.
+     */
+    @POST("/com.squareup.services.RxJavaService2/SendSomeMoreData")
```
swankjesse: Note that this is the one behavior change. The annotation now retains the same name as the proto service unconditionally, even if a suffix is added to the generated interface.

Specifically the `SomeEndpoints` suffix is added to the interface but not to the annotation's value.
```
-    @POST("/com.squareup.services.RxJavaService2SomeEndpoints/LetsData")
-    LetsDataResponse letsData(@Body LetsDataRequest request);
```

wire-runtime/src/test/java/com/squareup/services/RxJavaService.java
```diff
@@ -13,37 +14,19 @@
  * An example service.
  */
 public final class RxJavaService {
-
-  public interface Endpoint {
```
swankjesse: nested classes are now at the bottom of the class

wire-runtime/src/test/java/com/squareup/wire/protos/custom_options/FooBar.java
```diff
@@ -320,29 +364,33 @@ public More build() {
...
-    public final More complex_enum_value_option;
+
     public final Integer enum_value_option;
+
+    public final More complex_enum_value_option;
+
     public final Boolean foreign_enum_value_option;
 
-    FooBarBazEnum(int value, More complex_enum_value_option, Integer enum_value_option, Boolean foreign_enum_value_option) {
+    FooBarBazEnum(int value, Integer enum_value_option, More complex_enum_value_option, Boolean foreign_enum_value_option) {
```
swankjesse: The argument of these parameters changed; no user-visible impact.

wire-runtime/src/test/java/com/squareup/services/RxJavaService2SomeEndpoints.java
JakeWharton: Seems to be a bugfix since the remote service would be based on the proto name not the generated class name.

JakeWharton:
lgtm! interesting to see some of generated code get better and other regress a tiny bit.

PR #234, opened 2015-07-04
Complete the switch to JavaPoet by dropping JavaWriter.

swankjesse:
Plus all of the supporting code.


adriancole:
might want to permute JavaWriter and JavaPoet in the PR desc :)

swankjesse:
Yipes.

swankjesse:
Good to go. I think that landing this in master will kick off the start of a Wire 2.0 push.

JakeWharton:
Crashes mobile Chrome, I'll look in a bit.

JakeWharton:
:100:

PR #236, opened 2015-07-05
Land the JavaPoet Feature branch.

swankjesse:


JakeWharton:
LGTM. Can you add a commit where you `mvn versions:set` the number to "2.0.0-SNAPSHOT" so auto-snapshot deployment is on the right version?

PR #240, opened 2015-07-06
Introduce MessageAdapter for reading and writing instances.

### LINKED ISSUE #167: Add Message.writeTo(Sink)
Some libraries (like Tape) provide access to an outputstream but which may actually be a ByteArrayOutputStream (like Tape).  It'd be nice to avoid an extra byte array if we can swing it.

### LINKED ISSUE #209: Expose type adapter in public API
For libraries like Retrofit, it is beneficial to be able to do a one-time look-up of an adapter which is a reader of types from data as well a writer of types to data. Libraries like Gson, Moshi, Jackson, and even Google protobuf have similar concepts.

We should expose this as a first-party mechanism and API.

Not a strict proposal but merely an indicator as to what I'm looking for:
```java
<T extends Message> MessageAdapter<T> adapter(Class<T> cls);
```
```java
interface MessageAdapter<T extends Message> {
  T read(Source source);
  void write(T value, Sink sink);
}
```

JakeWharton:
This exposes a public API for an internal concept which is a class instance that is responsible for reading and writing protos of a specific type. This allows consumers to cache these instances for use while serializing to/from bytes. With this change all other means of reading and writing encoded data have been removed except for `toByteArray()` on instances.

Additionally, a few types and a lot of methods were removed from being exposed the public API as they are implementation details of serialization.

Closes #209. Closes #167.

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/Message.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/Message.java
@@ -101,11 +101,8 @@ public abstract class Message implements Serializable {
   /** Set to null until a field is added. */
   private transient UnknownFieldMap unknownFields;
 
-  /** False upon construction or deserialization. */
-  private transient boolean haveCachedSerializedSize;
-
-  /** If {@code haveCachedSerializedSize} is true, the serialized size of this message. */
-  private transient int cachedSerializedSize;
+  /** If not {@code -1} then the serialized size of this message. */
+  transient int cachedSerializedSize = -1;
 
   /** If non-zero, the hash code of this message. Accessed by generated code. */
   protected transient int hashCode = 0;
@@ -160,46 +157,13 @@ public abstract class Message implements Serializable {
     return adapter.fromInt(value);
   }
 
-  @SuppressWarnings("unchecked")
-  public byte[] toByteArray() {
-    return WIRE.messageAdapter((Class<Message>) getClass()).toByteArray(this);
-  }
-
-  public void writeTo(byte[] output) {
-    writeTo(output, 0, output.length);
-  }
-
-  public void writeTo(byte[] output, int offset, int count) {
-    write(WireOutput.newInstance(output, offset, count));
-  }
-
-  @SuppressWarnings("unchecked")
-  private void write(WireOutput output) {
-    MessageAdapter<Message> adapter = WIRE.messageAdapter((Class<Message>) getClass());
-    try {
-      adapter.write(this, output);
-    } catch (IOException e) {
-      throw new RuntimeException(e);
-    }
-  }
-
   void writeUnknownFieldMap(WireOutput output) throws IOException {
     if (unknownFields != null) {
       unknownFields.write(output);
     }
   }
 
-  @SuppressWarnings("unchecked")
-  public int getSerializedSize() {
-    if (!haveCachedSerializedSize) {
-      MessageAdapter<Message> adapter = WIRE.messageAdapter((Class<Message>) getClass());
-      cachedSerializedSize = adapter.getSerializedSize(this);
-      haveCachedSerializedSize = true;
-    }
-    return cachedSerializedSize;
-  }
-
-  public int getUnknownFieldsSerializedSize() {
+  int getUnknownFieldsSerializedSize() {
     return unknownFields == null ? 0 : unknownFields.getSerializedSize();
   }
 
--- a/wire-runtime/src/main/java/com/squareup/wire/MessageAdapter.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/MessageAdapter.java
@@ -16,6 +16,8 @@
 package com.squareup.wire;
 
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.io.Serializable;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
@@ -29,25 +31,25 @@ import java.util.List;
 import java.util.Map;
 import java.util.RandomAccess;
 import java.util.Set;
+import okio.Buffer;
+import okio.BufferedSink;
+import okio.BufferedSource;
 import okio.ByteString;
+import okio.Okio;
 
 import static com.squareup.wire.ExtendableMessage.ExtendableBuilder;
 import static com.squareup.wire.Message.Builder;
 import static com.squareup.wire.Message.Datatype;
 import static com.squareup.wire.Message.Label;
+import static com.squareup.wire.Preconditions.checkNotNull;
 
-/**
- * An adapter than can perform I/O on a given Message type.
- *
- * @param <M> the Message class handled by this adapter.
- */
-final class MessageAdapter<M extends Message> {
+public final class MessageAdapter<M extends Message> {
   // Unicode character "Full Block" (U+2588)
   private static final String FULL_BLOCK = "█";
   // The string to use when redacting fields from toString.
   private static final String REDACTED = FULL_BLOCK + FULL_BLOCK;
 
-  public static final class FieldInfo {
+  static final class FieldInfo {
     final int tag;
     final String name;
     final Datatype datatype;
@@ -120,7 +122,7 @@ final class MessageAdapter<M extends Message> {
     }
   }
 
-  public void setBuilderField(Builder<M> builder, FieldInfo fieldInfo, Object value) {
+  void setBuilderField(Builder<M> builder, FieldInfo fieldInfo, Object value) {
     try {
       fieldInfo.builderField.set(builder, value);
     } catch (IllegalAccessException e) {
@@ -128,7 +130,7 @@ final class MessageAdapter<M extends Message> {
     }
   }
 
-  public void setBuilderMethod(Builder<M> builder, FieldInfo fieldInfo, Object value) {
+  void setBuilderMethod(Builder<M> builder, FieldInfo fieldInfo, Object value) {
     try {
       fieldInfo.builderMethod.invoke(builder, value);
     } catch (IllegalAccessException e) {
@@ -302,8 +304,35 @@ final class MessageAdapter<M extends Message> {
     return size;
   }
 
-  /** Uses reflection to write {@code message} to {@code output} in serialized form. */
-  void write(M message, WireOutput output) throws IOException {
+  /** Encode {@code value} as a {@code byte[]}. */
+  public byte[] writeBytes(M value) {
+    checkNotNull(value, "value == null");
+    Buffer buffer = new Buffer();
+    try {
+      write(value, buffer);
+    } catch (IOException e) {
+      throw new AssertionError(e); // No I/O writing to Buffer.
+    }
+    return buffer.readByteArray();
+  }
+
+  /** Encode {@code value} and write it to {@code stream}. */
+  public void writeStream(M value, OutputStream stream) throws IOException {
+    checkNotNull(value, "value == null");
+    checkNotNull(stream, "stream == null");
+    BufferedSink buffer = Okio.buffer(Okio.sink(stream));
+    write(value, buffer);
+    buffer.emit();
+  }
+
+  /** Encode {@code value} and write it to {@code stream}. */
+  public void write(M value, BufferedSink sink) throws IOException {
+    checkNotNull(value, "value == null");
+    checkNotNull(sink, "sink == null");
+    write(value, new WireOutput(sink));
+  }
+
+  private void write(M message, WireOutput output) throws IOException {
     for (FieldInfo fieldInfo : getFields()) {
       Object value = getFieldValue(message, fieldInfo);
       if (value == null) {
@@ -371,19 +400,6 @@ final class MessageAdapter<M extends Message> {
     for (int i = 0, count = value.size(); i < count; i++) {
       writeValueNoTag(output, value.get(i), datatype);
     }
-  }
-
-  /**
-   * Serializes a given {@link Message} and returns the results as a byte array.
-   */
-  byte[] toByteArray(M message) {
-    byte[] result = new byte[getSerializedSize(message)];
-    try {
-      write(message, WireOutput.newInstance(result));
-    } catch (IOException e) {
-      throw new RuntimeException(e);
-    }
-    return result;
   }
 
   /**
@@ -477,8 +493,12 @@ final class MessageAdapter<M extends Message> {
 
   @SuppressWarnings("unchecked")
   private <MM extends Message> int getMessageSize(MM message) {
-    int messageSize = message.getSerializedSize();
-    return WireOutput.varint32Size(messageSize) + messageSize;
+    int size = message.cachedSerializedSize;
+    if (size == -1) {
+      MessageAdapter<MM> adapter = wire.messageAdapter((Class<MM>) message.getClass());
+      size = message.cachedSerializedSize = adapter.getSerializedSize(message);
+    }
+    return WireOutput.varint32Size(size) + size;
   }
 
   private void writeValue(WireOutput output, int tag, Object value, Datatype datatype)
@@ -521,8 +541,12 @@ final class MessageAdapter<M extends Message> {
 
   @SuppressWarnings("unchecked")
   private <MM extends Message> void writeMessage(MM message, WireOutput output) throws IOException {
-    output.writeVarint32(message.getSerializedSize());
     MessageAdapter<MM> adapter = wire.messageAdapter((Class<MM>) message.getClass());
+    int size = message.cachedSerializedSize;
+    if (size == -1) {
+      size = message.cachedSerializedSize = adapter.getSerializedSize(message);
+    }
+    output.writeVarint32(size);
     adapter.write(message, output);
   }
 
@@ -533,6 +557,29 @@ final class MessageAdapter<M extends Message> {
   }
 
   // Reading
+
+  /** Read an encoded message from {@code source}. */
+  public M read(BufferedSource source) throws IOException {
+    checkNotNull(source, "source == null");
+    return read(new WireInput(source));
+  }
+
+  /** Read an encoded message from {@code bytes}. */
+  public M readBytes(byte[] bytes) throws IOException {
+    checkNotNull(bytes, "bytes == null");
+    Buffer buffer = new Buffer();
+    M message = read(buffer.write(bytes));
+    if (!buffer.exhausted()) {
+      throw new IllegalArgumentException("");
+    }
+    return message;
+  }
+
+  /** Read an encoded message from {@code stream}. */
+  public M readStream(InputStream stream) throws IOException {
+    checkNotNull(stream, "stream == null");
+    return read(Okio.buffer(Okio.source(stream)));
+  }
 
   /** Uses reflection to read an instance from {@code input}. */
   M read(WireInput input) throws IOException {
--- a/wire-runtime/src/main/java/com/squareup/wire/MessageSerializedForm.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/MessageSerializedForm.java
@@ -27,15 +27,17 @@ final class MessageSerializedForm implements Serializable {
   private final Class<? extends Message> messageClass;
 
   public MessageSerializedForm(Message message, Class<? extends Message> messageClass) {
-    this.bytes = message.toByteArray();
+    //noinspection unchecked
+    this.bytes = Message.WIRE.adapter((Class<Message>) messageClass).writeBytes(message);
     this.messageClass = messageClass;
   }
 
   Object readResolve() throws ObjectStreamException {
+    MessageAdapter<? extends Message> adapter = Message.WIRE.adapter(messageClass);
     try {
       // Extensions are not supported at this time. Extension fields will be added to the
       // unknownFields map.
-      return Message.WIRE.parseFrom(bytes, messageClass);
+      return adapter.readBytes(bytes);
     } catch (IOException e) {
       throw new StreamCorruptedException(e.getMessage());
     }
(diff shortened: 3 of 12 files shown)


wire-runtime/src/main/java/com/squareup/wire/Wire.java
```diff
@@ -115,59 +115,6 @@ public Wire(List<Class<?>> extensionClasses) {
   }
 
   /**
-   * Reads a message of type {@code messageClass} from {@code bytes} and returns
-   * it.
-   */
-  public <M extends Message> M parseFrom(byte[] bytes, Class<M> messageClass) throws IOException {
```
danrice-square: Don't remove this method, it is very commonly used.

JakeWharton: By who? This means someone is buffering all their data instead of using a stream.

wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -160,46 +161,40 @@ protected void setBuilder(Builder builder) {
...
   }
 
+  /**
+   * Create a {@code byte[]} of the serialized form of the data.
+   * <p>
+   * Note: This is an inefficient means of serializing data. For best performance, call
+   * {@link Wire#adapter} and then {@link MessageAdapter2#write}.
+   */
   @SuppressWarnings("unchecked")
   public byte[] toByteArray() {
```
JakeWharton: I'd love to remove this method too. You can't serialize a JSON type by calling `toJson()` on it. This seems like a holdover from trying to be protobuf which we can't do based on the fundamental choices herein. It's a convenience method to inefficiency.

But it requires a lot of test changes, so... followup, if so!

wire-runtime/src/main/java/com/squareup/wire/MessageAdapter2.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+package com.squareup.wire;
+
+import java.io.IOException;
+import okio.BufferedSink;
+import okio.BufferedSource;
+
+/** An adapter for reading and writing Wire messages. */
+// NOTE: This class is suffixed with a "2" for git's sake. It is renamed in the next commit.
```
JakeWharton: Will be renamed in a follow-up. Otherwise you get horrid diffs due to me re-using a name.

wire-runtime/src/main/java/com/squareup/wire/Wire.java
danrice-square: You might have a byte array from a Parcel, for example. Or sometimes a proto message contains a `bytes[]` field that is an encrypted version of another proto message. You don't always have access to a stream.

I've specifically called out this method (and .toByteArray()) as examples of how simple it is to use Wire. If you have to also know something (even just a little) about Okio streams in order to do basic serialization/deserialization, the examples get much less compelling. It would be great to have comments on the convenience methods that explain exactly how they work and how you can gain efficiency by using lower-level primitives in your application code.

wire-runtime/src/main/java/com/squareup/wire/Message.java
danrice-square: I would prefer to leave it. The analogy with JSON is flawed, the only common supertype for JSON objects is `Object`. This provides a simple way in to the API. Once the developer gets Wire working in their app, they can learn about MessageAdapter, Okio streams, etc., to make things more efficient.

JakeWharton: It's simple but it's the wrong way into the API. This method is doubly inefficient. It requires a parallel `Wire` instance, its backing maps, and its own message adapter instances which are massive objects to keep around. This is a waste of memory when we can re-use the main instance which is also more knowledgable about the extensions. This method also buffers the output into a `byte[]` when you are almost always writing to something better observed as a stream.

As to the base class, it's a convenience for simplifying the generated code but I don't see how that requires this method be here–especially since it has no effect on the API for deserialization.

danrice-square: I could be willing to compromise on a new instance method `Wire#toByteArray(Message)` (so as not to duplicate the MessageAdapters). I want a way to serialize that does not require any knowledge of Okio sinks or anything like that as a way in to the API.

wire-runtime/src/main/java/com/squareup/wire/Wire.java
JakeWharton: Yeah, that's fair. What about `readBytes`/`writeBytes` and `readStream`/`writeStream` on the `MessageAdapter` and changing it to an abstract class with default implementations?

wire-runtime/src/main/java/com/squareup/wire/Message.java
swankjesse: I like where this is going. Drop these methods, add methods on MessageAdapter that make it foolishly easy to go from a protocol message to a byte[].

wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -160,46 +161,40 @@ protected void setBuilder(Builder builder) {
...
   @SuppressWarnings("unchecked")
-  public int getSerializedSize() {
+  int getSerializedSize() {
     if (!haveCachedSerializedSize) {
-      MessageAdapter<Message> adapter = WIRE.messageAdapter((Class<Message>) getClass());
+      WireMessageAdapter<Message> adapter = WIRE.messageAdapter((Class<Message>) getClass());
       cachedSerializedSize = adapter.getSerializedSize(this);
       haveCachedSerializedSize = true;
     }
     return cachedSerializedSize;
```
swankjesse: Is this useful for the future?

wire-runtime/src/main/java/com/squareup/wire/MessageAdapter2.java
```diff
@@ -0,0 +1,62 @@
...
+import okio.Okio;
+
+/** An adapter for reading and writing Wire messages. */
+// NOTE: This class is suffixed with a "2" for git's sake. It is renamed in the next commit.
+public abstract class MessageAdapter2<M extends Message> {
+  /** Read an encoded message from {@code source}. */
+  public abstract M read(BufferedSource source) throws IOException;
+
+  /** Read an encoded message from {@code bytes}. */
+  public M readBytes(byte[] bytes) throws IOException {
```
swankjesse: Should this explode if there are dangling extra bytes ?

swankjesse: final ? I see no upside to allowing overriding

wire-runtime/src/main/java/com/squareup/wire/MessageAdapter2.java
```diff
@@ -0,0 +1,62 @@
...
+  /** Encode {@code value} and write it to {@code sink}. */
+  public abstract void write(M value, BufferedSink sink) throws IOException;
+
+  /** Encode {@code value} as a {@code byte[]}. */
+  public byte[] writeBytes(M value) {
+    Buffer buffer = new Buffer();
+    try {
+      write(value, buffer);
+    } catch (IOException e) {
+      throw new AssertionError(e); // No I/O writing to Buffer.
```
swankjesse: Nice.

wire-runtime/src/main/java/com/squareup/wire/MessageAdapter2.java
```diff
@@ -0,0 +1,62 @@
...
+import java.io.InputStream;
+import java.io.OutputStream;
+import okio.Buffer;
+import okio.BufferedSink;
+import okio.BufferedSource;
+import okio.Okio;
+
+/** An adapter for reading and writing Wire messages. */
+// NOTE: This class is suffixed with a "2" for git's sake. It is renamed in the next commit.
+public abstract class MessageAdapter2<M extends Message> {
```
swankjesse: Tempted to rethink a lot of the names in Wire. Maybe ProtobufAdapter?

wire-runtime/src/main/java/com/squareup/wire/Wire.java
```diff
@@ -41,7 +36,7 @@
   private final Map<Class<? extends ProtoEnum>, EnumAdapter<? extends ProtoEnum>> enumAdapters =
       new LinkedHashMap<Class<? extends ProtoEnum>, EnumAdapter<? extends ProtoEnum>>();
 
-  // Visible to MessageAdapter
+  // Visible to WireMessageAdapter
```
swankjesse: MessageAdapter2 ?!

wire-runtime/src/main/java/com/squareup/wire/MessageAdapter2.java
swankjesse: I don't get it. Why do we have both this and WireMessageAdapter? Can't those be one type?

wire-runtime/src/main/java/com/squareup/wire/WireOutput.java
```diff
@@ -291,26 +171,12 @@ void writeVarint64(long value) throws IOException {
...
-    writeRawByte((int) (value      ) & 0xFF);
-    writeRawByte((int) (value >>  8) & 0xFF);
-    writeRawByte((int) (value >> 16) & 0xFF);
-    writeRawByte((int) (value >> 24) & 0xFF);
-    writeRawByte((int) (value >> 32) & 0xFF);
-    writeRawByte((int) (value >> 40) & 0xFF);
-    writeRawByte((int) (value >> 48) & 0xFF);
-    writeRawByte((int) (value >> 56) & 0xFF);
-    // CHECKSTYLE.ON: ParenPad
+    sink.writeLongLe(value);
```
swankjesse: YES!

wire-runtime/src/main/java/com/squareup/wire/MessageAdapter2.java
JakeWharton: Probably. This was an interface before the convenience methods were added. Also I wanted to start small but it kind of bubbled into a bigger change.

wire-runtime/src/main/java/com/squareup/wire/WireOutput.java
```diff
@@ -50,11 +50,12 @@
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import java.io.IOException;
+import okio.BufferedSink;
 
 /**
  * Utilities for encoding and writing protocol message fields.
  */
-public final class WireOutput {
```
swankjesse: We can probably kill this now, right?

swankjesse: ... I guess we're lacking support for varint in `BufferedSink`. Still, I think it might be simpler to just have callsites call the static methods.

swankjesse: public static void writeVarint32(int v, BufferedSink sink) { ... }

swankjesse:
LGTM

wire-runtime/src/main/java/com/squareup/wire/WireOutput.java
JakeWharton: Yep. That's what I planned. Was trying to make this a small change but it kind of got out of control so I'll probably go all in. There's a couple small tweaks like this all over for reducing allocations.

wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -160,46 +161,40 @@ protected void setBuilder(Builder builder) {
...
 
-  public void writeUnknownFieldMap(WireOutput output) throws IOException {
+  void writeUnknownFieldMap(WireOutput output) throws IOException {
     if (unknownFields != null) {
       unknownFields.write(output);
     }
   }
 
   @SuppressWarnings("unchecked")
-  public int getSerializedSize() {
```
danrice-square: This seems like something that client code might need to know.

swankjesse: It's a lie, isn't it? Does the serialized size change depending on which `Wire` instance it's serialized with?

wire-runtime/src/main/java/com/squareup/wire/WireOutput.java
JakeWharton: So having done this, I'm not sure how much I like it. I think I'm going to back off and go the other way, turn it into `ProtoReader` and `ProtoWriter` (ala `JsonReader` and `JsonWriter`) and completely hide the semantics of the underlying `BufferedSink`.

wire-runtime/src/main/java/com/squareup/wire/MessageAdapter2.java
JakeWharton: Is "Protobuf" one word. It doesn't exactly roll off the tongue...

swankjesse: It's three syllables. WireAdapter ?

wire-runtime/src/main/java/com/squareup/wire/Message.java
JakeWharton: There's a two-phase write in which this helps. First the size is determined (and written) and then the actual data is written. Would be nice to encapsulate it as local state rather than instance state in the future.

wire-runtime/src/main/java/com/squareup/wire/Message.java
JakeWharton: Unsure if it's a lie, but I'm not sure who would something like this in the public API. The only use case I can think of is pre-sizing a byte[] but at that point you'd just use the method which creates the byte[] for you.

danrice-square: It's not a lie, there is enough information to compute the size even if you don't know about extensions.

One use case is to allocate space in some larger buffer that you manage. We do this - COGS has a message type that consists of a proto plus an extra byte to indicate whether the data is dirty:

```
    public byte[] toByteArray() {
      ObjectWrapper backingObject = build().getBackingObject();
      byte[] bytes = new byte[backingObject.getSerializedSize() + 1];
      bytes[0] = dirty ? (byte) 1 : (byte) 0;
      backingObject.writeTo(bytes, 1, bytes.length - 1);
      return bytes;
    }
```

JakeWharton:
PTAL. Much cleaner diff after all the smaller PRs I sent.

wire-runtime/src/main/java/com/squareup/wire/Message.java
JakeWharton: Would we still write code like that today though? That pre-dates Okio by quite a bit and I think we've come to know cogs as a design that's very far from normal use case.

Today we'd write that as:
```java
public ByteString toBytes() {
  Buffer buffer = new Buffer();
  buffer.writeByte(dirty ? 1 : 0);
  build().getBackingObject().write(buffer);
  return buffer.readByteString();
}
```
which is much easier to understand.

wire-runtime/src/main/java/com/squareup/wire/WireOutput.java
JakeWharton: Doing this in a follow-up. Early prototypes look good though.

wire-runtime/src/main/java/com/squareup/wire/MessageAdapter.java
```diff
@@ -534,6 +552,26 @@ private void writeValueNoTag(WireOutput output, Object value, Datatype datatype)
...
+  public M read(BufferedSource source) throws IOException {
+    return read(new WireInput(source));
+  }
+
+  /** Read an encoded message from {@code bytes}. */
+  public M readBytes(byte[] bytes) throws IOException {
+    Buffer buffer = new Buffer();
+    M message = read(buffer.write(bytes));
+    if (!buffer.exhausted()) {
+      throw new IllegalArgumentException("");
```
JakeWharton: Forgot a message and test. But this was your question @swankjesse about left bytes.

wire-runtime/src/main/java/com/squareup/wire/MessageAdapter.java
```diff
@@ -534,6 +552,26 @@ private void writeValueNoTag(WireOutput output, Object value, Datatype datatype)
...
+    }
+    return message;
+  }
+
+  /** Read an encoded message from {@code stream}. */
+  public M readStream(InputStream stream) throws IOException {
+    return read(Okio.buffer(Okio.source(stream)));
```
JakeWharton: Bytes could also be left in this buffer :cry: . Not sure what to do about that. `java.io` for the lose.

wire-runtime/src/main/java/com/squareup/wire/MessageAdapter.java
```diff
@@ -534,6 +558,29 @@ private void writeValueNoTag(WireOutput output, Object value, Datatype datatype)
...
+    Buffer buffer = new Buffer();
+    M message = read(buffer.write(bytes));
+    if (!buffer.exhausted()) {
+      throw new IllegalArgumentException("");
+    }
+    return message;
+  }
+
+  /** Read an encoded message from {@code stream}. */
+  public M readStream(InputStream stream) throws IOException {
```
swankjesse: This method is slightly hazardous, in that it can read more bytes than it's supposed to. Imagine a stream that has protocol buffers stacked end-on-end. Sigh.

wire-runtime/src/main/java/com/squareup/wire/MessageSerializedForm.java
```diff
@@ -27,15 +27,16 @@
   private final Class<? extends Message> messageClass;
 
   public MessageSerializedForm(Message message, Class<? extends Message> messageClass) {
-    this.bytes = message.toByteArray();
+    this.bytes = Message.WIRE.adapter((Class<Message>) messageClass).writeBytes(message);
```
swankjesse: What's with the cast?

swankjesse: ahhh, I think I see it. You don't know that Message and messageClass agree. Ugh.

wire-runtime/src/main/java/com/squareup/wire/WireOutput.java
```diff
@@ -139,26 +108,48 @@ static int varint64Size(long value) {
     return 10;
   }
 
-  /** Write a single byte, represented by an integer value. */
+  /**
+   * Encode a ZigZag-encoded 32-bit value.  ZigZag encodes signed integers
```
swankjesse: Two spaces after a period won't make it to the generated Javadoc.

wire-runtime/src/main/java/com/squareup/wire/MessageSerializedForm.java
JakeWharton: This class could probably be parameterized. Was just copy/paste replacing `toByteArray` implementations and didn't look too hard.

swankjesse:
LGTM

wire-runtime/src/main/java/com/squareup/wire/MessageAdapter.java
JakeWharton: Yep. Don't know what to do here... What does Gson do?

JakeWharton: Also, a 0 byte InputStream is an infinitely long stream of any message which has all optional fields!

JakeWharton: * Gson requires a `Reader` or `JsonReader`.
 * Moshi requires a `BufferedSource` or `JsonReader`.
 * Wire 1.x was broken in the same way as this PR.
 * Protobuf documents it consumes the whole `InputStream` or requires you to limit or use `CodedInputStream`.
 * Jackson appears to be broken in the same way but I stopped following the ridiculously abstracted code after a while to verify.

JakeWharton: I'm inclined to leave it "broken" here since it's technically not a behavior change. We can address in a follow-up, potentially with a public `ProtobufReader` type (ala `JsonReader`).

JakeWharton:
Moving forward here. Happy to take more feedback if there is any.

wire-runtime/src/main/java/com/squareup/wire/MessageAdapter.java
swankjesse: Yep, leaving it “broken” is our only reasonable choice because of the 0-byte messages.

swankjesse: Actually, probably the right behavior is to check if the buffer is exhausted, and if it isn't we report an error. In that case the caller shouldn't have been using this API.

JakeWharton: As discussed but to document for others, Wire exhausts input so this isn't necessary.

drmichi:
I do understand your intentions, but consider the following:
1) Protocol buffers are binary byte representations, hence this is an essential feature
2) traditional ProtoBuf users are familiar with "byte[] toByteArray()" and "parseFrom(byte[] data)"
https://developers.google.com/protocol-buffers/docs/javatutorial

I can imagine the following use case:
A) Existing users of google ProtoBuf plan to switch to Wire
B) Existing transport layer is written to handle raw byte[] array

Let's compare the different implementations..

Google ProtoBuf Java:
byte[] serializedMsg = msg.toByteArray();

Google ProtoBuf JavaNano:
byte[] serializedMsg = MessageNano.toByteArray(msg);

Wire 1.8:
byte[] serializedMsg = msg.toByteArray();

Wire 2.0:
// i get it, the recommendation is to have a singleton for it, but just to prove a point
Wire wire = new Wire();
byte[] serializedMsg = wire.adapter(MyMessage.class).writeBytes(msg);

swankjesse:
You can make a static method, `toByteArray(Message)`if that suits you.

The reason we do it this way is that it's faster! The adapter is stateful and computing once & reusing will be a performance win.

PR #244, opened 2015-07-06
Remove a lot of implicit iterator allocations by loops.

JakeWharton:


swankjesse:
LGTM

JakeWharton:
These add up to quite a few per-message serialization. I was surprised. Every bit counts, especially in a library!

PR #251, opened 2015-07-07
Use ByteString for writing raw bytes. Remove redundant overloads.

JakeWharton:


wire-runtime/src/main/java/com/squareup/wire/MessageAdapter.java
```diff
@@ -501,14 +501,14 @@ private void writeValueNoTag(WireOutput output, Object value, Datatype datatype)
       case BOOL: output.writeRawByte((Boolean) value ? 1 : 0); break;
       case ENUM: writeEnum((ProtoEnum) value, output); break;
       case STRING:
-        final byte[] bytes = ((String) value).getBytes("UTF-8");
-        output.writeVarint32(bytes.length);
+        ByteString bytes = ByteString.encodeUtf8((String) value);
```
JakeWharton: Sadly this doesn't get the speedy UTF encoding :(

swankjesse: Actually it does. It uses the platform's built-in encoder without unnecessary extra allocations.

wire-runtime/src/main/java/com/squareup/wire/WireOutput.java
```diff
@@ -214,31 +215,22 @@ static int varint64Size(long value) {
...
+    buffer[position++] = (byte) value;
   }
 
-  /** Write part of an array of bytes. */
-  void writeRawBytes(byte[] value, int offset, int length) throws IOException {
-    if (limit - position >= length) {
+  /** Write a string of bytes. */
+  void writeRawBytes(ByteString bytes) throws IOException {
+    byte[] value = bytes.toByteArray();
+    if (limit - position >= value.length) {
```
swankjesse: Nice 2 GiB bounds check!

PR #253, opened 2015-07-09
Remove old ByteString test.

JakeWharton:
The Okio ByteString in use is well tested in its library.


danrice-square:
LGTM if equivalent test cases are present in Okio.

PR #255, opened 2015-07-09
Drop the Wire prefix where it's not essential.

swankjesse:


wire-compiler/src/main/java/com/squareup/wire/Type.java
```diff
@@ -0,0 +1,210 @@
...
+            if (name.isScalar) {
+              scalarTypes.put(name.names.get(0), name);
+            }
+          }
+        }
+        SCALAR_TYPES = scalarTypes.build();
+      } catch (IllegalAccessException e) {
+        throw new AssertionError();
+      }
+    }
```
JakeWharton: This number of LOC vs just listing them in `add()` calls is probably the same.

wire-compiler/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -58,72 +57,72 @@
...
+          .put(Type.Name.FIXED64, TypeName.LONG.box())
+          .put(Type.Name.INT32, TypeName.INT.box())
+          .put(Type.Name.INT64, TypeName.LONG.box())
+          .put(Type.Name.SFIXED32, TypeName.INT.box())
+          .put(Type.Name.SFIXED64, TypeName.LONG.box())
+          .put(Type.Name.SINT32, TypeName.INT.box())
+          .put(Type.Name.SINT64, TypeName.LONG.box())
+          .put(Type.Name.STRING, ClassName.get(String.class))
+          .put(Type.Name.UINT32, TypeName.INT.box())
+          .put(Type.Name.UINT64, TypeName.LONG.box())
```
JakeWharton: See you do it here!

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -29,21 +29,20 @@
 import com.squareup.javapoet.TypeSpec;
 import com.squareup.javapoet.TypeVariableName;
 import com.squareup.javapoet.WildcardTypeName;
+import com.squareup.wire.EnumConstant;
+import com.squareup.wire.EnumType;
+import com.squareup.wire.Extend;
+import com.squareup.wire.Field;
 import com.squareup.wire.Message;
+import com.squareup.wire.MessageType;
```
JakeWharton: I think I dislike that the compiler and runtime share a namespace.

wire-compiler/src/main/java/com/squareup/wire/Type.java
swankjesse: Good call. Can fix.

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
swankjesse: I'll do a schema project in a follow up; that'll get it's own package!

wire-compiler/src/main/java/com/squareup/wire/Type.java
swankjesse: https://github.com/square/wire/pull/260

PR #256, opened 2015-07-12
Extract schema loading, linking, pruning into its own module.

swankjesse:


wire-schema/src/test/java/com/squareup/wire/schema/OptionsTest.java
```diff
@@ -13,8 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.squareup.wire;
+package com.squareup.wire.schema;
 
+import com.squareup.wire.schema.Options;
```
danrice-square: Why an import from the same package?

wire-schema/src/test/java/com/squareup/wire/schema/TypeNameTest.java
```diff
@@ -13,8 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.squareup.wire;
+package com.squareup.wire.schema;
 
+import com.squareup.wire.schema.Type;
```
danrice-square: Why an import from the same package?

danrice-square:
LGTM

wire-schema/src/test/java/com/squareup/wire/schema/OptionsTest.java
swankjesse: Caught in refactoring. Fixed.

wire-schema/src/test/java/com/squareup/wire/schema/TypeNameTest.java
swankjesse: Caught in refactoring. Fixed.

PR #257, opened 2015-07-13
Import ProtoParser into Wire.

swankjesse:


danrice-square:
Why?

swankjesse:
Here's the Wire tech stack:

1. `protoparser`: basic syntax parsing, no linking.
2. `wire-schema`: a linked model over `protoparser`.
3. `wire-compiler` emit code based on that linked model.

The standalone utility of #1 is limited, and it's awkward to have an API boundary between the parsed models and the linked models. I'm attempting to bring layers 1 and 2 closer together, with the ultimate goal that layer #1 is just an implementation detail of `wire-schema` instead of a standalone dependency.

Current clients of `protoparser` should use `wire-schema` instead!

danrice-square:
Sounds great, can we add something to the `README.md` over in `protoparser`?

swankjesse:
Yes, definitely. Do you mind if I save that for a follow-up? This change is just about importing the code & retaining git history.

danrice-square:
LGTM

PR #264, opened 2015-07-17
New Schema and Location models.

swankjesse:
The new Schema class allows Linker and Pruner to be hidden as implementation
details of the wire-schema module. Other users of the wire-schema library
shouldn't need to use these directly.

The new Location class locates a message, enum, extension or service within
a .proto file. Later we can add support for fields, options, and RPCs, which
will allow for better error messages when a link cannot be resolved.


wire-schema/src/main/java/com/squareup/wire/schema/Loader.java
```diff
@@ -15,62 +15,86 @@
...
+  public Loader(IO io) {
     this.io = io;
   }
 
-  /** Recursively add {@code protoFile} and its dependencies. */
-  public void add(String protoFileName) throws IOException {
-    if (!protoFileNames.add(protoFileName)) {
+  /** Returns a loader that loads all files from {@code base}. */
+  public static Loader forBaseDirectory(final String base) {
+    IO io = new IO() {
```
JakeWharton: Needs a `toString()` as you implicitly use it in error messages.

swankjesse: Yes! Fixed.

JakeWharton:
lgtm

PR #265, opened 2015-07-17
Use recursive generics when referring to extendable message.

JakeWharton:


JakeWharton:
Oh, great. Java 7 and old Java 8 compilers inference fail. Works fine on _45 locally... UGH

swankjesse:
LGTM

PR #266, opened 2015-07-17
Better error messages when linking fails.

swankjesse:
Also move some type validation from ProtoParser into the schema package.
In particular, we no longer enforce that an RPC's type is not a scalar
in the protoparser layer; that happens in the linker.

In the future we can move more validation into the linker (unique tag
names, etc.) which should improve usability.


wire-schema/src/main/java/com/squareup/wire/internal/protoparser/DataType.java
```diff
@@ -1,141 +0,0 @@
...
-
-import java.util.Locale;
-
-import static com.google.common.base.Preconditions.checkNotNull;
-
-/**
- * Representation of a scalar, map, or named type. While this class is an interface, only the
- * included implementations are supported.
- */
-public interface DataType {
```
swankjesse: Replaced with `String` at this layer.

wire-schema/src/test/java/com/squareup/wire/schema/SchemaTest.java
```diff
@@ -0,0 +1,263 @@
...
+    try {
+      new SchemaBuilder()
+          .add("message.proto", ""
+              + "message Message {\n"
+              + "  optional foo_package.Foo unknown = 1;\n"
+              + "}\n")
+          .build();
+      fail();
+    } catch (SchemaException expected) {
+      assertThat(expected).hasMessage("unable to resolve foo_package.Foo\n"
```
swankjesse: Advice on the structure of these error messages is welcome.

wire-schema/src/main/java/com/squareup/wire/schema/SchemaException.java
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire.schema;
+
+import com.google.common.base.Joiner;
+import java.util.List;
+
+final class SchemaException extends RuntimeException {
```
JakeWharton: `public` so it's catchable by external users?

JakeWharton:
lgtm

wire-schema/src/test/java/com/squareup/wire/schema/SchemaTest.java
JakeWharton: I think they're fine. If I had one change it would be to change the subsequent "for"s to " in" and to indent those lines two spaces.
```
unable to resolve foo_package.Foo
  for field unknown (message.proto at 2:3)
  in message Message (message.proto at 1:1)
```

wire-schema/src/main/java/com/squareup/wire/schema/SchemaException.java
swankjesse: Good idea. Fixed.

wire-schema/src/test/java/com/squareup/wire/schema/SchemaTest.java
swankjesse: I like. Fixd.

PR #267, opened 2015-07-17
Update Gson's test protos in gen_tests script.

JakeWharton:
Also:
 * Affix an extension so IDEA will syntax highlight in Bash.
 * Fail the Travis CI build when the gen_tests.sh script creates uncommited changes.


.travis.yml
```diff
@@ -4,6 +4,10 @@ jdk:
   - oraclejdk7
   - oraclejdk8
 
+before_script:
+  - ./gen_tests.sh
+  - if [ ! -z "$(git status --porcelain)" ]; then echo "Test files changed. Did you run gen_tests.sh?"; exit 1; fi
```
JakeWharton: screen shot 2015-07-17 at 1 08 09 pm

PR #274, opened 2015-07-20
Use AutoValue.Builder in protoparser.

swankjesse:


danrice-square:
Awesome. Do you happen to know if autovalue classes "just work" with gson?

danrice-square:
LGTM

adriancole:
They don't OOB, as they need a type adapter for the class hierarchy.

Eventhough there's only one subclass, gson doesn't AutoValue_ is for
deserialization.

There's a number of solutions to this, but some config is needed.

wire-schema/src/main/java/com/squareup/wire/internal/protoparser/EnumConstantElement.java
```diff
@@ -18,28 +18,25 @@
...
+        .documentation("")
+        .options(ImmutableList.<OptionElement>of());
   }
 
   public abstract Location location();
   public abstract String name();
   public abstract int tag();
   public abstract String documentation();
-  public abstract List<OptionElement> options();
+  public abstract ImmutableList<OptionElement> options();
```
JakeWharton: Ha. I was going to say, "let's do immutable collections in the parser next!"

wire-schema/src/main/java/com/squareup/wire/internal/protoparser/RpcElement.java
```diff
@@ -62,57 +58,14 @@ public final String toSchema() {
...
-    public RpcElement build() {
-      checkNotNull(name, "name");
-      checkNotNull(requestType, "requestType");
-      checkNotNull(responseType, "responseType");
-
-      return new AutoValue_RpcElement(location, name, documentation, requestType, responseType,
-          ImmutableList.copyOf(options));
-    }
+  @AutoValue.Builder
+  public abstract static class Builder {
```
JakeWharton: These can all be interfaces if I remember correctly. Save you from typing `abstract static class` and `public abstract` everywhere.

JakeWharton:
Here's a really old example: https://gist.github.com/JakeWharton/0d67d01badcee0ae7bc9. You can also `@JsonAdapter` each type.

None of these types are being serialized though, or were you asking for different reasons?

danrice-square:
I have some unrelated code that relies on Gson serialization and I was wondering if I should try to use AutoValue on it.

PR #280, opened 2015-07-21
WIP: Switch code generation to instance visitor and static reader. 

JakeWharton:
This changes Wire to be reflection-free in every path except for the initial message adapter creation where a static field lookup has to be done.

Other wins:
 * Debuggable operations with stack traces that match object hierarchy.
 * ProtoReader and ProtoWriter syntax for 100% dynamic reading and writing of messages.
 * Massive upgrade in ProGuard compatibility with the ADAPTER field being the only thing needing to be kept.
 * Massive reduction in temporary objects during _every_ operation.

Note: This doesn't completely work yet. 7 tests are failing, but they should be passing in another day. Just wanted to get this out for early review. There's some tiny design problems in the API that I need to rectify, but it works in 99.9% of cases as-is!

The Gson module is disabled. It made it 90% of the way until I put the nail in the coffin of the `@ProtoField` annotation. I don't think it's viable as a reflection adapter and we should instead leverage the schema to code-gen Gson or Moshi adapters instead.


JakeWharton:
:boom:

JakeWharton:
Sorry I couldn't do this more incrementally. It's a pretty fundamental deviation which necessitated changes that were hard to isolate. You probably are better off checking out the branch and exploring.

JakeWharton:
Down to 5 failures! Check out `SimpleMessage.java` and `ExternalMessage.java` for representative changes.

JakeWharton:
I thought of some boundaries for which separate, more digestable PRs can be made. Will get one out soon.

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -153,25 +168,105 @@ private TypeSpec toTypeSpec(EnumType type) {
...
+    builder.addMethod(enumFromValue(type, typeName));
+
+    return builder.build();
+  }
+
+  // Example:
+  //
+  // @Override
+  // public int serializedSize(AnEnum value) {
+  //   return ProtoType.UINT32.serializedSize(value.value);
```
swankjesse: `ProtoType` -> `TypeAdapter` ?

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -506,6 +550,285 @@ private MethodSpec messageHashCode(MessageType type) {
...
+
+    return result.build();
+  }
+
+  // Example:
+  //
+  // @Override
+  // public Person read(ProtoReader reader) throws IOException {
+  //   Builder builder = new Builder();
+  //   while (reader.hasNext()) {
```
swankjesse: love this

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -506,6 +550,285 @@ private MethodSpec messageHashCode(MessageType type) {
...
+  //     int tag = reader.nextTag();
+  //     switch (tag) {
+  //       case 1:
+  //         builder.name = reader.value(ProtoType.STRING);
```
danrice-square: Why not `reader.string()`?

danrice-square:
I have concerns about this whole approach. We have always kept the amount of generated code quite minimal, and this seems like a substantial increase in the generated code size and complexity.

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -153,25 +168,105 @@ private TypeSpec toTypeSpec(EnumType type) {
...
+        .addAnnotation(Override.class)
+        .addModifiers(PUBLIC)
+        .returns(typeName)
+        .addParameter(ProtoReader.class, "reader")
+        .addException(IOException.class)
+        .addStatement("int value = $T.UINT32.read(reader)", TypeAdapter.class)
+        .addStatement("$T constant = fromValue(value)", typeName)
+        .beginControlFlow("if (constant == null)")
+        .addStatement("throw new $T($S + value)", IllegalStateException.class,
+            "No constant for value ")
```
swankjesse: missing the + value

swankjesse: nope, nevermind

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -153,25 +168,105 @@ private TypeSpec toTypeSpec(EnumType type) {
...
+        .endControlFlow()
+        .addStatement("return constant")
+        .build();
+  }
+
+  // Example:
+  //
+  // public static Binary fromValue(int tag) {
```
swankjesse: nit: tag –> value

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -476,29 +528,21 @@ private MethodSpec messageHashCode(MessageType type) {
       String name = sanitize(field.name());
       result.addStatement("int result = hashCode");
       result.addStatement(
-          "return result != 0 ? result : (hashCode = $L != null ? $L.hashCode() : $L)",
-          addThisIfOneOf(name, "result"), addThisIfOneOf(name, "result"), nullHashValue(field));
+          "return result != 0 ? result : (hashCode = $1L != null ? $1L.hashCode() : $2L)",
```
swankjesse: Nice `$1L`

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -476,29 +528,21 @@ private MethodSpec messageHashCode(MessageType type) {
...
       return result.build();
     }
 
     result.addStatement("int result = hashCode");
     result.beginControlFlow("if (result == 0)");
-    boolean afterFirstAssignment = false;
     if (!type.extensions().isEmpty()) {
-      result.addStatement("result = extensionsHashCode()");
-      afterFirstAssignment = true;
+      result.addStatement("result = result * 37 + extensionsHashCode()");
```
swankjesse: Proto3 drops support for extensions. Should we bundle that with wire2 ?

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -506,6 +550,285 @@ private MethodSpec messageHashCode(MessageType type) {
     return result.build();
   }
 
+  // Example:
+  //
+  // @Override
+  // public String toString() {
```
swankjesse: nit:

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -506,6 +550,285 @@ private MethodSpec messageHashCode(MessageType type) {
...
+          ? "toStringRedacted" : "toStringValue";
+      result.addStatement("$L(sb, $S, $L)", methodName, field.name(),
+          sanitize(field.name()));
+    }
+
+    if (!type.extensions().isEmpty()) {
+      result.addStatement("toStringValue(sb, $S, extensionsToString())", "{extensions");
+    }
+
+    result.addStatement("sb.insert(0, $S)", typeName.simpleName() + "{");
```
swankjesse: very curious. . .  why not append this first?

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -506,6 +550,285 @@ private MethodSpec messageHashCode(MessageType type) {
...
+  //
+  // @Override
+  // public int serializedSize(SomeMessage message) {
+  //   int serializedSize = message.serializedSize;
+  //   if (serializedSize == -1) {
+  //     serializedSize = message.unknownFieldsSerializedSize()
+  //         + sizeOf(1, opt_int32, TagWriter.INT32)
+  //         + sizeOf(2, other_type, OtherType.ADAPTER);
+  //   }
+  //   message.serializedSize = serializedSize;
```
swankjesse: this should be in the `if` block

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -506,6 +550,285 @@ private MethodSpec messageHashCode(MessageType type) {
...
+          expression.add("sizeOfPacked($L, message.$N, $L)", tag, name, adapter);
+        } else {
+          expression.add("sizeOf($L, message.$N, $L)", tag, name, adapter);
+        }
+      }
+    }
+    result.addStatement("serializedSize = $L", expression.build());
+
+    result.endControlFlow();
+    result.addStatement("message.serializedSize = serializedSize");
```
swankjesse: ... and this!

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -506,6 +550,285 @@ private MethodSpec messageHashCode(MessageType type) {
...
+      String name = sanitize(field.name());
+      CodeBlock adapter = typeAdapter(field.type());
+      if (field.isRepeated()) {
+        if (field.isPacked()) {
+          result.addStatement("writer.packed($L, message.$N, $L)", tag, name, adapter);
+        } else {
+          result.addStatement("writer.repeated($L, message.$N, $L)", tag, name, adapter);
+        }
+      } else {
+        result.addStatement("writer.value($L, message.$N, $L)", tag, name, adapter);
```
swankjesse: null check?

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -506,6 +550,285 @@ private MethodSpec messageHashCode(MessageType type) {
...
+        TypeName fieldName = javaGenerator.typeName(fieldType);
+        result.addStatement("int value = reader.value($T.UINT32)", TypeAdapter.class)
+            .addStatement("$1T constant = $1T.fromValue(value)", fieldName)
+            .beginControlFlow("if (constant != null)", name, fieldName)
+            .addStatement("builder.$L = constant", name)
+            .nextControlFlow("else")
+            .addStatement("builder.addVarint($L, value)", tag)
+            .endControlFlow();
+      } else {
+        result.addStatement("long cursor = reader.beginLengthDelimited()");
```
swankjesse: cursor is clever

wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
```diff
@@ -54,40 +55,117 @@
...
+    }
+    currentLimit = newLimit;
+    return oldLimit;
+  }
+
+  public boolean hasNext() throws IOException {
+    return pos < currentLimit && !source.exhausted();
+  }
+
+  public void endLengthDelimited(long cursor) throws IOException {
```
swankjesse: the cursor is genius

wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
```diff
@@ -0,0 +1,326 @@
+package com.squareup.wire;
```
swankjesse: nit: copyright

wire-runtime/src/test/java/com/squareup/wire/protos/foreign/ForeignEnum.java
```diff
@@ -2,22 +2,51 @@
...
+  public final int value;
 
   ForeignEnum(int value) {
     this.value = value;
   }
 
-  @Override
-  public int getValue() {
-    return value;
+  public static ForeignEnum fromValue(int value) {
```
swankjesse: Maybe these should be a member on the adapter? 

```
public class EnumAdapter<T extends Enum<T>> extends MessageAdapter<T> {
  public T fromValue(int value);
  public int toValue();
```

swankjesse: Definitely not worth addressing until follow up

swankjesse:
I think the reader/writer classes are an unambiguous improvement, especially the clever cursor feature.

I think the additional generated code is also a win, but only because it has a fixed cost per type: one extra anonymous inner class, plus 3 methods. I think the specific code you're generating is good, though I'd like to play with it a bit to see if there's further opportunities for simplification/compression.

I'm mostly indifferent to the `toString()` change. The generated code isn't very complex, but `toString()` isn't anything anyone is calling in their inner loops, so it's a wash. I think it's a win for consistency with the other APIs.

The great thing about this approach is that if we ever wanted to save some methods, this can easily be API-compatible with an alternate generated code that is very method-lean: the generated `ADAPTER` instance could be replaced with a reflective instance without any API change. Heck, you could even do it without bringing back the annotations:

```
  public static final MessageAdapter<PhoneNumber> ADAPTER = new MessageAdapterBuilder()
      .addPacked("name", 1, INT32)
      .addOptional("number", 2, STRING)
      .build();
```

... we don't need to do this yet. But my point is that this big change doesn't permanently commit us to a larger method count.

danrice-square:
@swankjesse I don't see the advantage of generating `MessageAdapter<PhoneNumber> ADAPTER` as you show above -- it is just a reification of information that is available rom the @ProtoField annotations. You have to do a reflective pass anyway in order to locate the fields by name.

I would be open to an AnnotationProcessor approach that generates the read/write/serializedSize methods in a way that is invisible to the developer. I am not in favor of this approach as it stands.

wire-runtime/src/test/java/com/squareup/wire/protos/redacted/NotRedacted.java
```diff
@@ -50,13 +90,22 @@ public boolean equals(Object other) {
   public int hashCode() {
     int result = hashCode;
     if (result == 0) {
-      result = a != null ? a.hashCode() : 0;
+      result = result * 37 + (a != null ? a.hashCode() : 0);
```
danrice-square: Prefer the old way, saves a useless multiplication.

wire-runtime/src/test/java/com/squareup/wire/protos/redacted/NotRedacted.java
```diff
@@ -50,13 +90,22 @@ public boolean equals(Object other) {
...
       result = result * 37 + (b != null ? b.hashCode() : 0);
       hashCode = result;
     }
     return result;
   }
 
+  @Override
+  public String toString() {
+    StringBuilder sb = new StringBuilder();
+    toStringValue(sb, "a", a);
```
danrice-square: If you are generating code anyway, you should generate type-specific code for each field. But I don't think we should generate toString() at all.

wire-runtime/src/test/java/com/squareup/wire/protos/redacted/NotRedacted.java
```diff
@@ -50,13 +90,22 @@ public boolean equals(Object other) {
...
+    toStringValue(sb, "b", b);
+    sb.insert(0, "NotRedacted{");
```
danrice-square: Why not just generate this at the top?

wire-runtime/src/test/java/com/squareup/wire/protos/redacted/NotRedacted.java
JakeWharton: > If you are generating code anyway, you should generate type-specific code for each field.

Why? The type information is thrown away with an implicit call to `String.valueOf` which is overloaded for every type already.

> But I don't think we should generate toString() at all.

Why do we generate equals and hashCode? Our app has never used them but we do use `toString()` all the time.

wire-runtime/src/test/java/com/squareup/wire/protos/redacted/NotRedacted.java
JakeWharton: `toStringValue` checks against 0 to determine whether to comma-separate. Without this it has no way of knowing whether to add the comma without passing in the length of this initial string.

wire-runtime/src/test/java/com/squareup/wire/protos/redacted/NotRedacted.java
JakeWharton: This might be the microist of mico optimizations and I doubt any modern CPU will actually issue a multiply operation on a zero operand but for the sake of the diff I'll revert.

wire-runtime/src/test/java/com/squareup/wire/protos/foreign/ForeignEnum.java
JakeWharton: `fromValue` needs to be `static` since it's used without an instance. I originally had `TypeAdapter.ENUM` for all enums which did this very thing.

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
JakeWharton: Then I need 30 methods, one for each scalar + message + enum and then repeated and packed variants of each. The visitor approach needs only 3, one for each arrangement of bytes.

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
JakeWharton: `toStringValue` checks against `.length() > 0` to determine if it prepends a comma. If this was appended first it would need to take an argument with the length of this initial string to compare against (which was the first implementation, actually, before switching to this).

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
JakeWharton: I determined that it cluttered the generated code too much. It's much easier to check in one place inside these methods.

JakeWharton:
> You have to do a reflective pass anyway in order to locate the fields by name.

There is zero reflection over fields.

wire-runtime/src/test/java/com/squareup/wire/protos/redacted/NotRedacted.java
danrice-square: If consistency is the high-order bit, I would rather get rid of equals and hashCode.

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
danrice-square: Why? This is like protoc code. I thought we explicitly decided not to generate code like this two years ago.

danrice-square:
> There is zero reflection over fields.

I'm referring to @swankjesse's suggestion above:

```
  public static final MessageAdapter<PhoneNumber> ADAPTER = new MessageAdapterBuilder()
      .addPacked("name", 1, INT32)
      .addOptional("number", 2, STRING)
      .build();
```

This would need to do a field lookup to make sense of `"name"`, etc.

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
JakeWharton: Yeah. They'll just only be used in the proto2 syntax.

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
JakeWharton: done?

JakeWharton:
Still 5 failing tests... but PTAL! Completely new approach!

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -445,65 +439,117 @@ private MethodSpec messageEquals(MessageType type) {
...
+      } else if (fieldType.isScalar()) {
+        result.addStatement("case $L: builder.$L = reader.value($L); break", tag, name, adapter);
       } else {
-        result.addStatement("result = $L != null ? $L.hashCode() : $L",
-            name, name, nullHashValue(field));
-        afterFirstAssignment = true;
+        TypeName fieldName = javaGenerator.typeName(fieldType);
+        if (isEnum) {
+          result.addStatement(
+              "case $1L: builder.$2L = enumOrUnknown($1L, reader, $3T.ADAPTER, builder); break",
```
swankjesse: not sure I track why `unknown` and enums go together, but I suspect I'll get to that

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -445,65 +439,117 @@ private MethodSpec messageEquals(MessageType type) {
...
+    }
+
+    result.endControlFlow(); // Switch
+    result.endControlFlow(); // While
+    result.addStatement("return builder.build()");
+
     return result.build();
   }
 
+  private CodeBlock typeAdapter(Type.Name fieldName) {
```
swankjesse: it says fieldName but I think it's fieldType

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -791,55 +826,38 @@ public TypeSpec extensionsType(ClassName javaTypeName, ProtoFile protoFile) {
...
-
-    if (field.isRepeated()) {
-      fieldType = JavaGenerator.listOf(fieldType);
-    }
-
-    return FieldSpec.builder(JavaGenerator.extensionOf(extendType, fieldType), field.name())
+  private FieldSpec extensionField(TypeName extendType, Field field) {
+    int tag = field.tag();
+    Type.Name fieldType = field.type();
+    TypeName fieldName = javaGenerator.typeName(fieldType);
```
swankjesse: This is just the fieldJavaType right?

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
JakeWharton: An enum value for which a constant is not present in the generated code gets put in the unknown field map and `null` returned (thus indicating to use the default). This ensures the value is retained if the message is re-serialized.

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
JakeWharton: Maybe we should rename `Type.Name` to `Type.Named` so the noun is unambiguously "type" rather than "name" (which is probably what I kept thinking during refactors).

wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -143,41 +83,57 @@ protected void setBuilder(Builder builder) {
...
+    if (o == this) return true;
+    if (o == null || o.getClass() != getClass()) return false;
+
+    EqualsVisitor thisVisitor = new EqualsVisitor();
+    visitFields(thisVisitor);
+
+    EqualsVisitor otherVisitor = new EqualsVisitor();
+    ((Message<?>) o).visitFields(otherVisitor);
+
+    return thisVisitor.isEqualTo(otherVisitor);
```
swankjesse: neat

wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -301,4 +261,244 @@ protected IllegalStateException missingRequiredFields(Object... args) {
...
+    int value = reader.value(TypeAdapter.UINT32);
+    M constant = adapter.fromValue(value);
+    if (constant != null) {
+      return constant;
+    }
+    builder.addVarint(tag, value);
+    return null;
+  }
+
+  protected static <M extends Enum<M> & WireEnum> List<M> repeatedEnumOrUnknown(int tag,
```
swankjesse: Does this method handle enums and unknowns both together because they're the same? Or just because the code to support them is the same? Maybe would be neater to have separate methods like `repeatedEnum()`, `repeatedUnknown()` and for those to delegate to this, which would be private?

wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -301,4 +261,244 @@ protected IllegalStateException missingRequiredFields(Object... args) {
...
+    }
+    int size = 0;
+    for (int i = 0, count = value.size(); i < count; i++) {
+      size += adapter.serializedSize(value.get(i));
+    }
+    return tagSize(tag) + varint32Size(size) + size;
+  }
+
+  protected interface Visitor {
+    <T> void value(int tag, String name, T value, TypeAdapter<T> adapter, boolean redacted);
```
swankjesse: For consideration later: packed and redacted could be flags?

wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -301,4 +261,244 @@ protected IllegalStateException missingRequiredFields(Object... args) {
...
+    private final List<Object> instances = new ArrayList<Object>();
+
+    @Override
+    public <T> void value(int tag, String name, T value, TypeAdapter<T> adapter, boolean redacted) {
+      instances.add(value);
+    }
+
+    @Override public <T> void repeated(int tag, String name, List<T> list, TypeAdapter<T> adapter,
+        boolean redacted) {
+      instances.add(list);
```
swankjesse: Bug!

foo: `{ a: [1, 2], b: [3, 4] }`
bar: `{ a: [1], b: [2, 3, 4] }`

foo.equals(bar)

swankjesse: Nope, nevermind. You're calling `add()`, not `addAll()`. All good.

wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -301,4 +261,244 @@ protected IllegalStateException missingRequiredFields(Object... args) {
...
+
+    @Override public void unknowns(Message<?> message) {
+    }
+
+    @Override public void extensions(ExtendableMessage<?> message) {
+      instances.add(message.extensionMap);
+    }
+
+    boolean isEqualTo(EqualsVisitor other) {
+      return instances.equals(other.instances);
```
swankjesse: Probably quite inefficient, since each nested message's `equals()` call is going to allocate two lists. Possible to implement this recursively with a single list?

wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
```diff
@@ -53,41 +54,113 @@
       "Protocol message end-group tag did not match expected tag.";
   private static final String ENCOUNTERED_A_MALFORMED_VARINT =
       "WireInput encountered a malformed varint.";
+  private static final int RECURSION_LIMIT = 64;
+
+  private final BufferedSource source;
+  private final ExtensionRegistry extensionRegistry;
+
+  /** The current position in the source, starting at 0 and increasing monotonically. */
```
swankjesse: Is this the number of bytes read from the source?

wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
```diff
@@ -53,41 +54,113 @@
...
+    return adapter.read(this);
+  }
+
+  public <E> List<E> repeated(List<E> existing, TypeAdapter<E> adapter) throws IOException {
+    return Message.repeated(existing, adapter.read(this));
+  }
+
+  public <E> List<E> packed(List<E> existing, TypeAdapter<E> adapter) throws IOException {
+    long cursor = beginLengthDelimited();
+    while (hasNext()) {
```
swankjesse: Neat.

wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
```diff
@@ -53,41 +54,113 @@
...
+
+  public <E> E value(TypeAdapter<E> adapter) throws IOException {
+    return adapter.read(this);
+  }
+
+  public <E> List<E> repeated(List<E> existing, TypeAdapter<E> adapter) throws IOException {
+    return Message.repeated(existing, adapter.read(this));
+  }
+
+  public <E> List<E> packed(List<E> existing, TypeAdapter<E> adapter) throws IOException {
```
swankjesse: ... and this method probably wants to just allocate a list and return it, and let the caller concatenate lists in the very rare double-packed case.

wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
```diff
@@ -53,41 +54,113 @@
...
+  ProtoReader(BufferedSource source, ExtensionRegistry extensionRegistry) {
+    this.source = source;
+    this.extensionRegistry = extensionRegistry;
+  }
+
+  public <E> E value(TypeAdapter<E> adapter) throws IOException {
+    return adapter.read(this);
+  }
+
+  public <E> List<E> repeated(List<E> existing, TypeAdapter<E> adapter) throws IOException {
```
swankjesse: Not sure if this method belongs here. Is it used frequently? I think it should be the caller's job to do the list voodoo

wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
```diff
@@ -53,41 +54,113 @@
...
+      throw new IOException("Wire recursion limit exceeded");
+    }
+    int length = readVarint32();
+    if (length < 0) {
+      throw new IOException(ENCOUNTERED_A_NEGATIVE_SIZE);
+    }
+    long newLimit = length + pos;
+    long oldLimit = currentLimit;
+    if (length > oldLimit) {
+      throw new EOFException(INPUT_ENDED_UNEXPECTEDLY);
```
swankjesse: Love this.

wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
```diff
@@ -53,41 +54,113 @@
...
+    }
+    endLengthDelimited(cursor);
+    return existing;
+  }
+
+  public <T extends ExtendableMessage<T>> Extension<T, ?, ?> getExtension(Class<T> type, int tag) {
+    return extensionRegistry == null ? null : extensionRegistry.getExtension(type, tag);
+  }
+
+  public long beginLengthDelimited() throws IOException {
```
swankjesse: _“Returns an opaque token...”_

wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
```diff
@@ -53,41 +54,113 @@
...
+    return oldLimit;
+  }
+
+  public boolean hasNext() throws IOException {
+    return pos < currentLimit && !source.exhausted();
+  }
+
+  public void endLengthDelimited(long cursor) throws IOException {
+    if (pos != currentLimit) {
+      throw new IOException("Expected to end at " + currentLimit + " but was " + pos);
```
swankjesse: Love this.

wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
```diff
@@ -53,41 +54,113 @@
...
+    }
+    currentLimit = cursor;
+    --recursionDepth;
```
swankjesse: Also check this? Bad code could call `endLengthDelimited()` too many times!

wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
swankjesse: ... maybe this should be called `token` so that it's opaque to the caller?

wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
```diff
@@ -97,18 +170,23 @@ public String readString() throws IOException {
...
     int count = readVarint32();
+    source.require(count); // Throws EOFException on insufficient bytes.
     pos += count;
-    source.require(count); // Throws EOFException if insufficient bytes are available.
     return source.readByteString(count);
   }
 
+  int readRawByte() throws IOException {
+    pos++;
+    return source.readByte() & 0xFF;
```
swankjesse: nit: lowercase hex literals

wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
```diff
@@ -165,131 +243,26 @@ public long readVarint64() throws IOException {
...
-  public void skipGroup() throws IOException {
-    while (true) {
-      int tag = readTag();
-      if (tag == 0 || skipField(tag)) {
-        return;
-      }
-    }
+  void skipGroup() throws IOException {
+    //noinspection StatementWithEmptyBody
+    while (skipField(nextTag()));
```
swankjesse: neat

swankjesse: confirm that the tag is a group?

wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
```diff
@@ -0,0 +1,369 @@
...
+
+  static final int FIXED_BOOL_SIZE = 1;
+  static final int FIXED_32_SIZE = 4;
+  static final int FIXED_64_SIZE = 8;
+
+  static int utf8Length(String s) {
+    int count = 0;
+    for (int i = 0, length = s.length(); i < length; i++) {
+      char ch = s.charAt(i);
+      if (ch <= 0x7F) {
```
swankjesse: nit: lowercase

wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
```diff
@@ -0,0 +1,369 @@
...
+        count++;
+      } else if (ch <= 0x7FF) {
+        count += 2;
+      } else if (Character.isHighSurrogate(ch)) {
+        count += 4;
+        ++i;
```
swankjesse: this scares me

swankjesse: (though unrelated)

wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
```diff
@@ -0,0 +1,369 @@
...
+   * unsigned support.
+   * @return A signed 64-bit integer.
+   */
+  static long decodeZigZag64(long n) {
+    return (n >>> 1) ^ -(n & 1);
+  }
+
+  /** Compute the number of bytes that would be needed to encode a varint. */
+  static int varint64Size(long value) {
+    if ((value & (0xffffffffffffffffL <<  7)) == 0) return 1;
```
swankjesse: off topic: I wonder if these are computed statically

wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
```diff
@@ -0,0 +1,369 @@
...
+
+    @Override public void write(Boolean value, ProtoWriter writer) throws IOException {
+      writer.writeRawByte(value ? 1 : 0);
+    }
+
+    @Override public Boolean read(ProtoReader reader) throws IOException {
+      int value = reader.readRawByte();
+      if (value == 0) return Boolean.FALSE;
+      if (value == 1) return Boolean.TRUE;
+      throw new IllegalStateException(String.format("Invalid boolean value 0x%02x", value));
```
swankjesse: nice error. Should this be an IOException?

swankjesse: (probably follow-up)

wire-runtime/src/main/java/com/squareup/wire/WriteIOException.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire;
+
+import java.io.IOException;
+
+final class WriteIOException extends RuntimeException {
+  public IOException cause;
```
swankjesse: final

wire-runtime/src/test/java/com/squareup/wire/protos/custom_options/FooBar.java.noOptions
```diff
@@ -117,7 +120,42 @@ public final class FooBar extends ExtendableMessage<FooBar> {
     return result;
   }
 
+<<<<<<< HEAD
```
swankjesse: uh oh

swankjesse:
I like where it's going.

wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
danrice-square: There are badly-formed strings that will cause problems here. I'm not sure what a proto serializer is supposed to do about such strings, though.

pom.xml
```diff
@@ -46,7 +46,7 @@
   <modules>
     <module>wire-compiler</module>
     <module>wire-runtime</module>
-    <module>wire-gson-support</module>
+    <!--<module>wire-gson-support</module>-->
```
danrice-square: Need to make sure this get re-enabled.

wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -24,90 +24,30 @@
...
 
   /** If not {@code -1} then the serialized size of this message. */
-  transient int cachedSerializedSize = -1;
+  protected transient int serializedSize = -1;
 
   /** If non-zero, the hash code of this message. Accessed by generated code. */
   protected transient int hashCode = 0;
 
-  protected Message() {
+  private transient final String name;
```
danrice-square: Add a comment indicating the name is for use by `toString()`.

wire-runtime/src/test/java/com/google/protobuf/DescriptorProto.java
```diff
@@ -93,35 +66,35 @@ private DescriptorProto(Builder builder) {
...
+  protected void visitFields(Message.Visitor visitor) {
+    visitor.value(1, "name", name, TypeAdapter.STRING, false);
+    visitor.value(8, "doc", doc, TypeAdapter.STRING, false);
+    visitor.repeated(2, "field", field, FieldDescriptorProto.ADAPTER, false);
+    visitor.repeated(6, "extension", extension, FieldDescriptorProto.ADAPTER, false);
+    visitor.repeated(3, "nested_type", nested_type, DescriptorProto.ADAPTER, false);
+    visitor.repeated(4, "enum_type", enum_type, EnumDescriptorProto.ADAPTER, false);
+    visitor.repeated(5, "extension_range", extension_range, ExtensionRange.ADAPTER, false);
+    visitor.value(7, "options", options, MessageOptions.ADAPTER, false);
+    visitor.unknowns(this);
```
danrice-square: What if the caller handled the call to `unknowns`?

wire-runtime/src/test/java/com/google/protobuf/DescriptorProto.java
```diff
@@ -93,35 +66,35 @@ private DescriptorProto(Builder builder) {
...
-    return equals(name, o.name)
-        && equals(doc, o.doc)
-        && equals(field, o.field)
-        && equals(extension, o.extension)
-        && equals(nested_type, o.nested_type)
-        && equals(enum_type, o.enum_type)
-        && equals(extension_range, o.extension_range)
-        && equals(options, o.options);
+  protected void visitFields(Message.Visitor visitor) {
+    visitor.value(1, "name", name, TypeAdapter.STRING, false);
```
danrice-square: Sort by tag number.

wire-runtime/src/main/java/com/squareup/wire/Message.java
danrice-square: While this is truly clever, it's pretty inefficient since there is no possibility of early exit. On the other hand, I would imaging that most calls to `equals` happen only when there is a hash code collision, which is extremely rare. We could always go back to codegen for `equals` if this is a problem.

danrice-square: An approach that is somewhat more efficient is to have an `EqualsVisitor` as above for the first message, then an `IsEqualToVisitor` that is initialized with the array from the first visitor. It can set a flag to `false` as soon as it finds a difference in a call to `value`, etc., and avoid further work (beyond the method calls) when the flag is already false.

wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
```diff
@@ -445,65 +439,117 @@ private MethodSpec messageEquals(MessageType type) {
   // Example:
   //
   // @Override
-  // public int hashCode() {
-  //   if (hashCode == 0) {
-  //     int result = super.extensionsHashCode();
-  //     result = result * 37 + (f != null ? f.hashCode() : 0);
-  //     hashCode = result;
+  // public Person read(ProtoReader reader) throws IOException {
```
danrice-square: Code example needs to be updated

wire-gson-support/src/main/java/com/squareup/wire/MessageTypeAdapter.java
```diff
@@ -22,10 +22,6 @@
 import com.google.gson.stream.JsonReader;
```
danrice-square: This class seems like it is in a weird state, is it supposed to go away?

wire-runtime/src/test/java/com/squareup/wire/RedactedTest.java
```diff
@@ -15,73 +15,16 @@
...
 public class RedactedTest {
-
   @Test
   public void testRedacted() throws IOException {
     assertThat(new Redacted.Builder().a("a").b("b").c("c").build().toString())
         .isEqualTo("Redacted{a=██, b=b, c=c}");
   }
-
-  @Test
-  public void testRedactor() {
```
danrice-square: So are we losing the ability to make a redacted copy of a message (not just to redact the `toString()` output)? We should be able to do redaction using the `visitor` interface.

wire-runtime/src/test/java/com/squareup/wire/protos/simple/Ext_simple_message.java
```diff
@@ -3,46 +3,23 @@
...
-      .setName("squareup.protos.simple.nested_enum_ext")
-      .setTag(129)
-      .buildOptional();
-
-  public static final Extension<ForeignMessage, Integer> j = Extension
-      .int32Extending(ForeignMessage.class)
-      .setName("squareup.protos.simple.j")
-      .setTag(100)
-      .buildOptional();
+  public static final Extension<ExternalMessage, Integer, List<Integer>> fooext = Extension.repeated(ExternalMessage.class, 125, TypeAdapter.INT32);
```
danrice-square: We are losing the extension name, any reason why?

wire-runtime/src/main/java/com/squareup/wire/Extension.java
```diff
@@ -15,314 +15,114 @@
...
-    return tag;
-  }
-
-  public Datatype getDatatype() {
-    return datatype;
-  }
-
-  public Label getLabel() {
-    return label;
+  @Override public final int compareTo(Extension<M, T, R> o) {
```
danrice-square: Can we get equals(), hashCode(), and toString() back?

wire-runtime/src/test/java/com/squareup/wire/WireTest.java
```diff
@@ -208,20 +208,6 @@ public void testExtensionEnum() throws Exception {
   }
 
   @Test
-  public void extensionToString() {
```
danrice-square: We should restore this.

wire-runtime/src/main/java/com/squareup/wire/ExtendableMessage.java
```diff
@@ -42,87 +40,61 @@ protected void setBuilder(ExtendableBuilder<T, ?> builder) {
     }
   }
 
-  /**
-   * Returns an immutable list of the extensions on this message in tag order.
-   */
-  public List<Extension<T, ?>> getExtensions() {
```
danrice-square: Restore this.

wire-runtime/src/test/java/com/squareup/wire/protobuf/TestAllTypes.java
```diff
@@ -350,12 +339,6 @@ public void testReadSource() throws IOException {
     assertThat(allTypes.getExtension(Ext_all_types.ext_opt_bool)).isEqualTo(Boolean.TRUE);
     assertThat(allTypes.getExtension(Ext_all_types.ext_rep_bool)).isEqualTo(list(true));
     assertThat(allTypes.getExtension(Ext_all_types.ext_pack_bool)).isEqualTo(list(true));
-
-    List<Extension<AllTypes, ?>> extensions = parsed.getExtensions();
```
danrice-square: Restore this.

danrice-square:
I'm liking the new approach much more. What I'd like to see is for the new code to be outside the public API as much as possible so we can potentially generate it using annotation processing in the future.

wire-runtime/src/test/java/com/squareup/wire/protos/simple/Ext_simple_message.java
JakeWharton: It's not used

wire-runtime/src/test/java/com/squareup/wire/RedactedTest.java
JakeWharton: I dropped it as an unused feature. What was it for? I found no occurrences of its use.

wire-runtime/src/test/java/com/google/protobuf/DescriptorProto.java
JakeWharton: Not sure what you're asking here. Who is the caller? And how would they handle unknowns?

wire-runtime/src/test/java/com/squareup/wire/protos/custom_options/FooBar.java.noOptions
JakeWharton: This crap isn't generated by the test gen script. UGH.

wire-runtime/src/test/java/com/squareup/wire/RedactedTest.java
danrice-square: It's not used inside of Wire but it is a public API that we ship.

wire-runtime/src/test/java/com/google/protobuf/DescriptorProto.java
danrice-square: Ignore my comment.

wire-runtime/src/test/java/com/squareup/wire/RedactedTest.java
JakeWharton: We don't have to care about public API since this is for a v2. I see that Register uses it on code search. I'll see if we can add a `toRedacted()` or something using the visitor.

JakeWharton: Although not sure how to reference the builder fields/methods without reflection. Ugh...

danrice-square: Maybe it could go through a serialization/deserialization path.

I'm not worried about the specific API (since as you say this is v2) but I want to maintain the capability.

wire-runtime/src/test/java/com/squareup/wire/protos/simple/Ext_simple_message.java
danrice-square: It was used for toString.

wire-runtime/src/test/java/com/squareup/wire/RedactedTest.java
JakeWharton: Yep. Not sure why it didn't turn up in codesearch before, but I see it in use now. Might have mistaken it for protoc's version.

wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
JakeWharton: Yes

JakeWharton:
No need to keep this open. The work is being decomposed to multiple PRs.

PR #281, opened 2015-07-23
Optional and required for proto3 syntax.

swankjesse:


wire-runtime/src/test/java/com/squareup/wire/protos/oneof/OneOfMessage.java
```diff
@@ -21,8 +21,7 @@
    */
   @ProtoField(
       tag = 1,
-      type = Message.Datatype.INT32,
-      label = Message.Label.ONE_OF
```
JakeWharton: Can you delete this constant?

JakeWharton:
LGTM

wire-runtime/src/test/java/com/squareup/wire/protos/oneof/OneOfMessage.java
swankjesse: I'm going to wait for your `@ProtoField` runtime changes to land. Right now it's in an awkward intermediate state where we have `OPTIONAL` and `REQUIRED` but they don't really apply to proto3.

PR #290, opened 2015-08-03
Move name-to-tag lookup map into Gson.

JakeWharton:
This is the only place it was used so there was no point in wasting the Map allocation on every MessageAdapter for the proto-only case.


JakeWharton:
This is part 6 of 52 in the Jake Moves Things Around™ series of pull requests.

swankjesse:
LGTM

PR #291, opened 2015-08-03
Move field binding type to top-level.

JakeWharton:
It is a very logical place to put field-specific behavior and future changes will leverage this.


wire-runtime/src/main/java/com/squareup/wire/ReflectiveFieldBinding.java
```diff
@@ -0,0 +1,122 @@
+package com.squareup.wire;
```
swankjesse: copyright

PR #292, opened 2015-08-04
Re-arrange reading logic to be more clear, simpler.

JakeWharton:


JakeWharton:
Appending `?w=1` to the URL!

danrice-square:
LGTM

PR #293, opened 2015-08-04
Throw ProtocolException for protocol problems.

JakeWharton:


swankjesse:
Or perhaps, WireDataException, for symmetry with Moshi?

JakeWharton:
Yep. Future change.

PR #294, opened 2015-08-04
Push read termination and nesting logic into the reader.

JakeWharton:
This API will be exposed to callers who are parsing dynamically. A representative use is in our own reflection-based reading.


wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
```diff
@@ -54,41 +54,128 @@
...
+    return tagAndType >> WireType.TAG_TYPE_BITS;
+  }
+
+  /** The type of the field value. {@link #readTag()} must be called before this method. */
+  public WireType peekType() throws IOException {
+    return nextType;
+  }
+
+  /** Skips a section of the input delimited by START_GROUP/END_GROUP type markers. */
+  int skipGroup() throws IOException {
```
JakeWharton: Generic `skipValue()` will come later.

danrice-square:
LGTM

PR #297, opened 2015-08-04
Expose ProtoReader/Writer and add type-safe TypeAdapter.

JakeWharton:
This allows partial dynamic reading and writing of single-value types (including messages). Reading and writing enums and repeated/packed values are currently not supported, but will be soon.

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
@@ -42,7 +42,7 @@ import okio.ByteString;
 /**
  * Reads and decodes protocol message fields.
  */
-final class ProtoReader {
+public final class ProtoReader {
 
   private static final String ENCOUNTERED_A_NEGATIVE_SIZE =
       "Encountered a negative size";
@@ -94,7 +94,7 @@ final class ProtoReader {
     if (length < 0) {
       throw new ProtocolException(ENCOUNTERED_A_NEGATIVE_SIZE);
     }
-    long newLimit = length + pos;
+    long newLimit = pos + length;
     long oldLimit = limit;
     if (newLimit > oldLimit) {
       throw new EOFException(INPUT_ENDED_UNEXPECTEDLY);
@@ -132,6 +132,10 @@ final class ProtoReader {
     }
     nextType = WireType.valueOf(tagAndType);
     return tagAndType >> WireType.TAG_TYPE_BITS;
+  }
+
+  public <T> T value(TypeAdapter<T> adapter) throws IOException {
+    return adapter.read(this);
   }
 
   /** The type of the field value. {@link #readTag()} must be called before this method. */
@@ -174,11 +178,10 @@ final class ProtoReader {
     source.skip(count);
   }
 
-  /** Reads a {@code string} field value from the stream. */
-  String readString() throws IOException {
-    int count = readVarint32();
-    pos += count;
-    return source.readUtf8(count);
+  int readByte() throws IOException {
+    source.require(1); // Throws EOFException if insufficient bytes are available.
+    pos++;
+    return source.readByte() & 0xff;
   }
 
   /**
@@ -187,9 +190,17 @@ final class ProtoReader {
    */
   ByteString readBytes() throws IOException {
     int count = readVarint32();
-    pos += count;
     source.require(count); // Throws EOFException if insufficient bytes are available.
+    pos += count;
     return source.readByteString(count);
+  }
+
+  /** Reads a {@code string} field value from the stream. */
+  String readString() throws IOException {
+    int count = readVarint32();
+    source.require(count); // Throws EOFException if insufficient bytes are available.
+    pos += count;
+    return source.readUtf8(count);
   }
 
   /**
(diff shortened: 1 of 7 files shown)


wire-runtime/src/main/java/com/squareup/wire/ReflectiveMessageAdapter.java
```diff
@@ -280,44 +280,25 @@ private int getSerializedSize(int tag, Object value, Datatype datatype) {
...
-      case SINT64: return ProtoWriter.varint64Size(ProtoWriter.zigZag64((Long) value));
-      case BOOL: return 1;
-      case ENUM: return getEnumSize((ProtoEnum) value);
-      case STRING:
-        int utf8Length = utf8Length((String) value);
-        return ProtoWriter.varint32Size(utf8Length) + utf8Length;
-      case BYTES:
-        int length = ((ByteString) value).size();
-        return ProtoWriter.varint32Size(length) + length;
+      case INT32: return TypeAdapter.INT32.serializedSize((Integer) value);
```
JakeWharton: All this stuff is just temporary until it's wired through more and these switches disappear completely!

wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
```diff
@@ -0,0 +1,319 @@
...
+
+    @Override public void write(Boolean value, ProtoWriter writer) throws IOException {
+      writer.writeByte(value ? 1 : 0);
+    }
+
+    @Override public Boolean read(ProtoReader reader) throws IOException {
+      int value = reader.readByte();
+      if (value == 0) return Boolean.FALSE;
+      if (value == 1) return Boolean.TRUE;
+      throw new IllegalStateException(String.format("Invalid boolean value 0x%02x", value));
```
JakeWharton: Should be IOException

wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
```diff
@@ -134,6 +134,10 @@ public int readTag() throws IOException {
     return tagAndType >> WireType.TAG_TYPE_BITS;
   }
 
+  public <T> T value(TypeAdapter<T> adapter) throws IOException {
```
swankjesse: this is a weird API

wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
```diff
@@ -174,11 +178,10 @@ private void skip(long count) throws IOException {
...
 
-  /** Reads a {@code string} field value from the stream. */
-  String readString() throws IOException {
-    int count = readVarint32();
-    pos += count;
-    return source.readUtf8(count);
+  int readByte() throws IOException {
+    source.require(1); // Throws EOFException if insufficient bytes are available.
+    pos++;
+    return source.readByte() & 0xff;
```
swankjesse: Tempting to flip this.
```
int result = source.readByte() & 0xff;
pos;
return result;
```

JakeWharton: `require` guarantees the byte will be there. plus we (currently) make no claims about the state after an exception.

wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
JakeWharton: It makes a bit more sense when you have `repeated(TypeAdapter)` next to it

wire-runtime/src/main/java/com/squareup/wire/ProtoWriter.java
```diff
@@ -107,45 +94,24 @@ static int varint64Size(long value) {
...
   ProtoWriter(BufferedSink sink) {
     this.sink = sink;
   }
 
-  void writeRawByte(int value) throws IOException {
+  public <T> void write(TypeAdapter<T> adapter, T value) throws IOException {
+    if (adapter.type == WireType.LENGTH_DELIMITED) {
+      writeVarint32(adapter.serializedSize(value));
+    }
+    adapter.write(value, this);
```
swankjesse: should this happen inside the adapter?

wire-runtime/src/main/java/com/squareup/wire/ProtoWriter.java
```diff
@@ -107,45 +94,24 @@ static int varint64Size(long value) {
...
   private final BufferedSink sink;
 
   ProtoWriter(BufferedSink sink) {
     this.sink = sink;
   }
 
-  void writeRawByte(int value) throws IOException {
+  public <T> void write(TypeAdapter<T> adapter, T value) throws IOException {
+    if (adapter.type == WireType.LENGTH_DELIMITED) {
+      writeVarint32(adapter.serializedSize(value));
```
swankjesse: should this happen inside the adapter?

JakeWharton: There's a problem with that which is the root message, a length-delimited object, should NOT prefix itself with its length. This could be solved by adding state to `ProtoWriter` to ignore the first call to `writeVarint32` or something... but that didn't seem great either.

wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
```diff
@@ -0,0 +1,319 @@
...
+    }
+  };
+  public static final TypeAdapter<String> STRING = new TypeAdapter<String>(WireType.LENGTH_DELIMITED) {
+    @Override public int serializedSize(String value) {
+      return utf8Length(value);
+    }
+
+    @Override public void write(String value, ProtoWriter writer) throws IOException {
+      ByteString bytes = ByteString.encodeUtf8(value);
+      writer.writeBytes(bytes);
```
swankjesse: Later we'll want writer to support this directly!

wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
```diff
@@ -0,0 +1,319 @@
...
+    @Override public void write(ByteString value, ProtoWriter writer) throws IOException {
+      writer.writeBytes(value);
+    }
+
+    @Override public ByteString read(ProtoReader reader) throws IOException {
+      return reader.readBytes();
+    }
+  };
+
+  static int utf8Length(String s) {
```
swankjesse: probably want this in okio?!

wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
swankjesse: Cool. Later we should make those claims! (Currently it calls `require()` internally anyway).

wire-runtime/src/main/java/com/squareup/wire/ProtoWriter.java
swankjesse: That's broken.

swankjesse:
LGTM

wire-runtime/src/main/java/com/squareup/wire/ProtoWriter.java
swankjesse: ...the spec, that is.

wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
JakeWharton: Not for `Buffer`, which throw IAE (to avoid the throws IO/EOF checked declaration)

swankjesse: Yep. SIGH!

wire-runtime/src/main/java/com/squareup/wire/ProtoWriter.java
JakeWharton: The code becomes a lot nicer when this happens in the adapter. So I'm going to look for an approach that's not hacky for this.

JakeWharton:
Landing this to keep forward progress. I expect things to change as it gets more integrated and used.

PR #298, opened 2015-08-05
Use type adapters in unknown map for a big simplification.

JakeWharton:


JakeWharton:
This falls into: "It's going to get a little worse before it can get better."

JakeWharton:
I have a plan for the tag / size nonsense that's proliferating.. and I'll resist posting another batman image. It's all part of the plan!

swankjesse:
Yeah, well, the good thing about the mob is they keep giving you second chances.

PR #300, opened 2015-08-06
Added support for multiple --proto_path arguments

ThalmicRevane:
protoc supports multiple --proto_path arguments to specify extra paths to search for .proto files in. This commit adds similar functionality to wire-compiler.

New tests haven't been added as of yet. Wanted to make sure this commit was functionally ok with the project before spending time on them.

I have plans to update wire-gradle-plugin as well.

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/CommandLineOptions.java
+++ b/wire-compiler/src/main/java/com/squareup/wire/CommandLineOptions.java
@@ -38,7 +38,8 @@ final class CommandLineOptions {
   public static final String QUIET_FLAG = "--quiet";
   public static final String DRY_RUN_FLAG = "--dry_run";
 
-  final String protoPath;
+  final List<String> protoPaths;
+
   final File javaOut;
   final List<String> sourceFileNames;
   final List<String> roots;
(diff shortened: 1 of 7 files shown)


wire-compiler/src/main/java/com/squareup/wire/CommandLineOptions.java
```diff
@@ -202,10 +203,10 @@ private ServiceFactory loadServiceFactory(String className) throws WireException
     return Arrays.asList(arg.substring(flagLength).split(","));
   }
 
-  public String protoPath() {
-    String result = protoPath;
+  public List<String> protoPaths() {
+    List<String> result = protoPaths;
     if (result == null) {
```
swankjesse: I'm guessing this is going to fail, cause the value will be an empty list instead of null

wire-schema/src/main/java/com/squareup/wire/schema/Loader.java
```diff
@@ -34,19 +36,25 @@ public Loader(IO io) {
...
-        return Location.get(base, path);
+      @Override public Location locate(String file) throws IOException {
+        for (String path : paths) {
+          File f = new File(path, file);
+          if (f.exists()) {
+            return Location.get(path, file);
+          }
+        }
+
+        throw new IOException();
```
swankjesse: Detailed message please. What wasn't found, and where did we look!

swankjesse:
Awesome. Add a test or two?

ThalmicRevane:
Will address comments and provide tests in an amendment. Thanks for the quick review.

swankjesse:
Added copyrights, fixed formatting & merged manually. Thanks!

PR #306, opened 2015-08-08
Field bindings cache their adapters.

JakeWharton:
This front-loads resolution of adapters so that read, write, and size calculation can all be done quickly. This also cleans up the duplicated read/write/size code for determining the type.

Due to the lack of root messages omitting a length-prefix, there's a minor hack in this PR which reads the length prefix manually for nested messages (since their type adapters omit it). There are a few options for eliminating this in the future that will be explored. This also isn't the end of the refactoring to make type adapters completely first-class citizens but it gets them much farther along.


swankjesse:
I love this.

PR #307, opened 2015-08-08
Split MessageAdapter from being a TypeAdapter.

JakeWharton:
This formalizes (and enforces in the type system) the semantic difference between root messages which do not length-prefix themselves vs. nested messages which do. The message adapters for nested messages are wrapped in a type adapter which adds the length-prefixing behavior to reading and writing.

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
@@ -28,26 +28,50 @@ public abstract class TypeAdapter<E> {
 
   final WireType type;
 
-  protected TypeAdapter(WireType type) {
+  TypeAdapter(WireType type) {
     this.type = type;
   }
 
-  /** The serialized size of non-null {@code value}. */
-  public abstract int serializedSize(E value);
+  /**
+   * The size of the non-null data {@code value}. This does not include the size required for
+   * a length-delimited prefix (should the type require one).
+   */
+  abstract int dataSize(E value);
+
+  /**
+   * The size of {@code tag} and non-null {@code value} in the wire format. This size includes the
+   * tag, type, length-delimited prefix (should the type require one), and value.
+   */
+  public final int serializedSize(int tag, E value) {
+    int size = dataSize(value);
+    if (type == WireType.LENGTH_DELIMITED) {
+      size += varint32Size(size);
+    }
+    return size + ProtoWriter.tagSize(tag);
+  }
 
   /** Write non-null {@code value} to {@code writer}. */
-  public abstract void write(E value, ProtoWriter writer) throws IOException;
+  abstract void writeData(E value, ProtoWriter writer) throws IOException;
+
+  /** Write {@code tag} and non-null {@code value} to {@code writer}. */
+  public final void write(int tag, E value, ProtoWriter writer) throws IOException {
+    writer.writeTag(tag, type);
+    if (type == WireType.LENGTH_DELIMITED) {
+      writer.writeVarint32(dataSize(value));
+    }
+    writeData(value, writer);
+  }
 
   /** Read a non-null value from {@code reader}. */
   public abstract E read(ProtoReader reader) throws IOException;
 
 
   public static final TypeAdapter<Boolean> BOOL = new TypeAdapter<Boolean>(WireType.VARINT) {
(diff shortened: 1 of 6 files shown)


wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
```diff
@@ -208,6 +232,25 @@ protected TypeAdapter(WireType type) {
...
+      }
+
+      @Override public M read(ProtoReader reader) throws IOException {
+        long token = reader.beginLengthDelimited();
+        M value = adapter.read(reader);
+        reader.endLengthDelimited(token);
+        return value;
+      }
+    };
+  }
```
JakeWharton: Coming soon:
```java
  public static <T> TypeAdapter<List<T>> createPacked(final TypeAdapter<T> adapter) {
    if (adapter.type == WireType.LENGTH_DELIMITED) {
      throw new IllegalArgumentException("Unable to pack a length-delimited type.");
    }
    return new TypeAdapter<List<T>>(WireType.LENGTH_DELIMITED) {
      @Override int dataSize(List<T> value) {
        int size = 0;
        for (int i = 0, count = value.size(); i < count; i++) {
          size += adapter.dataSize(value.get(i));
        }
        return size;
      }

      @Override void writeData(List<T> value, ProtoWriter writer) throws IOException {
        for (int i = 0, count = value.size(); i < count; i++) {
          adapter.writeData(value.get(i), writer);
        }
      }

      @Override public List<T> read(ProtoReader reader) throws IOException {
        List<T> items = new ArrayList<T>();
        long token = reader.beginLengthDelimited();
        while (reader.hasNext()) {
          items.add(adapter.read(reader));
        }
        reader.endLengthDelimited(token);
        return items;
      }
    };
  }
```

wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
```diff
@@ -28,26 +28,50 @@
...
-  protected TypeAdapter(WireType type) {
+  TypeAdapter(WireType type) {
     this.type = type;
   }
 
-  /** The serialized size of non-null {@code value}. */
-  public abstract int serializedSize(E value);
+  /**
+   * The size of the non-null data {@code value}. This does not include the size required for
+   * a length-delimited prefix (should the type require one).
```
swankjesse: maybe instead of `should the type require one` it's `should the type require one in its current context`. Because the message type requires one, except when it doesn't!

wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
```diff
@@ -28,26 +28,50 @@
...
+    }
+    return size + ProtoWriter.tagSize(tag);
+  }
 
   /** Write non-null {@code value} to {@code writer}. */
-  public abstract void write(E value, ProtoWriter writer) throws IOException;
+  abstract void writeData(E value, ProtoWriter writer) throws IOException;
+
+  /** Write {@code tag} and non-null {@code value} to {@code writer}. */
+  public final void write(int tag, E value, ProtoWriter writer) throws IOException {
```
swankjesse: Nice.

wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
```diff
@@ -208,6 +232,25 @@ protected TypeAdapter(WireType type) {
         }
       };
 
+  public static <M extends Message> TypeAdapter<M> forMessage(final MessageAdapter<M> adapter) {
```
swankjesse: This is wonderful.

wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
swankjesse: Asymmetry between read and write isn't perfect here. Though I'll take it!

wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
JakeWharton: True, but the context never matters for type adapters. It always wants the prefix to be accounted for (or in this case, ignored). They remain blissfully unaware of the root message nonsense.

PR #308, opened 2015-08-10
Push more read/write/size functionality into the type adapters.

JakeWharton:
Unfortunately, due to the behavior of erroneous enum retaining in the unknown field map, reflective reading still requires manual parsing using the raw type adapter for a single value.

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
@@ -16,6 +16,9 @@
 package com.squareup.wire;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import okio.ByteString;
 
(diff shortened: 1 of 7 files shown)


wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
```diff
@@ -251,6 +254,73 @@ public final void write(int tag, E value, ProtoWriter writer) throws IOException
...
+      }
+
+      @Override void write(int tag, List<T> value, ProtoWriter writer) throws IOException {
+        for (int i = 0, count = value.size(); i < count; i++) {
+          adapter.write(tag, value.get(i), writer);
+        }
+      }
+
+      @Override List<T> read(ProtoReader reader) throws IOException {
+        throw new UnsupportedOperationException(); // TODO
```
JakeWharton: Future change. Needs a `peekTag()` on `ProtoReader` to know when to stop.

PR #310, opened 2015-08-12
Detect and forbid conflicts on extension tags.

swankjesse:
Ths breaks one test case which was trying to keep this working. There's no reason
a well-formed program would do this; it would crash attempting to parse the data
if it were ever encoded.


wire-schema/src/test/java/com/squareup/wire/schema/SchemaTest.java
```diff
@@ -476,7 +476,7 @@
       assertThat(expected).hasMessage("multiple fields share tag 1:\n"
           + "  1. name1 (message.proto at 4:3)\n"
           + "  2. name2 (message.proto at 5:3)\n"
-          + "  for extend Message (message.proto at 3:1)");
+          + "  for message Message (message.proto at 1:1)");
```
JakeWharton: I initially was hesitant about this message change because it seems worse, but rather the fields point directly to their declaration which is what you'd want since they would reside in different locations.

PR #315, opened 2015-08-13
Fold runtime data redacting into the runtime adapter.

JakeWharton:


wire-runtime/src/main/java/com/squareup/wire/RuntimeFieldBinding.java
```diff
@@ -105,6 +105,12 @@ private RuntimeFieldBinding(int tag, String name, Message.Datatype datatype,
     this.messageField = messageField;
     this.builderField = builderField;
     this.builderMethod = builderMethod;
+
+    if (redacted && label == Message.Label.REQUIRED) {
```
JakeWharton: This check should be in the compiler.

swankjesse: There’s 2 reasons to use redacted:

 * To obscure values in `toString()`
 * To strip fields

Presumably redacted+required is safe for the first case?

wire-runtime/src/main/java/com/squareup/wire/RuntimeFieldBinding.java
```diff
@@ -152,6 +158,23 @@ void setBuilderField(Object builder, Object value) {
     }
   }
 
+  public void redactBuilderField(Message.Builder<?> builder) {
+    if (!redacted && datatype != Message.Datatype.MESSAGE) return;
```
swankjesse: not sure I follow this: who redacts scalar fields?

swankjesse: Ahh. . . it's polymorphic. Duh.

swankjesse:
LGTM

wire-runtime/src/main/java/com/squareup/wire/RuntimeFieldBinding.java
JakeWharton: Ah, good catch. Will move to `redact`.

PR #316, opened 2015-08-14
Remove wasteful StringBuilder allocations which just call String.valueOf anyway.

JakeWharton:


JakeWharton:
`</petPeeve>`

danrice-square:
LGTM

PR #318, opened 2015-08-14
okio -> 1.5.0 (latest)

vanvlack:
update it, due to 1.2.0 being 8 months old

ran `mvn clean verify` without issues. Also signed the contributing agreement.


swankjesse:
Yes!

PR #320, opened 2015-08-15
TagBindings.

swankjesse:
This make message fields and extensions symmetric, and aligned with
a single API.

Some behavior changes:
 - toString now mixes regular fields and extensions together. I think
   this is fine.
 - we observe the set of extensions at MessageAdapter creation time, not
   lazily when an extension is encountered. This means it's going to work
   much better if extensions are registered first, particularly when
   parsing.


wire-runtime/src/main/java/com/squareup/wire/ExtensionTagBinding.java
```diff
@@ -0,0 +1,43 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire;
+
+import com.squareup.wire.ExtendableMessage.ExtendableBuilder;
+
+final class ExtensionTagBinding<M extends Message> extends TagBinding<M, Message.Builder<M>> {
```
swankjesse: We should consider making `Extension` and this class the same thing.

wire-runtime/src/main/java/com/squareup/wire/RuntimeFieldBinding.java
```diff
@@ -1,189 +0,0 @@
-package com.squareup.wire;
-
-import java.io.IOException;
-import java.lang.reflect.Field;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
-import java.util.List;
-
-final class RuntimeFieldBinding {
```
swankjesse: Mostly renamed to `FieldTagBinding`.

wire-runtime/src/main/java/com/squareup/wire/RuntimeMessageAdapter.java
```diff
@@ -44,7 +42,7 @@
   private final Class<M> messageType;
   private final Class<Builder<M>> builderType;
   private final Constructor<Builder<M>> builderCopyConstructor;
-  private final TagMap<RuntimeFieldBinding> fieldBindingMap;
+  private final Map<Integer, TagBinding<M, Builder<M>>> tagBindings;
```
swankjesse: Not currently using `TagMap` 'cause it's awkward to create a derived TagMap. May revisit later; or potentially we don't need TagMap anymore because it's optimizing an uncommon case. (We should only have 1 TagMap in memory per message type in the common case.)

wire-runtime/src/main/java/com/squareup/wire/RuntimeMessageAdapter.java
```diff
@@ -53,20 +51,69 @@
...
+    // Create tag bindings for registered extensions.
+    for (Map.Entry<Integer, Extension<?, ?>> entry
+        : wire.registry.getExtensions(messageType).entrySet()) {
+      Extension<?, ?> extension = entry.getValue();
+      TypeAdapter<?> singleAdapter = TypeAdapter.get(wire, extension.getDatatype(),
+          extension.getMessageType(), extension.getEnumType());
+      tagBindings.put(entry.getKey(), new ExtensionTagBinding<M>(extension, singleAdapter));
+    }
+
+    this.tagBindings = Collections.unmodifiableMap(tagBindings);
```
swankjesse: This is the magic: a single map that contains both message tags and extension tags.

wire-runtime/src/main/java/com/squareup/wire/ExtensionRegistry.java
```diff
@@ -42,16 +43,8 @@
...
-  public <T extends ExtendableMessage<T>, E> Extension<T, E>
-      getExtension(Class<T> messageClass, int tag) {
-    Map<Integer, Extension<?, ?>> map = extensionsByTag.get(messageClass);
-    return map == null ? null : (Extension<T, E>) map.get(tag);
-  }
-
-  @SuppressWarnings("unchecked")
-  public <T extends ExtendableMessage<T>, E> Extension<T, E>
-      getExtension(Class<T> messageClass, String name) {
-    Map<String, Extension<?, ?>> nameMap = extensionsByName.get(messageClass);
```
JakeWharton: I think you can kill `extensionsByName` map and associated population in the loop.

swankjesse:
If I'm lucky, the fallout of this change will be that I'll be able to change `RuntimeMessageAdapter` into a generic `MessageAdapter` and only the tagbindings will be reflective or non-reflective.

And that'll allow me to use that generic `MessageAdapter` from the schema package, without any generated code.

wire-runtime/src/main/java/com/squareup/wire/MessageAdapter.java
```diff
@@ -26,11 +26,14 @@
 import static com.squareup.wire.Preconditions.checkNotNull;
 
 public abstract class MessageAdapter<M> {
+  /** Returns the raw type of the messages this adapter can read and write. */
+  public abstract Class<?> messageType();
```
JakeWharton: `Class<M>`

wire-runtime/src/main/java/com/squareup/wire/ExtensionRegistry.java
swankjesse: Oh indeed. I was able to drop the entire registry down to one multimap field:
```
Map<Class<? extends Message>, List<Extension<?, ?>>>
```

wire-runtime/src/main/java/com/squareup/wire/MessageAdapter.java
swankjesse: Sadly that doesn't work for Lists. I could change it to return a `Type`, but that wouldn't be better.

wire-runtime/src/main/java/com/squareup/wire/ExtensionRegistry.java
```diff
@@ -15,43 +15,29 @@
  */
 package com.squareup.wire;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 
 final class ExtensionRegistry {
```
JakeWharton: We should probably just fold this type into `Wire` as both are now tiny versions of their previous selves. Will do in a follow-up.

swankjesse: Yes. Please do.

wire-runtime/src/main/java/com/squareup/wire/MessageAdapter.java
JakeWharton: Gotcha. I hadn't gotten that far down. I saw `Class<M>` returned in the override and thought it might've just been an oversight.

wire-runtime/src/main/java/com/squareup/wire/RuntimeMessageAdapter.java
```diff
@@ -53,20 +51,67 @@
...
     for (Field messageField : messageType.getDeclaredFields()) {
-      // Process fields annotated with '@ProtoField'
       ProtoField protoField = messageField.getAnnotation(ProtoField.class);
       if (protoField != null) {
-        int tag = protoField.tag();
-        map.put(tag, RuntimeFieldBinding.create(wire, protoField, messageField, builderType));
+        TypeAdapter<?> singleAdapter = singleTypeAdapter(wire, messageField, protoField);
+        Class<?> singleType = singleAdapter.javaType;
+        FieldTagBinding<M> tagBinding = new FieldTagBinding<M>(protoField,
+            protoField.type(), singleAdapter, singleType, messageField, builderType);
```
JakeWharton: As this is the only instantiation site, passing `protoField.type()` is redundant when you're already passing the whole `protoField`.

wire-runtime/src/main/java/com/squareup/wire/RuntimeMessageAdapter.java
```diff
@@ -297,28 +276,17 @@ public String toString(M message) {
         }
         if (label.isRepeated()) {
           storage.add(tag, value != null ? value : Collections.emptyList());
-        } else if (extension != null) {
```
JakeWharton: So nice to see all these conditionals here and above go away!

wire-runtime/src/main/java/com/squareup/wire/TagBinding.java
```diff
@@ -0,0 +1,79 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire;
+
+import java.io.IOException;
+
+/**
+ * Read, write and describe a tag within a message. This class knows how to assign fields to a
```
JakeWharton: missing oxford comma... _/me \*ducks*_

wire-runtime/src/main/java/com/squareup/wire/TypeAdapter.java
```diff
@@ -270,11 +306,20 @@ void write(int tag, E value, ProtoWriter writer) throws IOException {
     };
   }
 
-  public static <T> TypeAdapter<List<T>> createPacked(final TypeAdapter<T> adapter) {
+  public TypeAdapter<?> withLabel(Message.Label label) {
```
JakeWharton: :+1:

JakeWharton:
LGTM

wire-runtime/src/main/java/com/squareup/wire/RuntimeMessageAdapter.java
swankjesse: Fixed!

PR #321, opened 2015-08-15
Fold extension registry into Wire.

swankjesse:


wire-runtime/src/main/java/com/squareup/wire/Wire.java
```diff
@@ -54,13 +55,11 @@ public Wire(Class<?>... extensionClasses) {
    * and start with the "Ext_" prefix.
    */
   public Wire(List<Class<?>> extensionClasses) {
-    this.registry = new ExtensionRegistry();
     for (Class<?> extensionClass : extensionClasses) {
       for (Field field : extensionClass.getDeclaredFields()) {
         if (field.getType().equals(Extension.class)) {
```
JakeWharton: I feel like we should just generate a `List<Extension>` static constant on each extension class and require those to be passed so we didn't need to reflect. It'd just be passing `FooBar.extensions` (or something) instead of `FooBar.class` as we skip all this.

swankjesse: Great idea.

PR #323, opened 2015-08-15
Limit responsibilities of TagBinding.

swankjesse:
It just reads and writes values.


wire-runtime/src/test/java/com/squareup/wire/RuntimeMessageAdapterRedactTest.java
```diff
@@ -60,6 +62,21 @@
     assertThat(wire.adapter(RedactedChild.class).redact(message)).isEqualTo(expected);
   }
 
+  @Test public void redactedExtensions() {
```
swankjesse: This didn't used to work!

Issue #301, opened 2015-08-07
Message extensions cannot have required fields


swankjesse:
We don't enforce this. We should.

JakeWharton:
I _never_ understood this in the code and always meant to follow-up but would forget.

PR #326, opened 2015-08-16
Extension fields may not be required.

swankjesse:
Closes https://github.com/square/wire/issues/301


JakeWharton:
LGTM. Build failure in schema.

swankjesse:
Yep. Tests were using required extensions!

PR #327, opened 2015-08-16
Don't return groups to callers of ProtoReader.

swankjesse:
This means alternate code that uses ProtoReader doesn't need to know
that groups even exist.

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
@@ -122,20 +122,30 @@ public final class ProtoReader {
   }
 
   /**
-   * Read the tag of the next field. Use {@link #peekType()} after calling this method to query its
-   * type.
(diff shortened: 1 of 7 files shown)


wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
```diff
@@ -122,10 +122,24 @@ public void endLengthDelimited(long token) throws IOException {
   }
 
   /**
-   * Read the tag of the next field. Use {@link #peekType()} after calling this method to query its
-   * type.
+   * Reads and returns the next tag of the message, or -1 if there are no further tags. This
+   * silently skips tags whose types are groups. Use {@link #peekType()} after calling this method
+   * to query its type.
```
JakeWharton: Switch the 2nd and 3rd sentences. Otherwise the "its" in the last sentence is ambiguous.

JakeWharton:
LGTM

JakeWharton:
Checkstyle crap failed the build.

PR #332, opened 2015-08-18
Fix bug in TestAllTypes sample data.

swankjesse:
The field encoding was VARINT when it should have been FIXED32.

### DIFF
--- a/wire-runtime/src/test/java/com/squareup/wire/protobuf/TestAllTypesData.java
+++ b/wire-runtime/src/test/java/com/squareup/wire/protobuf/TestAllTypesData.java
@@ -15,6 +15,8 @@
  */
 package com.squareup.wire.protobuf;
 
+import okio.ByteString;
+
 class TestAllTypesData {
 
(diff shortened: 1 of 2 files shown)


wire-runtime/src/test/java/com/squareup/wire/protobuf/TestAllTypesData.java
```diff
@@ -40,536 +42,472 @@
...
+      + "6f" // value = 111
+      + "f012" // tag = 302, type = 0
+      + "70" // value = 112
+      + "f012" // tag = 302, type = 0
+      + "70" // value = 112
+      + "f812" // tag = 303, type = 0
+      + "e201" // value = 226 (=113 zig-zag)
+      + "f812" // tag = 303, type = -
+      + "e201" // value = 226 (=113 zig-zag)
+      + "8513" // tag = 304, type = 5
```
swankjesse: Here's the fix

danrice-square:
LGTM

PR #334, opened 2015-08-18
Fold together MessageAdapter and TypeAdapter.

swankjesse:
Now that reading lengths is implicit in reading tags, we don't need
separate classes for TypeAdapters and MessageAdapters.


swankjesse:
Also rename `serializedSize` to `encodedSize`. I don't want to imply Java Serialization!

PR #338, opened 2015-08-21
Use 0 for the cached size to workaround Gson magic.

swankjesse:
Unfortunately we've got some users that rely on Gson's magic no-constructor-call
allocation of message instances, and these come back with 0 already cached when
that isn't the actual serialized size.

Work around by just changing the sentinel value. The drawback is that objects
that are actually 0 won't benefit from caching. This is an acceptable tradeoff;
those are the fastest to serialize anyway.


JakeWharton:
:frowning:

PR #341, opened 2015-08-22
Ugh. Defaults and arrays don't seem to get along.

swankjesse:


JakeWharton:
Can we detect the empty/null array in the run method and use the default path there? Having to specify each time is unfortunate.

swankjesse:
That's a bit closer to what you're requesting, and it also restores backwards-compatibility with the single-valued field. I couldn't figure a way to get the default value without defining a 2nd `@parameter` tag, which is sad. Learning the internals of Mojo might reveal a better solution, but the documentation wasn't helpful for my case.

wire-maven-plugin/src/test/resources/exemplar/pom.xml
```diff
@@ -62,6 +62,9 @@
               <goal>generate-sources</goal>
             </goals>
             <configuration>
+              <protoPaths>
+                <protoPath>${project.basedir}/src/main/proto</protoPath>
+              </protoPaths>
```
JakeWharton: Can be deleted now

JakeWharton:
Lgtm

Issue #278, opened 2015-07-20
Strict import checking


swankjesse:
Our current implementation permits `import` statements to be omitted. We should be more strict, and fail if a referenced type is not specifically imported.

PR #343, opened 2015-08-23
Strict import checking.

swankjesse:
Closes https://github.com/square/wire/issues/278

### DIFF
--- a/wire-schema/src/main/java/com/squareup/wire/schema/Linker.java
+++ b/wire-schema/src/main/java/com/squareup/wire/schema/Linker.java
@@ -33,6 +33,7 @@ final class Linker {
   private final ImmutableList<ProtoFile> protoFiles;
   private final Map<String, Type> protoTypeNames;
   private final Map<Type.Name, Map<String, Field>> extensionsMap;
+  private final Multimap<String, String> imports;
   private final List<String> errors;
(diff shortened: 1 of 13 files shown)


wire-schema/src/main/java/com/squareup/wire/schema/Linker.java
```diff
@@ -119,6 +137,16 @@ public Schema link() {
     return new Schema(protoFiles);
   }
 
+  /** Add all paths in {@code paths} to {@code sink}, plus their public imports, recursively. */
```
swankjesse: (this was surprisingly tricky to come up with)

wire-schema/src/test/java/com/squareup/wire/schema/SchemaTest.java
```diff
@@ -626,4 +626,177 @@
...
+              + "}\n")
+          .build();
+      fail();
+    } catch (SchemaException expected) {
+      assertThat(expected.getMessage()).isEqualTo("a.proto needs to import b.proto\n"
+          + "  for extend pb.B (a.proto at 2:1)");
+    }
+  }
+
+  @Test public void transitiveImportNotFollowed() throws Exception {
```
JakeWharton: Wow nice.

PR #346, opened 2015-08-30
Fix recursive messages in SchemaWireAdapterFactory.

swankjesse:
Also make the API public.


JakeWharton:
LGTM. Build failures.

PR #347, opened 2015-08-31
Drop Rx and Retrofit service factories.

swankjesse:
No known uses.


JakeWharton:
Also now broken with Retrofit 2.0b1 and presumptuous about things like using JSR 330.

LGTM. Build failed.

adriancole:
:shipit:

PR #348, opened 2015-08-31
New tag map class for both unknown and extension values.

swankjesse:
This isn't wired in yet, but I wanted to get this reviewed
independently. The implementation is interesting!

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/NewTagMap.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/NewTagMap.java
@@ -1,0 +1,134 @@
+/*
+ * Copyright 2015 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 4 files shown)


wire-runtime/src/main/java/com/squareup/wire/NewTagMap.java
```diff
@@ -0,0 +1,134 @@
...
+ * <p>Tags are stored in the order that they were added. Values that are semantically equal may be
+ * unequal according to this class because this implementation doesn't implement a strong order-
+ * independent equals. This is unlikely to be a problem in practice because encoders typically use a
+ * predictable encoding order.
+ */
+final class NewTagMap {
+  private int size = 0;
+
+  /** Tags. Defined only for [0..size). */
+  private int[] tags = new int[8];
```
JakeWharton: Assuming instances will be lazily initialized

swankjesse: For now.

swankjesse: (it's mutable, so anything else is a hassle!)

JakeWharton:
Cool tests. LGTM!

PR #354, opened 2015-09-04
Change how extensions are bound in RuntimeMessageAdapter.

swankjesse:
For a while we were using TagBinding which made extensions work just like
known fields.

With this change extensions get closer to unknown fields. This fits better
as currently we're manually filtering out extensions wherever we iterate
the tags map.


wire-runtime/src/main/java/com/squareup/wire/RegisteredExtension.java
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire;
+
+/** An extension and its type adapter. */
+final class RegisteredExtension {
```
swankjesse: this class is so negligible.

wire-runtime/src/main/java/com/squareup/wire/FieldBinding.java
```diff
@@ -20,8 +20,11 @@
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 
-final class FieldTagBinding<M extends Message<M>, B extends Message.Builder<M, B>>
-    extends TagBinding<M, Builder<M, B>> {
+/**
+ * Read, write, and describe a tag within a message. This class knows how to assign fields to a
```
JakeWharton: nice oxford comma :star:

wire-gson-support/src/main/java/com/squareup/wire/MessageTypeAdapter.java
```diff
@@ -55,19 +55,25 @@ public static UnknownFieldType of(String name) {
...
   @SuppressWarnings("unchecked")
   public MessageTypeAdapter(Wire wire, Gson gson, TypeToken<M> type) {
     this.gson = gson;
     this.messageAdapter = wire.messageAdapter((Class<M>) type.getRawType());
 
-    Map<String, TagBinding<M, Message.Builder<M, B>>> tagMap
-        = new LinkedHashMap<String, TagBinding<M, Message.Builder<M, B>>>();
-    for (TagBinding<M, Message.Builder<M, B>> tagBinding : messageAdapter.tagBindings().values()) {
-      tagMap.put(tagBinding.name, tagBinding);
+    Map<String, FieldBinding<M, B>> fieldBindings = new LinkedHashMap<String, FieldBinding<M, B>>();
```
JakeWharton: Next PR: switch language level to Java 7 for the diamond operator already!

wire-runtime/src/main/java/com/squareup/wire/RuntimeMessageAdapter.java
```diff
@@ -93,40 +92,47 @@
   private final Class<M> messageType;
   private final Class<Builder<M, B>> builderType;
```
JakeWharton: We should probably change all these to be `Class<B>`, `Constructor<B>`, etc. since `B` _is_ a `Builder<M, B>`. Follow up.

wire-runtime/src/main/java/com/squareup/wire/RuntimeMessageAdapter.java
```diff
@@ -93,40 +92,47 @@
...
 
-  Builder<M, B> newBuilder() {
+  Map<Integer, RegisteredExtension> extensions() {
+    return extensions;
+  }
+
+  B newBuilder() {
     try {
-      return builderType.newInstance();
+      return (B) builderType.newInstance();
```
JakeWharton: ...and it should remove this cast!

PR #355, opened 2015-09-04
Drop the Storage step when decoding.

swankjesse:
Now that tag maps handle multiplicity, just go right to the field.


JakeWharton:
Excellent.

PR #357, opened 2015-09-04
Improve pruning granularity to message fields and enum constants.

swankjesse:
Still need to think about nested types and oneofs.

### DIFF
--- a/wire-schema/src/main/java/com/squareup/wire/schema/Pruner.java
+++ b/wire-schema/src/main/java/com/squareup/wire/schema/Pruner.java
@@ -19,13 +19,21 @@ import com.google.common.collect.ImmutableList;
 import java.util.ArrayDeque;
 import java.util.Collection;
 import java.util.Deque;
-import java.util.LinkedHashSet;
-import java.util.Set;
+import java.util.NavigableSet;
+import java.util.TreeSet;
 
 /** Removes unused types and services. */
 final class Pruner {
-  /** Homogeneous identifiers including type names, service names, and RPC names. */
-  final Set<String> marks = new LinkedHashSet<>();
+  /**
+   * Homogeneous identifiers including type names, service names, field names, enum constants, and
+   * RPC names. This uses a navigable set so that we can easily check for the presence of member
+   * fields.
+   *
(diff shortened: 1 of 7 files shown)


wire-schema/src/main/java/com/squareup/wire/schema/Pruner.java
```diff
@@ -19,13 +19,21 @@
...
-  /** Homogeneous identifiers including type names, service names, and RPC names. */
-  final Set<String> marks = new LinkedHashSet<>();
+  /**
+   * Homogeneous identifiers including type names, service names, field names, enum constants, and
+   * RPC names. This uses a navigable set so that we can easily check for the presence of member
+   * fields.
+   *
+   * <p>Member names are omitted unless a specific subset is being retained. That is, if a type name
+   * {@code Foo} is present, all of Foo's members are retained unless some member name {@code
+   * Foo#Bar} is in the set.
```
szurbrigg: :+1:

szurbrigg:
:shipit:

wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.java
```diff
@@ -79,4 +79,105 @@
...
+    Schema schema = new SchemaBuilder()
+        .add("service.proto", ""
+            + "message MessageA {\n"
+            + "  optional string b = 1;\n"
+            + "  optional string c = 2;\n"
+            + "}\n"
+            + "message MessageB {\n"
+            + "}\n")
+        .build();
+    Schema pruned = schema.retainRoots(ImmutableList.of("MessageA#b"));
```
JakeWharton: This makes me want the inverse as well. The ability to prune fields (and thus transitive types).

JakeWharton: Mostly because it doesn't scale to say what you want at this granularity but rather what you don't want. Or at least in our case it doesn't.

swankjesse: https://github.com/square/wire/issues/361

PR #364, opened 2015-09-06
Omit MessageType from the @ProtoField annotation.

swankjesse:
It's not much more work to get it from the field itself, and it will
work better if we support custom wire adapters.

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
+++ b/wire-compiler/src/main/java/com/squareup/wire/java/TypeWriter.java
@@ -325,11 +325,6 @@ public final class TypeWriter {
       }
     }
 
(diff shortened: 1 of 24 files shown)


danrice-square:
@swankjesse We added the annotation parameter to work around a limitation of reflection support in j2objc, which we were attempting to use to translate Wire code. Unless someone is actively still trying to use Wire with obj2c I think it's fine to make this change.

swankjesse:
Ah, thanks Dan. I think the right solution for Objective C users is to teach Wire to emit nice Objective C code directly.

JakeWharton:
Or to use a `--no-reflection` flag to emit fully-formed adapters.

JakeWharton:
And I think we want to aspire to both.

PR #368, opened 2015-09-08
Use WireType instead of Message.Datatype.

swankjesse:
The nice upside of this change is that we now include the protobuf type name
like squareup.protos.person.Person in the annotation where it is known.

The other nice impact of this is we don't have the weird pseudotypes ENUM
and MESSAGE that don't correspond to anything on the wire.

The drawback is that the generated @ProtoField enum now uses a string instead
of an enum constant for known scalars.

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/WireAdapter.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/WireAdapter.java
@@ -19,7 +19,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import okio.Buffer;
 import okio.BufferedSink;
 import okio.BufferedSource;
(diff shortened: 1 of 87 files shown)


wire-runtime/src/main/java/com/squareup/wire/WireAdapter.java
```diff
@@ -347,6 +334,27 @@ public String toString(E value) {
     }
   };
 
+  private static final Map<WireType, WireAdapter<?>> TYPE_TO_ADAPTER;
+  static {
+    Map<WireType, WireAdapter<?>> map = new LinkedHashMap<WireType, WireAdapter<?>>();
+    map.put(WireType.BOOL, WireAdapter.BOOL);
```
swankjesse: tempting follow up: API on WireType to return a TypeAdapter if it is a scalar. Trick is what it does otherwise.

swankjesse: ... in which case, we could drop the 15 constants from `WireAdapter` that 100% correspond to the 15 constants in `WireType`.

PR #371, opened 2015-09-09
Move internal under schema package, the place of its use.

JakeWharton:
Also rename 'protoparser' to 'parser' since the qualifier is no longer useful in an environment which only deal with protos.


JakeWharton:
Just some left behind refactoring before things like 'schema' existed.

PR #372, opened 2015-09-09
Use FileSystem and Path for loading data.

JakeWharton:
The compiler is still a convoluted mess that needs fixed.


wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -68,14 +83,16 @@ public static void main(String... args) throws IOException {
...
-    Schema schema;
-    try {
-      schema = loader.load(options.sourceFileNames);
-    } catch (IOException e) {
-      throw new WireException(e);
+  public void compile() throws IOException {
+    List<String> sourceFileNames = options.sourceFileNames;
+    List<Path> sourcePaths = new ArrayList<>(sourceFileNames.size());
+    for (String sourceFileName : sourceFileNames) {
+      Path path = fs.getPath(sourceFileName);
```
swankjesse: This is relative, right?

wire-schema/src/main/java/com/squareup/wire/schema/Loader.java
```diff
@@ -28,86 +30,58 @@
 
 /** Load proto files and their transitive dependencies, parse them, and link them together. */
 public final class Loader {
-  private final IO io;
-  private final Map<String, ProtoFile> loaded = new LinkedHashMap<>();
+  private final List<Path> dirs;
```
swankjesse: paths?

wire-schema/src/main/java/com/squareup/wire/schema/Loader.java
```diff
@@ -28,86 +30,58 @@
...
       return;
     }
 
-    Location location;
-    try {
-      location = io.locate(path);
-    } catch (IOException e) {
-      throw new IOException("Failed to locate " + path + " with " + io, e);
+    ProtoFileElement element = null;
+    for (Path dir : dirs) {
```
swankjesse: ahhh, that `paths` is taken!

swankjesse: Maybe name 'em roots ?

swankjesse:
LGTM

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
JakeWharton: Yep

wire-schema/src/main/java/com/squareup/wire/schema/Loader.java
JakeWharton: We use the term "root" to denote types. I expect this class to undergo significant change in a follow-up I have started.

JakeWharton: I renamed to `protoDirs` and `protoFiles` for now. We can bikeshed the names in the follow-up where we'll expose these names in public API.

PR #374, opened 2015-09-09
Expose more user-friendly loading API for Schema.

JakeWharton:


wire-schema/src/main/java/com/squareup/wire/schema/SchemaLoader.java
```diff
@@ -0,0 +1,155 @@
...
+
+  public Schema load() throws IOException {
+    ImmutableList<Path> directories = this.directories.build();
+    if (directories.isEmpty()) {
+      throw new IllegalStateException("No directories added.");
+    }
+
+    Deque<Path> protos = new ArrayDeque<>(this.protos.build());
+    if (protos.isEmpty()) {
+      // TODO traverse all files in every directory.
```
JakeWharton: Follow-up

wire-schema/src/main/java/com/squareup/wire/schema/SchemaLoader.java
```diff
@@ -0,0 +1,155 @@
...
+  private final ImmutableList.Builder<Path> directories = ImmutableList.builder();
+  private final ImmutableList.Builder<Path> protos = ImmutableList.builder();
+
+  /** Add one or more directories from which proto files will be loaded. */
+  public SchemaLoader addDirectory(File file) {
+    return addDirectory(file.toPath());
+  }
+
+  /** Add directories from which proto files will be loaded. */
+  public SchemaLoader addDirectories(Collection<? extends File> files) {
```
JakeWharton: This (and with `Path` below) are erasure hacks against method overloading.

wire-schema/src/main/java/com/squareup/wire/schema/SchemaLoader.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.ArrayDeque;
+import java.util.Collection;
+import java.util.Deque;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import okio.Okio;
+import okio.Source;
+
+/** Load proto files and their transitive dependencies, parse them, and link them together. */
+public final class SchemaLoader {
```
swankjesse: Neato

wire-schema/src/main/java/com/squareup/wire/schema/SchemaLoader.java
swankjesse: Neat idea.

wire-schema/src/main/java/com/squareup/wire/schema/SchemaLoader.java
```diff
@@ -0,0 +1,155 @@
...
+        throw new FileNotFoundException("Failed to locate " + proto + " in " + directories);
+      }
+
+      ProtoFile protoFile = ProtoFile.get(element);
+      loaded.put(proto, protoFile);
+
+      // Queue dependencies to be loaded.
+      FileSystem fs = proto.getFileSystem();
+      for (String importPath : element.imports()) {
+        protos.addLast(fs.getPath(importPath));
```
swankjesse: Unsure the consequences of specifying a file system here. It shouldn't be file-system specific.

Actually not sure about using `Path` at all for the path part, since we also want to support `.jar` files in what is currently the directories list.

PR #375, opened 2015-09-10
Support loading proto files from ZIP files.

swankjesse:


wire-schema/src/main/java/com/squareup/wire/schema/SchemaLoader.java
```diff
@@ -107,6 +119,32 @@ public Schema load() throws IOException {
       throw new IllegalStateException("No directories added.");
     }
 
+    try {
+      // Attempt to load open each directory entry as a ZIP file. Most will fail; the rest are
+      // real ZIP files. We could ask first "is this a ZIP file" but trying and failing is easier!
+      for (Path directory : directories) {
+        if (Files.isRegularFile(directory)) {
+          try {
+            ZipFile zipFile = new ZipFile(directory.toFile());
```
swankjesse: `toFile()` breaks Jimfs! Not a problem here, but unfortunate that you can't have the `FileSystem` abstraction and random access simultaneously.

PR #376, opened 2015-09-11
Delete ServiceFactory.

swankjesse:
The new schema API does the same thing. It's also a little more flexible.


wire-schema/src/main/java/com/squareup/wire/schema/SchemaLoader.java
```diff
@@ -45,38 +42,26 @@
...
     directories.add(path);
     return this;
   }
 
-  /** Add directories from which proto files will be loaded. */
-  public SchemaLoader addDirectories(Iterable<? extends Path> paths) {
-    directories.addAll(paths);
-    return this;
+  /** Returns a mutable list of the directories that this loader will load from. */
+  public List<Path> directories() {
```
swankjesse: This is along the same pattern as OkHttp's interceptors, where a mutable class returns a mutable list. I was playing with the current API and programatically this is more powerful, especially for checks like "do we have _any_ directories?"

wire-compiler/src/test/java/com/squareup/wire/CommandLineOptionsTest.java
```diff
@@ -1,13 +1,15 @@
 package com.squareup.wire;
 
 import com.google.common.collect.Lists;
-import com.squareup.wire.java.SimpleServiceFactory;
-import org.junit.Test;
-
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.PrintWriter;
-import java.util.*;
```
JakeWharton: I smell merge conflict!

JakeWharton:
Needs a rebase.

LGTM

PR #377, opened 2015-09-11
Use java.nio.FileSystem instead of JavaGenerator.IO.

swankjesse:


wire-compiler/src/test/java/com/squareup/wire/WireCompilerErrorTest.java
```diff
@@ -38,49 +35,31 @@
...
+  FileSystem fileSystem = Jimfs.newFileSystem(Configuration.unix());
 
   /**
    * Compile a .proto containing in a String and returns the contents of each output file,
    * indexed by class name.
    */
-  private Map<String, String> compile(String source) throws Exception {
-    CommandLineOptions options = new CommandLineOptions("/",  new File("."),
+  private void compile(String source) throws Exception {
+    CommandLineOptions options = new CommandLineOptions("/source",  new File("/target"),
```
JakeWharton: Can we just make `javaOut` be a String like the others?

JakeWharton:
LGTM

PR #379, opened 2015-09-11
Fold TypeWriter into JavaGenerator.

swankjesse:
Now that this functionality is public API (for build tools to use directly) there's
only really one public-facing type. It does both type mapping (int32 to Integer) and
also generating TypeSpec instances.


swankjesse:
screen shot 2015-09-11 at 12 46 11 am
The full API of the Java Generator.

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -82,9 +83,8 @@ public void compile() throws IOException {
       schema = schema.retainRoots(options.roots);
     }
 
-    JavaGenerator javaGenerator = JavaGenerator.get(schema);
-    TypeWriter typeWriter = new TypeWriter(
-        javaGenerator, options.emitOptions, options.enumOptions);
+    JavaGenerator javaGenerator = JavaGenerator.get(schema)
+        .withOptions(options.emitOptions, options.enumOptions);
```
swankjesse: (I think this is a really cute way to make it easy to get one of these without thinking about options, while still making options available to those of us who need 'em)

PR #380, opened 2015-09-12
Make Extension.unknown() and TagMap public.

swankjesse:


wire-runtime/src/test/java/com/squareup/wire/TagMapTest.java
```diff
@@ -319,9 +344,28 @@
...
-    TagMap map = new TagMap();
-    map.add(90, FieldEncoding.LENGTH_DELIMITED, ByteString.decodeHex("d904da04"));
+    TagMap map = new TagMap.Builder()
+        .add(90, FieldEncoding.LENGTH_DELIMITED, ByteString.decodeHex("d904da04"))
+        .build();
 
     assertThat(map.get(extension)).isEqualTo(Arrays.asList(601, 602));
   }
+
+  /** We had a bug where the capacity was being included in {@link TagMap#equals}. */
```
swankjesse: This was failing before.

JakeWharton:
Like the immutability parity with the rest.

PR #381, opened 2015-09-13
Move extensions from adapters to readers/writers.

swankjesse:
The big upside of this is that adapters can now be completely static
and stateless.


wire-runtime/src/main/java/com/squareup/wire/FieldBinding.java
```diff
@@ -48,27 +49,49 @@ private static Method getBuilderMethod(Class<?> builderType, String name, Class<
...
-  public final ProtoAdapter<?> singleAdapter;
-  public final ProtoAdapter<?> adapter;
-
   private final Field messageField;
   private final Field builderField;
   private final Method builderMethod;
 
-  FieldBinding(WireField wireField, ProtoAdapter<?> singleAdapter,
-      Field messageField, Class<B> builderType) {
+  // Delegate adapters are created lazily; otherwise we could stack overflow!
```
swankjesse: (no more deferred adapter. this laziness goes away when we have static constants!)

wire-runtime/src/main/java/com/squareup/wire/ProtoAdapter.java
```diff
@@ -138,20 +134,35 @@ public final void encode(OutputStream stream, E value) throws IOException {
...
-    return decode(Okio.buffer(Okio.source(stream)));
+    return decode(Okio.buffer(Okio.source(stream)), extensionRegistry);
   }
 
   /** Read an encoded message from {@code source}. */
   public final E decode(BufferedSource source) throws IOException {
+    return decode(source, ExtensionRegistry.NO_EXTENSIONS);
+  }
+
+  public final E decode(BufferedSource source, ExtensionRegistry extensionRegistry)
```
swankjesse: now there's way too many overloads. I can actually cheat a bit here and put a little state back, `WireAdapter.withExtensions()` but that operation would be transitive for calls on that instance.

wire-runtime/src/main/java/com/squareup/wire/Extension.java
```diff
@@ -91,158 +84,153 @@ private Builder(Class<T> extendedType, Class<? extends Message> messageType,
...
   enumExtending(String type, Class<E> enumType, Class<T> extendedType) {
-    return new Builder<>(extendedType, null, enumType, ProtoType.get(type));
+    return new Builder<>(extendedType, enumType, ProtoType.get(type));
   }
 
   public static <T extends Message<T>, M extends Message> Builder<T, M> messageExtending(
       String type, Class<M> messageType, Class<T> extendedType) {
-    return new Builder<>(extendedType, messageType, null, ProtoType.get(type));
+    return new Builder<>(extendedType, messageType, ProtoType.get(type)
+    );
```
JakeWharton: You dropped this.

wire-runtime/src/main/java/com/squareup/wire/RuntimeMessageAdapter.java
```diff
@@ -246,32 +202,19 @@ public String toString(M message) {
...
-        } catch (RuntimeEnumAdapter.EnumConstantNotFoundException e) {
-          // An unknown Enum value was encountered, store it as an unknown field
-          builder.addVarint(tag, e.value);
+        } else {
+          Extension<?, ?> extension = reader.getExtension(messageType, tag);
+          Object value = extension.getAdapter().decode(reader);
+          builder.ensureTagMap().add(extension, value);
         }
-        continue;
+      } catch (RuntimeEnumAdapter.EnumConstantNotFoundException e) {
```
swankjesse: We're down to exactly one catch block!

wire-runtime/src/main/java/com/squareup/wire/Extension.java
swankjesse: Fixed!

wire-runtime/src/main/java/com/squareup/wire/Wire.java
```diff
@@ -15,151 +15,12 @@
...
-    @Override public void encode(ProtoWriter writer, M value) throws IOException {
-      if (delegate == null) throw new IllegalStateException("Type adapter isn't ready");
-      delegate.encode(writer, value);
-    }
-
-    @Override public M decode(ProtoReader reader) throws IOException {
-      if (delegate == null) throw new IllegalStateException("Type adapter isn't ready");
-      return delegate.decode(reader);
-    }
+    return RuntimeMessageAdapter.create(type);
```
JakeWharton: Kick. Ass.

wire-runtime/src/main/java/com/squareup/wire/RuntimeMessageAdapter.java
```diff
@@ -246,32 +202,19 @@ public String toString(M message) {
...
+        // An unknown Enum value was encountered, store it as an unknown field
+        builder.setExtension(Extension.unknown(messageType, tag, FieldEncoding.VARINT), e.value);
       }
-
-      FieldEncoding fieldEncoding = reader.peekFieldEncoding();
-      Object value = fieldEncoding.rawProtoAdapter().decode(reader);
-      builder.ensureTagMap().add(tag, fieldEncoding, value);
     }
     reader.endMessage(token);
```
JakeWharton: This file used to be 1000+ lines of complex logic. Now it's like ~80 of simple logic. A long battle, a big win.

PR #383, opened 2015-09-13
Remove method indirection to value field.

JakeWharton:


JakeWharton:
Buying developers everywhere an extra 15 minutes until they hit their dex method count...

JakeWharton:
Also saving 3 microseconds at runtime.

PR #386, opened 2015-09-13
Rename FieldEncoding to ProtoEncoding to match convention.

JakeWharton:
This represents an encoded format in protocol buffers and thus should be prefixed with 'Proto'.


swankjesse:
LGTM, but I broke your rebase

swankjesse:
(would we ever confuse this for the text/binary encoding thing that Google offers? This is mostly so that decoders know how to skip unknown fields)

JakeWharton:
Do they have a type of similar name? It's been so long since I've seen their public API.

swankjesse:
`WireFormat.WIRETYPE_FIXED64`

PR #390, opened 2015-09-13
Change WireField to use the adapter, not the proto type.

swankjesse:
This opens the door to later generating custom adapters in other files,
or writing custom binding code to mix protos with built-in types.

### DIFF
--- a/wire-gson-support/src/test/java/com/squareup/wire/GsonTest.java
+++ b/wire-gson-support/src/test/java/com/squareup/wire/GsonTest.java
@@ -98,7 +98,7 @@ public class GsonTest {
 
   private static final String JSON_EXTENSIONS =
       ",\"squareup.protos.alltypes.ext_opt_int32\":2147483647,"
-          + "\"squareup.protos.alltypes.ext_opt_int64\":4611686018427388081,"
+          + "\"squareup.protos.alltypes.ext_opt_int64\":-4611686018427387726,"
           + "\"squareup.protos.alltypes.ext_opt_uint64\":13835058055282163890,"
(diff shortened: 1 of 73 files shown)


wire-gson-support/src/test/java/com/squareup/wire/GsonTest.java
```diff
@@ -98,7 +98,7 @@
 
   private static final String JSON_EXTENSIONS =
       ",\"squareup.protos.alltypes.ext_opt_int32\":2147483647,"
-          + "\"squareup.protos.alltypes.ext_opt_int64\":4611686018427388081,"
```
swankjesse: I changed this test because previously we could use either hte adapter for int64 or uint64 and both would have been correct. With this different test value, the adapter for int64 returns a negative value, whereas uint64 is positive...

wire-runtime/src/main/java/com/squareup/wire/ProtoAdapter.java
```diff
@@ -272,7 +272,24 @@ public String toString(E value) {
       return reader.readVarint64();
     }
   };
-  public static final ProtoAdapter<Long> UINT64 = INT64;
+  /**
+   * Like INT64, but negative longs are interpreted as large positive values, and encoded that way
+   * in JSON.
+   */
+  public static final ProtoAdapter<Long> UINT64 = new ProtoAdapter<Long>(
```
swankjesse: ... And here it matters! Because we're using the type adapter to imply what the JSON encoding should be. The binary encoding for both of these types is the same, but the JSON encoding is different.

PR #394, opened 2015-09-16
Restore Wire.get().

swankjesse:
This was lost by accident when we moved type adapters out of Wire.java.

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/Wire.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/Wire.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2013 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


wire-runtime/src/main/java/com/squareup/wire/Wire.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire;
+
+public final class Wire {
```
JakeWharton: needs private ctor

JakeWharton:
LGTM

PR #397, opened 2015-09-18
Expose repeated and packed adapter variant methods.

JakeWharton:


wire-runtime/src/test/java/com/squareup/wire/ProtoReaderTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.io.IOException;
+import okio.Buffer;
+import okio.ByteString;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public final class ProtoReaderTest {
+  @Test public void packedExposedAsRepeated() throws IOException {
```
JakeWharton: This test isn't for the changes in the runtime, but verifies behavior that the changes in the runtime now assume (the ability to decode a packed field as single repeated values).

wire-runtime/src/main/java/com/squareup/wire/ProtoAdapter.java
```diff
@@ -381,71 +381,75 @@ public String toString(E value) {
...
 
-  private static <T> ProtoAdapter<List<T>> createRepeated(final ProtoAdapter<T> adapter) {
-    return new ProtoAdapter<List<T>>(adapter.fieldEncoding, List.class) {
-      @Override public int encodedSize(List<T> value) {
-        throw new UnsupportedOperationException();
+  /** Returns an adapter for {@link E} but as a repeated value. */
+  public ProtoAdapter<List<E>> asRepeated() {
+    return new ProtoAdapter<List<E>>(fieldEncoding, List.class) {
+      @Override public int encodedSize(List<E> value) {
+        throw new UnsupportedOperationException("Repeated values can only be sized with a tag.");
```
swankjesse: nice.

PR #404, opened 2015-09-19
Change the maven plugin to not delegate to the CLI.

swankjesse:
The Java API is now good enough to use directly.


swankjesse:
Now with native logging.
```
[INFO] --- wire-maven-plugin:2.0.0-SNAPSHOT:generate-sources (default) @ wire-sample ---
[INFO] Loaded 2 proto files in 62.17 ms
[INFO] Generated com.squareup.dinosaurs.Dinosaur in 32.54 ms
[INFO] Generated com.squareup.geology.Period in 4.266 ms
```

JakeWharton:
Seems like there's a good amount of duplication here and in the CLI.

swankjesse:
Yes, currently lots of duplication. I suspect in the future these might diverge, especially if we want to add more sophisticated options (custom types) that aren't exposed through the CLI.

PR #409, opened 2015-09-19
Add unknown decoding helper to TagMap.Builder.

JakeWharton:
Also useful for `default` case in code gen:
```java
@Override public Person decode(ProtoReader reader) throws IOException {
  Person.Builder builder = new Person.Builder();
  long token = reader.beginMessage();
  for (int tag; (tag = reader.nextTag()) != -1;) {
    switch (tag) {
      case 1: builder.name(nameAdapter.decode(reader)); break;
      case 2: builder.id(idAdapter.decode(reader)); break;
      case 3: builder.email(emailAdapter.decode(reader)); break;
      default: builder.tagMap().decodeUnknown(reader, tag, Person.class); break;
  }
  reader.endMessage(token);
  return builder.build();
}
```


JakeWharton:
Actually this should probably be on the `Builder`

JakeWharton:
Maybe not. Open to opinions. There's a bunch of helpers that generated code will want, they might not all be appropriate for the runtime API

JakeWharton:
Actually this `reader.getExtension` method is pretty good.

PR #410, opened 2015-09-19
Lookup adapter for toString instead of creating.

JakeWharton:


swankjesse:
LGTM. Test failure is unrelated!

PR #414, opened 2015-09-19
Fix field resolution for options.

swankjesse:


wire-schema/src/test/java/com/squareup/wire/schema/SchemaBuilder.java
```diff
@@ -35,17 +35,18 @@
   final SchemaLoader schemaLoader = new SchemaLoader().addDirectory(root);
 
   public SchemaBuilder add(String name, String protoFile) {
-    Path path = root.resolve(name);
+    Path relativePath = fs.getPath(name);
```
swankjesse: Previously this was an absolute path, which caused duplicates on imports

PR #418, opened 2015-09-20
Try to lookup adapter before creating.

JakeWharton:
This avoids the costly recomputation of the constant cache on every call.

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/Wire.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/Wire.java
@@ -46,7 +46,12 @@ public final class Wire {
    * @param <E> the enum class type
    */
   public static <E extends Enum & WireEnum> E enumFromInt(Class<E> enumClass, int value) {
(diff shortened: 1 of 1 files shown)


wire-runtime/src/main/java/com/squareup/wire/Wire.java
```diff
@@ -46,7 +46,12 @@ private Wire() {
    * @param <E> the enum class type
    */
   public static <E extends Enum & WireEnum> E enumFromInt(Class<E> enumClass, int value) {
-    RuntimeEnumAdapter<E> adapter = ProtoAdapter.newEnumAdapter(enumClass);
+    RuntimeEnumAdapter<E> adapter;
+    try {
+      adapter = (RuntimeEnumAdapter<E>) enumClass.getField("ADAPTER").get(null);
+    } catch (IllegalAccessException | NoSuchFieldException | ClassCastException e) {
+      adapter = ProtoAdapter.newEnumAdapter(enumClass);
```
swankjesse: I'm having trouble imagining when this occurs!

PR #424, opened 2015-09-20
Optimize list movement for repeated fields.

JakeWharton:
We incorrectly mixed mutable (copy constructor) and immutable (default constructor) list values when initializing the builder. In the mutable case, we did an entire copy of the immutable list only to make another immutable copy when building. In the immutable case, we allowed it to pass through when building as an optimization. A mutable list is always used in the builder for consistency, but with some special behavior for efficiency:

* When the copy constructor is used, retain the immutable instance inside a mutable container from the original value and re-use it if it goes unchanged when building.
* When the default constructor is used, wrap an emptyList() in a mutable container and use it if no values were added when building.
* If the mutable container is mutated in any way, make a copy of the immutable data and into a proper mutable list and allow mutation.

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/Message.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/Message.java
@@ -139,19 +139,30 @@ public abstract class Message<T extends Message<T>> implements Serializable {
     public abstract T build();
   }
 
+  /** <b>For generated code only.</b> */
+  protected static <T> List<T> newMutableList() {
+    return new MutableOnWriteList<>(Collections.<T>emptyList());
+  }
+
   /** <b>For generated code only.</b> Utility method to return a mutable copy of {@code list}. */
   protected static <T> List<T> copyOf(List<T> list) {
     if (list == null) throw new NullPointerException("list == null");
+    if (list == Collections.emptyList() || list instanceof ImmutableList) {
+      return new MutableOnWriteList<>(list);
+    }
     return new ArrayList<>(list);
   }
 
   /** <b>For generated code only.</b> Utility method to return an immutable copy of {@code list}. */
   protected static <T> List<T> immutableCopyOf(List<T> list) {
     if (list == null) throw new NullPointerException("list == null");
+    if (list instanceof MutableOnWriteList) {
+      list = ((MutableOnWriteList<T>) list).mutableList;
+    }
     if (list == Collections.emptyList() || list instanceof ImmutableList) {
       return list;
     }
-    return Collections.unmodifiableList(new ArrayList<>(list));
+    return new ImmutableList<>(list);
   }
 
   /** <b>For generated code only.</b> */
--- a/wire-runtime/src/main/java/com/squareup/wire/MutableOnWriteList.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/MutableOnWriteList.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2015 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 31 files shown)


JakeWharton:
AWS (and thus Travis) is down.

```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO]
[INFO] wire ............................................... SUCCESS [  1.188 s]
[INFO] wire-runtime ....................................... SUCCESS [  5.967 s]
[INFO] wire-schema ........................................ SUCCESS [  3.659 s]
[INFO] wire-java-generator ................................ SUCCESS [  0.891 s]
[INFO] wire-maven-plugin .................................. SUCCESS [  2.711 s]
[INFO] wire-sample ........................................ SUCCESS [  0.713 s]
[INFO] wire-compiler ...................................... SUCCESS [  3.726 s]
[INFO] wire-gson-support .................................. SUCCESS [  1.245 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 20.359 s
[INFO] Finished at: 2015-09-20T10:41:09-04:00
[INFO] Final Memory: 46M/597M
[INFO] ------------------------------------------------------------------------
```

wire-runtime/src/main/java/com/squareup/wire/MutableOnWriteList.java
```diff
@@ -0,0 +1,52 @@
+package com.squareup.wire;
```
swankjesse: nit: copyright

wire-runtime/src/main/java/com/squareup/wire/MutableOnWriteList.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.io.ObjectStreamException;
+import java.io.Serializable;
+import java.util.AbstractList;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.RandomAccess;
+
+/** A wrapper around an empty/immutable list which only switches to mutable on first mutation. */
+final class MutableOnWriteList<T> extends AbstractList<T> implements RandomAccess, Serializable {
```
swankjesse: Good on remembering RandomAccess!

wire-runtime/src/main/java/com/squareup/wire/MutableOnWriteList.java
JakeWharton: done

wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -139,19 +139,30 @@ public TagMap buildTagMap() {
...
+      return new MutableOnWriteList<>(list);
+    }
     return new ArrayList<>(list);
   }
 
   /** <b>For generated code only.</b> Utility method to return an immutable copy of {@code list}. */
   protected static <T> List<T> immutableCopyOf(List<T> list) {
     if (list == null) throw new NullPointerException("list == null");
+    if (list instanceof MutableOnWriteList) {
+      list = ((MutableOnWriteList<T>) list).mutableList;
```
swankjesse: This is going to work great. Really neat.

wire-runtime/src/test/java/com/squareup/wire/WireTest.java
```diff
@@ -346,4 +347,38 @@ public void testBadEnum() throws IOException {
...
+    // Round-trip these instances through the builder and ensure the lists are the same instances.
+    assertThat(new Person.Builder(personWithPhone).build().phone).isSameAs(personWithPhone.phone);
+    assertThat(new Person.Builder(personNoPhone).build().phone).isSameAs(personNoPhone.phone);
+  }
+
+  @Test public void builderListsAreAlwaysMutable() {
+    PhoneNumber phone = new PhoneNumber.Builder().number("555-1212").type(PhoneType.WORK).build();
+
+    Person.Builder newBuilder = new Person.Builder();
+    newBuilder.phone.add(phone);
```
swankjesse: Win!

PR #425, opened 2015-09-20
Message.newBuilder() and withoutUnknownFields().

swankjesse:
The generics on newBuilder() are brutal, and I'm unhappy with them.

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/Message.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/Message.java
@@ -17,6 +17,8 @@ package com.squareup.wire;
 
 import java.io.ObjectStreamException;
 import java.io.Serializable;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -69,6 +71,39 @@ protected final Object writeReplace() throws ObjectStreamException {
     return new MessageSerializedForm<>(this, (Class<Message>) getClass());
   }
 
+  /** Returns a new builder for this class, initialized with the data in this message. */
+  public final <B extends Builder<T, B>> B newBuilder() {
```
swankjesse: should this be public? Useful for metaprogramming. Regular programming can just use the existing Builder constructor.

wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -69,6 +71,39 @@ protected final Object writeReplace() throws ObjectStreamException {
...
+      Constructor<B> constructor = RuntimeMessageAdapter.getBuilderCopyConstructor(
+          builderType, messageType);
+      return constructor.newInstance(this);
+    } catch (InstantiationException | IllegalAccessException | InvocationTargetException e) {
+      throw new IllegalStateException("Failed to create builder for " + messageType);
+    }
+  }
+
+  /** Returns a copy of this message with unknown fields removed. */
+  public final T withoutUnknownFields() {
```
swankjesse: Also withoutExtensions?

wire-runtime/src/main/java/com/squareup/wire/Message.java
JakeWharton: We could eliminate the builder's copy constructor and replace it with an abstract instance method here which was implemented on each message as the contents of the copy constructor.

JakeWharton: That gives you metaprogramming ability, same method count, and a bonus of no reflection.

JakeWharton: Downside is you'll double-initialize the builder fields, once with a `newMutableList()` allocation only to overwrite with the version from `copyOf`. Perhaps `copyOf` could peek inside the `MutableOnWriteList` and replace its immutable list with the passed in version to avoid the extra allocation.

swankjesse:
Obsoleted by https://github.com/square/wire/pull/456

PR #426, opened 2015-09-20
Generate value-to-constant mappers on each enum type.

JakeWharton:
Low cost, high leverage method.


JakeWharton:
Alternative: enum-specific `ProtoAdapter` subclass (e.g., `ProtoAdapter.EnumAdapter`) which exposes this method.

PR #427, opened 2015-09-20
Upgrade to JavaPoet 1.3 and use its NameAllocator.

swankjesse:
The most interesting fallout is that _package is now package_ in a few
places.

This may break backwards compatibility for certain JSON encoding
of proto messages that use Java keywords.


wire-gson-support/src/test/java/com/squareup/wire/protos/alltypes/AllTypes.java
```diff
@@ -1742,7 +1742,7 @@ public int hashCode() {
       return result;
     }
 
-    public static final class Builder extends com.squareup.wire.Message.Builder<NestedMessage, Builder> {
+    public static final class Builder extends com.squareup.wire.Message.Builder<NestedMessage, NestedMessage.Builder> {
```
swankjesse: JavaPoet behavior change!

JakeWharton: Is it a good one? The use of `Builder` should never be ambiguous here, would it?

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -337,6 +337,20 @@ public TypeSpec generateEnum(EnumType type) {
 
   /** Returns the generated code for {@code type}, which may be a top-level or a nested type. */
   public TypeSpec generateMessage(MessageType type) {
+    NameAllocator nameAllocator = new NameAllocator();
+    nameAllocator.newName("serialVersionUID", "serialVersionUID");
+    nameAllocator.newName("ADAPTER", "ADAPTER");
+    nameAllocator.newName("MESSAGE_OPTIONS", "MESSAGE_OPTIONS");
+    nameAllocator.newName("tagMap", "tagMap");
+    nameAllocator.newName("hashCode", "hashCode");
```
JakeWharton: Any reason these two are above the field names? I can't think of any.

JakeWharton: Also `hashCode` could just re-use `result`

JakeWharton: `hashCode` is from `Message`, `tagMap` is named in the public API.

JakeWharton: Still though, we could always `super.hashCode` qualify and do we care if `tagMap` gets renamed? I think users would prefer their message field name retained `tagMap` instead of some argument in the ctor they'll never use.

wire-gson-support/src/test/java/com/squareup/wire/protos/alltypes/AllTypes.java
swankjesse: Nope, it's a small regression. I wanna fix it by improving the resolution algorithm.

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
swankjesse: I'll use `super.` prefixes.

JakeWharton:
This is so cool. I was using `addThisIfOneOf` a lot in the code-gen stuff and now I can kill it all!

LGTM

PR #430, opened 2015-09-20
Redact messages inside repeated fields.

JakeWharton:


wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -166,6 +166,13 @@ public TagMap buildTagMap() {
   }
 
   /** <b>For generated code only.</b> */
+  protected static <T> void redactElements(List<T> list, ProtoAdapter<T> adapter) {
```
JakeWharton: This is here instead of inlined at the only call site for future code-gen purposes.

swankjesse:
I wonder if we'd be better off with a general purpose builder thing which redacted could build upon.

PR #433, opened 2015-09-21
Improve performance of TagMap in toString, expose helper API.

JakeWharton:
Calling `extensions(true)` and `get(Extension)` are expensive in terms of computation and allocation. This changes the `TagMap` to do an effectively-single-pass traversal of the underlying array appending to an existing `StringBuilder`. This method will also be used by generated code, and the implementation of `toString` in the message adapter has been updated to reflect more closely the behavior of what generated code would look like.

Without a method like this, generating `toString` in code is very verbose.

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/RuntimeMessageAdapter.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/RuntimeMessageAdapter.java
@@ -28,6 +28,8 @@ import static com.squareup.wire.Message.Builder;
 
 final class RuntimeMessageAdapter<M extends Message<M>, B extends Builder<M, B>>
     extends ProtoAdapter<M> {
+  private static final String REDACTED = "\u2588\u2588";
+
   static <M extends Message<M>, B extends Builder<M, B>> RuntimeMessageAdapter<M, B> create(
       Class<M> messageType) {
     Class<B> builderType = getBuilderType(messageType);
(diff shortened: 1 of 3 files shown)


wire-runtime/src/test/java/com/squareup/wire/WireTest.java
```diff
@@ -211,7 +211,7 @@ public void testExtensionEnum() throws Exception {
     // Original value shows up as an extension.
     assertThat(msg.toString()).contains("squareup.protos.simple.nested_enum_ext=BAZ");
     // New value is unknown in the tag map.
-    assertThat(newMsg.toString()).contains("ExternalMessage{129=[17]}");
+    assertThat(newMsg.toString()).contains("ExternalMessage{129=17}");
```
danrice-square: Why this change?

wire-runtime/src/main/java/com/squareup/wire/RuntimeMessageAdapter.java
```diff
@@ -165,31 +165,22 @@ B newBuilder(M value) {
...
-      if (seenValue) sb.append(", ");
-      if (extension.isUnknown()) {
-        sb.append(extension.getTag());
-      } else {
-        sb.append(extension.getName());
+      if (value != null) {
+        sb.append(", ")
+            .append(fieldBinding.name)
+            .append('=')
+            .append(fieldBinding.redacted ? "██" : value);
```
danrice-square: Define a constant for the redacted string with a comment as to what char is being used.

wire-runtime/src/test/java/com/squareup/wire/WireTest.java
JakeWharton: The extension is not a repeated one. It represents a single enum value.

wire-runtime/src/main/java/com/squareup/wire/RuntimeMessageAdapter.java
```diff
@@ -165,31 +165,22 @@ B newBuilder(M value) {
...
       }
-      sb.append('=').append(message.tagMap.get(extension));
-      seenValue = true;
     }
-    sb.append('}');
-    return sb.toString();
+    message.tagMap.appendToString(sb);
+    sb.ensureCapacity(name.length() + 2); // Make sure next two chars fit, for empty objects.
+    sb.setCharAt(name.length() - 1, name.charAt(name.length() - 1));
```
danrice-square: -1, handling of initial comma is too tricky.

wire-runtime/src/main/java/com/squareup/wire/RuntimeMessageAdapter.java
```diff
@@ -166,30 +168,20 @@ B newBuilder(M value) {
...
       }
-      sb.append('=').append(message.tagMap.get(extension));
-      seenValue = true;
     }
-    sb.append('}');
-    return sb.toString();
+    message.tagMap.appendToString(sb);
+
+    // Replace leading comma with class name and opening brace.
+    sb.replace(0, 2, messageType.getSimpleName() + '{');
```
swankjesse: what if the message is empty?

wire-runtime/src/main/java/com/squareup/wire/TagMap.java
```diff
@@ -227,6 +227,31 @@ static void transcode(List<Object> list, Extension<?, ?> sourceExtension,
...
+          .append(extension.isUnknown() ? extension.getTag() : extension.getName())
+          .append("=");
+      int end = runEnd(i);
+      if (end > i + 2) {
+        builder.append('[');
+        for (int start = i; i < end; i += 2) {
+          if (i > start) builder.append(", ");
+          builder.append(array[i + 1]);
+        }
+        i -= 2; // Undo double-increment.
```
swankjesse: instead of this, move the `i += 2` above to line 250?

wire-runtime/src/main/java/com/squareup/wire/RuntimeMessageAdapter.java
JakeWharton: `replace` will ensure there is enough capacity. I can write a test that shows it.

wire-runtime/src/main/java/com/squareup/wire/TagMap.java
swankjesse: ("double increment" is confusing because it's incrementing by 2)

swankjesse:
LGTM

Issue #413, opened 2015-09-19
Android code generation


JakeWharton:
* [x] Implement `Parcelable`, likely by delegating to the adapter for `byte[]` (#435)
* [ ] 'support-annotations' `@Nullable`? (#225) 

Triggered by JavaGenerator options.

EddieRingle:
@JakeWharton The Parcelable code that is generated for `createFromParcel()` fails to compile due to an unhandled checked exception here: https://github.com/square/wire/blob/master/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java#L467

JakeWharton:
Can you file an issue or send a PR?

EddieRingle:
@JakeWharton see #470

PR #435, opened 2015-09-21
Provide flag to emit Android-compatible Parcelable code.

JakeWharton:
Refs #413.


wire-runtime/src/test/proto-java/com/squareup/wire/protos/person/Person.java.android
```diff
@@ -0,0 +1,352 @@
...
+
+  private static final long serialVersionUID = 0L;
+
+  public static final String DEFAULT_NAME = "";
+
+  public static final Integer DEFAULT_ID = 0;
+
+  public static final String DEFAULT_EMAIL = "";
+
+  public static final Parcelable.Creator<Person> CREATOR = new Parcelable.Creator<Person>() {
```
swankjesse: Are `ADAPTER` and `CREATOR` brothers?!

Issue #437, opened 2015-09-21
Don't list proto files when running the compiler


swankjesse:
Instead just scan the directories and do the right thing.

JakeWharton:
Fixed by #449.

PR #443, opened 2015-09-26
Load zip files as their own FileSystems.

JakeWharton:
Java 7 ships with a built-in FileSystem that can load zip files. Booya!

(Note: #437 is going to kill all the `addProto` stuff which I switched to Strings for use across different FileSystems)

### DIFF
--- a/wire-maven-plugin/src/main/java/com/squareup/wire/mojo/WireGenerateSourcesMojo.java
+++ b/wire-maven-plugin/src/main/java/com/squareup/wire/mojo/WireGenerateSourcesMojo.java
@@ -154,10 +154,10 @@ public class WireGenerateSourcesMojo extends AbstractMojo {
 
     SchemaLoader schemaLoader = new SchemaLoader();
     for (String directory : directories) {
-      schemaLoader.addDirectory(new File(directory));
+      schemaLoader.addSource(new File(directory));
     }
     for (String proto : protos) {
-      schemaLoader.addProto(new File(proto));
+      schemaLoader.addProto(proto);
     }
     Schema schema = schemaLoader.load();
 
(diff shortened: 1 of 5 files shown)


JakeWharton:
Bonus: If you have some other obscure file format that you want to use (tar, gz, 7z, rar, etc.), you can toss its FileSystemProvider jar onto the classpath in which you're running Wire and It Just Works™!

wire-maven-plugin/src/main/java/com/squareup/wire/mojo/WireGenerateSourcesMojo.java
```diff
@@ -154,10 +154,10 @@ private Schema loadSchema(List<String> directories, List<String> protos) throws
 
     SchemaLoader schemaLoader = new SchemaLoader();
     for (String directory : directories) {
-      schemaLoader.addDirectory(new File(directory));
+      schemaLoader.addSource(new File(directory));
```
swankjesse: I like this name much better!

wire-schema/src/main/java/com/squareup/wire/schema/SchemaLoader.java
```diff
@@ -29,118 +32,89 @@
...
-        try {
-          zipFile.close();
-        } catch (IOException ignored) {
+    try (Closer closer = Closer.create()) {
+      List<Path> directories = new ArrayList<>();
+      for (Path source : sources) {
+        if (Files.isRegularFile(source)) {
+          FileSystem sourceFs = FileSystems.newFileSystem(source, getClass().getClassLoader());
+          closer.register(sourceFs);
+          Iterables.addAll(directories, sourceFs.getRootDirectories());
```
swankjesse: In practice this will have exactly one root, right?

swankjesse:
Wow, that's way better.

wire-schema/src/main/java/com/squareup/wire/schema/SchemaLoader.java
JakeWharton: Yes. I'm not sure why they chose multiple, but every FS I've seen always
returns one.

PR #444, opened 2015-09-26
Load all protos in the supplied paths.

### LINKED ISSUE #437: Don't list proto files when running the compiler

JakeWharton:
Closes #437.


wire-runtime/src/test/proto-java/com/squareup/wire/protos/ProtoRegistry.java
```diff
@@ -12,8 +12,8 @@
 public final class ProtoRegistry {
   @SuppressWarnings("unchecked")
   public static final List<Class<?>> EXTENSIONS = Collections.unmodifiableList(Arrays.asList(
-      Ext_simple_message.class,
-      Ext_foreign.class));
+      Ext_foreign.class,
+      Ext_simple_message.class));
```
JakeWharton: https://github.com/square/wire/issues/445

gen_tests.sh
```diff
@@ -4,66 +4,86 @@ set -e
...
+
+cp wire-runtime/src/test/proto-java.android/com/squareup/wire/protos/person/Person.java \
+   wire-runtime/src/test/proto-java/com/squareup/wire/protos/person/Person.java.android
 
-cp src/test/proto-java.android/com/squareup/wire/protos/person/Person.java \
-   src/test/proto-java/com/squareup/wire/protos/person/Person.java.android
+rm -r wire-runtime/src/test/proto-java.android
 
-rm -r src/test/proto-java.android
+rm -r temp_gen_tests
```
JakeWharton: This file is getting a bit crazy. I might port it to a file in `wire-runtime`'s tests with a main method so it's a little less crazy.

wire-runtime/src/test/proto-java/com/squareup/wire/protos/ProtoRegistry.java
JakeWharton: Hah. This non-determinism is actually the cause of the build failure!

gen_tests.sh
JakeWharton: Or we probably want to switch to a more partitioned integration test model where we just have multiple proto folders and check in their expected outputs.

swankjesse: I like self-contained tests with the big ugly strings. Though that's impractical for some integration-sized tests.

wire-schema/src/main/java/com/squareup/wire/schema/Location.java
```diff
@@ -26,23 +25,15 @@
...
+    return new AutoValue_Location(path, -1, -1);
   }
 
   public Location at(int line, int column) {
-    return new AutoValue_Location(base(), path(), line, column);
+    return new AutoValue_Location(path(), line, column);
   }
 
-  /** Returns the base of this location; typically a directory or .jar file. */
-  public abstract String base();
```
swankjesse: Hmm . . . I liked this, particularly for things like IDE integration

JakeWharton: It can stay, there just has to be some hackery in the tests which use the real filesystem and now require copying a subset of protos to a separate folder for processing.

I basically have to do `content.replace(giantTemoraryFolderString, "../wire-runtime/things/are/pretty");` to make it work. The current test setup is sorely in need of overhaul.

swankjesse:
LGTM

I'm a little anxious that this will be too slow for very large `.proto` datasets. In particular, imagine an organization that puts all their protos into a single repo. Would it make sense to make this behavior the default, but also allow the application developer to provide a seed set of `.proto` files and follow imports?

swankjesse:
Also a problem because previously an unused-and-broken `.proto` wouldn't break the build; now it could. For large shared repos this could be a big hassle.

JakeWharton:
> Would it make sense to make this behavior the default, but also allow the application developer to provide a seed set of .proto files and follow imports?

This is the behavior I wanted, but I was merely going off of #437 and our discussions. It's an easy change to make, you just do the tree walk when `protos.isEmpty()`.

swankjesse:
Cool. Let's do that, with the optional tree walk.

JakeWharton:
Going to break this up.

PR #446, opened 2015-09-26
Show diff when gen_tests has changes.

JakeWharton:
screen shot 2015-09-26 at 1 27 33 pm


JakeWharton:
TBR. I need this for diagnosing the failure in #444.

PR #449, opened 2015-09-27
Load all files in sources if none explicitly specified.

JakeWharton:


wire-schema/src/test/java/com/squareup/wire/schema/SchemaLoaderTest.java
```diff
@@ -33,6 +34,19 @@
   @Rule public final TemporaryFolder tempFolder1 = new TemporaryFolder();
   @Rule public final TemporaryFolder tempFolder2 = new TemporaryFolder();
 
+  @Test public void loadAllFilesWhenNoneSpecified() throws IOException {
```
swankjesse: I like.

swankjesse:
LGTM

PR #453, opened 2015-10-02
Drop extensions from the Wire runtime.

swankjesse:


wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -587,7 +578,7 @@ private MethodSpec messageAdapterEncode(NameAllocator nameAllocator, MessageType
             adapter, fieldTag, fieldName);
       }
     }
-    result.addStatement("value.tagMap().encode(writer)");
+    result.addStatement("writer.writeBytes(message.$L)", nameAllocator.get("unknownFields"));
```
JakeWharton: This should just be `.write(`, not `.writeBytes(`

JakeWharton: It should also be `.$L()` since the field is package scoped

JakeWharton: Nevermind on the former. It's a `ProtoWriter`! Not a `BufferedSink`...

JakeWharton:
LGTM! Just missing those two parenthesis.

PR #456, opened 2015-10-02
Replace copy constructor with newBuilder() API.

swankjesse:
This new method is polymorphic.

### DIFF
--- a/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
+++ b/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
@@ -448,6 +448,7 @@ public final class JavaGenerator {
 
     builder.addMethod(messageFieldsConstructor(nameAllocator, type));
     builder.addMethod(messageFieldsAndUnknownFieldsConstructor(nameAllocator, type));
+    builder.addMethod(newBuilder(nameAllocator, type));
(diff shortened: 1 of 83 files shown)


wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -630,7 +631,7 @@ private MethodSpec messageAdapterDecode(NameAllocator nameAllocator, MessageType
     result.beginControlFlow("default:");
     result.addStatement("$T fieldEncoding = reader.peekFieldEncoding()", FieldEncoding.class);
     result.addStatement("$T value = fieldEncoding.rawProtoAdapter().decode(reader)", Object.class);
-    result.addStatement("builder.addUnknownField(tag, fieldEncoding, value);");
```
JakeWharton: Should JavaPoet detect and log a warning or throw when you do this?

JakeWharton:
I think this make the right trade-offs. Plus one less thing using reflection!

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
swankjesse: https://github.com/square/javapoet/issues/333

swankjesse:
Yep. Migrating to this will be a lot of work!

PR #457, opened 2015-10-03
Add convenience method for returning the instance without unknowns.

### LINKED ISSUE #398: Unknown fields included in equals/hashCode when extensions present.
Right now this logic is broken, and I'm about to submit a PR which makes unknown fields always accounted for in equals/hashCode. However, this behavior is often undesireable. We should look into a method which returns a copy of the message without unknowns.

JakeWharton:
Closes #398.

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/Message.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/Message.java
@@ -56,6 +56,12 @@ public abstract class Message<T extends Message<T>> implements Serializable {
    */
   public abstract Builder newBuilder();
 
+  /** Returns this message with any unknown fields removed. */
+  public final T withoutUnknownFields() {
+    //noinspection unchecked
+    return (T) newBuilder().clearUnknownFields().build();
+  }
+
   @SuppressWarnings("unchecked")
   @Override public String toString() {
     return ProtoAdapter.get((Class<Message>) getClass()).toString(this);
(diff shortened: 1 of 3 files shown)


wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -56,6 +56,12 @@ public final ByteString unknownFields() {
    */
   public abstract Builder newBuilder();
```
JakeWharton: This returns a raw type instead of a correctly parameterized one. Unfortunately to fill it in properly we'd need to add `B extends Builder<T, B>` to `Message`. Otherwise the best we can do is `Builder<T, ?>` which is unfortunate...

wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -56,6 +56,12 @@ public final ByteString unknownFields() {
...
 
+  /** Returns this message with any unknown fields removed. */
+  public final T withoutUnknownFields() {
+    //noinspection unchecked
+    return (T) newBuilder().clearUnknownFields().build();
```
JakeWharton: The cast and warning here are a result of the raw type above. When they are correctly parameterized both go away.

wire-runtime/src/main/java/com/squareup/wire/Message.java
swankjesse: Ugh!

PR #458, opened 2015-10-03
Add generic ceremony to correctly parameterize newBuilder return type.

JakeWharton:


wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -54,12 +55,11 @@ public final ByteString unknownFields() {
...
    */
-  public abstract Builder newBuilder();
+  public abstract Builder<M, B> newBuilder();
 
   /** Returns this message with any unknown fields removed. */
-  public final T withoutUnknownFields() {
-    //noinspection unchecked
-    return (T) newBuilder().clearUnknownFields().build();
+  public final M withoutUnknownFields() {
+    return newBuilder().clearUnknownFields().build();
```
swankjesse: Nice that this is safe.

swankjesse:
Thanks!

Issue #454, opened 2015-10-02
Name --full something else


JakeWharton:
Per comments on #450.

swankjesse:
Maybe we should make _full_ the default and offer a `--compact` option?

JakeWharton:
No complaints here for that approach.

JakeWharton:
This happened.

PR #460, opened 2015-10-04
Make full the default and add --compact.

swankjesse:
https://github.com/square/wire/issues/454

### DIFF
--- a/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
+++ b/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
@@ -116,32 +116,32 @@ public final class JavaGenerator {
   private final boolean emitOptions;
   private final ImmutableSet<String> enumOptions;
   private final boolean emitAndroid;
-  private final boolean emitFull;
+  private final boolean emitCompact;
 
   private JavaGenerator(Schema schema, ImmutableMap<ProtoType, TypeName> nameToJavaName,
       boolean emitOptions, ImmutableSet<String> enumOptions, boolean emitAndroid,
-      boolean emitFull) {
+      boolean emitCompact) {
     this.schema = schema;
     this.nameToJavaName = nameToJavaName;
     this.emitOptions = emitOptions;
     this.enumOptions = enumOptions;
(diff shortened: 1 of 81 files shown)


wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -159,7 +159,7 @@ public static JavaGenerator get(Schema schema) {
     }
 
     return new JavaGenerator(schema, nameToJavaName.build(), false,
-        ImmutableSet.<String>of(), false, false);
+        ImmutableSet.<String>of(), false, true);
```
JakeWharton: Shouldn't this be `false` still? So that full is the default?

wire-compiler/src/test/java/com/squareup/wire/WireCompilerErrorTest.java
```diff
@@ -43,7 +43,7 @@
   private void compile(String source) throws Exception {
     CommandLineOptions options = new CommandLineOptions("/source",  "/target",
         singletonList("test.proto"), new ArrayList<String>(), true,
-        Collections.<String>emptySet(), false, false, false, false);
+        Collections.<String>emptySet(), false, false, false, true);
```
JakeWharton: Similarly, shouldn't this be `false`?

JakeWharton:
lgtm, just questions about defaults

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
swankjesse: Yes!

swankjesse: (I used IntelliJ to invert the boolean, and didn't go back to flip these)

wire-compiler/src/test/java/com/squareup/wire/WireCompilerErrorTest.java
swankjesse: Fixed.

PR #469, opened 2015-10-07
Omit the base directory from the generated code.

### LINKED ISSUE #462: Extension fields should document which file they came from

swankjesse:
This is dependent on the build directory, which makes it difficult to
do hermetic builds.

Closes #462.


wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -422,6 +422,9 @@ public TypeSpec generateMessage(MessageType type) {
       if (!field.documentation().isEmpty()) {
         fieldBuilder.addJavadoc("$L\n", sanitizeJavadoc(field.documentation()));
       }
+      if (field.isExtension()) {
```
swankjesse: FYI @danrice-square

JakeWharton:
Is there any reason to keep the base in `Location` anymore? Seems like we just populate it only to throw it away.

JakeWharton:
LGTM

swankjesse:
The base is still used in error reporting.

PR #470, opened 2015-10-07
Fix unhandled IOException in generated createFromParcel()

EddieRingle:
Not quite sure how you best wanted to handle this, but I'm just assuming if an IOException is thrown here something horrible has already gone wrong.


swankjesse:
Thanks!

PR #471, opened 2015-10-08
Guard against null elements in lists.

swankjesse:
Also provide more context when a null element is found.


wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -142,8 +142,8 @@ public ByteString buildUnknownFields() {
   }
 
   /** <b>For generated code only.</b> Utility method to return a mutable copy of {@code list}. */
-  protected static <T> List<T> copyOf(List<T> list) {
-    if (list == null) throw new NullPointerException("list == null");
+  protected static <T> List<T> copyOf(String name, List<T> list) {
+    if (list == null) throw new NullPointerException(name + " == null");
```
JakeWharton: This one really isn't needed, but I guess it doesn't hurt, and it makes explicit the contract.

JakeWharton:
Awesome

wire-runtime/src/main/java/com/squareup/wire/Message.java
swankjesse: Yeah, I think it's foolish consistency.

Issue #172, opened 2014-11-21
Shade the compiler


JakeWharton:
The compiler should be self-contained so as to not cause dependency resolution problems with our upstream library usage choices.

Anecdotally, we have seen this happen twice with Wire already when used in a shared classloader such as a Gradle build script.

swankjesse:
Not taking action on this at this time. Might reopen if we run into problems in practice.

PR #473, opened 2015-10-08
Shade the compiler.

swankjesse:
Closes https://github.com/square/wire/issues/172

### DIFF
--- a/wire-compiler/pom.xml
+++ b/wire-compiler/pom.xml
@@ -62,23 +62,37 @@
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


swankjesse:
This only shades the CLI interface. Is that the right target?

swankjesse:
Notably, it doesn't shade the Wire internals. So if you want to use the Wire CLI and the Wire Java API in the same gradle plugin, you might still suffer a conflict.

JakeWharton:
This actually might no longer be necessary due to all the architectural changes that were made.

JakeWharton:
We can always revisit post-2.0 if it turns out to be necessary.

swankjesse:
Deal.

Issue #475, opened 2015-10-08
Name conflict when generating with --android


EddieRingle:
When generating code with the `--android` flag set, generated messages import `android.os.Parcelable`.

I have a message called `Creator`, which now conflicts with `Parcelable.Creator`. In particular, this creates an issue in the Builder class that is generated:

`public static final class Builder extends com.squareup.wire.Message.Builder<Creator, Builder>`

In the type parameters, `Creator` should reference the newly generated outer message class `Creator`, but instead resolves to `Parcelable.Creator`, causing a whole chain of errors when trying to compile.

EddieRingle:
@JakeWharton Does JavaPoet has some sort of flag to force the use of fully-qualified names? Would that be the way to go here or would you able to make JavaPoet smarter when it comes to resolving conflicts like this?

JakeWharton:
We would prefer to just fix JavaPoet to do the correct thing.

swankjesse:
I'm punting this to 2.1 'cause it's difficult & I don't want to block the Wire release.

swankjesse:
Blocked by https://github.com/square/javapoet/issues/77

PR #476, opened 2015-10-08
Add test to check for naming conflict with android.os.Parcelable.Creator

EddieRingle:
Not sure how useful this is, but I've taken the current generated output for a `Creator` message and replaced simple usages of `Creator` with the fully-qualified name of the message instead and added it as a test.

RE: square/wire#475


JakeWharton:
Cool. I think bumping to the latest JavaPoet (which would need a release) should fix this. You can test if you want by referencing 1.4.0-SNAPSHOT

EddieRingle:
@JakeWharton It doesn't appear that JavaPoet 1.4 resolves this.

JakeWharton:
Doh. Well that's unfortunate.

swankjesse:
So the problem here is that JavaPoet doesn’t consider that `Creator` is nested in the implemented interface `Parcelable`, and _that_ `Creator` takes precedence over the self-type `Creator`. We need to fix this in JavaPoet.

swankjesse:
https://github.com/square/javapoet/issues/77

swankjesse:
(This test belongs in JavaPoet. If you'd like to submit there, please do.)

PR #477, opened 2015-10-08
Handle extension name collisions on field options.

swankjesse:


JakeWharton:
I think you need a rebase on master? The SHA itself passes but the build for testing the merge SHA isn't.

JakeWharton:
LGTM

swankjesse:
Strange. The thing it's failing on is this:
```
Test files changed. Did you run gen_tests.sh?
```
But gen_tests is up-to-date.

swankjesse:
I think Travis was onto something. I think my fields are reordered in different JVMs. Will investigate.

PR #478, opened 2015-10-10
Improve pruning edge cases for options.

swankjesse:
One side effect of this change is that extensions are not always
emitted unconditionally. They must be retained.

### DIFF
--- a/wire-schema/src/main/java/com/squareup/wire/schema/IdentifierSet.java
+++ b/wire-schema/src/main/java/com/squareup/wire/schema/IdentifierSet.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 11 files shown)


wire-schema/src/main/java/com/squareup/wire/schema/IdentifierSet.java
```diff
@@ -0,0 +1,87 @@
...
+ * Actor#birth_date} or {@code Actor#oscar_count}.
+ */
+public final class IdentifierSet {
+  final NavigableSet<String> set = new TreeSet<>();
+
+  public boolean isEmpty() {
+    return set.isEmpty();
+  }
+
+  /** Adds a type name or member name. Returns true if the set was modified. */
```
JakeWharton: Calling `add("Foo#bar")` and then `add("Foo#baz")` will return true and then false, despite the set being modified the second time.

JakeWharton:
LGTM. Just one little bug I think. The consequences would be that options on `Foo#baz` wouldn't have been retained I think.

wire-schema/src/main/java/com/squareup/wire/schema/IdentifierSet.java
swankjesse: I disagree! I've added some tests that hopefully make it more obvious.

PR #479, opened 2015-10-11
Allow pruning to exclude members.

swankjesse:
This splits the IdentifierSet class into two:
 - IdentifierSet is the list of includes and excludes. It is immutable.
 - MarkSet keeps track of what has been marked in the marking phase. It
   is a working object that is used once and discarded.

Options are not currently subject to pruning; that needs to come in a follow up.
Also interesting will be excluding by package. Currently only types and members
can be excluded; this is likely insufficient.

### DIFF
--- a/wire-schema/src/main/java/com/squareup/wire/schema/IdentifierSet.java
+++ b/wire-schema/src/main/java/com/squareup/wire/schema/IdentifierSet.java
@@ -15,73 +15,100 @@
  */
 package com.squareup.wire.schema;
 
-import java.util.NavigableSet;
-import java.util.TreeSet;
+import com.google.common.collect.ImmutableSet;
 
 /**
- * A heterogeneous set of type names and type members. If a member is added to the set, its type is
- * implicitly also added. A type that is added without a specific member implicitly contains all
- * of that type's members.
+ * A heterogeneous set of type names and type members. If a member is included in the set, its type
+ * is implicitly also included. A type that is included without a specific member implicitly
+ * includes all of that type's members.
  *
- * <p>For example, an identifiers set populated with {@code Movie} and {@code Actor#name} contains
- * all members of {@code Movie} (such as {@code Movie#name} and {@code Movie#release_date}). It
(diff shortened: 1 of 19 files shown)


wire-schema/src/main/java/com/squareup/wire/schema/IdentifierSet.java
```diff
@@ -15,73 +15,100 @@
...
+ * <p>This set is initialized with <i>included identifiers</i> and <i>excluded identifiers</i>, with
+ * excludes taking precedence over includes. That is, if a type {@code Movie} is in both the
+ * includes and the excludes, it is not contained in the set. Any attempt to mark an excluded
+ * identifier will return false, preventing the identifier's transitive dependencies from also being
+ * marked.
+ *
+ * <p>If the includes set is empty, that implies that all elements should be included. Use this to
+ * exclude unwanted types and members without also including everything else.
  */
 public final class IdentifierSet {
```
swankjesse: public API! And a good place to document the roots syntax.

(Will probably rename roots to includes and excludes, in follow-up.)

wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.java
```diff
@@ -309,10 +339,89 @@
...
+    assertThat(((MessageType) pruned.getType("MessageA")).field("b")).isNotNull();
+    assertThat(((MessageType) pruned.getType("MessageA")).field("c")).isNull();
+  }
+
+  @Test public void excludeTypeExcludesField() throws Exception {
+    Schema schema = new SchemaBuilder()
+        .add("service.proto", ""
+            + "message MessageA {\n"
+            + "  optional MessageB b = 1;\n"
+            + "  optional MessageC c = 2;\n"
```
swankjesse: This is very interesting. Excluding `MessageC` impacts callsites.

JakeWharton:
Took me a while to get through, but this looks great.

PR #480, opened 2015-10-12
Prune options.

swankjesse:
This got a little out of hand. In order to correctly prune options we need
to have a proper definition of what the options are, and for that we need
to consult Google's descriptor.proto, particularly for things like deprecated
and java_package, otherwise those get list.

And then descriptor.proto isn't quite complete, it doesn't model the field
default option. It's special and goes directly on the field.

Pruning the options themselves isn't so bad, but it means we need to call
retainAll() on a bunch of things we weren't previously.

This also fixes pruning of oneofs, previously they were missing.

### DIFF
--- a/wire-schema/src/main/java/com/squareup/wire/schema/SchemaLoader.java
+++ b/wire-schema/src/main/java/com/squareup/wire/schema/SchemaLoader.java
@@ -21,6 +21,7 @@ import com.squareup.wire.schema.internal.parser.ProtoParser;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.file.FileSystem;
 import java.nio.file.FileSystems;
 import java.nio.file.FileVisitResult;
@@ -34,6 +35,7 @@ import java.util.Deque;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import okio.BufferedSource;
 import okio.Okio;
 import okio.Source;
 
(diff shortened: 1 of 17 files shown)


swankjesse:
GitHub is reporting 1000 new lines of code. That's a bit sensational; ~700 of these come from Google's `descriptor.proto` which I've imported.

wire-schema/src/main/java/com/squareup/wire/schema/SchemaLoader.java
```diff
@@ -156,9 +160,29 @@ private Schema loadFromDirectories(Map<Path, Path> directories) throws IOExcepti
...
+      loaded.put(DESCRIPTOR_PROTO, loadDescriptorProto());
+    }
+
     return new Linker(loaded.values()).link();
   }
 
+  /**
+   * Returns Google's protobuf descriptor, which defines standard options like default, deprecated,
+   * and java_package. If the user has provided their own version of the descriptor proto, that is
+   * preferred.
```
swankjesse: Should we prefer the `descriptor.proto` that comes with Wire?!

JakeWharton: Can we just require that it be included in the set when running the parser? It seems strange to pull a copy from somewhere other than what's being supplied.

JakeWharton:
LGTM. More great tests.

wire-schema/src/main/java/com/squareup/wire/schema/SchemaLoader.java
swankjesse: Unfortunately most calling code won't have it — and shouldn't have it. It's only necessary if you're defining custom extensions on things.

PR #481, opened 2015-10-13
Change options to reference values by member instead of by field.

swankjesse:
The problem with fields is that we replace instances with different
fields when pruning, and that causes options to become stale.

### DIFF
--- a/wire-schema/src/main/java/com/squareup/wire/schema/EnumType.java
+++ b/wire-schema/src/main/java/com/squareup/wire/schema/EnumType.java
@@ -22,7 +22,11 @@ import com.squareup.wire.schema.internal.parser.EnumElement;
 import java.util.Collection;
 import java.util.Map;
 
+import static com.squareup.wire.schema.Options.ENUM_OPTIONS;
+
 public final class EnumType extends Type {
+  static final ProtoMember ALLOW_ALIAS = ProtoMember.get(ENUM_OPTIONS, "allow_alias");
+
   private final ProtoType protoType;
   private final EnumElement element;
   private final ImmutableList<EnumConstant> constants;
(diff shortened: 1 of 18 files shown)


wire-schema/src/main/java/com/squareup/wire/schema/EnumType.java
```diff
@@ -22,7 +22,11 @@
 import java.util.Collection;
 import java.util.Map;
 
+import static com.squareup.wire.schema.Options.ENUM_OPTIONS;
+
 public final class EnumType extends Type {
+  static final ProtoMember ALLOW_ALIAS = ProtoMember.get(ENUM_OPTIONS, "allow_alias");
```
JakeWharton: So awesome

JakeWharton: Should we make all these public API?

JakeWharton: Actually we already do in `isDeprecated()` and `isPacked()` for a field, which is good enough for me.

wire-schema/src/main/java/com/squareup/wire/schema/ProtoMember.java
```diff
@@ -0,0 +1,73 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire.schema;
+
+/**
+ * Identifies a field, enum or RPC on a declaring type. Members are encoded as strings containing a
```
JakeWharton: oxford comma :grinning:

wire-schema/src/main/java/com/squareup/wire/schema/ProtoMember.java
```diff
@@ -0,0 +1,73 @@
...
+ * type name, a hash, and a member name, like {@code squareup.dinosaurs.Dinosaur#length_meters}.
```
JakeWharton: ...you got it down here!

PR #482, opened 2015-10-16
Expose --excludes in the command line API.

swankjesse:


wire-compiler/src/main/java/com/squareup/wire/CommandLineOptions.java
```diff
@@ -28,6 +28,8 @@
   public static final String JAVA_OUT_FLAG = "--java_out=";
   public static final String FILES_FLAG = "--files=";
   public static final String ROOTS_FLAG = "--roots=";
```
swankjesse: roots is now a non-documented feature of the command line API

swankjesse: I'll delete it completely, but not urgently 'cause now we have plugins that expect this to exist

wire-compiler/src/main/java/com/squareup/wire/CommandLineOptions.java
```diff
@@ -64,7 +66,8 @@
    * <pre>
    * java WireCompiler --proto_path=&lt;path&gt; --java_out=&lt;path&gt;
```
JakeWharton: All this Javadoc here is completely useless as this isn't a public type. We should merge this class with `WireCompiler`.

JakeWharton: Maybe rename it `WireCommandLine` or `WireCli` while we're at it

JakeWharton: Or just `com.squareup.wire.compiler.Main`

wire-compiler/src/main/java/com/squareup/wire/CommandLineOptions.java
JakeWharton: Mark with a TODO to remove before 2.0 final?

wire-compiler/src/main/java/com/squareup/wire/CommandLineOptions.java
swankjesse: https://github.com/square/wire/pull/484

PR #483, opened 2015-10-16
Switch the Maven plugin to includes and excludes instead of roots.

swankjesse:


wire-maven-plugin/src/main/java/com/squareup/wire/mojo/WireGenerateSourcesMojo.java
```diff
@@ -79,8 +82,9 @@
       List<String> protoFilesList = Arrays.asList(protoFiles);
       Schema schema = loadSchema(directories, protoFilesList);
 
-      if (roots != null && roots.length > 0) {
-        schema = retainRoots(schema);
+      IdentifierSet identifierSet = identifierSet();
+      if (!identifierSet.includesEverything() || !identifierSet.excludesNothing()) {
```
JakeWharton: `isEmpty()` on `IdentifierSet`?

swankjesse: Yeah, maybe. The naming is so awkward there.

PR #484, opened 2015-10-16
Fold CommandLineOptions and WireCompiler.

swankjesse:
Now that WireCompiler lets other classes do most of the heavy lifting,
its convenient to define options near to where they're used.


wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -20,32 +20,120 @@
...
+  final List<String> protoPaths;
+  final String javaOut;
+  final List<String> sourceFileNames;
+  final IdentifierSet identifierSet;
+  final boolean emitOptions;
+  final boolean dryRun;
+  final boolean emitAndroid;
+  final boolean emitCompact;
+
+  public WireCompiler(FileSystem fs, WireLogger log, List<String> protoPaths, String javaOut,
```
JakeWharton: private or package scope this guy? historically it's changed over the course of a major version when new major features come in.

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -53,43 +141,98 @@ public static void main(String... args) throws IOException {
     }
   }
 
-  static WireCompiler create(CommandLineOptions options) throws WireException {
-    return new WireCompiler(options, FileSystems.getDefault(),
-        new ConsoleWireLogger(options.quiet));
+  static WireCompiler forArgs(String... args) throws WireException {
```
JakeWharton: and `public` here instead

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
swankjesse: Definitely package-private. The only API this thing offers is `main()`.

wire-compiler/src/test/java/com/squareup/wire/CommandLineOptionsTest.java
```diff
@@ -1,11 +1,9 @@
 package com.squareup.wire;
```
JakeWharton: Class name should be renamed

JakeWharton: Oh there already is a `WireCompilerTest`

JakeWharton:
LGTM!

swankjesse:
Fixed so only `main` is public. Making anything else public isn't that useful, the only thing you can do with a `WireCompiler` is what main offers.

This class mostly just serves as a model for plugin authors.

Also, worth considering switching this to use one of the JCommander or something.

PR #485, opened 2015-10-16
Only parse protos when traversing all files in a source.

JakeWharton:


wire-schema/src/main/java/com/squareup/wire/schema/SchemaLoader.java
```diff
@@ -114,7 +114,9 @@ private Schema loadFromDirectories(Map<Path, Path> directories) throws IOExcepti
         Files.walkFileTree(entry.getValue(), new SimpleFileVisitor<Path>() {
           @Override public FileVisitResult visitFile(Path file, BasicFileAttributes attrs)
               throws IOException {
-            protos.add(entry.getValue().relativize(file).toString());
+            if (file.getFileName().toString().endsWith(".proto")) {
```
swankjesse: Would anyone ever have `.PROTO`? `AUTOEXEC.BAT` ?

swankjesse: (particularly tricky on file systems like the one I have on my Mac)

PR #486, opened 2015-10-16
Add folder-tree parsing and linking litmus tester.

JakeWharton:
This lets you evaluate changes to the parser and linker (via SchemaLoader) on a large swath of files at once.

### DIFF
--- a/wire-schema/src/test/java/com/squareup/wire/schema/ParsingTester.java
+++ b/wire-schema/src/test/java/com/squareup/wire/schema/ParsingTester.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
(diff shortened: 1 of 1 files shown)


wire-schema/src/test/java/com/squareup/wire/schema/ParsingTester.java
```diff
@@ -0,0 +1,50 @@
+package com.squareup.wire.schema;
```
swankjesse: copyright

swankjesse:
I love this.

swankjesse:
LGTM

wire-schema/src/test/java/com/squareup/wire/schema/ParsingTester.java
JakeWharton: one of these days it'll stick!

PR #488, opened 2015-10-16
Allow text format option separator to be omitted when value is a map.

### LINKED ISSUE #487: Nested object in text format omitting separator currently fails to parse
```
option (foo) = {
  bar {
    baz: 2
  }
};
```

JakeWharton:
This feels dirty. Closes #487.


swankjesse:
Coverage FTW.

swankjesse:
Alternately, we fix the offending files?

JakeWharton:
They're not technically broken, and I think I actually prefer it without
the colon!

PR #490, opened 2015-10-17
Refactor test API to avoid all the one-off methods.

JakeWharton:


wire-compiler/src/test/java/com/squareup/wire/WireCompilerTest.java
```diff
@@ -346,106 +277,132 @@ private void testProtoWithRoots(
         "com/squareup/wire/protos/roots/J.java",
         "com/squareup/wire/protos/roots/K.java",
     };
-    testProto(sources, outputs);
+    assertOutputs(outputs);
   }
 
   @Test public void testExcludes() throws Exception {
```
JakeWharton: I seem to have broken this test, but i'm not sure how...

JakeWharton: Ah, it wasn't actually asserting contents before!

JakeWharton: Fixed. One step back (adding to `gen_tests.sh`) and two steps forward!

PR #491, opened 2015-10-17
Emit all files when no files are explicitly specified.

JakeWharton:


swankjesse:
LGTM

I wanna emit all files always, and use excludes for filtering outputs.

JakeWharton:
:+1:

JakeWharton:
This was blocking us internally since I'm moving away from needing to specify a file list. Before it would just silently emit nothing!

PR #494, opened 2015-10-17
Wildcards on includes and excludes.

swankjesse:


JakeWharton:
This is awesome. Tests took a long time to read, but they're great.

PR #498, opened 2015-10-19
Start tracking unused identifiers.

swankjesse:
Not yet surfaced in the compiler or maven plugin.

### DIFF
--- a/wire-schema/src/main/java/com/squareup/wire/schema/IdentifierSet.java
+++ b/wire-schema/src/main/java/com/squareup/wire/schema/IdentifierSet.java
@@ -16,6 +16,9 @@
 package com.squareup.wire.schema;
 
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
+import java.util.LinkedHashSet;
+import java.util.Set;
 
 /**
  * A heterogeneous set of type names and type members. If a member is included in the set, its type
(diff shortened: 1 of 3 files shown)


wire-schema/src/main/java/com/squareup/wire/schema/IdentifierSet.java
```diff
@@ -78,13 +86,26 @@ public boolean includes(ProtoMember protoMember) {
...
 
-    boolean included = false;
+    String includeMatch = null;
+    String excludeMatch = null;
     while (identifier != null) {
-      if (excludes.contains(identifier)) return false;
-      included = included || includes.contains(identifier);
+      if (excludes.contains(identifier)) {
+        excludeMatch = identifier;
+      }
```
JakeWharton: Shouldn't this `break`? Otherwise you get last-matched rule instead of the first.

wire-schema/src/main/java/com/squareup/wire/schema/IdentifierSet.java
```diff
@@ -78,13 +86,26 @@ public boolean includes(ProtoMember protoMember) {
...
+      if (includes.contains(identifier)) {
+        includeMatch = identifier;
```
JakeWharton: Somewhat similarly, I would think this should only replace the value if it's `null`

wire-schema/src/main/java/com/squareup/wire/schema/IdentifierSet.java
JakeWharton: Hold up I'm thinking about this backwards. We're looking at an identifier and checking in a set of rules.

JakeWharton: Ok so overwriting will let the least-specific rule be reported.

JakeWharton:
LGTM

PR #504, opened 2015-10-24
New --named_files_only option. It is off by default.

swankjesse:
This causes Wire to emit everything unless specifically directed. We would
prefer to use pruning rather than selecting which files are included; that's
less likely to emit something that doesn't build.


JakeWharton:
LGTM lacking any better names

PR #505, opened 2015-10-24
Fix JavaGenerator to cleanly emit the latest descriptor.proto.

swankjesse:
We'd been relying on an old descriptor.proto. The new one has a problem with the
comment delimiter. The fix is ugly, but I can't find anything neater.


wire-runtime/src/test/proto-java/com/google/protobuf/EnumValueOptions.java
```diff
@@ -1,5 +1,5 @@
 // Code generated by Wire protocol buffer compiler, do not edit.
-// Source file: google/protobuf/descriptor.proto at 389:1
+// Source file: google/protobuf/descriptor.proto at 549:1
```
swankjesse: @mattprecious

wire-runtime/src/test/proto-java/com/google/protobuf/SourceCodeInfo.java
```diff
@@ -301,14 +314,72 @@ public Location redact(Location value) {
...
+     *   // Detached comment for corge. This is not leading or trailing comments
+     *   // to qux or corge because there are blank lines separating it from
+     *   // both.
+     *
+     *   // Detached comment for corge paragraph 2.
+     *
+     *   optional string corge = 5;
+     *   /* Block comment attached
+     *    * to corge.  Leading asterisks
+     *    * will be removed. *_/
```
swankjesse: `*_/`

wire-runtime/src/test/proto-java/com/google/protobuf/EnumValueOptions.java
JakeWharton: karma!

wire-runtime/src/test/proto-java/com/google/protobuf/SourceCodeInfo.java
JakeWharton: Interesting. Can we just turn one of the chars into an `&#0NN;`?

JakeWharton:
LGTM

wire-runtime/src/test/proto-java/com/google/protobuf/SourceCodeInfo.java
swankjesse: Good idea. Done.

PR #509, opened 2015-11-01
Do not emit case statements for aliased enum constants.

### LINKED ISSUE #508: Option allow_alias = true causes enum exception
Hey all, loving the 2.0.0 library... I have a piece of code where I cannot change the protobuf unfortunately. It uses this repeated enum:

    enum System_Enum{
      option allow_alias = true;
      SYSTEM_ENUM_NONE  = 0;
      SYSTEM_NAME_SIZE  = 8;
      SYSTEM_SN_SIZE    = 17;
      SYSTEM_FW_SIZE    = 6;
      SYSTEM_HW_SIZE    = 6;
      SYSTEM_PCB_SIZE   = 17;
    }

After running it through Wire, I get:

    public static System_Enum fromValue(int value) {
      switch (value) {
        case 0: return SYSTEM_ENUM_NONE;
        case 8: return SYSTEM_NAME_SIZE;
        case 17: return SYSTEM_SN_SIZE;
        case 6: return SYSTEM_FW_SIZE;
        case 6: return SYSTEM_HW_SIZE;
        case 17: return SYSTEM_PCB_SIZE;
        default: return null;
      }
    }

Which then errors out during compilation as: 
Error:(44, 7) error: duplicate case label

Do, my question is whether this should fail out during the wire-compiler stage, or whether it should be allowed? Is this a bug with the repeated cases in a switch?

Again... Not my .proto file - so no power to change.

Also, please note that it is apparently valid according to Google: https://developers.google.com/protocol-buffers/docs/proto?hl=en#enum

JakeWharton:
Closes #508.

### DIFF
--- a/wire-runtime/src/test/proto-java/com/squareup/wire/protos/simple/SimpleMessage.java
+++ b/wire-runtime/src/test/proto-java/com/squareup/wire/protos/simple/SimpleMessage.java
@@ -554,6 +554,7 @@ public final class SimpleMessage extends Message<SimpleMessage, SimpleMessage.Bu
 
     BAZ(3, null),
 
+    @Deprecated
     BUZ(3, true);
 
     public static final ProtoAdapter<NestedEnum> ADAPTER = ProtoAdapter.newEnumAdapter(NestedEnum.class);
--- a/wire-runtime/src/test/proto-java/com/squareup/wire/protos/simple/SimpleMessage.java
+++ b/wire-runtime/src/test/proto-java/com/squareup/wire/protos/simple/SimpleMessage.java
@@ -2,6 +2,7 @@
 // Source file: simple_message.proto at 24:1
 package com.squareup.wire.protos.simple;
 
+import com.google.protobuf.EnumOptions;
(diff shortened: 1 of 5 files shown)


wire-runtime/src/test/proto-java/com/squareup/wire/protos/simple/SimpleMessage.java
```diff
@@ -546,18 +548,27 @@ public NestedMessage build() {
...
-    FOO(1),
+    FOO(1, null),
 
-    BAR(2),
+    BAR(2, null),
 
-    BAZ(3);
+    BAZ(3, null),
+
+    BUZ(3, true);
```
JakeWharton: We're not emitting `@Deprecated` on enum constants. Working on that now.

swankjesse:
Nice.

swankjesse:
Would be interesting to not omit the value altogether, and to hook up something to Gson or Moshi to permit either value to be decoded. As-is it's a bit weird for things like equals and map keys.

Issue #461, opened 2015-10-04
Gson support broken without @WireField annotations


swankjesse:
Using Gson without customization _almost_ works. What's missing:

 - unknown types
 - uint64

The first is easy. The second is quite difficult. I'm tempted to punt on it.

JakeWharton:
I removed them for brevity. It would be nice to continue to omit them, but
if not it's not a huge deal if they return.

EddieRingle:
@swankjesse @JakeWharton Is this why I'm getting a NumberFormatException at runtime when trying to parse JSON into a Protobuf with wire-gson-support?

Exception throws at this line: https://github.com/square/wire/blob/master/wire-gson-support/src/main/java/com/squareup/wire/MessageTypeAdapter.java#L175

Message I'm trying to parse from JSON:
```proto
message CommentsResponse {
    ResponseMeta meta = 1;
    repeated Comment comments = 2;
}
```

`ResponseMeta` is another message defined above that one:
```proto
message ResponseMeta {
    string uid = 1;
    string type = 2;
    string request_url = 3;
    uint64 request_time_ms = 4;
    string cursor = 5;
    string window_token = 6;
}
```

If so, what would a good workaround be for now?

swankjesse:
@EddieRingle does using `--compact` fix for you?

EddieRingle:
@swankjesse Yes sir!

EddieRingle:
@swankjesse Though I just discovered this: If there's a field in the JSON that isn't specified in the Protobuf, it'll try to parse it as an Integer and fail (in this case it was a String). This could be tricky for me as our back-end guys could add new fields at any time. :\

swankjesse:
Well that's definitely not acceptable. We should discard unrecognized fields, and probably drop support for int field IDs also. I doubt anybody wants that, and it won't work with proto3 anyway.

EddieRingle:
@swankjesse Does that mean that fixing this would be as simple as removing [the call](https://github.com/square/wire/blob/master/wire-gson-support/src/main/java/com/squareup/wire/MessageTypeAdapter.java#L175) to `parseUnknownField()` in the `read()` method? If so I'll happily submit a PR.

JakeWharton:
Closed by #518.

JakeWharton:
Also, :cry: to have these back in our build to _only_ be used by a sibling module that I'm not using.

PR #515, opened 2015-11-20
Drop unknown fields when parsing JSON

EddieRingle:
Fixes an issue where JSON with unknown fields would result in an exception being thrown due to `MessageTypeAdapter` trying to parse that field as an Integer. Further discussed [here](https://github.com/square/wire/issues/461#issuecomment-158203475).


swankjesse:
Looks good! Want to submit a follow up that omits unknown fields when encoding?

Issue #507, opened 2015-10-27
Enums embedded in Messages - my proto files now must import themselves.


asarazan:
I tried to switch to wire 2.0 and it's spewing errors. Maybe I'm just getting something wrong with the proto_path setting?

```
~/level/brotofuff/resources/proto $ java -jar ~/wire-compiler-2.0.0-jar-with-dependencies.jar --java_out=~/wiretmp --proto_path=. `find . -name '*.proto'`
```

yields a whole bunch of stuff like this:

```
./users.proto needs to import users.proto
  for field error2 (././users.proto at 24:3)
  in message com.levelmoney.proto.users.SendPasswordResetResponse (././users.proto at 17:1)
```

The proto in question looks like this:

```
message SendPasswordResetResponse {
  enum SendPasswordResetError { ... }
  optional common.Error error = 1;
  optional common.Events events = 2;
  optional SendPasswordResetError error2 = 3;
}
```

swankjesse:
I think it’s the `././` in the path that's causing it confusion. It ends up not recognizing that `users.proto` and `././users.proto` are the same thing. That's a bug that we should fix.

In the interim, does this work?
```
~/level/brotofuff/resources/proto $ java -jar ~/wire-compiler-2.0.0-jar-with-dependencies.jar --java_out=~/wiretmp --proto_path=`pwd`
```

asarazan:
Yep that works!

asarazan:
Looks like I spoke too soon. It's erroring out again. Not sure if it's something about running it from a shell script or what, but everything just exploded.

swankjesse:
Uh oh. Should be an easy fix.

PR #516, opened 2015-11-21
Exclude unknown fields when encoding JSON also.

swankjesse:
Follow up to https://github.com/square/wire/pull/515


swankjesse:
FYI @EddieRingle

EddieRingle:
@swankjesse Ah, you beat me to it. Looks good to me!

EddieRingle:
@swankjesse @JakeWharton Would it be possible to do another bugfix release soon to get this and #515 into the repos without me going through JitPack?

JakeWharton:
Yeah we can do one. I was hoping to take a crack at #507 before the next bugfix, but don't have the time today.

PR #518, opened 2015-12-11
Make ProtoAdapter.get() more liberal. Always include @WireField in generated code.

crazybob:
1. "Be liberal in what you accept and conservative in what you return." Changed ProtoAdapter.get() to be more accepting. This code is typically called in reflective situations where you don't know the types until run time. With the old design, one would basically have to resort to raw types to get things to compile.

2. Always include @WireField in generated code. This way, the Gson adapter will work with non-compact generated code.

### DIFF
--- a/wire-compiler/src/test/java/com/squareup/wire/WireCompilerTest.java
+++ b/wire-compiler/src/test/java/com/squareup/wire/WireCompilerTest.java
@@ -15,6 +15,7 @@
  */
 package com.squareup.wire;
 
+import com.google.common.io.Files;
 import java.io.File;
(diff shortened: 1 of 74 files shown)


wire-compiler/src/test/java/com/squareup/wire/WireCompilerTest.java
```diff
@@ -15,6 +15,7 @@
  */
 package com.squareup.wire;
 
+import com.google.common.io.Files;
```
swankjesse: unused?

swankjesse:
Thanks!

PR #525, opened 2015-12-16
Add a little sugar to get an Adapter from a Message instance

pforhan:
Helps avoid messy warning-filled code like this:

```
  public byte[] getBytes(Message message) {
    Class<? extends Message> messageClass = message.getClass();
    // Here we're knowingly dropping type information to make generics
    // workable.
    // ProtoAdapter.get for a message's class will give us that
    // message's adapter.
    ProtoAdapter<Message> protoAdapter = (ProtoAdapter<Message>)
ProtoAdapter.get(messageClass);
    byte[] messageBody = protoAdapter.encode(message);
...
```

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/ProtoAdapter.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/ProtoAdapter.java
@@ -66,7 +66,12 @@ public abstract class ProtoAdapter<E> {
     return new RuntimeEnumAdapter<>(type);
   }
 
-  /** Returns the default adapter for {@code type}. */
+  /** Returns the adapter for the type of {@code Message}. */
+  public static <M extends Message> ProtoAdapter<M> get(M message) {
+    return (ProtoAdapter<M>) get(message.getClass());
+  }
+
+  /** Returns the adapter for {@code type}. */
   public static <M> ProtoAdapter<M> get(Class<M> type) {
     try {
       return (ProtoAdapter<M>) type.getField("ADAPTER").get(null);
--- a/wire-runtime/src/test/java/com/squareup/wire/ProtoAdapterTest.java
+++ b/wire-runtime/src/test/java/com/squareup/wire/ProtoAdapterTest.java
@@ -35,6 +35,18 @@ public final class ProtoAdapterTest {
     assertThat(personAdapter.decode(encoded.toByteArray())).isEqualTo(person);
   }
 
+  @Test public void getFromInstanceSameAsFromClass() throws Exception {
+    Person person = new Person.Builder()
+        .id(99)
+        .name("Omar Little")
+        .build();
+
+    ProtoAdapter<Person> instanceAdapter = ProtoAdapter.get(person);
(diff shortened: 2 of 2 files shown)


wire-runtime/src/test/java/com/squareup/wire/ProtoAdapterTest.java
```diff
@@ -35,6 +35,18 @@
...
+  @Test public void getFromInstance() throws Exception {
+    Person person = new Person.Builder()
+        .id(99)
+        .name("Omar Little")
+        .build();
+    ByteString encoded = ByteString.decodeHex("0a0b4f6d6172204c6974746c651063");
+
+    ProtoAdapter<Person> personAdapter = ProtoAdapter.get(person);
+    assertThat(ByteString.of(personAdapter.encode(person))).isEqualTo(encoded);
+    assertThat(personAdapter.decode(encoded.toByteArray())).isEqualTo(person);
```
JakeWharton: I think the only assertion you need is that the same instance as `Person.ADAPTER` was returned. That's the contract of the method and encoding is covered elsewhere.

wire-runtime/src/main/java/com/squareup/wire/ProtoAdapter.java
```diff
@@ -66,6 +66,11 @@ public ProtoAdapter(FieldEncoding fieldEncoding, Class<?> javaType) {
     return new RuntimeEnumAdapter<>(type);
   }
 
+  /** Returns the default adapter for the type of {@code Message}. */
```
JakeWharton: Default? I don't think we use that term anywhere. I think just "the adapter" works.

swankjesse:
Thumbs up on the API. Nice .

wire-runtime/src/test/java/com/squareup/wire/ProtoAdapterTest.java
pforhan: Good point, will fix.

wire-runtime/src/main/java/com/squareup/wire/ProtoAdapter.java
pforhan: I'll change on the javadoc below, then.

pforhan:
Updated based on recommendations.

wire-runtime/src/test/java/com/squareup/wire/ProtoAdapterTest.java
```diff
@@ -35,6 +35,18 @@
...
+  @Test public void getFromInstanceSameAsFromClass() throws Exception {
+    Person person = new Person.Builder()
+        .id(99)
+        .name("Omar Little")
+        .build();
+
+    ProtoAdapter<Person> instanceAdapter = ProtoAdapter.get(person);
+    ProtoAdapter<Person> classAdapter = ProtoAdapter.get(Person.class);
+
+    assertThat(instanceAdapter).isSameAs(classAdapter);
```
JakeWharton: FYI you can just call `Person.ADAPTER ` which is the source of truth as the adapter. Calling `get` just does reflection on this field.

PR #529, opened 2015-12-28
Promote some utility methods to public to avoid synthetic accessors.

JakeWharton:
These methods are called from Message.Builder and ProtoAdapter subclasses which are defined inside a Message subclass. Since neither is directly a Message subclass synthetic accessor methods needed to be generated on the enclosing types. By making these public they can be called from anywhere.


swankjesse:
LGTM

JakeWharton:
Moved to `WireInternal` type.

JakeWharton:
This eliminates 596 method references in our app :grinning:

swankjesse:
LGTM

PR #536, opened 2016-01-16
Added Message.encode().

crazybob:
Now, instead of:

Response.ADAPTER.encode(out, response);

I can write:

response.encode(out)

which is shorter, more fluent and less repetitive.

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/Message.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/Message.java
@@ -18,9 +18,11 @@ package com.squareup.wire;
 import com.squareup.wire.internal.Internal;
 import java.io.IOException;
 import java.io.ObjectStreamException;
+import java.io.OutputStream;
 import java.io.Serializable;
 import java.util.List;
 import okio.Buffer;
+import okio.BufferedSink;
 import okio.ByteString;
 
 /** A protocol buffer message. */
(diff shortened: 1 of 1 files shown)


wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -574,6 +575,14 @@ private FieldSpec messageAdapterField(NameAllocator nameAllocator, ClassName jav
     return result.build();
   }
 
+  private MethodSpec messageAdapterMethod(ClassName javaType) {
+    return MethodSpec.methodBuilder("adapter")
+        .addModifiers(PUBLIC)
+        .returns(adapterOf(javaType))
+        .addStatement("return ADAPTER")
```
JakeWharton: +`addAnnotation(Override.class)`

JakeWharton:
For what it's worth, 2.0.3 added `ProtoAdapter.get(response).encode(out)`.

swankjesse:
Yeah, I think that new API in 2.0.3 wins:

```
ProtoAdapter.get(response).encode(response, out)
```
vs.
```
response.adapter().encode(response, out)
```

crazybob:
ProtoAdapter.get() uses reflection and is less readable. You repeat "response" twice and it's less fluent.

crazybob:
Also, it's:

response.adapter().encode(out)

not:

response.adapter().encode(response, out)

swankjesse:
How does that work?

crazybob:
Oops. Never mind.

crazybob:
Maybe we could add Message.encode(...)?

swankjesse:
Yeah, maybe `Message.toByteString()` ?

swankjesse:
Actually `encode/decode` have nice symmetry. But I actually don’t mind the API we have now. It’s admittedly more verbose, but structurally simple.

crazybob:
Yeah, but I could write:

  response.encode(out)

instead of:

  Response.ADAPTER.encode(out, response);

I'm not worried about the decode case because I don't have to repeat "response" every time.

crazybob:
This doesn't affect the generated code, and it makes my code quite a bit prettier.

JakeWharton:
What about passing the adapter down through the constructor super call (to
avoid adding a method to each generated type) and then adding the adapter()
and convenience methods to Message. This costs 0 methods in the generated
code and still gives you the nice fluent API you want.

crazybob:
Actually, the latest version just adds methods to Message. PTAL.

crazybob:
I fixed the style error.

JakeWharton:
:+1:

I still might do the adapter-through-super change for 2.1 since, like you said, the current approach not only requires reflection but it isn't memoized.

PR #540, opened 2016-01-18
Provide Message instances with their adapter.

### LINKED ISSUE #537: Consider piping the adapter down to the base class through ctor super.
This lets the base class do things on the instance's corresponding adapter without needing reflection each time.

Deprecate `ProtoAdapter.get(Message)` and add `adapter()` on `Message` which returns the instance as well.

JakeWharton:
Closes #537.


swankjesse:
Works. It’s unfortunate that we’re increasing the in-memory size of every single message with data that could be a static field.

Issue #541, opened 2016-01-18
serialization issue with repeated fields and [packed = true] when used in embedded messages


SkinnerAnonymous:
Hi Guys,
Never posted something on github before, so if this is the wrong place to post issues on the topic, please let me know. And before I start, let me say thank you for the wire protobuf library. It’s a great utility for android app development!

Now to the issue: The current versions of Wire seem to have problems with the serialization/deserialization of repeated fields inside embedded messages when using the [packed = true] option. When the repeated field is not set but other fields inside the embedded message are, serialization/deserialization either does not work (IOException, see below) or fails at (de-)serializing the message correctly (values assigned to wrong fields or dropping fields completely). The problems are easily recreated, so I included below three cases that I stumbled across. I tested them with 2.0.1, 2.0.2 and 2.0.3. The result is always the same. Wire versions prior to 2.0.0 seem not to have this problem/these problems. I am using wire-compiler-VERSION-jar-with-dependencies.jar to compile the proto files and the gradle dependency 'com.squareup.wire:wire-runtime:VERSION' in Android Studio.

CASE 1: - IOException thrown -

Proto File:
package com.test;
message EmbeddedMessage {
    repeated int32 inner_repeated_number = 1 [packed=true];
    optional int32 inner_number_after = 2;
}
message OuterMessage {
  optional int32 outer_number_before = 1;
  optional EmbeddedMessage embedded_message = 2;
}

Java code:
```java
try {
    EmbeddedMessage.Builder embeddedMessageBuilder = new EmbeddedMessage.Builder();
    embeddedMessageBuilder.inner_number_after(1);
    OuterMessage.Builder outerMessageBuilder = new OuterMessage.Builder();
    outerMessageBuilder
        .outer_number_before(2)
        .embedded_message(embeddedMessageBuilder.build());
    byte[] outerMessagesBytes = OuterMessage.ADAPTER.encode(outerMessageBuilder.build());
    OuterMessage outerMessagesAfterSerialisation = OuterMessage.ADAPTER.decode(outerMessagesBytes);
} catch (Exception e) {
    e.printStackTrace();
}
```

Error Message:
```java
01-18 17:22:29.999 4344-4344/com.test W/System.err: java.io.IOException: Expected to end at 8 but was 6
01-18 17:22:30.039 4344-4344/com.test W/System.err:     at com.squareup.wire.ProtoReader.endMessage(ProtoReader.java:117)
[...]

PR #542, opened 2016-01-18
Fix a critical bug where we were encoding empty packed values.

swankjesse:
By not overriding encodedSizeWithTag() we were including the size for
empty packed lists, which is definitely not a good idea.

Closes: https://github.com/square/wire/issues/541


JakeWharton:
Needs rebased on `master` for the adapter change.

JakeWharton:
LGTM

PR #546, opened 2016-01-27
Add support for single quoted strings.

JakeWharton:
https://developers.google.com/protocol-buffers/docs/reference/proto3-spec#string_literals


swankjesse:
Nice.

Issue #283, opened 2015-07-24
New proto3 keyword: reserved


JakeWharton:
https://github.com/google/protobuf/releases/tag/v3.0.0-alpha-3

danrice-square:
Did you mean "reserved"?

PR #548, opened 2016-01-27
Add 'reserved' keyword parsing in element model.

JakeWharton:
Refs #283.


wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/MessageElement.java
```diff
@@ -32,7 +32,8 @@ public static Builder builder(Location location) {
         .oneOfs(ImmutableList.<OneOfElement>of())
         .nestedTypes(ImmutableList.<TypeElement>of())
         .extensions(ImmutableList.<ExtensionsElement>of())
-        .options(ImmutableList.<OptionElement>of());
+        .options(ImmutableList.<OptionElement>of())
+        .reserveds(ImmutableList.<ReservedElement>of());
```
JakeWharton: Could be "reservations" if this word offends.

PR #549, opened 2016-01-27
Reserved requires proto3.

JakeWharton:


wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ProtoParser.java
```diff
@@ -351,7 +354,7 @@ private Object readField(String documentation, Location location, String word) {
         break;
 
       default:
-        if (syntax == ProtoFile.Syntax.PROTO_2) {
+        if (syntax != ProtoFile.Syntax.PROTO_3) {
```
JakeWharton: null is implicit proto2

PR #550, opened 2016-01-27
Implement groups at the element level, throw at model level.

JakeWharton:
This lets us test parsing against any valid syntax, but not implement in the linker/schema.


JakeWharton:
Actually holdup. These can be in `oneof`s as well...

JakeWharton:
Pro-tip: re-open before force-pushing branch.

PR #551, opened 2016-01-27
Implement groups at the element level, throw at model level.

JakeWharton:
This lets us test parsing against any valid syntax, but not implement in the linker/schema.

### DIFF
--- a/wire-schema/src/test/java/com/squareup/wire/schema/SchemaTest.java
+++ b/wire-schema/src/test/java/com/squareup/wire/schema/SchemaTest.java
@@ -1041,10 +1041,50 @@ public final class SchemaTest {
               + "  optional .b.MessageB message_b = 1;\n"
               + "}\n")
           .build();
+      fail();
     } catch (SchemaException expected) {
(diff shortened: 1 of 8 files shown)


wire-schema/src/test/java/com/squareup/wire/schema/SchemaTest.java
```diff
@@ -1047,4 +1047,41 @@
...
+      new SchemaBuilder()
+          .add("test.proto", ""
+              + "message SearchResponse {\n"
+              + "  repeated group Result = 1 {\n"
+              + "    required string url = 2;\n"
+              + "    optional string title = 3;\n"
+              + "    repeated string snippets = 4;\n"
+              + "  }\n"
+              + "}\n")
+          .build();
```
swankjesse: Fail()

JakeWharton: Aha. At least this time it was copy/paste. Fixed along with the origin.

PR #552, opened 2016-01-27
Reserved is actually available for proto2 syntax.

JakeWharton:
Oops!


JakeWharton:
This is basically a revert of #549 so going to merge on green.

swankjesse:
LGTM

Should we build an integration test or something that downloads a set of files and tried to parse em?

PR #559, opened 2016-01-28
Defer enum method lookup to not conflict with classloading.

JakeWharton:


jlabanca:
LGTM

wire-runtime/src/main/java/com/squareup/wire/RuntimeEnumAdapter.java
```diff
@@ -24,14 +24,20 @@
  */
 final class RuntimeEnumAdapter<E extends WireEnum> extends ProtoAdapter<E> {
   private final Class<E> type;
-  private final Method fromValueMethod;
+  private Method fromValueMethod; // Loaded lazily to avoid reflection during class loading.
```
pyricau: `volatile` ? Probably doesn't heart much if we retrieve the method twice though.

JakeWharton: Right, and don't want to pay the cost of the volatile read when in all likelihood it will make no difference.

PR #565, opened 2016-03-24
Unexpected is supposed to return, not throw

mattprecious:


mattprecious:
Thought this was a mistake, but I think it's done on purpose.

JakeWharton:
Is it? It shouldn't matter to the calling code whether it throws in the method call or throws at the call site, right?

swankjesse:
Really doesn’t matter either way.

PR #566, opened 2016-03-26
Always create a no-arg constructor

jlabanca:
The no-arg constructor is invoked during deserialization and ensures that fields (e.g. Lists) are initialized properly.

### DIFF
--- a/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
+++ b/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
@@ -477,55 +477,16 @@ public final class JavaGenerator {
 
     builder.addMethod(messageFieldsConstructor(nameAllocator, type));
     builder.addMethod(messageFieldsAndUnknownFieldsConstructor(nameAllocator, type));
+    // If there are no fields, then we already created a public no-args constructor.
+    if (!type.fieldsAndOneOfFields().isEmpty()) {
+      builder.addMethod(messageNoArgsConstructor(type));
+    }
     builder.addMethod(newBuilder(nameAllocator, type));
     builder.addMethod(messageEquals(nameAllocator, type));
     builder.addMethod(messageHashCode(nameAllocator, type));
 
     if (emitAndroid) {
-      String adapterName = nameAllocator.get("ADAPTER");
-
-      builder.addSuperinterface(PARCELABLE);
-
-      builder.addMethod(MethodSpec.methodBuilder("writeToParcel")
-          .addAnnotation(Override.class)
-          .addModifiers(PUBLIC)
-          .addParameter(PARCEL, "out")
-          .addParameter(int.class, "flags")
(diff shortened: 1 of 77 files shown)


JakeWharton:
> It's possible for a repeated field to have a null value if you instantiate the class by deserializing from a source that does not define the field.

Shouldn't this be the bug to fix? We disallow null lists everywhere and there's other bits of code which assume non-null for every list.

jlabanca:
I'll take a look at WireTypeAdapterFactory.  Maybe I can modify that to always ensure that repeated fields are not null.

jlabanca:
Turns out this is a simple fix.  GSON will use the no-arg constructor if one is available, and it's even recommended:
> Well-behaved  classes that are meant for serialization and deserialization should have a no-argument constructor

Adding a private no-arg constructor that initializes repeated fields to empty lists ensures that the lists are never null.

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -477,55 +477,16 @@ public TypeSpec generateMessage(MessageType type) {
...
+    // If there are no fields, then we already created a public no-args constructor.
+    if (!type.fieldsAndOneOfFields().isEmpty()) {
+      builder.addMethod(messageNoArgsConstructor(type));
+    }
     builder.addMethod(newBuilder(nameAllocator, type));
     builder.addMethod(messageEquals(nameAllocator, type));
     builder.addMethod(messageHashCode(nameAllocator, type));
 
     if (emitAndroid) {
-      String adapterName = nameAllocator.get("ADAPTER");
```
jlabanca: I didn't change this code.  I just moved it because checkstyle complained that this method was more than 150 lines.

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -922,6 +931,29 @@ private MethodSpec messageFieldsAndUnknownFieldsConstructor(
...
+  /** The private no-arg constructor is used when an object is deserialized. */
+  // Example:
+  //
+  // private SimpleMessage() {
+  //   this(null, Internal.newMutableList(), ByteString.EMPTY);
+  // }
+  //
+  private MethodSpec messageNoArgsConstructor(MessageType type) {
+    MethodSpec.Builder result = MethodSpec.constructorBuilder().addModifiers(PRIVATE);
+    result.addJavadoc("Used for deserialization.\n");
```
JakeWharton: This comment is misleading. The constructor is not used for proto deserialization but only JSON deserialization via Gson and only when you are not using the supplied type adapter.

JakeWharton:
I'm still not clear on why we need this when the Gson support module provides the same effect (via using each message's builder to construct instances).

jlabanca:
Let me take a closer look at it.  Would be great if the adapter does everything we need.

PR #569, opened 2016-04-15
Do not throw when parsing a null repeated field

jlabanca:
MessageTypeAdapter throws an IllegalArgumentException when attempting to parse a repeated field that is set to null in a JSON string.  Ideally, the null field would never have been included in the JSON to begin with, but the source may not be as smart as Wire.  MessageTypeAdapter should handle it anyway.


danrice-square:
LGTM

wire-gson-support/src/main/java/com/squareup/wire/MessageTypeAdapter.java
```diff
@@ -136,8 +136,10 @@ private JsonElement parse(JsonReader in) {
   private Object parseValue(Label label, Type valueType, JsonElement valueElement) {
     if (label.isRepeated()) {
       List<Object> valueList = new ArrayList<>();
-      for (JsonElement element : valueElement.getAsJsonArray()) {
-        valueList.add(readJson(valueType, element));
+      if (!valueElement.isJsonNull()) {
```
JakeWharton: Maybe flip this check and early return `Collections.emptyList()`?

JakeWharton: Avoids the allocation.

JakeWharton:
Looks good

Issue #279, opened 2015-07-20
proto3 Working Document


swankjesse:
As kicked off by [issue 210](https://github.com/square/wire/issues/210), here’s what we need for proto3:
- [x] Forbid `optional` / `required` labels
- [x] First enum value with tag 0 (#575)
  - [x] Default Java enums to that value
  - [x] Forbid null for Java enums
- [x] `Any`
  - [x] support in parser (TODO oldergod: update)
  - [x] support in schema
  - [x] code gen and runtime API (requires `google/protobuf/any.proto`)
  - [x] serialization & deserialization
- [x] `map`
  - [x] support in parser
  - [x] support in schema (#578)
  - [x] code gen and runtime API (#596)
  - [x] serialization & deserialization (#596)
- [x] `Timestamp` and `Duration` (#1382)
  - [x] code gen and runtime API (`timestamp.proto,` `duration.proto`)
  - [x] serialization & deserialization
- [x] Packed is default for repeated scalars (#1383)
- [x] `reserved`
  - [x] support in parser (#548)
  - [x] support enforcement in schema (#283)
- [x] `stream` keyword for RPC response types (#282, #580)
- [x] No extensions unless it's a custom option. (#1384)
- [x] ~No unknown fields~ They got reverted in 3.5.
- [x] Forbid user-defined default values (#1385)
- [x] Scalars become non-nullable, but message fields remain nullable. (#1386)
  - Test case: empty string, default enum, 0 int in a repeated field should all be encoded.
  - Test case: map entries that have default values.
- [x] Support in Java.
  - [x] Crash if we pass a null to a non-nullable type (like `int32`)
  - [x] Default Java strings to ""
  - [x] Forbid null for Java strings
- [x] Support in Gson.
- [x] Tests for Gson and proto3.
- [x] Tests for Java and proto3.
- [x] AnyMessage documentation. (@oldergod)
- [x] Wrappers. (@swankjesse and @oldergod)
  - [x] Runtime.
  - [x] JSON.
- [x] JSON support.
  - [x] Moshi release for custom keys (@swankjesse to cut Moshi release) 
  - [x] MapAdapters. (@swankjesse)
    - [x] Unignore `WireJsonTest#wrappers()`.
    - [x] Map values in Gson (@swankjesse).
- [x] Create a doc to list differences we took in designing our generated code differently from Google. (@oldergod)
- [x] Proto2 and Proto3 interoperability tests. (@oldergod)
[...]

PR #578, opened 2016-06-03
Pipe map support through the schema.

JakeWharton:
This allows for linking and pruning. The schema-based proto adapter and code generation are not supported.

Refs #279.

### DIFF
--- a/wire-schema/src/main/java/com/squareup/wire/schema/ProtoType.java
+++ b/wire-schema/src/main/java/com/squareup/wire/schema/ProtoType.java
@@ -19,10 +19,11 @@ import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
- * Names a protocol buffer message, enumerated type, service, or a scalar. This class models a
+ * Names a protocol buffer message, enumerated type, service, map, or a scalar. This class models a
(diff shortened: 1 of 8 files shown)


wire-schema/src/main/java/com/squareup/wire/schema/ProtoType.java
```diff
@@ -65,11 +66,33 @@
...
+    this.valueType = null;
+  }
+
+  /** Creates a map type. */
+  ProtoType(ProtoType keyType, ProtoType valueType, String string) {
+    checkNotNull(keyType, "keyType == null");
+    checkNotNull(valueType, "valueType == null");
+    checkNotNull(string, "string == null");
+    checkArgument(
+        keyType.isScalar() && !keyType.equals(BYTES) && !keyType.equals(DOUBLE) && !keyType.equals(
```
swankjesse: interesting!

swankjesse: not enums eh?

wire-schema/src/main/java/com/squareup/wire/schema/ProtoType.java
```diff
@@ -65,11 +66,33 @@
...
+            FLOAT), "map key must be non-byte, non-floating point scalar: %s", keyType);
+    this.isScalar = false;
+    this.string = string;
+    this.isMap = true;
+    this.keyType = keyType; // TODO restrict what's allowed here
```
swankjesse: is the TODO done with the checkArgument above?

PR #579, opened 2016-06-05
Add direct link to Wire compiler and some documentation

tevjef:
I had dive through closed issues to find these two bits of information. I thought it would be great for it to be in the README.


swankjesse:
I don’t know how this took forever. Thank you.

ricky2400:
Yo

tevjef:
Better late than never!

PR #581, opened 2016-06-07
Eliminate null checks for repeated fields.

JakeWharton:
Note: There's a slight behavior change here in that empty lists are no longer included in toString output for a message.


JakeWharton:
Noticed this while working on maps, since maps will behave the same way as lists in most places.

swankjesse:
LGTM!

PR #585, opened 2016-06-09
Consolidate all Android-related support into a dedicated base class.

JakeWharton:
There's two primary benefits here which ultimately have the same effect: method reduction.

1. Switching from a per-message generated `Creator` implementation to one implementation which can work for all types by leveraging their `ProtoAdapter`.
2. Consolidating the two `Parcelable` methods onto this class which also uses the available `ProtoAdpater` for its implementation.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,6 +60,11 @@
   <dependencyManagement>
     <dependencies>
       <dependency>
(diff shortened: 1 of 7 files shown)


swankjesse:
LGTM. Love it.

PR #587, opened 2016-06-10
Retain redacted information even if options are excluded.

### LINKED ISSUE #586: Retain the correct default and redacted values even when stripping options.
This is an optimization for generating less code while retaining correct behavior. Our clients don't care about options at runtime and thus want to remove them, but we also want to retain the correct default values and redacted fields in the generated code at generation time.

JakeWharton:
Closes #586.


wire-runtime/src/test/proto-java/com/squareup/wire/protos/custom_options/FooBar.java.noOptions
```diff
@@ -77,12 +78,13 @@ public final class FooBar extends Message<FooBar, FooBar.Builder> {
   @WireField(
       tag = 7,
       adapter = "com.squareup.wire.protos.custom_options.FooBar#ADAPTER",
-      label = WireField.Label.REPEATED
+      label = WireField.Label.REPEATED,
+      redacted = true
```
JakeWharton: Here's the result!

JakeWharton:
Oh, great. Non-determinism based on filesystem iteration order or something.

JakeWharton:
Not sure why this change decided to expose the non-determinism in filesystems... but I'm really surprised it took this long.

swankjesse:
Yup, this is the way to go.

PR #588, opened 2016-06-10
Remove things which are no longer in scope for Wire's parsing

JakeWharton:
* Converting a schema back to elements.
* Rendering elements to proto.


JakeWharton:
We're going to use this in Future Things™

PR #594, opened 2016-06-11
Add support for empty enclosing types.

### LINKED ISSUE #570: Produce empty enclosing types when not explicitly marked.
Currently enclosing types are implicitly kept when a descendent is needed which also transitively keeps types of its members.

An example:
```proto
message FoodAndDrink {
  enum Bar { BEER_HALL }
  enum Food { MC_DONALDS }

  optional Food food = 1;
  optional Bar drink = 2;
}
message Person {
  optional FoodAndDrink.Bar drinking_at = 1;
}
```
When instructed to include only `Person` the generated code will contain `Food` as a result of `Bar`'s enclosing type `FoodAndDrink`'s member dependency.

Since `FoodAndDrink` is only needed for structure and not actually as a proto I would like it to become an empty type and its members not contribute dependencies. The generated code would instead look like this:
```java
public final class FoodAndDrink {
  private FoodAndDrink() {
    throw new AssertionError();
  }

  public enum Bar {
    BEER_HALL
  }
}
```

JakeWharton:
Closes #570.


wire-maven-plugin/src/main/java/com/squareup/wire/mojo/WireGenerateSourcesMojo.java
```diff
@@ -95,10 +93,8 @@
 
         for (Type type : protoFile.types()) {
           Stopwatch stopwatch = Stopwatch.createStarted();
+          TypeSpec typeSpec = javaGenerator.generateType(type);
           ClassName javaTypeName = (ClassName) javaGenerator.typeName(type.type());
-          TypeSpec typeSpec = type instanceof MessageType
```
swankjesse: Wow, we had this a lot more than I realized

wire-schema/src/main/java/com/squareup/wire/schema/EnclosingType.java
```diff
@@ -0,0 +1,89 @@
...
+  @Override public Location location() {
+    return location;
+  }
+
+  @Override public ProtoType type() {
+    return type;
+  }
+
+  @Override public String documentation() {
+    return documentation;
```
swankjesse: Should this be used?

wire-schema/src/main/java/com/squareup/wire/schema/EnclosingType.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  /** Always null. */
```
swankjesse: Not empty?

wire-schema/src/main/java/com/squareup/wire/schema/MessageType.java
```diff
@@ -229,10 +229,14 @@ void validate(Linker linker) {
...
-    // If this type is not retained, and none of its nested types are retained, prune it.
     ImmutableList<Type> retainedNestedTypes = retainedNestedTypesBuilder.build();
-    if (!markSet.contains(protoType) && retainedNestedTypes.isEmpty()) {
-      return null;
+    if (!markSet.contains(protoType)) {
+      // If this type is not retained, and none of its nested types are retained, prune it.
+      if (retainedNestedTypes.isEmpty()) {
+        return null;
+      }
+      // If this type is not retained but retained nested types, replace it with an enclosing type.
```
swankjesse: Dig.

swankjesse:
LGTM! Nice model.

wire-schema/src/main/java/com/squareup/wire/schema/EnclosingType.java
swankjesse: Otherwise throw an unsupported operation exception?

JakeWharton: Will throw

wire-schema/src/main/java/com/squareup/wire/schema/EnclosingType.java
JakeWharton: I had it originally (which is why the field and returning here exists) but removed it. I can add it back. I didn't have a strong opinion.

JakeWharton: I updated to include it along with the warning.

PR #595, opened 2016-06-11
Explicitly call superclass implementation.

JakeWharton:
This avoids Java generating a method reference to the desired method name on the calling class which at runtime resolves to the superclass. Instead, be explicit about the location of this method so that the method reference is directly to its implementation which eliminates one reference per Builder.

This saves us 237 methods.


wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -151,7 +148,7 @@ public Builder() {
      * Returns a byte string with this message's unknown fields. Returns an empty byte string if
      * this message has no unknown fields.
      */
-    public ByteString buildUnknownFields() {
+    public final ByteString buildUnknownFields() {
```
JakeWharton: These changes aren't strictly needed, but they feel correct since the superclass is the only one which manages the unknown field data.

swankjesse: Yeah, this seems safer

swankjesse: I like final on types wherever we can, and if we can’t final on methods is a good fallback

PR #596, opened 2016-06-13
Add support for map in the runtime.

JakeWharton:
Refs #279.


wire-gson-support/src/main/java/com/squareup/wire/MessageTypeAdapter.java
```diff
@@ -110,51 +111,63 @@ private void emitUint64(Long value, JsonWriter out) throws IOException {
...
-    return gson.fromJson(in, JsonElement.class);
-  }
-
-  private Object parseValue(Label label, Type valueType, JsonElement valueElement) {
-    if (label.isRepeated()) {
-      if (valueElement.isJsonNull()) {
+  private Object parseValue(FieldBinding<?, ?> fieldBinding, JsonElement element) {
+    if (fieldBinding.label.isRepeated()) {
+      if (element.isJsonNull()) {
         return Collections.emptyList();
```
swankjesse: off topic: this is sad.

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -219,10 +219,17 @@ public TypeName typeName(ProtoType protoType) {
   }
 
   private CodeBlock singleAdapterFor(Field field) {
-    ProtoType type = field.type();
+    return field.type().isMap()
+        ? CodeBlock.of("$N", field.name())
```
swankjesse: This is curious. I wonder how this works...

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -787,7 +816,13 @@ private String fieldName(ProtoType type, Field field) {
   }
 
   private TypeName fieldType(Field field) {
-    TypeName messageType = typeName(field.type());
+    ProtoType type = field.type();
+    if (type.isMap()) {
+      return ParameterizedTypeName.get(ClassName.get(Map.class), //
```
swankjesse: is `//` still useful? I nolove `//`

wire-runtime/src/main/java/com/squareup/wire/FieldBinding.java
```diff
@@ -45,48 +46,67 @@ private static Method getBuilderMethod(Class<?> builderType, String name, Class<
...
+    this.keyAdapterString = wireField.keyAdapter();
     this.adapterString = wireField.adapter();
     this.redacted = wireField.redacted();
     this.messageField = messageField;
     this.builderField = getBuilderField(builderType, name);
     this.builderMethod = getBuilderMethod(builderType, name, messageField.getType());
   }
 
+  boolean isMap() {
+    return !keyAdapterString.isEmpty();
```
swankjesse: empty instead of null? sigh

swankjesse: I’m guessing this is a consequence of annotations. . .

wire-runtime/src/main/java/com/squareup/wire/ProtoAdapter.java
```diff
@@ -66,6 +67,12 @@ public ProtoAdapter(FieldEncoding fieldEncoding, Class<?> javaType) {
     return new RuntimeEnumAdapter<>(type);
   }
 
+  /** Creates a new proto adapter for a map using {@code keyAdapter} and {@code valueAdapter}. */
+  public static <K, V> ProtoAdapter<Map<K, V>> createMapAdapter(ProtoAdapter<K> keyAdapter,
```
swankjesse: weirdly inconsistent for the line above `newEnumAdapter` and this `createMapAdapter` . . . maybe `newMapAdapter` ?

wire-runtime/src/main/java/com/squareup/wire/ProtoAdapter.java
```diff
@@ -485,4 +492,84 @@ public String toString(E value) {
...
+    }
+
+    @Override public void encodeWithTag(ProtoWriter writer, int tag, Map<K, V> value)
+        throws IOException {
+      for (Map.Entry<K, V> entry : value.entrySet()) {
+        entryAdapter.encodeWithTag(writer, tag, entry);
+      }
+    }
+
+    @Override public Map<K, V> decode(ProtoReader reader) throws IOException {
```
swankjesse: Maybe add a comment that this returns a single element map? That’s potentially surprising

wire-runtime/src/main/java/com/squareup/wire/ProtoAdapter.java
```diff
@@ -485,4 +492,84 @@ public String toString(E value) {
...
+          case 1: key = entryAdapter.keyAdapter.decode(reader); break;
+          case 2: value = entryAdapter.valueAdapter.decode(reader); break;
+          default: // Ignore unknown tags in map entries.
+        }
+      }
+      reader.endMessage(token);
+
+      if (key == null) throw new IllegalStateException("Map entry with null key");
+      if (value == null) throw new IllegalStateException("Map entry with null value");
+      return Collections.singletonMap(key, value);
```
swankjesse: Not sure about this call. I don’t think we ever do singleton lists do we?

swankjesse: Would it be possible for this to return just a map entry?

swankjesse: Or is this supposed to build a fullsize map and just doing a bad job?

wire-runtime/src/main/java/com/squareup/wire/ProtoAdapter.java
```diff
@@ -485,4 +492,84 @@ public String toString(E value) {
...
+    }
+
+    @Override public Map<K, V> redact(Map<K, V> value) {
+      return Collections.emptyMap();
```
swankjesse: Alternately, a map where we retain the keys and have redacted values?

swankjesse: (though that doesn’t work if it’s the keys we’re redacting)

wire-runtime/src/main/java/com/squareup/wire/RuntimeMessageAdapter.java
```diff
@@ -167,7 +167,10 @@ B newBuilder() {
       FieldBinding<M, B> fieldBinding = fieldBindings.get(tag);
       try {
         if (fieldBinding != null) {
-          Object value = fieldBinding.singleAdapter().decode(reader);
+          ProtoAdapter<?> adapter = fieldBinding.isMap()
+              ? fieldBinding.adapter()
```
swankjesse: This is a bit surprising. I would have expected maps and lists to be symmetric.

wire-runtime/src/main/java/com/squareup/wire/internal/ImmutableMap.java
```diff
@@ -0,0 +1,54 @@
+package com.squareup.wire.internal;
```
swankjesse: nit: copyright

wire-runtime/src/main/java/com/squareup/wire/internal/ImmutableMap.java
```diff
@@ -0,0 +1,54 @@
...
+
+  @Override public Set<K> keySet() {
+    return map.keySet();
+  }
+
+  @Override public Collection<V> values() {
+    return map.values();
+  }
+
+  @Override public Set<Entry<K, V>> entrySet() {
```
swankjesse: Sadly the entries are mutable

swankjesse: You want `http://docs.oracle.com/javase/7/docs/api/java/util/AbstractMap.SimpleImmutableEntry.html`

wire-runtime/src/main/java/com/squareup/wire/internal/ImmutableMap.java
```diff
@@ -0,0 +1,54 @@
...
+  @Override public boolean containsValue(Object value) {
+    return map.containsValue(value);
+  }
+
+  @Override public V get(Object key) {
+    return map.get(key);
+  }
+
+  @Override public Set<K> keySet() {
+    return map.keySet();
```
swankjesse: Sadly mutable

wire-runtime/src/main/java/com/squareup/wire/internal/ImmutableMap.java
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  @Override public Collection<V> values() {
+    return map.values();
```
swankjesse: Sadly mutable.

wire-runtime/src/main/java/com/squareup/wire/internal/Internal.java
```diff
@@ -53,12 +67,37 @@ private Internal() {
...
+    return result;
+  }
+
   public static <T> void redactElements(List<T> list, ProtoAdapter<T> adapter) {
     for (int i = 0, count = list.size(); i < count; i++) {
       list.set(i, adapter.redact(list.get(i)));
     }
   }
 
+  public static <T> void redactElements(Map<?, T> map, ProtoAdapter<T> adapter) {
```
swankjesse: Cool

wire-runtime/src/main/java/com/squareup/wire/internal/Internal.java
```diff
@@ -53,12 +67,37 @@ private Internal() {
...
+    for (Map.Entry<?, T> entry : map.entrySet()) {
+      entry.setValue(adapter.redact(entry.getValue()));
```
swankjesse: This can be mutating an immutable map?

swankjesse: (upon inspection, yes, and working as intended)

wire-runtime/src/main/java/com/squareup/wire/internal/Internal.java
```diff
@@ -94,6 +133,17 @@ public static void checkElementsNotNull(List<?> list) {
     }
   }
 
+  /** Throw {@link NullPointerException} if {@code map} or one of its keys or values are null. */
+  public static void checkElementsNotNull(Map<?, ?> map) {
+    if (map == null) throw new NullPointerException("map == null");
+    if (map.containsKey(null)) throw new NullPointerException("map.containsKey(null)");
```
swankjesse: probably simpler to just check in the loop?  You’re already going through entry-by-entry?

wire-runtime/src/main/java/com/squareup/wire/internal/MutableOnWriteMap.java
```diff
@@ -0,0 +1,90 @@
...
+  private final Map<K, V> immutableMap;
+  Map<K, V> mutableMap;
+
+  MutableOnWriteMap(Map<K, V> immutableMap) {
+    this.immutableMap = immutableMap;
+    this.mutableMap = immutableMap;
+  }
+
+  @Override public Set<Entry<K, V>> entrySet() {
+    return mutableMap.entrySet();
```
swankjesse: These entries are immutable, and mutating them will crash

wire-runtime/src/test/java/com/squareup/wire/MapTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+import org.junit.runners.Parameterized.Parameters;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(Parameterized.class)
+public final class MapTest {
+  private static final ByteString BYTES = ByteString.decodeHex(
+      "0a0c0a036f6e6512050a034f6e650a0c0a0374776f12050a0354776f0a100a05746872656512070a055468726565");
```
swankjesse: nice.

wire-runtime/src/test/proto-java/com/squareup/wire/map/Mappy.java
```diff
@@ -0,0 +1,140 @@
...
+import okio.ByteString;
+
+public final class Mappy extends Message<Mappy, Mappy.Builder> {
+  public static final ProtoAdapter<Mappy> ADAPTER = new ProtoAdapter_Mappy();
+
+  private static final long serialVersionUID = 0L;
+
+  @WireField(
+      tag = 1,
+      keyAdapter = "com.squareup.wire.ProtoAdapter#STRING",
```
swankjesse: dig.

wire-runtime/src/test/proto-java/com/squareup/wire/map/Mappy.java
```diff
@@ -0,0 +1,140 @@
...
+  )
+  public final Map<String, Thing> things;
+
+  public Mappy(Map<String, Thing> things) {
+    this(things, ByteString.EMPTY);
+  }
+
+  public Mappy(Map<String, Thing> things, ByteString unknownFields) {
+    super(ADAPTER, unknownFields);
+    this.things = Internal.immutableCopyOf("things", things);
```
swankjesse: Dig.

wire-runtime/src/test/proto-java/com/squareup/wire/map/Mappy.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @Override
+    public Mappy build() {
+      return new Mappy(things, super.buildUnknownFields());
+    }
+  }
+
+  private static final class ProtoAdapter_Mappy extends ProtoAdapter<Mappy> {
+    private final ProtoAdapter<Map<String, Thing>> things = ProtoAdapter.createMapAdapter(ProtoAdapter.STRING, Thing.ADAPTER);
```
swankjesse: Ahhh, so `things` can be the field name here. Makes sense.

wire-runtime/src/test/proto-java/com/squareup/wire/map/Mappy.java
```diff
@@ -0,0 +1,140 @@
...
+      writer.writeBytes(value.unknownFields());
+    }
+
+    @Override
+    public Mappy decode(ProtoReader reader) throws IOException {
+      Builder builder = new Builder();
+      long token = reader.beginMessage();
+      for (int tag; (tag = reader.nextTag()) != -1;) {
+        switch (tag) {
+          case 1: builder.things.putAll(things.decode(reader)); break;
```
swankjesse: ... and this is the code that decodes a single entry. Still tempted to make `things` a `ProtoAdapter<Map.Entry<String, Thing>>`

wire-runtime/src/test/proto/map.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto2";
+
+package com.squareup.wire.map;
+
+message Mappy {
+  map<string, Thing> things = 1;
```
swankjesse: sweet

swankjesse:
# nice!

Most of my comments are around how horribly painful it is to implement an immutable map. ([shoutout to 2008](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/ImmutableMap.java#L47)!)

wire-runtime/src/main/java/com/squareup/wire/ProtoAdapter.java
JakeWharton: This is because of how the encoding works.

Think of how we do repeated messages: we create an empty list and append a single message at a time as we see them. This is because the proto adapter only knows how to decode the message but the parent adapter is the one that matches on the tag and delegates.

Maps are effectively a repeated message of key/value pairs. Unlike lists, however, there are two types that we need to decode along with the enclosing message. This means that we either can re-use the `Map<K, V>` adapter which is needed for encoding and getting the encoded size for reading single-element `Map<K, V>` instances, or we can create a separate adapter for just `Map.Entry<K, V>`.

If you compare the implementation of `Collections.singletonMap` with `SimpleMapEntry` they're effectively the same class. The difference is that `Map` has a `putAll(Map)` method where there is no `put(Map.Entry)`. This means that if we want the calling code to be a one-liner, the singleton map + `putAll` is the easiest way to go.

If we use an adapter for `Map.Entry` instead, you also have a problem of encoding. If you call `asRepeated()` which returns a `ProtoAdapter<List<Map.Entry<K, V>>>` we have no way to represent the list as that type without a copy of `.entrySet()`.

Simply put: neither `ProtoAdapter<Map<K, V>>` nor `ProtoAdapter<Map.Entry<K, V>>` ends up being ergonomic for all three methods, but I think `Map` is more ergonomic (and there's parity with `List`).

wire-runtime/src/main/java/com/squareup/wire/ProtoAdapter.java
JakeWharton: Yep. You either mark the field as redacted and get empty or we ask the values to redact themselves.

wire-runtime/src/main/java/com/squareup/wire/internal/Internal.java
JakeWharton: This is only called from generated code where we're assured it's a mutable map.

JakeWharton:
Dropped the immutable/mutable-on-write map implementations for now. Updated and responded to everything else.

wire-runtime/src/main/java/com/squareup/wire/RuntimeMessageAdapter.java
JakeWharton: In this case `singleAdapter()` delegates to call `ProtoAdatper.get` of `WireField.adapter` which in `Map`'s case is the adapter for the value type only.

I can look to make it more logical. The runtime adapter and field binding classes are a bit of a mess right now.

wire-runtime/src/main/java/com/squareup/wire/ProtoAdapter.java
JakeWharton: I can add on `newMapAdapter` but it's also true for `asRepeated()` and `List`. Because of how adapters work we can only ever decode one element at a time. In the case of `List` we delegate directly to the item adapter and skip the singleton list. Unfortunately because of the complex type here we can't do that.

JakeWharton: Here's the implementation for `asRepeated`:
```java
@Override public List<E> decode(ProtoReader reader) throws IOException {
  E value = ProtoAdapter.this.decode(reader);
  return Collections.singletonList(value);
}
```

wire-gson-support/src/main/java/com/squareup/wire/MessageTypeAdapter.java
```diff
@@ -110,51 +111,63 @@ private void emitUint64(Long value, JsonWriter out) throws IOException {
...
+      return result;
     }
-  }
 
-  private Object readJson(Type valueType, JsonElement element) {
-    return gson.fromJson(element, valueType);
-  }
+    if (fieldBinding.isMap()) {
+      if (element.isJsonNull()) {
+        return Collections.emptyList();
```
swankjesse: emptyMap ?

wire-runtime/src/main/java/com/squareup/wire/ProtoAdapter.java
```diff
@@ -66,6 +67,18 @@ public ProtoAdapter(FieldEncoding fieldEncoding, Class<?> javaType) {
     return new RuntimeEnumAdapter<>(type);
   }
 
+  /**
+   * Creates a new proto adapter for a map using {@code keyAdapter} and {@code valueAdapter}.
+   * <p>
```
swankjesse: For what it’s worth, putting `<p>` on its own line like this makes CMD+ALT+L work less well.

swankjesse:
LGTM. Just one issue with emptyList().

wire-gson-support/src/main/java/com/squareup/wire/MessageTypeAdapter.java
JakeWharton: Good catch. Adding tests.

wire-runtime/src/main/java/com/squareup/wire/ProtoAdapter.java
JakeWharton: Hmm it doesn't have any effect for me either way. It stays in the same location and the rendered output looks the same.

PR #599, opened 2016-06-15
Generate Parcelable.Creator for enums on Android.

### LINKED ISSUE #598: Enums should use a runtime Creator when generating for Android
Similar to what we do for messages now. Currently `Serializable` has to be used which is gross city.

JakeWharton:
Closes #598.


JakeWharton:
This encodes the value as a varint32 encoded to bytes. We should just use an integer.

PR #601, opened 2016-06-15
Add concurrent java generation.

matthewdu:
@swankjesse


wire-compiler/src/test/java/com/squareup/wire/StringWireLogger.java
```diff
@@ -12,18 +12,22 @@
...
+      buffer.append(outputPath);
+      buffer.append(" ");
+      buffer.append(javaFile.packageName);
+      buffer.append(".");
+      buffer.append(javaFile.typeSpec.name);
+      buffer.append('\n');
+    }
   }
 
   @Override public void info(String message) {
```
swankjesse: just change this method to be `synchronized` ?

swankjesse:
Nice. What’s the runtime improvement?

matthewdu:
2.5s down to 1.5s

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -225,39 +232,73 @@ void compile() throws IOException {
...
         if (namedFilesOnly || protoFile.location().path().equals(DESCRIPTOR_PROTO)) continue;
       }
+      types.addAll(protoFile.types());
+    }
 
-      for (Type type : protoFile.types()) {
-        TypeSpec typeSpec = javaGenerator.generateType(type);
-        ClassName javaTypeName = (ClassName) javaGenerator.typeName(type.type());
-        writeJavaFile(javaTypeName, typeSpec, type.location());
+    List<Future<Void>> futures = new ArrayList<>(8);
```
JakeWharton: nit: use `MAX_WRITE_CONCURRENCY` for argument here

JakeWharton: Or just don't bother because the default size is 10 or 16 anyway

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -225,39 +232,73 @@ void compile() throws IOException {
...
+    for (int i = 0; i < MAX_WRITE_CONCURRENCY; ++i) {
+      futures.add(i, executor.submit(new JavaFileWriter(javaGenerator, types)));
+    }
+    executor.shutdown();
+
+    try {
+      for (Future<Void> future : futures) {
+        future.get();
```
JakeWharton: Any reason to prefer this over `executor.awaitTermination(1, MINUTES);`?

matthewdu: `get()` will throw exceptions thrown by the thread.

JakeWharton: Got it. Futures are such boilerplate-ridden messes in pre-Java 8... we have Guava and I was looking into what it affords but perhaps we just want to enforce JDK 8 in the compiler.

Guava has a couple helpers on `Futures` that might make this simpler, but it also requires the use of `ListenableFuture`.

With Java 8 this looks like it could be done trivially with `.parallelStream()` , although I'm less certain how that deals with exceptions. There's also `CompleteableFuture` in that version which is like `ListenableFuture`++.

swankjesse:
Nice!

PR #603, opened 2016-06-21
pojo be gone

matthewdu:
DO NOT MERGE

### Add `--proto_adapter` flag
`--proto_adapter` takes 3 comma separated arguments: `proto message name`, `java class` to be mapped to, and an `adapter` for encoding/decoding

ie `
--proto_adapter=squareup.dinosaur.common.Tyrannosaurus, com.squareup.wire.Tyrannosaurus, com.squareup.wire.DinosaurAdapter#TYRANNOSAURUS_ADAPTER`

A helper class `[proto_message_name]ProtoFields` is generated instead of the normal pojo class. This class has information about the tags and corresponding `ProtoAdapter`s for each proto field.

@swankjesse


wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -154,24 +155,40 @@
...
+        emitAndroid, emitCompact);
   }
 
   public JavaGenerator withCompact(boolean compactGeneration) {
-    return new JavaGenerator(schema, nameToJavaName, emitAndroid, compactGeneration);
+    return new JavaGenerator(schema, nameToWireJavaName, nameToJavaName, protoTypeToAdapter,
+        emitAndroid, compactGeneration);
+  }
+
+  public JavaGenerator withCustomProtoAdapter(Map<ProtoType, TypeName> nameToJavaName,
```
JakeWharton: * Name should be plural since you're accepting multiple
* Should we use Guava's `Table` here? @swankjesse might have opinions.
* I think we probably want to use `ClassName` here just so no one does anything stupid and passes like a generic or primitive or something.

wire-runtime/src/main/java/com/squareup/wire/ProtoField.java
```diff
@@ -0,0 +1,12 @@
+package com.squareup.wire;
+
+public class ProtoField<E> {
+  public final int tag;
+  public final ProtoAdapter<E> protoAdapter;
+  private E defaultValue;
```
JakeWharton: unused?

wire-runtime/src/main/java/com/squareup/wire/ProtoField.java
```diff
@@ -0,0 +1,12 @@
+package com.squareup.wire;
+
+public class ProtoField<E> {
```
JakeWharton: nit: `final`

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -94,6 +98,7 @@
   public static final String NAMED_FILES_ONLY = "--named_files_only";
   public static final String ANDROID = "--android";
   public static final String COMPACT = "--compact";
+  public static final String CUSTOM_PROTO_ADAPTER = "--custom_proto_adapter=";
```
swankjesse: hmm . . . maybe omit `custom`  and just call this `--proto_adapter` ? (All options are customizations, right?)

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -111,10 +116,13 @@
...
   final boolean emitAndroid;
   final boolean emitCompact;
+  final Map<String, String> protoPathToClassPath;
+  final Map<String, String> protoPathToAdapterClassPath;
 
   WireCompiler(FileSystem fs, WireLogger log, List<String> protoPaths, String javaOut,
       List<String> sourceFileNames, IdentifierSet identifierSet, boolean dryRun,
-      boolean namedFilesOnly, boolean emitAndroid, boolean emitCompact) {
+      boolean namedFilesOnly, boolean emitAndroid, boolean emitCompact,
+      Map<String, String> protoPathToClassPath, Map<String, String> protoPathToAdapterClassPath) {
```
swankjesse: Classpath is a special word in Java . . . I assume that’s not what this is

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -176,6 +188,13 @@ static WireCompiler forArgs(
         for (String identifier : splitArg(arg, EXCLUDES_FLAG.length())) {
           identifierSetBuilder.exclude(identifier);
         }
+      } else if (arg.startsWith(CUSTOM_PROTO_ADAPTER)) {
+        List<String> customProtoAdapterArgs = splitArg(arg, CUSTOM_PROTO_ADAPTER.length());
+        String protoPath = customProtoAdapterArgs.get(0).trim();
```
swankjesse: what does _path_ mean here? Is it the name? Typically _path_ is a directory, or a list of directories, but this is just a fully qualified name.

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -176,6 +188,13 @@ static WireCompiler forArgs(
...
+        String classPath = customProtoAdapterArgs.get(1).trim();
```
swankjesse: classname ?

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -176,6 +188,13 @@ static WireCompiler forArgs(
...
+        String adapterClassPath = customProtoAdapterArgs.get(2).trim();
```
swankjesse: I don’t understand this name. Is this the field?

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -176,6 +188,13 @@ static WireCompiler forArgs(
         for (String identifier : splitArg(arg, EXCLUDES_FLAG.length())) {
           identifierSetBuilder.exclude(identifier);
         }
+      } else if (arg.startsWith(CUSTOM_PROTO_ADAPTER)) {
+        List<String> customProtoAdapterArgs = splitArg(arg, CUSTOM_PROTO_ADAPTER.length());
```
swankjesse: this probably should confirm that it gets exactly 3 args back

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -125,6 +133,8 @@
     this.namedFilesOnly = namedFilesOnly;
     this.emitAndroid = emitAndroid;
     this.emitCompact = emitCompact;
+    this.protoPathToClassPath = protoPathToClassPath;
```
swankjesse: Please make a defensive copy!

swankjesse: Even better, declare it as an `ImmutableMap` and use `ImmutableMap.copyOf()`

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -228,9 +248,23 @@ void compile() throws IOException {
       }
     }
 
+    LinkedHashMap<ProtoType, TypeName> protoTypeToJavaType = new LinkedHashMap<>();
+    LinkedHashMap<ProtoType, TypeName> protoTypeToAdapterClass = new LinkedHashMap<>();
```
swankjesse: Is this a class or a field? I was thinking this would be a field constant.

swankjesse: You could save some work and change `WireCompiler` to use `Map<ProtoType, TypeName>` immediately instead of starting with `Map<String, String>` and then later converting.

swankjesse: Can you change the left hand side to be a `Map` instead of a `LinkedHashMap` ? `LinkedHashMap` is an implementation detail!

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -228,9 +248,23 @@ void compile() throws IOException {
...
+    for (Map.Entry<String, String> entry : protoPathToAdapterClassPath.entrySet()) {
+      ProtoType protoType = ProtoType.get(entry.getKey());
+      ClassName className = ClassName.bestGuess(entry.getValue());
+      protoTypeToAdapterClass.put(protoType, className);
+    }
+
     JavaGenerator javaGenerator = JavaGenerator.get(schema)
-        .withAndroid(emitAndroid)
-        .withCompact(emitCompact);
+            .withCustomProtoAdapter(protoTypeToJavaType, protoTypeToAdapterClass)
```
swankjesse: nit: +4 on wrapped lines. Fix your IntelliJ to do this for you!

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -277,7 +311,7 @@ public Void call() throws IOException {
         }
 
         TypeSpec typeSpec = javaGenerator.generateType(type);
-        ClassName javaTypeName = (ClassName) javaGenerator.typeName(type.type());
+        ClassName javaTypeName = (ClassName) javaGenerator.wireTypeName(type.type());
```
swankjesse: why this change?

swankjesse: ... this is a backwards-incompatible change to a public API!

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -154,24 +155,40 @@
   });
 
   private final Schema schema;
+  private final ImmutableMap<ProtoType, TypeName> nameToWireJavaName;
```
swankjesse: Do we need both this and the next thing? I hope we don’t!

swankjesse: . . . instead you could probably change the code to prefer the configured name if it exists

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -154,24 +155,40 @@
...
   private final ImmutableMap<ProtoType, TypeName> nameToJavaName;
+  private final ImmutableMap<ProtoType, TypeName> protoTypeToAdapter;
   private final boolean emitAndroid;
   private final boolean emitCompact;
 
-  private JavaGenerator(Schema schema, ImmutableMap<ProtoType, TypeName> nameToJavaName,
+  private JavaGenerator(Schema schema, ImmutableMap<ProtoType,
```
swankjesse: nit: if you can avoid it, never wrap lines in the middle of a type name like `ImmutableMap<Prototype,`
`TypeName>`. It makes the code difficult to understand.

Even better, never wrap within a single parameter.

Even this is awkward: `ImmutableMap<ProtoType, TypeName>`
`nameToJavaName`.

swankjesse: Some guidance: https://google.github.io/styleguide/javaguide.html#s4.5-line-wrapping

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -154,24 +155,40 @@
...
 
   public JavaGenerator withCompact(boolean compactGeneration) {
-    return new JavaGenerator(schema, nameToJavaName, emitAndroid, compactGeneration);
+    return new JavaGenerator(schema, nameToWireJavaName, nameToJavaName, protoTypeToAdapter,
+        emitAndroid, compactGeneration);
+  }
+
+  public JavaGenerator withCustomProtoAdapter(Map<ProtoType, TypeName> nameToJavaName,
+      Map<ProtoType, TypeName> protoTypeToAdapter) {
+    Map<ProtoType, TypeName> clone = new LinkedHashMap<>(this.nameToJavaName);
```
swankjesse: dig.

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -218,6 +236,12 @@ public TypeName typeName(ProtoType protoType) {
     return candidate;
   }
 
+  public TypeName wireTypeName(ProtoType protoType) {
```
swankjesse: not sure how I feel about this. Do we really need it? I’d prefer if `typeName` just worked and the caller didn’t need to decide.

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -231,7 +255,11 @@ private CodeBlock singleAdapterFor(ProtoType type) {
     } else if (type.isMap()) {
       throw new IllegalArgumentException("Cannot create single adapter for map type " + type);
     } else {
-      result.add("$T.ADAPTER", typeName(type));
+      if (protoTypeToAdapter.containsKey(type)) {
```
swankjesse: Here’s an example where making the caller choose is awkward

swankjesse: By the way, this is a bad antipattern because you’re doing 2 map lookups. Much more efficient to do a single map lookup and check for null, like so:

```
ClassName className = protoTypeToAdapter.get(type);
if (className != null) {
  result.add("$T.ADAPTER", className);
} else {
...
```
```

File: wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java [hunk 19]
```diff
@@ -301,6 +329,9 @@ static String sanitizeJavadoc(String documentation) {
   /** Returns the generated code for {@code type}, which may be a top-level or a nested type. */
   public TypeSpec generateType(Type type) {
     if (type instanceof MessageType) {
+      if (protoTypeToAdapter.containsKey(type.type())) {
+        return generateProtoFields((MessageType) type);
+      }
       //noinspection deprecation: Only deprecated as a public API.
       return generateMessage((MessageType) type);
```
swankjesse: should this be in an else clause?

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -825,7 +892,7 @@ private TypeName fieldType(Field field) {
           typeName(type.valueType()));
     }
     TypeName messageType = typeName(type);
-    return field.isRepeated() ? listOf(messageType) : messageType;
+    return field.isRepeated() || field.isPacked() ? listOf(messageType) : messageType;
```
swankjesse: I’m curious . . . do we ever have fields that are packed but not repeated? That’s a horrible bug that we should fix. I don’t like this change.

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -874,9 +941,13 @@ private AnnotationSpec wireFieldAnnotation(Field field) {
   }
 
   private String adapterString(ProtoType type) {
-    return type.isScalar()
-          ? ProtoAdapter.class.getName() + '#' + type.toString().toUpperCase(Locale.US)
-          : reflectionName((ClassName) typeName(type)) + "#ADAPTER";
+    if (type.isScalar()) {
```
swankjesse: Dig.

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -874,9 +941,13 @@ private AnnotationSpec wireFieldAnnotation(Field field) {
...
+      return ProtoAdapter.class.getName() + '#' + type.toString().toUpperCase(Locale.US);
+    } else if (protoTypeToAdapter.containsKey(type)) {
```
swankjesse: Can you change this so that the `#ADAPTER` is in the map too? Otherwise I can’t do this:

```
public static final class FranklinWireAdapters {
  public static final ProtoAdapter<Money> MONEY_ADAPTER = ...
  public static final ProtoAdapter<Date> DATE_ADAPTER = ...
  public static final ProtoAdapter<Cell> CELL_ADAPTER = ...

}
```

wire-java-generator/src/test/java/com/squareup/wire/java/JavaGeneratorTest.java
```diff
@@ -37,4 +41,36 @@
     String expected = "/* comment inside comment. &#42;/";
     assertThat(JavaGenerator.sanitizeJavadoc(input)).isEqualTo(expected);
   }
+
+  @Test public void generateProtoFields() {
```
swankjesse: Love this

wire-java-generator/src/test/java/com/squareup/wire/java/JavaGeneratorTest.java
```diff
@@ -37,4 +41,36 @@
...
+            + "message Bar {\n"
+            + "}\n")
+        .build();
+
+    MessageType message = (MessageType) schema.getType("Message");
+
+    JavaGenerator javaGenerator = JavaGenerator.get(schema);
+    TypeSpec typeSpec = javaGenerator.generateProtoFields(message);
+
+    assertThat(typeSpec.toString()).isEqualTo(""
```
swankjesse: There’s a slightly better way to do the `toString()` thing that gets you imports. It’ll make this way more readable.

You probably want this method:
https://github.com/square/javapoet/blob/master/src/test/java/com/squareup/javapoet/TypeSpecTest.java#L1511

wire-runtime/src/main/java/com/squareup/wire/ProtoField.java
```diff
@@ -0,0 +1,12 @@
+package com.squareup.wire;
```
swankjesse: needs a copyright statement

wire-runtime/src/main/java/com/squareup/wire/ProtoField.java
swankjesse: final

swankjesse: Maybe a sentence or two of documentation explaining what this models

wire-runtime/src/main/java/com/squareup/wire/ProtoField.java
swankjesse: remove this til we actually use it?

swankjesse:
# Really nice change.

Though I have a bunch of comments, they’re all pretty shallow. You built the right stuff in the right places despite lots of danger. Nice work.

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -600,6 +600,17 @@ public TypeSpec generateProtoFields(MessageType type) {
     TypeSpec.Builder builder = TypeSpec.classBuilder(javaType.simpleName() + "ProtoFields");
     builder.addModifiers(PUBLIC, FINAL);
 
+    for (Type nestedType : type.nestedTypes()) {
+      if (!protoTypeToAdapter.containsKey(nestedType.type())) {
+        throw new IllegalArgumentException("Missing custom proto adapter for "
```
swankjesse: oooh interesting case.

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -111,10 +118,14 @@
...
   final boolean emitCompact;
+  final ImmutableMap<ProtoType, TypeName> protoTypeToJavaClassName;
+  final ImmutableMap<ProtoType, Map.Entry<ClassName, String>> protoTypeToAdapter;
 
   WireCompiler(FileSystem fs, WireLogger log, List<String> protoPaths, String javaOut,
       List<String> sourceFileNames, IdentifierSet identifierSet, boolean dryRun,
-      boolean namedFilesOnly, boolean emitAndroid, boolean emitCompact) {
+      boolean namedFilesOnly, boolean emitAndroid, boolean emitCompact,
+      Map<ProtoType, TypeName> protoTypeToJavaClassName,
+      Map<ProtoType, Map.Entry<ClassName, String>> protoTypeToAdapter) {
```
swankjesse: Are you using `Map.Entry` as a pair? That’s a bit weird. I think it’s probably worth the effort to define a little class `AdapterMember` or something that contains the two fields, ClassName and String.

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -176,6 +191,21 @@ static WireCompiler forArgs(
         for (String identifier : splitArg(arg, EXCLUDES_FLAG.length())) {
           identifierSetBuilder.exclude(identifier);
         }
+      } else if (arg.startsWith(PROTO_ADAPTER)) {
+        List<String> customProtoAdapterArgs = splitArg(arg, PROTO_ADAPTER.length());
+        if (customProtoAdapterArgs.size() != 3) {
+          throw new IllegalArgumentException("Wrong number of arguments in '" + arg + "'.");
```
swankjesse: dig.

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
swankjesse: Alternately, just leave it as a `String` ?

swankjesse: As is this is gross ’cause it’s using Map.Entry to do something that Map.Entry isn’t intended for.

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -277,7 +309,11 @@ public Void call() throws IOException {
         }
 
         TypeSpec typeSpec = javaGenerator.generateType(type);
-        ClassName javaTypeName = (ClassName) javaGenerator.typeName(type.type());
+        TypeName typeName = javaGenerator.protoFieldsTypeName(type.type());
```
swankjesse: Hang on, why is this still necessary? I was expecting javaGenerator.typeName() to always do the right thing.

swankjesse: Ahh, I get it. It’s because `generateType` does a different thing in the two different cases. Ahh.

swankjesse: That works for me.

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -154,24 +155,47 @@
...
 
   public JavaGenerator withCompact(boolean compactGeneration) {
-    return new JavaGenerator(schema, nameToJavaName, emitAndroid, compactGeneration);
+    return new JavaGenerator(schema, nameToProtoFieldsJavaName, nameToJavaName, protoTypeToAdapter,
+        emitAndroid, compactGeneration);
+  }
+
+  public JavaGenerator withCustomProtoAdapter(ImmutableMap<ProtoType, TypeName> nameToJavaName,
+      ImmutableMap<ProtoType, Map.Entry<ClassName, String>> protoTypeToAdapter) {
+    Map<ProtoType, TypeName> clone = new LinkedHashMap<>(this.nameToJavaName);
```
swankjesse: nit: the word `clone` is a bit distracting here. There’s a method called `Object.clone()` that does a specific thing, and this isn’t it. Maybe you instead want to just shadow the target? Ie. name this `nameToJavaName` ?

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -154,24 +155,47 @@
...
+        emitAndroid, emitCompact);
   }
 
   public JavaGenerator withCompact(boolean compactGeneration) {
-    return new JavaGenerator(schema, nameToJavaName, emitAndroid, compactGeneration);
+    return new JavaGenerator(schema, nameToProtoFieldsJavaName, nameToJavaName, protoTypeToAdapter,
+        emitAndroid, compactGeneration);
+  }
+
+  public JavaGenerator withCustomProtoAdapter(ImmutableMap<ProtoType, TypeName> nameToJavaName,
```
swankjesse: Should this method explode if `this.nameToProtoFieldsJavaName` is not empty?

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -154,24 +155,47 @@
...
+      ImmutableMap<ProtoType, Map.Entry<ClassName, String>> protoTypeToAdapter) {
+    Map<ProtoType, TypeName> clone = new LinkedHashMap<>(this.nameToJavaName);
+    Map<ProtoType, TypeName> nameToProtoFieldsJavaName = new LinkedHashMap<>();
+
+    for (Map.Entry<ProtoType, TypeName> entry : nameToJavaName.entrySet()) {
+      TypeName protoFieldsJavaName = clone.put(entry.getKey(), entry.getValue());
```
swankjesse: Right now this is going to have a name like `com.squareup.Point` and we’re going to use it to generate names like `com.squareup.PointProtoFields`. Should the value in the map be `Point` or `PointProtoFields` ? I suspect that might be a bit better?

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -218,6 +244,11 @@ public TypeName typeName(ProtoType protoType) {
     return candidate;
   }
 
+  public TypeName protoFieldsTypeName(ProtoType protoType) {
+    TypeName candidate = nameToProtoFieldsJavaName.get(protoType);
```
swankjesse: you can probably omit the variable declaration and just return the immediate value?

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -218,6 +244,11 @@ public TypeName typeName(ProtoType protoType) {
     return candidate;
   }
 
+  public TypeName protoFieldsTypeName(ProtoType protoType) {
```
swankjesse: Add a short doc explaining what this returns and what it means if the result is null?

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -231,7 +262,12 @@ private CodeBlock singleAdapterFor(ProtoType type) {
     } else if (type.isMap()) {
       throw new IllegalArgumentException("Cannot create single adapter for map type " + type);
     } else {
-      result.add("$T.ADAPTER", typeName(type));
+      Map.Entry<ClassName, String> adapter = protoTypeToAdapter.get(type);
+      if (adapter != null) {
+        result.add("$T.$L", adapter.getKey(), adapter.getValue());
```
swankjesse: neato

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -553,6 +592,61 @@ private TypeSpec generateEnclosingType(EnclosingType type) {
     return builder.build();
   }
 
+  /** Returns generated code that maps field names to field tags. */
+  public TypeSpec generateProtoFields(MessageType type) {
+    NameAllocator nameAllocator = nameAllocators.getUnchecked(type);
+
+    ClassName javaType = (ClassName) protoFieldsTypeName(type.type());
+    TypeSpec.Builder builder = TypeSpec.classBuilder(javaType.simpleName() + "ProtoFields");
```
swankjesse: .  . . . yeah yeah, what if _this_ name is the name that you put in the map?

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -301,6 +337,9 @@ static String sanitizeJavadoc(String documentation) {
   /** Returns the generated code for {@code type}, which may be a top-level or a nested type. */
   public TypeSpec generateType(Type type) {
     if (type instanceof MessageType) {
+      if (protoTypeToAdapter.containsKey(type.type())) {
+        return generateProtoFields((MessageType) type);
```
swankjesse: it’s lame that you made the map lookup on line 340, and then the very first thing the implementation of this method must do is also do the map lookup. Maybe instead this method should take the ClassName that it’s going to generate?

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -553,6 +592,61 @@ private TypeSpec generateEnclosingType(EnclosingType type) {
...
+    for (Field field : type.fieldsAndOneOfFields()) {
+      String fieldName = nameAllocator.get(field);
+
+      TypeName typeName =
+          ParameterizedTypeName.get(ClassName.get(ProtoField.class), fieldType(field));
+      FieldSpec.Builder fieldBuilder =
+          FieldSpec.builder(typeName, fieldName, PUBLIC, STATIC, FINAL);
+
+      CodeBlock adapter;
+      if (field.type().isMap()) {
```
swankjesse: can you make the existing `adapterFor()` method smarter and handle this? (I’m unsure)

swankjesse: As-is seems like unnecessary duplication

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -553,6 +592,61 @@ private TypeSpec generateEnclosingType(EnclosingType type) {
...
+            singleAdapterFor(field.type().valueType()));
+      } else if (field.isPacked()) {
+        adapter = singleAdapterFor(field).toBuilder().add(".asPacked()").build();
+      } else if (field.isRepeated()) {
+        adapter = singleAdapterFor(field).toBuilder().add(".asRepeated()").build();
+      } else {
+        adapter = singleAdapterFor(field);
+      }
+
+      CodeBlock defaultVal = CodeBlock.of("null");
```
swankjesse: nit: `defaultVal` –> `defaultValue`

matthewdu: `defaultValue` is already defined else where

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -553,6 +592,61 @@ private TypeSpec generateEnclosingType(EnclosingType type) {
...
+      if ((field.type().isScalar() || isEnum(field.type()))
+          && !field.isRepeated()
+          && !field.isPacked()) {
+        defaultVal = defaultValue(field);
+      }
+
+      fieldBuilder.initializer("new $T($L, $L, $L)", ClassName.get(ProtoField.class), field.tag(),
```
swankjesse: sweet

wire-java-generator/src/test/java/com/squareup/wire/java/JavaGeneratorTest.java
```diff
@@ -37,4 +49,60 @@
...
+    Map.Entry<ClassName, String> adapter =
+        new AbstractMap.SimpleImmutableEntry<>(className, "ADAPTER");
+    nameToAdapter.put(protoType, adapter);
+
+    JavaGenerator javaGenerator = JavaGenerator.get(schema)
+        .withCustomProtoAdapter(ImmutableMap.copyOf(nameToJavaName),
+            ImmutableMap.copyOf(nameToAdapter));
+    TypeSpec typeSpec = javaGenerator.generateProtoFields(message);
+
+    assertThat(JavaFile.builder("com.squareup.message", typeSpec).build().toString()).isEqualTo(""
```
swankjesse: this is wondeful

swankjesse:
# LGTM

Please address the feedback above, rebase and squash.

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
matthewdu: Another class is probably better. String is ugly for this case since we'd have to parse the string everytime we wanted to use it.

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
matthewdu: Yup, calling it twice with would cause weird output.

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
matthewdu: I can try to change this, but I have to pass in the `MessageType type`.

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -111,10 +118,14 @@
   final boolean namedFilesOnly;
   final boolean emitAndroid;
   final boolean emitCompact;
+  final ImmutableMap<ProtoType, TypeName> protoTypeToJavaClassName;
+  final ImmutableMap<ProtoType, AdapterEntry> protoTypeToAdapter;
```
swankjesse: better!

swankjesse: Though why _entry_ ?

swankjesse: nit: protoTypeToAdapter –> protoTypeToAdapterEntry ? (otherwise the name is a lie?)

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -176,6 +191,17 @@ static WireCompiler forArgs(
...
+        List<String> customProtoAdapterArgs = splitArg(arg, PROTO_ADAPTER.length());
+        if (customProtoAdapterArgs.size() != 3) {
+          throw new IllegalArgumentException("Wrong number of arguments in '" + arg + "'.");
+        }
+
+        ProtoType protoType = ProtoType.get(customProtoAdapterArgs.get(0).trim());
+        ClassName javaClassName = ClassName.bestGuess(customProtoAdapterArgs.get(1).trim());
+
+        protoTypeToJavaClass.put(protoType, javaClassName);
+        protoTypeToAdapter.put(protoType, new AdapterEntry(customProtoAdapterArgs.get(2).trim()));
```
swankjesse: nice

wire-java-generator/src/main/java/com/squareup/wire/java/AdapterEntry.java
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire.java;
+
+import com.squareup.javapoet.ClassName;
+
+public class AdapterEntry {
```
swankjesse: nit: final

wire-java-generator/src/main/java/com/squareup/wire/java/AdapterEntry.java
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
```
swankjesse: nope, 2016

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
swankjesse: +1

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
swankjesse: Protip– never use Map.containsKey() followed by Map.get(). It’s 2x as much work

wire-runtime/src/main/java/com/squareup/wire/ProtoField.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
```
swankjesse: 2016

swankjesse:
Merged manually!

Two small changes I made before merging:
 - renamed `Adapter` to `AdapterConstant`. 
 - changed types from `ImmutableMap<ProtoType, TypeName>` to `ImmutableMap<ProtoType, ClassName>` to avoid casting

Thanks @matthewdu !

PR #609, opened 2016-07-04
Use NameAllocator in Message.Builder.build()

rjcase:
Use the NameAllocator when inserting string literals, don't use the field
name directly.


JakeWharton:
Lgtm

swankjesse:
Thanks neighbour!

PR #617, opened 2016-07-27
Begin to parse build.wire configuration files.

swankjesse:
This extracts the general-purpose parts of the parser out of ProtoParser
and into a standalone file, Parser.java.

It then uses that file to build a new parser for build.wire files, which
currently only support type configs. The type configs do not yet support
any deep validation; that will come in a follow up that will hook these
up to SchemaLoader.

### DIFF
--- a/wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ProtoParser.java
+++ b/wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ProtoParser.java
@@ -29,13 +29,13 @@ import java.util.Map;
 
 /** Basic parser for {@code .proto} schema declarations. */
 public final class ProtoParser {
+  private final SyntaxReader reader;
+
   /** Parse a named {@code .proto} schema. */
   public static ProtoFileElement parse(Location location, String data) {
     return new ProtoParser(location, data.toCharArray()).readProtoFile();
   }
 
-  private final Location location;
-  private final char[] data;
   private final ProtoFileElement.Builder fileBuilder;
   private final ImmutableList.Builder<String> publicImports = ImmutableList.builder();
   private final ImmutableList.Builder<String> imports = ImmutableList.builder();
@@ -46,14 +46,10 @@ public final class ProtoParser {
 
   /** The number of declarations defined in the current file. */
   private int declarationCount = 0;
+
   /** The syntax of the file, or null if none is defined. */
   private ProtoFile.Syntax syntax;
-  /** Our cursor within the document. {@code data[pos]} is the next character to be read. */
-  private int pos;
-  /** The number of newline characters encountered thus far. */
-  private int line;
-  /** The index of the most recent newline character. */
-  private int lineStart;
+
   /** Output package name, or null if none yet encountered. */
   private String packageName;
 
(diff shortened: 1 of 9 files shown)


wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ProtoParser.java
```diff
@@ -29,13 +29,13 @@
 
 /** Basic parser for {@code .proto} schema declarations. */
 public final class ProtoParser {
+  private final Parser p;
```
JakeWharton: Oof these names. What about `SchemaReader`?

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ProtoParser.java
```diff
@@ -275,18 +245,17 @@ private ExtendElement readExtend(Location location, String documentation) {
...
   /** Reads a service declaration and returns it. */
   private ServiceElement readService(Location location, String documentation) {
-    String name = readName();
+    String name = p.readName();
     ServiceElement.Builder builder = ServiceElement.builder(location)
         .name(name)
         .documentation(documentation);
 
-    if (readChar() != '{') throw unexpected("expected '{'");
+    if (p.readChar() != '{') throw p.unexpected("expected '{'");
```
JakeWharton: This pattern happens 100s of times. Maybe make an `expectChar(char)`, `expectWord(String)`, etc. set of methods inside the parser.

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/BuildConfigFileElement.java
```diff
@@ -0,0 +1,98 @@
...
+ * <pre>   {@code
+ *
+ *   syntax = "wire2";
+ *   package squareup.dinosaurs;
+ *
+ *   import "squareup/geology/period.proto";
+ *
+ *   // Roar!
+ *   type squareup.dinosaurs.Dinosaur {
+ *     target com.squareup.dino.Dinosaur using com.squareup.dino.Dinosaurs#DINO_ADAPTER;
```
JakeWharton: Not sure how I feel about this syntax. The nesting seems arbitrary considering you can only have a single declaration inside the block.

JakeWharton:
lgtm

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ProtoParser.java
swankjesse: Done. Also fixed callers to use `reader` instead of `p`.

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ProtoParser.java
swankjesse: Added `require(char c)` which works out wonderfully. Good call. Required words isn’t a big win so I’m not gonna add it.

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/BuildConfigFileElement.java
swankjesse: Discussed offline. Extensibility FTW!

PR #619, opened 2016-07-27
ProfileLoader, a class that loads android.wire files.

swankjesse:
Or java.wire files, or whatever. It discovers files that are in the same
directories as the project .proto files, and in their parent directories.


szurbrigg:
LGTM!

PR #620, opened 2016-07-28
Validate that profiles are well-formed.

swankjesse:


szurbrigg:
LGTM

wire-java-generator/src/main/java/com/squareup/wire/java/ProfileLoader.java
```diff
@@ -21,51 +21,69 @@
...
-    }
-    return this;
+  public ProfileLoader(String name) {
+    this(FileSystems.getDefault(), name);
   }
 
-  public ProfileLoader addProtoLocation(Location location) {
-    protoLocations.add(location);
+  public ProfileLoader schema(Schema schema) {
+    checkState(this.schema == null);
```
JakeWharton: should be `checkNotNull` or perhaps `checkArgument`. Getting an ISE for passing `null` is weird.

swankjesse: this is an ISE for calling `schema()` 2x

JakeWharton: Ah, so it is!

PR #621, opened 2016-07-28
Honor profiles for Wire configuration.

swankjesse:
Previously we used command line parameters like --proto-adapter. With this
change we instead discover android.wire or java.wire files and use those
to customize the generated .java code.

This renames SchemaBuilder to RepoBuilder. The new name is preferred because
it now builds both Schema objects and also Profile objects. The class now
represents a source tree.

This renames wire-schema-tests to wire-tests. Much of the movement in this
PR is to handle this move.

This also moves some JavaGenerator tests from wire-java-generator to wire-tests.
The reason is that I want to reuse the new RepoBuilder type.


wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
```diff
@@ -257,7 +236,7 @@ void compile() throws IOException {
     }
 
     JavaGenerator javaGenerator = JavaGenerator.get(schema)
-        .withCustomProtoAdapter(protoTypeToJavaClassName, protoTypeToAdapterConstant)
+        .withProfile(profile)
         .withAndroid(emitAndroid)
```
JakeWharton: We should be able to omit this, right?

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -163,59 +163,30 @@
...
-    this.nameToAbstractAdapterName = nameToAbstractAdapterName;
-    this.nameToJavaName = nameToJavaName;
-    this.protoTypeToAdapterConstant = protoTypeToAdapterConstant;
+    this.nameToJavaName = ImmutableMap.copyOf(nameToJavaName);
+    this.profile = profile;
     this.emitAndroid = emitAndroid;
     this.emitCompact = emitCompact;
   }
 
   public JavaGenerator withAndroid(boolean emitAndroid) {
```
JakeWharton: `@Deprecated`?

wire-maven-plugin/src/main/java/com/squareup/wire/mojo/WireGenerateSourcesMojo.java
```diff
@@ -84,23 +78,17 @@
...
 
-      Map<ProtoType, ClassName> protoTypeToCustomJavaName = new LinkedHashMap<>();
-      Map<ProtoType, AdapterConstant> protoTypeToAdapterConstant = new LinkedHashMap<>();
-      for (ProtoAdapterParameter parameter : protoAdapters) {
-        protoTypeToCustomJavaName.put(parameter.protoType(), parameter.javaName());
-        protoTypeToAdapterConstant.put(parameter.protoType(), parameter.adapter());
-      }
-
       JavaGenerator javaGenerator = JavaGenerator.get(schema)
           .withAndroid(emitAndroid)
```
JakeWharton: Same

JakeWharton:
lgtm

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
swankjesse: sadly no . . . this change hasn’t yet promoted the `android` flag into the Profile. That’ll come soon.

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
swankjesse: I’ll roll it into the profile in the follow up. I’m tempted to make this a full breaking change. Maybe call it Wire 3.0 ?

wire-maven-plugin/src/main/java/com/squareup/wire/mojo/WireGenerateSourcesMojo.java
swankjesse: Yep.

PR #623, opened 2016-08-04
Fix a bug on messages generated into the default package.

swankjesse:


JakeWharton:
lgtm

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -1023,7 +1023,9 @@ private String adapterString(ProtoType type) {
   }
 
   private String reflectionName(ClassName className) {
```
JakeWharton: Seems like this could be upstreamed into a member of `ClassName`.

JakeWharton: `toString()` + `toCanonicalString()` or whatever

swankjesse: Yeah, we want that and something for members.

PR #624, opened 2016-08-04
Fix bug in Pruner where maps were not being marked as transitively re…

nayeemzen:
…achable

### DIFF
--- a/wire-schema/src/main/java/com/squareup/wire/schema/Pruner.java
+++ b/wire-schema/src/main/java/com/squareup/wire/schema/Pruner.java
@@ -175,8 +175,12 @@ final class Pruner {
   }
 
   private void mark(ProtoType type) {
-    // Map key type is always scalar. No need to mark it.
-    if (type.isMap()) type = type.valueType();
+    // Mark the map type as it's non-scalar and transitively reachable.
+    if (type.isMap()) {
+      marks.mark(type);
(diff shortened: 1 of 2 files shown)


wire-schema/src/main/java/com/squareup/wire/schema/Pruner.java
```diff
@@ -176,7 +176,11 @@ private void markReachable() {
 
   private void mark(ProtoType type) {
     // Map key type is always scalar. No need to mark it.
```
JakeWharton: Can you move this comment to be above the `type = type.valueType()` line?

nayeemzen: will do!

wire-schema/src/main/java/com/squareup/wire/schema/Pruner.java
```diff
@@ -175,8 +175,12 @@ private void markReachable() {
   }
 
   private void mark(ProtoType type) {
-    // Map key type is always scalar. No need to mark it.
-    if (type.isMap()) type = type.valueType();
+    // Mark the map type as it's non-scalar and transitively reachable.
+    if (type.isMap()) {
+      marks.mark(type);
```
swankjesse: I wonder if this is necessary . . .

swankjesse:
LGTM!

PR #625, opened 2016-08-05
Fix AbstractAdapters for field order and missing enums.

swankjesse:
Previously the order of parameters declared by fromProto() was not
consistent with the arguments in decode().

Also missing enums were attempting to dereference an inexistent
builder field.


wire-tests/src/test/java/com/squareup/wire/java/JavaGeneratorTest.java
```diff
@@ -74,15 +74,20 @@
             + "import \"foo.proto\";\n"
             + "message ProtoMessage {\n"
             + "  optional foo.proto.Foo field = 1;\n"
-            + "  map<string, foo.proto.Bar> bars = 2;\n"
             + "  repeated int32 numbers = 3;\n"
+            + "  optional foo.proto.CoinFlip coin_flip = 4;\n"
+            + "  map<string, foo.proto.Bar> bars = 2;\n"
```
swankjesse: deliberately use an order inconsistent with tag order

PR #631, opened 2016-09-09
Do not use reflection for enum in non-compact mode.

congt:
Making wire protos j2objc-friendly. 

CC @swankjesse

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/EnumAdapter.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/EnumAdapter.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2016 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire;
+
+import java.io.IOException;
+
+/**
+ * An abstract {@link ProtoAdapter} that converts values of an enum to and from integers.
+ */
+public abstract class EnumAdapter<E extends WireEnum> extends ProtoAdapter<E> {
+  protected EnumAdapter(Class<E> type) {
+    super(FieldEncoding.VARINT, type);
+  }
+
+  @Override public final int encodedSize(E value) {
+    return ProtoWriter.varint32Size(value.getValue());
+  }
+
+  @Override public final void encode(ProtoWriter writer, E value) throws IOException {
+    writer.writeVarint32(value.getValue());
+  }
+
+  @Override public final E decode(ProtoReader reader) throws IOException {
+    int value = reader.readVarint32();
+    E constant = fromValue(value);
+    if (constant == null) {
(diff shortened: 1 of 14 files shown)


wire-runtime/src/main/java/com/squareup/wire/EnumAdapter.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * An abstract {@link ProtoAdapter} that encodes values of an enum into integers.
+ */
+public abstract class EnumAdapter<E extends WireEnum> extends ProtoAdapter<E> {
+  protected EnumAdapter(Class<E> type) {
+    super(FieldEncoding.VARINT, type);
+  }
+
+  @Override public int encodedSize(E value) {
```
JakeWharton: `final`

wire-runtime/src/main/java/com/squareup/wire/EnumAdapter.java
```diff
@@ -0,0 +1,35 @@
...
+    return ProtoWriter.varint32Size(value.getValue());
+  }
+
+  @Override public void encode(ProtoWriter writer, E value) throws IOException {
```
JakeWharton: `final`

wire-runtime/src/test/proto-java/com/squareup/wire/protos/person/Person.java
```diff
@@ -219,6 +220,22 @@ public static PhoneType fromValue(int value) {
     public int getValue() {
       return value;
     }
+
+    private static final class ProtoAdapter_PhoneType extends EnumAdapter<PhoneType> {
```
danrice-square: Do we use this naming convention elsewhere? If not, I would prefer something like PhoneTypeEnumAdapter that's more human-friendly.

congt: Yes, it follows the naming convention of message adapter.

wire-runtime/src/test/proto-java/com/google/protobuf/FieldDescriptorProto.java
```diff
@@ -408,6 +409,22 @@ public static Type fromValue(int value) {
...
       return value;
     }
+
+    private static final class ProtoAdapter_Type extends EnumAdapter<Type> {
+      ProtoAdapter_Type() {
+        super(Type.class);
+      }
+
+      @Override
+      public Type decode(ProtoReader reader) throws IOException {
```
JakeWharton: How about pushing this implementation down into `EnumAdapter` and just adding `public abstract T fromValue(int)` which gets overridden. Then this implementation just becomes:
```java
@Override public Type fromValue(int value) {
  return Type.fromValue(value);
}
```

wire-runtime/src/test/proto-java/com/squareup/wire/protos/person/Person.java
JakeWharton: These classes are private and explicitly not for humans.

wire-runtime/src/test/proto-java/com/google/protobuf/FieldDescriptorProto.java
congt: Good point!

wire-runtime/src/test/proto-java/com/squareup/wire/protos/person/Person.java
danrice-square: That's fine, I just hate having to look at adapter code, but I guess that is the price we paid to get rid of reflection.

wire-runtime/src/test/proto-java/com/google/protobuf/FieldDescriptorProto.java
congt: Done

JakeWharton:
LGTM!

PR #636, opened 2016-10-24
Make unknown fields on Message.Builder transient.

swankjesse:
If we're using frameworks like Moshi on JsonWriter, it really hurts to
have these types included in the serialized form.


JakeWharton:
You're serializing a `Builder`?

swankjesse:
Deserializing. Makes it so that it Just Works on empty lists, etc.

PR #637, opened 2016-11-10
Upgrade to JavaPoet 1.8.0.

swankjesse:


swankjesse:
(this won’t build until JavaPoet 1.8 is cut)

PR #639, opened 2016-12-12
Clears the unknown fields of a message builder in Message#decode.

congt:
Buffer has a circularly-linked list of segments, which should be manually cleared to release resources. However, the Buffer for unknown fields in a message builder isn't cleared, which will cause memory leak in J2Objc transpiled code.

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/Message.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/Message.java
@@ -102,7 +102,18 @@ public abstract class Message<M extends Message<M, B>, B extends Message.Builder
    * Superclass for protocol buffer message builders.
    */
   public abstract static class Builder<T extends Message<T, B>, B extends Builder<T, B>> {
-    // Lazily-instantiated buffer and writer of this message's unknown fields.
+    /**
+     * Caches unknown fields as a {@link ByteString} when {@link #buildUnknownFields()} is called.
+     * When the caller adds an additional unknown field after that, it will be written to the new
+     * {@link #unknownFieldsBuffer} to ensure that all unknown fields are retained between calls to
+     * {@link #buildUnknownFields()}.
+     */
+    transient ByteString unknownFieldsByteString = ByteString.EMPTY;
+    /**
+     * {@link Buffer} of the message's unknown fields that is lazily instantiated between calls to
+     * {@link #buildUnknownFields()}. It's automatically cleared in {@link #buildUnknownFields()},
+     * and can also be manually cleared by calling {@link #clearUnknownFields()}.
+     */
     transient Buffer unknownFieldsBuffer;
     transient ProtoWriter unknownFieldsWriter;
 
@@ -111,10 +122,7 @@ public abstract class Message<M extends Message<M, B>, B extends Message.Builder
 
     public final Builder<T, B> addUnknownFields(ByteString unknownFields) {
       if (unknownFields.size() > 0) {
-        if (unknownFieldsWriter == null) {
-          unknownFieldsBuffer = new Buffer();
-          unknownFieldsWriter = new ProtoWriter(unknownFieldsBuffer);
-        }
+        prepareForNewUnknownFields();
         try {
           unknownFieldsWriter.writeBytes(unknownFields);
         } catch (IOException e) {
@@ -125,10 +133,7 @@ public abstract class Message<M extends Message<M, B>, B extends Message.Builder
     }
 
     public final Builder<T, B> addUnknownField(int tag, FieldEncoding fieldEncoding, Object value) {
-      if (unknownFieldsWriter == null) {
-        unknownFieldsBuffer = new Buffer();
-        unknownFieldsWriter = new ProtoWriter(unknownFieldsBuffer);
-      }
+      prepareForNewUnknownFields();
       try {
         ProtoAdapter<Object> protoAdapter = (ProtoAdapter<Object>) fieldEncoding.rawProtoAdapter();
         protoAdapter.encodeWithTag(unknownFieldsWriter, tag, value);
@@ -139,8 +144,12 @@ public abstract class Message<M extends Message<M, B>, B extends Message.Builder
     }
 
     public final Builder<T, B> clearUnknownFields() {
+      unknownFieldsByteString = ByteString.EMPTY;
+      if (unknownFieldsBuffer != null) {
+        unknownFieldsBuffer.clear();
+        unknownFieldsBuffer = null;
+      }
       unknownFieldsWriter = null;
-      unknownFieldsBuffer = null;
       return this;
     }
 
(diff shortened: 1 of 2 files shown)


swankjesse:
I’m not seeing the relationship between unknown fields (a `ByteString`) and the corresponding `Buffer` of unknown fields.

congt:
@swankjesse The generated `Builder#build()` calls `Builder#buildUnknownFields()` to build the unknownFields. Here is a simple example:
```
public final class Foo extends Message<Foo, Foo.Builder> {
  public NoFields(ByteString unknownFields) {
    super(ADAPTER, unknownFields);
  }
...
  public static final class Builder extends Message.Builder<Foo, Builder> {
    public Builder() {
    }

    @Override
    public Foo build() {
      return new Foo(super.buildUnknownFields());
    }
  }
...
}
```

where `Builder#buildUnknownFields` is:
```
public final ByteString buildUnknownFields() {
      return unknownFieldsBuffer != null
          ? unknownFieldsBuffer.clone().readByteString()
          : ByteString.EMPTY;
    }
```

The `unknownFieldsBuffer` field of the `Builder` is a `Buffer`which is never cleared, (i.e., `unknownFieldsBuffer.clone()` won't release it resources which makes sense because `Builder#build` may be called again. ).

swankjesse:
Ahh, got it. And with this, if `build()` is called a second time it’ll be without the unknown fields. Hmm.

What about this?

```
    ByteString unknownFieldsByteString = ByteString.EMPTY;
    Buffer unknownFieldsBuffer;
    ProtoWriter unknownFieldsWriter;


    public final Builder<T, B> addUnknownFields(ByteString unknownFields) {
      ...
        if (unknownFieldsWriter == null) {
          unknownFieldsBuffer = new Buffer()
              .writeByteString(unknownFieldsByteString);
          unknownFieldsWriter = new ProtoWriter(unknownFieldsBuffer);
        }
        ...
    }

    public final ByteString buildUnknownFields() {
      if (unknownFieldsBuffer != null) {
        unknownFieldsByteString = unknownFieldsBuffer.readByteString();
        unknownFieldsBuffer = null;
        unknownFieldsWriter = null;
      }
      return unknownFieldsByteString;
    }
```

swankjesse:
... basically caching the byte string for future calls. Overall the whole thing is a bit weird though. We’re writing Java code to defend against memory leaks which is weird!

congt:
Good idea! Updated according to your suggestion. We don't need to manually clear the unknown fields from the Builder after building a message. 

Agreed that it's a little weird, but given the current J2Objc status it's essential for transpiling Wire into Objc.

wire-runtime/src/test/java/com/squareup/wire/UnknownFieldsTest.java
```diff
@@ -89,4 +90,73 @@ public void testUnknownFields() throws IOException {
     assertThat(v2C.v2_f64).isEqualTo(new Long(98765L));
     assertThat(v2C.v2_rs).containsExactly("1", "2");
   }
+
+  @Test
+  public void testBuilderAddingAndClearingUnknownFields() throws IOException {
```
swankjesse: Overall I generally don’t like these tests that focus on exercise the general API and not any particular features.

Would you consider writing focused tests to cover specific inputs?

Maybe a single, smalltest-possible test to confirm the following:
 - repeated calls to `build()` retain unknown fields
 - unknown fields can be added between calls to build
 - clear unknown fields works

Ideally if a test fails it’s clear from the test name which specific behavior is broken

wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -149,12 +147,30 @@ protected Builder() {
      * this message has no unknown fields.
      */
     public final ByteString buildUnknownFields() {
-      return unknownFieldsBuffer != null
-          ? unknownFieldsBuffer.clone().readByteString()
-          : ByteString.EMPTY;
+      if (unknownFieldsBuffer != null) {
+        // Reads and caches the unknown fields from the buffer.
+        unknownFieldsByteString = unknownFieldsBuffer.readByteString();
+        unknownFieldsBuffer = null;
```
danrice-square: Why are you clearing `unknownFieldsBuffer` and `unknownFieldsBuffer`? What if the caller adds an additional field and calls this method a second time?

wire-runtime/src/main/java/com/squareup/wire/Message.java
```diff
@@ -102,7 +102,7 @@ public final void encode(OutputStream stream) throws IOException {
    * Superclass for protocol buffer message builders.
    */
   public abstract static class Builder<T extends Message<T, B>, B extends Builder<T, B>> {
-    // Lazily-instantiated buffer and writer of this message's unknown fields.
+    transient ByteString unknownFieldsByteString = ByteString.EMPTY;
```
danrice-square: Not sure why these are `transient`, it seems that at least `unknownFieldsBuffer` is part of the object state. But I can't really imagine anyone using Java serialization on this class.

wire-runtime/src/main/java/com/squareup/wire/Message.java
congt: The reason for clearing unknownFieldBuffer is to ensure that it has no unreleased segments in the buffer which will cause memory leak when translating wire into objc.

The unknown fields are cached in unknownFieldsByteString. Next time when the caller adds an additional field, the method prepareForNewUnknownFields () will be called, which will read unknownFieldsByteString into the new Buffer again. We won't lose any unknown field.

danrice-square: I think my preference in that case would be to have `unknownFieldsByteString` be the sole source of truth, and when an unknown field is added, create a new `Buffer`, write `unknownFieldsByteString` into it, and write the results back into `unknownFieldsByteString`. Although it's slower, the intent will be clearer since there are no cache fields needed.

swankjesse: I think that’s probably a bad fit in practice. It means handling unknown fields is O(N*N), which is a trivial DoS on our servers.

wire-runtime/src/main/java/com/squareup/wire/Message.java
swankjesse: It’s for naive use of Gson.

danrice-square:
Fair enough. How about some additional comments to clarify how the caching
scheme works.

wire-runtime/src/main/java/com/squareup/wire/Message.java
danrice-square: Fair enough. How about some additional comments to clarify how the caching scheme works.

danrice-square:
LGTM

wire-runtime/src/main/java/com/squareup/wire/Message.java
congt: Sure. Added JavaDoc to `unknownFieldsByteString` and `unknownFieldsBuffer`.

PR #642, opened 2017-02-14
Add error-prone compiler.

JakeWharton:


JakeWharton:
Oh of course Travis is using a version of JDK 8 from _two years_ ago.

Issue #651, opened 2017-05-30
inline star comment should allow multiple lines


knoguchi:
I'm trying to parse .proto files that contain inline multiple line star comment.

I got exception "trailing comment must be closed on the same line", however Protobuf syntax specification has no mentioning of that limitation.  In fact protoc can compile it without problem.

The exception was from this line.  
https://github.com/square/wire/blob/master/wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/SyntaxReader.java#L328

Can we allow inline multiple line star comments?

PR #652, opened 2017-05-30
allow inline multiline comments

knoguchi:
This PR addresses #651.
The parser should allow inline multiline start comment.


swankjesse:
Doesn’t harm anyone else and makes your life easier. Works for me.

PR #656, opened 2017-07-13
Generate an empty message for an enclosing type.

JakeWharton:
There's no tests for any of this model->element stuff so I didn't add any.


wire-schema/src/main/java/com/squareup/wire/schema/EnclosingType.java
```diff
@@ -85,4 +86,11 @@
     }
     return new EnclosingType(location, type, documentation, retainedNestedTypes);
   }
+
+  MessageElement toElement() {
```
swankjesse: document

JakeWharton: It's not public, and there's about 20 of these methods on every Schema type that also lack documentation...

PR #660, opened 2017-07-16
Add the beginnings of the proto pruner tool.

JakeWharton:
Still needs a bit of testing and the output schema from the AST likely isn't 100% perfect.

### DIFF
--- a/proto-pruner/src/main/java/com/squareup/wire/prune/ProtoPruner.kt
+++ b/proto-pruner/src/main/java/com/squareup/wire/prune/ProtoPruner.kt
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2017 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


proto-pruner/src/main/java/com/squareup/wire/prune/ProtoPruner.kt
```diff
@@ -0,0 +1,103 @@
...
+
+  companion object {
+    @JvmStatic fun main(vararg args: String) {
+      var outPath: String? = null
+      val inPaths = mutableListOf<String>()
+      val identifierSetBuilder = IdentifierSet.Builder()
+
+      for (arg in args) {
+        if (arg == "--help") {
+          println("""
```
swankjesse: nice

proto-pruner/src/main/java/com/squareup/wire/prune/ProtoPruner.kt
```diff
@@ -0,0 +1,103 @@
...
+        } else if (arg.startsWith("--in=")) {
+          inPaths.add(arg.substringAfter('='))
+        } else if (arg.startsWith("--includes=")) {
+          identifierSetBuilder.include(arg.substringAfter('=').split(',').map(String::trim))
+        } else if (arg.startsWith("--excludes=")) {
+          identifierSetBuilder.exclude(arg.substringAfter('=').split(',').map(String::trim))
+        } else if (arg.startsWith("--")) {
+          throw IllegalArgumentException("Unknown argument: $arg")
+        } else {
+          File(arg).forEachLine { line ->
```
swankjesse: this is a nice API

PR #662, opened 2017-07-22
Test ProfileFileElement.toSchema().

swankjesse:
Also start using @Nullable in wire-java-generator.


wire-java-generator/src/main/java/com/squareup/wire/java/internal/ProfileFileElement.java
```diff
@@ -69,7 +69,7 @@ public static Builder builder(Location location) {
   public final String toSchema() {
     StringBuilder builder = new StringBuilder();
     builder.append("// ").append(location()).append('\n');
-    builder.append("syntax \"wire2\";\n");
+    builder.append("syntax = \"wire2\";\n");
```
JakeWharton: Didn't I already change this? Did you rebase too far or something?

swankjesse:
That one fixed .proto files but not .wire files.

Issue #602, opened 2016-06-15
Extensible enums


swankjesse:
It’d be handy to configure a proto enum to _not_ map to a Java enum. Instead it would map to something that simulates a Java enum, but that allows new values to be defined at runtime.

```
enum MenuItem {
  option wire.extensible = true;
  TACO = 1;
  BURRITO = 2;
}
```

Which generates this:

```
public final class MenuItem {
  public static final MenuItem TACO = new MenuItem("TACO", 1);
  public static final MenuItem BURRITO = new MenuItem("BURRITO", 2);

  public static MenuItem get(int tag) {
    ...
  }
}
```

Most importantly, if we read a menu item with tag 3 from the network, we can represent that in the model. It won’t work in switch statements, but it will work everywhere else.

PR #663, opened 2017-07-22
Implement support for custom enum types.

swankjesse:
This makes it possible to decode proto enums as regular user-defined Java
types.

Closes: https://github.com/square/wire/issues/602

### DIFF
--- a/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
+++ b/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
@@ -59,6 +59,7 @@ import com.squareup.wire.schema.Service;
 import com.squareup.wire.schema.Type;
 import java.io.IOException;
 import java.math.BigInteger;
+import java.net.ProtocolException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -145,23 +146,39 @@
...
-      }
-      Set<String> collidingNames = collidingFieldNames(type.fieldsAndOneOfFields());
-      for (Field field : type.fieldsAndOneOfFields()) {
-        String suggestion = collidingNames.contains(field.name())
-            ? field.qualifiedName()
-            : field.name();
-        nameAllocator.newName(suggestion, field);
+
+      if (type instanceof MessageType) {
+        nameAllocator.newName("serialVersionUID", "serialVersionUID");
```
JakeWharton: I'm pretty sure I added a one-arg version of this... I'll fix in a follow-up.

wire-tests/src/test/java/com/squareup/wire/java/JavaGeneratorTest.java
```diff
@@ -200,6 +205,85 @@
...
+        + "    this.TAILS = TAILS;\n"
+        + "  }\n"
+        + "\n"
+        + "  protected int toValue(JavaCoinFlip value) {\n"
+        + "    if (value.equals(HEADS)) return 1;\n"
+        + "    if (value.equals(TAILS)) return 2;\n"
+        + "    return -1;\n"
+        + "  }\n"
+        + "\n"
+        + "  protected JavaCoinFlip fromValue(int value) {\n"
```
JakeWharton: These two can be `private` as they're in a `final` class.

swankjesse: Whoops, class shouldn’t be final! These are for overriding.

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
swankjesse: That one doesn’t have a tag. We’re cheating and using the suggestion as a tag.

JakeWharton: Ah, right. It's for a one-shot name where you use the return value and don't care about it again.

PR #665, opened 2017-07-24
Allow custom types to be constrained with a 'with' clause.

swankjesse:
This will be used to target fields like 'int64 created_at' that carry
extra type information in options '[is_date = true]' to be matched for
custom types.


swankjesse:
Thoughts on syntax?

In `date_formats.proto`:
```
enum DateFormat {
  RFC3339
}

extend google.protobuf.FieldOptions {
  optional DateFormat date_format;
}
```

in `java.wire`:
```
type string {
  with date_format = RFC3339
  target java.time.Instant using com.squareup.time.WireAdapters#INSTANT_ADAPTER;
}
```

JakeWharton:
Will try to review tomorrow!

JakeWharton:
I'm tempted to just call it `option`

swankjesse:
Going to keep as `with` just in case we later decide we want to put options on the custom type bindings themselves.

PR #668, opened 2017-07-27
Use Okio's Utf8.size.

NightlyNexus:


wire-runtime/src/main/java/com/squareup/wire/ProtoAdapter.java
```diff
@@ -375,7 +375,7 @@ public String toString(E value) {
   public static final ProtoAdapter<String> STRING = new ProtoAdapter<String>(
       FieldEncoding.LENGTH_DELIMITED, String.class) {
     @Override public int encodedSize(String value) {
-      return utf8Length(value);
+      return (int) Utf8.size(value);
```
NightlyNexus: if this is even a change you want, should this assert that it's an int?

swankjesse: ooh, good question. Lemme add a task to test that.

swankjesse: https://github.com/square/wire/issues/670

swankjesse:
Much better!

PR #672, opened 2017-08-03
Don't crash when a comment has a dollar sign

swankjesse:


adriancole:
Was this one a bug found at $quare?

swankjesse:
How’d you guess?!

PR #677, opened 2017-10-14
Make ProtoReader non-final

ksperling:
My specific use-case for this is a `FilteringProtoReader` that overrides `nextTag` and delegates to a policy to decide whether to process or `skip()` each value. This is can speed up reading of large protos from persistent storage quite dramatically if only a subset of the data is required.

Another option would be to extract the public API into a `ProtoSource` interface which `ProtoReader` implements, but that's of course a much larger change.

### DIFF
--- a/wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
+++ b/wire-runtime/src/main/java/com/squareup/wire/ProtoReader.java
@@ -42,9 +42,9 @@ import okio.ByteString;
 /**
  * Reads and decodes protocol message fields.
  */
-public final class ProtoReader {
+public class ProtoReader {
   /** The standard number of levels of message nesting to allow. */
-  private static final int RECURSION_LIMIT = 65;
+  protected static final int RECURSION_LIMIT = 65;
 
   private static final int FIELD_ENCODING_MASK = 0x7;
   static final int TAG_FIELD_ENCODING_BITS = 3;


JakeWharton:
Interesting use case. I assume you want to then hand that customized `ProtoReader` back to Wire to get an object back?

ksperling:
Yes, via `ProtoAdapter.decode(ProtoReader)` which is already public. It all ends up pretty tidy actually, along these lines:
```
    final Set<List<Integer>> excludes = ImmutableSet.of(
        ImmutableList.of(/* sensor_details */ 4),
        ImmutableList.of(/* frame */ 100, /* orientation */ 8));
    
    try (BufferedSource src = Okio.buffer(Okio.source(ts.srcFile))) {
      return SensorSegment.ADAPTER.decode(new FilteringProtoReader(src, (p) -> !excludes.contains(p)));
    }
```

It would be nice to have constants available for the tags, but in practice the exclusion sets are managed by a class that represents a particular "projection" and that can also strip out the same fields from a fully parsed message object, so it's pretty easy to have some tests to validate that the decode-time filtering produces the same result. (Normally I would prefer to use a white-list rather than a black-list, but the policy would then need to support prefix or wildcard matches if I want to include a nested message in its entirety without repeating it's entire structure in the include list.)

The `FilteringProtoReader` itself is also straightforward: https://gist.github.com/ksperling/bbde499437b7251d52e60b131930126f

JakeWharton:
This class wasn't designed for inheritance so i'm extremely wary of such a
simple change. We'll have to re-evaluate its entire API before making such
a change, and we'll likely want to investigate alternative means of
accomplishing this.

ksperling:
I understand your reluctance in terms of opening up classes that are mostly an implementation detail to sub-classing, in terms of the burden of maintaining compatibility going forward. At the same time it can also be frustrating when otherwise straight-forward use cases of a library are blocked by restrictive access modifiers and/or final modifiers on classes. Personally I feel like a good solution to that dilemma is to just document what's considered to be public API and what isn't, and then anyone who chooses to go beyond the public API accepts the risk that things may break even on a point release.

A way to lessen the impact of making `ProtoReader` non-final while still enabling my use-case would be to make all methods except for `nextTag`, `beginMessage` and `endMessage` final. If there was an accessor for `recursionDepth` this could be restricted further to allow only `nextTag` to be overridden.

Another approach would be to make `ProtoReader` (and for symmetry`ProtoWriter`) interfaces; currently they are both instantiated from the `ProtoAdapter` base class, so such a change should be binary compatible with previously generated message classes and require no change to the code generator. However it would break any user code that instantiates `ProtoReader` directly -- probably not a very common thing to do, not least because it is currently final. However this is a much  bigger change with more potential for immediate compatibility issues.

Yet another approach would be to add support for skipping values based on a policy delegate directly into `ProtoReader` and leave it as final. I haven't benchmarked this but I imagine the performance impact for code that doesn't use the feature should be minimal.

Let me know if either of these (or another approach that I haven't though of) sound like something that would be OK to merge in principle, and I'll update the PR. (Or if it's definitely out of scope for Wire that's of course OK too, but would be good to know one way or another.)

swankjesse:
I don’t think I’m ready to commit to the complexity of ProtoReader subclasses. It is a simple change, but it means maintenance to that class gets difficult. 

I think if you do want this you should do it as a different layer; perhaps something that operates on the bytes before they make it to the proto reader.

Closing for now.

PR #682, opened 2018-02-10
Raise error when option not defined.

mc-lovin:


jrodbx:
Closing, since outdated

PR #685, opened 2018-02-15
Add support for Kotlin 

tomoima525:
# Overview

- In Proto3, a field value returns the default value. However, if a scalar message field is set to its default, the value [will not be serialized on the wire](https://developers.google.com/protocol-buffers/docs/proto3#default), hence fields can be `null`. 
- This causes an issue in a Kotlin project because we expect value exists but can be null in some cases. 
- This PR will help accessing fields safely from Kotlin by:
  - Adding Nullable annotation when a field can be null. 
  - Adding `get*OrDefault()` method which allows returning the default value when the value does not exist

### Usage

```java
public String getEmailOrDefault() {
   return com.squareup.wire.Wire.get(email, Person.DEFAULT_EMAIL); 
}
```

```kotlin
val person = ...
val emailOptional: String? = person.email 
val email = person.emailOrDefault // empty string if email is null
person.address?.let { it.zipcode ... } 
```


JakeWharton:
Not a fan of this change for a lot of reasons.

 1. This hardly adds support for Kotlin. If we were doing that we would generate data classes and kt files. 
 2. Naming this flag `--kotlin` is inaccurate. It's more like `--default-getters`.
 3. One of the driving motivations of Wire was to avoid these methods. If Square were to use this flag it would add on the order of _thousands_ of methods. They are trivial to ProGuard out, though.
 4. This flag implies `--android` which is not always true. The whole reason there aren't nullability annotations by default is because Java screwed up here and it requires a third-party dependency for which there are many. Actually generating Kotlin eliminates this problem since the JetBrains annotations are a dependency of the Kotlin stdlib and are used by its compiler.

tomoima525:
@JakeWharton thanks for the comments.

1. True, I think generating Kotlin might be the best solution. But I feel it would take some time to implement that feature.

2. Agree, `--default-getters` makes more sense. I'm happy to change the option name.

3.4. I understand Wire's motivation to reduce the generated code. However, it is cumbersome when we access proto file from Kotlin; we always have to wrap with `Wire.get()` method which, in my opinion, is killing the benefit of using Kotlin. I think this option is not too harmful until Wire actually supports Kotlin.

`proto-lite` has getter(but no default getter) and I played around with it, but I did not like its generated code. x3 times more LOC than Wire, low readability. So I want to stick to Wire.

JakeWharton:
We/they at Square feel your pain, as we/they have a lot of Kotlin code consuming Wire protos.

I don't think it would actually be that hard to get Kotlin code generation working. The biggest problem is probably the base `Message` class which is optimized for minimizing method count in Java. I don't know how well it would play in Kotlin.

I can probably find time to prototype something in the next week, but I'm not sure I'd have the time to push it to completion.

swankjesse:
Thanks for the PR, and thanks for reminding us that we’re doing a bad job on Kotlin support. But this isn’t how we’re going to do it.

tomoima525:
@JakeWharton @swankjesse Thanks folks, I'll wait for the update with KotlinPoet

PR #692, opened 2018-05-10
Update google/protobuf/descriptor.proto.

eventualbuddha:
Copied from [v3.5.2](https://github.com/google/protobuf/blob/v3.5.2/src/google/protobuf/descriptor.proto).

Without this, any project that uses Wire and incorporates messages/fields/etc from google/protobuf/descriptor.proto may not work as expected. For example, this update adds `google.protobuf.OneofOptions`. Before this update, if a project had an extension for that message its proto build would fail. Wire _always_ uses its own copy of this file rather than one that might be included by the project being built, which is itself unexpected and should potentially be re-examined.


eventualbuddha:
@JakeWharton the version I used is identical to the one that was released with v3.5.2, so I updated the PR description to link to the file using the tag name.

eventualbuddha:
@swankjesse what do you think?

eventualbuddha:
Thanks!

PR #694, opened 2018-05-18
Add support for Android annotations without Parcelable

### LINKED ISSUE #693: Nullable annotations for Android without Parcelable
Currently, Wire supports Nullable via the "--android" flag, which also bundles in Parcelable. Parcelable introduces some unfortunate dependencies, however, whereas Nullable tends to be usable by a lot of tools (https://github.com/uber/NullAway, for instance) with minimal added overhead/dependencies (just the Android support annotations).

https://github.com/square/wire/issues/225 runs down some of the different options around Nullability annotations - before opening a PR, I wanted to see if people had thought about supporting a flag to only include android support annotations, but not the rest of Android (right now, this just looks like Parcelable and Nullable to me, after scanning for emitAndroid, though I could've missed something. 

Would personally prefer to specify --android-annotations to indicate that we'd add android-specific annotations vs full Android support.

Happy to submit this PR here - it felt to me like the jsr305 route wasn't as worth the added annoyance for consumers, and that scoping to Android for now was extremely low cost (supporting another flag).

LukeStClair:
This PR adds support for a flag that includes Android annotations only (no
parcelable support). This is to enable folks to benefit from useful static
analysis (https://github.com/uber/NullAway, for instance) without incurring
dependencies on the wider Android framework. Decided to forgo more generalized
annotation support for now.

A couple of notes:

- Didn't catch where the Android functionality was unit tested previously, so I
  added small tests with and without this flag.
- The approach of adding more flags in this way won't scale well forever, and I
  don't love flags implying each other, but in this case, I think it's
  reasonably straightforward and hard to be surprised.
- I'm not attached to the awkward but descriptive flag name - happy to change
  the semantics.

Fixes #693.

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
+++ b/wire-compiler/src/main/java/com/squareup/wire/WireCompiler.java
@@ -53,6 +53,7 @@ import static com.google.common.base.Preconditions.checkState;
  *     [--quiet]
  *     [--dry_run]
  *     [--android]
+ *     [--android-annotations]
  *     [--compact]
  *     [file [file...]]
  * </pre>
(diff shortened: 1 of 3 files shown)


JakeWharton:
They're tested with golden files currently. There's a script which generates them. I don't think it's important to do the same here.

PR #696, opened 2018-05-25
Possible to add package prefix

vandmo:
This is needed since we need since we need to coexist with Java classes already on the class path but generated with protoc.

For example, if we want to generate code with Wire that uses google/protobuf/Timestamp we will get two com.google.protobuf.Timestamp classes on the class path were one is generated with protoc and one with wire.

With this change it is possible to generate se.vandmo.somelibrary.com.google.protobuf.Timestamp

### DIFF
--- a/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
+++ b/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
@@ -211,10 +211,17 @@ public final class JavaGenerator {
   }
 
   public static JavaGenerator get(Schema schema) {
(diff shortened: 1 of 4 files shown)


swankjesse:
I’d like to do this, but with our new _profiles_ feature.

vandmo:
Fair enough, do you have any ETA for that? Will it be possible to do it by configuring the Maven plugin? We do not have control over the .proto files so we can not put .wire files beside them.

vandmo:
Sounds great, glad to hear that 2.3 is approaching :) keep up the good work!

oldergod:
The new `wire_package` since 3.3.0 might help!

PR #699, opened 2018-06-22
Wire support for Moshi

oldergod:
### Counterpart of `wire-gson-support` for [Moshi](https://github.com/square/moshi).

It is similar to `wire-gson-support`; e.g. tests are exactly the same.

The 2 major differences were around parsing numbers, specially `long`s.
`Moshi`, by default, parses `Number` as `Double`. But if the data was formerly a `long`, this would lead to data loss if we blindly followed the `gson` counterpart's code.

First, we needed an `JsonAdapter` to enable numbers segregation. I took inspiration from a [`gist` Jesse wrote'](https://gist.github.com/swankjesse/0331f24df1aa25454eaae04e49fe2f19) some [months ago to deal with this very problem](https://github.com/square/moshi/pull/314). This is what the `SegregatedNumbersAdapter` is about.

Second, we have to re-sign, formerly signed but unsigned for serialization, longs. This is what the `consumeUint64()` is about.


wire-moshi-support/src/test/java/com/squareup/wire/MoshiTest.java
```diff
@@ -0,0 +1,298 @@
...
+
+    AllTypes allTypes = builder.build();
+    String json = moshi.adapter(AllTypes.class).toJson(allTypes);
+    assertThat(json).isEqualTo(JSON);
+  }
+
+  @SuppressWarnings("ConstantConditions")
+  @Test public void nullRepeatedField() throws IOException {
+    RepeatedPackedAndMap parsed = moshi.adapter(RepeatedPackedAndMap.class).lenient()
+        .fromJson("{rep_int32=null,pack_int32=null,map_int32_int32=null}");
```
oldergod: The test is ported from `wire-gson-support`, not sure if it was better to use `lenient` or to rewrite the json here.

wire-moshi-support/src/main/java/com/squareup/wire/ByteStringJsonAdapter.java
```diff
@@ -0,0 +1,32 @@
...
+import javax.annotation.Nullable;
+import java.io.IOException;
+
+import static com.squareup.moshi.JsonReader.Token;
+
+class ByteStringJsonAdapter extends JsonAdapter<ByteString> {
+    @Override
+    public void toJson(JsonWriter out, @Nullable ByteString byteString) throws IOException {
+        if (byteString == null) {
+            out.nullValue();
```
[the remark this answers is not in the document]
oldergod: I'd need to check if `byteString` is null anyway to avoid NPE

wire-moshi-support/src/main/java/com/squareup/wire/MessageJsonAdapter.java
```diff
@@ -0,0 +1,168 @@
...
+    if (in.peek() == JsonReader.Token.NULL) {
+      in.nextNull();
+      return null;
+    }
+
+    B builder = messageAdapter.newBuilder();
+
+    in.beginObject();
+    while (in.peek() != JsonReader.Token.END_OBJECT) {
+      String name = in.nextName();
```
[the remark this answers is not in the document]
oldergod: I don't know. What would we gain?

wire-moshi-support/src/main/java/com/squareup/wire/MessageJsonAdapter.java
```diff
@@ -0,0 +1,168 @@
...
+        }
+        in.endArray();
+
+        return longs;
+      } else {
+        return consumeUint64(in);
+      }
+    }
+
+    Object object = moshi.adapter(Object.class).fromJson(in);
```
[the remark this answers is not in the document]
oldergod: Using `in.readJsonValue()` would parse `long`s using the `double` adapter.

wire-moshi-support/src/main/java/com/squareup/wire/ByteStringJsonAdapter.java
```diff
@@ -0,0 +1,31 @@
+package com.squareup.wire;
```
swankjesse: nit: copyright

wire-moshi-support/src/main/java/com/squareup/wire/ByteStringJsonAdapter.java
```diff
@@ -0,0 +1,31 @@
...
+import com.squareup.moshi.JsonAdapter;
+import com.squareup.moshi.JsonReader;
+import com.squareup.moshi.JsonWriter;
+import java.io.IOException;
+import javax.annotation.Nullable;
+import okio.ByteString;
+
+import static com.squareup.moshi.JsonReader.Token;
+
+class ByteStringJsonAdapter extends JsonAdapter<ByteString> {
```
swankjesse: if you declare this as a single instance you can call `.nullSafe()` to not need the null handling

wire-moshi-support/src/main/java/com/squareup/wire/ByteStringJsonAdapter.java
```diff
@@ -0,0 +1,31 @@
...
+  @Override
+  public void toJson(JsonWriter out, @Nullable ByteString byteString) throws IOException {
+    if (byteString == null) {
+      out.nullValue();
+    } else {
+      out.value(byteString.base64());
+    }
+  }
+
+  @Nullable
```
swankjesse: nit: this is a type annotation so it should immediately precede the type it applies to, `ByteString`

wire-moshi-support/src/main/java/com/squareup/wire/MessageJsonAdapter.java
```diff
@@ -0,0 +1,168 @@
...
+
+    if (fieldBinding.label.isRepeated()) {
+      if (object == null) {
+        return Collections.emptyList();
+      }
+
+      Class<?> itemType = protoAdapter.javaType;
+
+      return moshi.adapter(Types.newParameterizedType(List.class, itemType)).fromJsonValue(object);
```
swankjesse: gotta find a way to do these moshi.adapter lookups eagerly

wire-moshi-support/src/main/java/com/squareup/wire/SegregatedNumbersAdapter.java
```diff
@@ -0,0 +1,40 @@
...
+  SegregatedNumbersAdapter(JsonAdapter<Object> delegate) {
+    this.delegate = delegate;
+  }
+
+  @Override public @Nullable Object fromJson(JsonReader reader) throws IOException {
+    if (reader.peek() != JsonReader.Token.NUMBER) {
+      return delegate.fromJson(reader);
+    }
+
+    final BigDecimal decimal = new BigDecimal(reader.nextString());
```
swankjesse: this is not going to be fast

wire-moshi-support/src/main/java/com/squareup/wire/WireJsonAdapterFactory.java
```diff
@@ -0,0 +1,50 @@
...
+
+    if (Types.getRawType(type).equals(ByteString.class)) {
+      return new ByteStringJsonAdapter();
+    }
+    if (Message.class.isAssignableFrom(Types.getRawType(type))) {
+      return new MessageJsonAdapter(moshi, type);
+    }
+    if (Types.getRawType(type).equals(Object.class)) {
+      JsonAdapter<Object> delegate = moshi.nextAdapter(this, Object.class, annotations);
+      return new SegregatedNumbersAdapter(delegate);
```
swankjesse: I wonder if there’s a way we can do this only on recursive calls required by the MessageJsonAdapter

wire-moshi-support/src/test/java/com/squareup/wire/MoshiTest.java
```diff
@@ -0,0 +1,298 @@
+/*
+ * Copyright 2013 Square Inc.
```
swankjesse: nit: 2018?

wire-moshi-support/src/test/java/com/squareup/wire/MoshiTest.java
swankjesse: we want strict please

wire-moshi-support/src/test/java/com/squareup/wire/MoshiTest.java
oldergod: I remember reading somewhere that this should be aligned on all files. That new files should keep the date for the first copyrighted file.

wire-moshi-support/src/main/java/com/squareup/wire/MessageJsonAdapter.java
oldergod: Struggling with this one. Even if I had to target type via `itemType`, I don't know how to avoid a lookup for an adapter through `moshi.adapter()`

wire-moshi-support/src/main/java/com/squareup/wire/SegregatedNumbersAdapter.java
oldergod: This whole adapter is gone now.

wire-moshi-support/src/main/java/com/squareup/wire/MessageJsonAdapter.java
oldergod: I changed the logic so I don't do _two_ adapter lookup, but not sure that is enough.

wire-moshi-support/src/main/java/com/squareup/wire/WireJsonAdapterFactory.java
oldergod: not a problem, anymore

wire-moshi-support/src/main/java/com/squareup/wire/ByteStringJsonAdapter.java
oldergod: doing as what we're doing in `wire-gson-support` for now

wire-moshi-support/src/test/java/com/squareup/wire/MoshiTest.java
oldergod: I was wrong. Fixed it.

oldergod:
Good for review

wire-moshi-support/src/main/java/com/squareup/wire/MessageJsonAdapter.java
```diff
@@ -0,0 +1,190 @@
...
+
+    if (fieldBinding.label.isRepeated()) {
+      if (in.peek() == JsonReader.Token.NULL) {
+        in.nextNull();
+        return Collections.emptyList();
+      }
+
+      Class<?> itemType = protoAdapter.javaType;
+
+      return moshi.adapter(Types.newParameterizedType(List.class, itemType)).fromJson(in);
```
JakeWharton: Lots of room for future optimization like caching these adapters so we don't have to allocate each time we want to use it

wire-moshi-support/src/main/java/com/squareup/wire/MessageJsonAdapter.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  @Override public void toJson(JsonWriter out, @Nullable M message) throws IOException {
+    if (message == null) {
+      out.nullValue();
+      return;
+    }
+
+    out.beginObject();
+    for (FieldBinding<M, B> tagBinding : messageAdapter.fieldBindings().values()) {
```
JakeWharton: Use `fieldBindings` field

wire-moshi-support/src/main/java/com/squareup/wire/MessageJsonAdapter.java
```diff
@@ -0,0 +1,190 @@
...
+        out.endArray();
+      } else {
+        emitUint64((Long) value, out);
+      }
+    } else {
+      moshi.adapter(Object.class).toJson(out, value);
+    }
+  }
+
+  private void emitUint64(Long value, JsonWriter out) throws IOException {
```
JakeWharton: static

wire-moshi-support/src/main/java/com/squareup/wire/MessageJsonAdapter.java
```diff
@@ -0,0 +1,190 @@
...
+      Class<?> valueType = protoAdapter.javaType;
+
+      return moshi.adapter(Types.newParameterizedType(Map.class, keyType, valueType)).fromJson(in);
+    }
+
+    Class<?> elementType = protoAdapter.javaType;
+    return moshi.adapter(elementType).fromJson(in);
+  }
+
+  private long consumeUint64(JsonReader in) throws IOException {
```
JakeWharton: static

wire-moshi-support/src/main/java/com/squareup/wire/WireJsonAdapterFactory.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+public final class WireJsonAdapterFactory implements JsonAdapter.Factory {
+  @Override
+  public @Nullable JsonAdapter<?> create(Type type, Set<? extends Annotation> annotations,
+      Moshi moshi) {
+    if (!annotations.isEmpty()) {
+      return null;
+    }
+
+    if (Types.getRawType(type).equals(ByteString.class)) {
```
JakeWharton: pull rawType to local since it's computed twice

wire-moshi-support/src/main/java/com/squareup/wire/MessageJsonAdapter.java
oldergod: Isn't moshi doing the caching here?
https://github.com/square/moshi/blob/master/moshi/src/main/java/com/squareup/moshi/Moshi.java#L94-L148

JakeWharton: Yes but it requires allocating the parameterized type. We should be able to maintain an array of adapters which are parallel to the field bindings so that it's O(1) lookups. Plus it pushes validation to creation time instead of serialization time.

oldergod: Ok, I can do that in a follow up.
I believe I'll have to create a `3 * size` array of adapters so each field has its adapter, its list adapter, and the map one, does that sound correct?

JakeWharton: Hmm I thought we would only need one per field. I don't think it's at all urgent, we can play around in a follow-up.

oldergod: fixed

oldergod:
includes now all the goodies.

wire-moshi-support/src/main/java/com/squareup/wire/WireJsonAdapterFactory.java
```diff
@@ -0,0 +1,141 @@
...
+  }.nullSafe();
+
+  /**
+   * Wire uses the signed long type to store unsigned longs. Sigh. But when we encode as JSON we
+   * need to emit an unsigned value.
+   */
+  static final JsonAdapter<Long> UINT64_JSON_ADAPTER = new JsonAdapter<Long>() {
+    // 2^64, used to convert sint64 values >= 2^63 to unsigned decimal form
+    private final BigInteger power64 = new BigInteger("18446744073709551616");
+    private final BigInteger maxLong = BigInteger.valueOf(Long.MAX_VALUE);
```
oldergod: because checkstyle complains if we start with capital on a non-static field...

wire-moshi-support/src/main/java/com/squareup/wire/WireJsonAdapterFactory.java
```diff
@@ -0,0 +1,141 @@
...
+        writer.value(unsigned);
+      } else {
+        writer.value(value);
+      }
+    }
+  }.nullSafe();
+
+  /**
+   * Tragically Moshi doesn't know enough to follow a {@code @Uint64 List<Long>} really wants to be
+   * treated as a {@code List<@Uint64 Long>} and so we have to do it manually.
```
oldergod: https://github.com/square/moshi/issues/666

swankjesse: even better than linking on the codereview thread is linking in the source code. That way if somebody is editing this later they can see if that bug is closed and this code can be simplified!

wire-moshi-support/src/main/java/com/squareup/wire/MessageJsonAdapter.java
```diff
@@ -0,0 +1,190 @@
...
+        out.beginArray();
+        for (int i = 0, count = longs.size(); i < count; i++) {
+          emitUint64(longs.get(i), out);
+        }
+        out.endArray();
+      } else {
+        emitUint64((Long) value, out);
+      }
+    } else {
+      moshi.adapter(Object.class).toJson(out, value);
```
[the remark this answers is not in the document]
swankjesse: never mind me, this comment predated your fix!

wire-moshi-support/src/main/java/com/squareup/wire/MessageJsonAdapter.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  @Override public void toJson(JsonWriter out, @Nullable M message) throws IOException {
+    if (message == null) {
+      out.nullValue();
+      return;
+    }
+
+    out.beginObject();
+    for (FieldBinding<M, B> tagBinding : this.fieldBindings) {
```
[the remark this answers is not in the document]
swankjesse: already fixed!

wire-moshi-support/src/main/java/com/squareup/wire/MessageJsonAdapter.java
```diff
@@ -0,0 +1,190 @@
...
+        Object value = parseValue(fieldBinding, in);
+        fieldBinding.set(builder, value);
+      }
+    }
+
+    in.endObject();
+    return builder.build();
+  }
+
+  private Object parseValue(FieldBinding<?, ?> fieldBinding, JsonReader in) throws IOException {
```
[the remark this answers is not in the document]
swankjesse: already fixed!

pom.xml
```diff
@@ -49,6 +49,7 @@
     <module>wire-gson-support</module>
     <module>wire-java-generator</module>
     <module>wire-maven-plugin</module>
+    <module>wire-moshi-support</module>
```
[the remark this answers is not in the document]
oldergod: go for `wire-moshi-adapter`

wire-moshi-support/src/main/java/com/squareup/wire/WireJsonAdapterFactory.java
```diff
@@ -0,0 +1,62 @@
...
+ * the Moshi Json library.
+ *
+ * <pre>
+ * Moshi moshi = new Moshi.Builder()
+ *   .add(new WireJsonAdapterFactory())
+ *   .build();
+ * </pre>
+ *
+ * The resulting {@link Moshi} instance will be able to serialize and deserialize any Wire {@link
+ * Message} type, including extensions and unknown field values. The JSON encoding is intended to be
```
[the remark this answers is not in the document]
oldergod: I copied that from the `gson` module; I fixed it. For extensions, I'm not 100% sure but since I didn't see any code that would take care of that, I wrote we don't support them, and ignore unknown fields.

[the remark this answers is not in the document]
oldergod: We actually support extensions.

PR #701, opened 2018-07-04
File options constructor should take in a builder

mc-lovin:
Running `gen_tests.sh` generates this fileoptions which has a different format.

wonder why travis didn't catch it when the actual pr was merged.

### DIFF
--- a/wire-runtime/src/test/proto-java/com/google/protobuf/FileOptions.java
+++ b/wire-runtime/src/test/proto-java/com/google/protobuf/FileOptions.java
@@ -297,43 +297,27 @@ public final class FileOptions extends Message<FileOptions, FileOptions.Builder>
   )
   public final List<UninterpretedOption> uninterpreted_option;
 
-  public FileOptions(String java_package, String java_outer_classname, Boolean java_multiple_files,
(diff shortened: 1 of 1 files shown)


wire-runtime/src/test/proto-java/com/google/protobuf/FileOptions.java
```diff
@@ -297,43 +297,27 @@
...
+    this.py_generic_services = builder.py_generic_services;
+    this.php_generic_services = builder.php_generic_services;
+    this.deprecated = builder.deprecated;
+    this.cc_enable_arenas = builder.cc_enable_arenas;
+    this.objc_class_prefix = builder.objc_class_prefix;
+    this.csharp_namespace = builder.csharp_namespace;
+    this.swift_prefix = builder.swift_prefix;
+    this.php_class_prefix = builder.php_class_prefix;
+    this.php_namespace = builder.php_namespace;
+    this.uninterpreted_option = Internal.immutableCopyOf("uninterpreted_option", builder.uninterpreted_option);
```
JakeWharton: you can drop the `this`s if you want

mc-lovin: sure doing it in a follow up

PR #712, opened 2018-09-18
Configure integration testing for Kotlin

Egorand:
Improvement over `KotlinGeneratorTest`: those tests aren't very precise and rely on `assertTrue()` a lot, which spits out extremely unhelpful error messages.


Egorand:
Actually decided to avoid moving the files around, I started getting golden value test failures, possibly due to `Person.java` and `Person.kt` having the same FQ name, thus confusing the tests. Added code that allows configuring language targets making test setup more generic, the API is not very elegant, but it does the job. PTAL.

Egorand:
@JakeWharton can you take another look and merge?

mc-lovin:
lgtm

PR #713, opened 2018-09-19
Make generated Kotlin protos extend Message

### LINKED ISSUE #706: Generated classes should extend Message (or AndroidMessage)
Bumped into it when trying to make this work with `WireConverterFactory` for Retrofit, which expects response classes to be assignable from `Message`.

Egorand:
Fixes #706

### DIFF
--- a/wire-runtime/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt
+++ b/wire-runtime/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt
@@ -12,8 +12,8 @@ import com.squareup.wire.TagHandler
 import com.squareup.wire.WireEnum
 import com.squareup.wire.internal.Internal
 import kotlin.Deprecated
+import kotlin.DeprecationLevel
 import kotlin.Int
-import kotlin.ReplaceWith
 import kotlin.String
 import kotlin.collections.List
 import okio.ByteString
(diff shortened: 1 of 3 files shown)


wire-runtime/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt
```diff
@@ -21,7 +24,18 @@ data class Person(
     val email: String? = null,
     val phone: List<PhoneNumber> = emptyList(),
     val unknownFields: ByteString = ByteString.EMPTY
-) {
+) : Message<Person, Person.Builder>(ADAPTER, unknownFields) {
+    @Deprecated(
+            message = "Not relevant for data classes, use copy()",
```
JakeWharton: I don't find this to be true since there's no way to have multiple functions contribute to the building of one object.

JakeWharton: Also this is going to cause deprecations on the Java side for all these things where Builder is the only thing you have

Egorand: The idea is to generate different code for Kotlin-only/mixed codebases: for Kotlin-only, we don't really need Builders, but we still have to extend `Message` and provide a `Message.Builder` implementation, deprecation annotations should prevent people from using the Builder and prefer `copy()`; for mixed codebases we'll generate full versions of Builders to remain source-compatible with Java protos - this will be hidden behind a compiler flag.

JakeWharton: In that case they should be marked as HIDDEN without a ReplaceWith

oldergod: Is this `HIDDEN` the solution to https://youtrack.jetbrains.com/issue/KT-24769 ?

mc-lovin: Can kotlin functions with HIDDEN annotations be accessed from JAVA?

mc-lovin:
can we add a flag for this ?

Egorand:
@mc-lovin Currently protos have to extend `Message`, otherwise Retrofit's `WireConverterFactory` won't work.

Egorand:
@JakeWharton I removed `Deprecated` from `Builder` (there's no way to make it `HIDDEN`, cause then the proto itself doesn't compile) and added `HIDDEN` to `newBuilder()`. Not sure this is what makes the most sense, let's change this later if there are better ideas.

Issue #707, opened 2018-09-18
Kotlin- and Java-generated protos are incompatible


Egorand:
This presents an issue for gradual migration to Kotlin. For example, Java protos have `public static final ADAPTER` fields and access it as `Foo.ADAPTER`. In Kotlin, this has been changed to a nested `object ADAPTER`, which is great for Kotlin callers, but Java callers would need to use `Foo.ADAPTER.INSTANCE`. We need to make sure this kind of compatibility is taken into account. Also, ideally we'd like this to be an opt-in feature behind a compiler flag, so that we don't generate unnecessary code for 100% Kotlin projects.

Egorand:
`WireConverterFactory` also relies on `ADAPTER` being a `public static` field.

swankjesse:
I’d like for the compiled code to make `ADAPTER` look like a public static final field. Maybe we could do a `@JvmStatic` thing on a member of the companion?

Egorand:
@swankjesse #714

PR #710, opened 2018-09-18
Fix generation of the CREATOR field for Kotlin Android Parcelables

Egorand:
Fixes #705

PR #714, opened 2018-09-19
Move generated adapter to companion object

mc-lovin:
Fixes https://github.com/square/wire/issues/707


wire-runtime/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt
```diff
@@ -22,42 +23,46 @@ data class Person(
...
 ) {
-    object ADAPTER : ProtoAdapter<Person>(FieldEncoding.LENGTH_DELIMITED, Person::class.java) {
-        override fun encodedSize(value: Person): Int = ProtoAdapter.STRING.encodedSizeWithTag(1, value.name) +
-            ProtoAdapter.INT32.encodedSizeWithTag(2, value.id) +
-            ProtoAdapter.STRING.encodedSizeWithTag(3, value.email) +
-            PhoneNumber.ADAPTER.asRepeated().encodedSizeWithTag(4, value.phone) +
-            value.unknownFields.size()
+    companion object {
+        @JvmField
+        val ADAPTER: ProtoAdapter<Person> =
```
mc-lovin: @Egorand is there a way we can get rid of the explicit type here if we are initializing the property?

Egorand: No, property types are always declared explicitly.

JakeWharton:
Dupe of #710? Not that it matters...

Egorand:
@JakeWharton no, #710 fixes `Parcelable` implementation.

Egorand:
@mc-lovin I suggest getting rid of the test class containing failing tests (in a separate PR), that kind of tests is not very helpful. #712 sets up proper integration testing that is easier to maintain and provides better feedback on failures.

mc-lovin:
@Egorand PTAL and merge

PR #715, opened 2018-09-21
Make Kotlin Android protos extend AndroidMessage

Egorand:


swankjesse:
This is pretty slick!

PR #717, opened 2018-09-22
Migrate to Gradle

Egorand:
- [x] Migrate sample, wire-codegen-sample, wire-maven-plugin
- [x] Ensure Maven builder plugin works
- [x] Setup Checkstyle
- [x] Update docs
- [x] Cleanup
- [ ] Package proto-pruner and wire-compiler as fat jars

Followups:
- Import Gradle plugin
- ~Update `sample` to work with Gradle plugin~
- Kill `gen_tests`, replace with Gradle magic

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,14 +7,6 @@ classes
 .classpath
 .project
 
-# Maven
-target
-pom.xml.*
-release.properties
-
-# Maven shade plugin
-dependency-reduced-pom.xml
-
 out/
 
 # Gradle


gen_tests.sh
```diff
@@ -2,21 +2,21 @@
 
 set -e
 
-mvn clean package -pl wire-compiler -am -Dmaven.test.skip
+./gradlew -p wire-compiler clean installDist
```
JakeWharton: This whole script can die and be replaced with `JavaExec` tasks and actual compilation of the generated sources once this actually lands.

Egorand:
`sample` has to work with the Gradle plugin now, I suggest importing it in a followup PR, updating and publishing.

JakeWharton:
You can use a JavaExec task instead. And it had the bonus of being able to use the version on master instead of a released version.

Issue #721, opened 2018-09-25
Replace gen_tests with Gradle tasks


Egorand:
Followup to #717

Egorand:
This is done.

PR #724, opened 2018-09-27
Rewrite gen_tests in Gradle

Egorand:


Egorand:
Closes #721

JakeWharton:
This does not close #721 to the degree that I would like it to. We need to create source sets for each of these output folders so that they are actually compiled.

Egorand:
Cool, changed the message.

Egorand:
@JakeWharton we're already creating source sets for output folders, so not sure what you mean https://github.com/square/wire/blob/57eb24676edc2a391af9ebf14b45c5660b08f005/wire-runtime/build.gradle#L8-L11

JakeWharton:
That's for one set. What about Android? What about no options? What about Kotlin?

Egorand:
They all end up in either proto-java or proto-kotlin. Is your idea to get rid of copying and suffixes and have test source sets for each type? That would make sense (it would've been possible to do with Maven too though).

JakeWharton:
Yes. Separate configurations for classpath, separate compilation units, and the option for separate tests. That is extremely difficult in Maven and trivial in Gradle.

Egorand:
Hmm, what do we gain by separating them in different compilation units?

JakeWharton:
Separate JavaCompile tasks with separate classpaths. Some depend on Java.
Some depend on Android. Some depend on Kotlin now.

Egorand:
Got it, interesting. I'll give it a shot.

JakeWharton:
I'm happy to do it after October 4th at 11am GMT

Egorand:
Let's pair on this!

JakeWharton:
Works for me! Get your slides done first though.

Egorand:
@JakeWharton let's pair on this at KotlinConf 2019 (unless you wanna get it in before then)

PR #730, opened 2018-10-03
Generate correct default value for optional enum in Kotlin

### LINKED ISSUE #729: Optional enum field generates broken code
Proto:
```proto
message OptionalEnumUser {
  optional OptionalEnum optional_enum = 1;
}

enum OptionalEnum {
  FOO = 1;
  BAR = 2;
}
```

Kotlin:
```kotlin
companion object {
    @JvmField
    val ADAPTER: ProtoAdapter<OptionalEnumUser> =
     ...
        override fun decode(reader: ProtoReader): OptionalEnumUser {
            var optional_enum: OptionalEnum? = OptionalEnum.null
            ...
        }
    }
}
```

The `OptionalEnum.null` part doesn't compile.

Egorand:
Fixes #729


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -597,8 +597,11 @@ class KotlinGenerator private constructor(
   private fun getDefaultValue(field: Field): CodeBlock {
     val internalClass = ClassName("com.squareup.wire.internal", "Internal")
     return when {
-      isEnum(field) ->
+      isEnum(field) -> if (field.default != null) {
```
Egorand: @mc-lovin is it ever possible for the default value to not be null? If not - we can simplify this.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -612,8 +615,7 @@ class KotlinGenerator private constructor(
     val default = getDefaultValue(this)
     return when {
       isRepeated -> CodeBlock.of("var $allocatedName = mutableListOf<%T>()", baseClass)
-      default != null -> CodeBlock.of("var $allocatedName: %T = %L", fieldClass, default)
```
Egorand: Warning: "default can never be null"

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
mc-lovin: yes, we can provide a default value via the proto 

```
  enum PhoneType {
    MOBILE = 0;
    HOME = 1;
    /** Could be phone or fax. */
    WORK = 2;
  }

  message PhoneNumber {
    // The customer's phone number.
    required string number = 1;
    // The type of phone stored here.
    optional PhoneType type = 2 [default = HOME];
  }
```

PR #738, opened 2018-10-23
Add WireField annotation to Kotlin constructor parameters.

JGulbronson:
The generated Kotlin files current don't contain the `@WireField` annotation. This seems like a breaking change from Java, that could prohibit users from switching to Kotlin. One such example is [Misk](https://github.com/square/misk/blob/6e75ee147724997a09662bedd5c46b8ab91d8795/misk/src/main/kotlin/misk/moshi/wire/WireMessageAdapter.kt#L30). My general assumption is if it breaks Misk, it can break other usages.

It might also be worth only including the annotation if generated with Java interop enabled, but happy to have that conversation!

I've made sure it compiles and tests pass with `./gradlew clean build`, and hopefully have added enough test coverage. First time in the codebase, so let me know if anything else is needed!

### DIFF
--- a/wire-runtime/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/DeprecatedProto.kt.java.interop
+++ b/wire-runtime/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/DeprecatedProto.kt.java.interop
@@ -8,13 +8,14 @@ import com.squareup.wire.ProtoAdapter
 import com.squareup.wire.ProtoReader
 import com.squareup.wire.ProtoWriter
 import com.squareup.wire.TagHandler
+import com.squareup.wire.WireField
 import kotlin.Deprecated
 import kotlin.Int
 import kotlin.String
 import kotlin.jvm.JvmField
 import okio.ByteString
 
-data class DeprecatedProto(val foo: String? = null, val unknownFields: ByteString = ByteString.EMPTY) : Message<DeprecatedProto, DeprecatedProto.Builder>(ADAPTER, unknownFields) {
+data class DeprecatedProto(@WireField(tag = 1, adapter = "com.squareup.wire.ProtoAdapter.STRING") val foo: String? = null, val unknownFields: ByteString = ByteString.EMPTY) : Message<DeprecatedProto, DeprecatedProto.Builder>(ADAPTER, unknownFields) {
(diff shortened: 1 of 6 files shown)


JGulbronson:
Thanks @JakeWharton! After talking with @Egorand I think it makes sense to put this behind the Java interop flag, and potentially come up with a better solution for pure Kotlin code.

JakeWharton:
If we abandoned Wire's runtime and JSON serializers we could choose to generate kotlinx.serialization-compatible Kotlin code which will be serializable as both proto and JSON as one definition.

JakeWharton:
And it would also mean our generated models would also work for JS and iOS...

JakeWharton:
Honestly we should be doing that as an optional regardless. Will file an issue for it.

wire-runtime/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/DeprecatedProto.kt.java.interop
```diff
@@ -8,13 +8,14 @@ import com.squareup.wire.ProtoAdapter
...
 import com.squareup.wire.TagHandler
+import com.squareup.wire.WireField
 import kotlin.Deprecated
 import kotlin.Int
 import kotlin.String
 import kotlin.jvm.JvmField
 import okio.ByteString
 
-data class DeprecatedProto(val foo: String? = null, val unknownFields: ByteString = ByteString.EMPTY) : Message<DeprecatedProto, DeprecatedProto.Builder>(ADAPTER, unknownFields) {
+data class DeprecatedProto(@WireField(tag = 1, adapter = "com.squareup.wire.ProtoAdapter.STRING") val foo: String? = null, val unknownFields: ByteString = ByteString.EMPTY) : Message<DeprecatedProto, DeprecatedProto.Builder>(ADAPTER, unknownFields) {
```
Egorand: Not really happy with how this looks, should we put those behind a compiler flag maybe? 
@swankjesse Jeff mentioned the absence of the annotations breaks Misk, is there any functionality that relies on `WireField`?

JGulbronson: Moved it behind the java interop flag.

PR #743, opened 2018-11-02
Add One Of Support

mc-lovin:


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -125,6 +125,10 @@ class KotlinGenerator private constructor(
         .addFunction(generateNewBuilderMethod(type, builderClassName))
         .addType(generateBuilderClass(type, className, builderClassName))
 
+    if (type.oneOfs().size != 0) {
```
JakeWharton: isNotEmpty

wire-runtime/src/test/proto/kotlin/one_of.proto
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package squareup.protos.kotlin.oneof;
+
+option java_package = "com.squareup.wire.protos.kotlin";
+
+message OneOfMessage {
+
+  // Must have a foo or a bar.
```
Egorand: ..or a baz!

wire-tests/src/test/java/com/squareup/wire/java/KotlinGeneratorTest.kt
```diff
@@ -7,7 +7,7 @@ import org.junit.Test
 import kotlin.test.assertTrue
 
 class KotlinGeneratorTest {
-  @Test fun basic() {
+  @Test fun `basic`() {
```
Egorand: Don't think these test name changes are needed, and they're outside of the scope of the PR

wire-runtime/src/test/proto/kotlin/deprecated.proto
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013 Square Inc.
+ * Copyright 2018 Square Inc.
```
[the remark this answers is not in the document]
mc-lovin: this file was added in 2018 but the copyright was mistakenly copied, not sure what is the right thing to do here

mc-lovin:
@Egorand PTAL

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -629,7 +629,7 @@ public TypeSpec generateMessage(MessageType type) {
    * Decides if a constructor should take all fields or a builder as a parameter.
    */
   private boolean constructorTakesAllFields(MessageType type) {
-    return type.fields().size() < MAX_PARAMS_IN_CONSTRUCTOR;
+    return type.fieldsAndOneOfFields().size() < MAX_PARAMS_IN_CONSTRUCTOR;
```
Egorand: Does this fix an old bug? Looks out of scope of this PR, and needs a test

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -15,6 +15,7 @@
  */
 package com.squareup.wire.kotlin
 
+import com.google.common.collect.ImmutableList
```
Egorand: No need for Guava in Kotlin modules, just use `listOf()`

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -130,6 +131,14 @@ class KotlinGenerator private constructor(
       addAndroidCreator(type, companionObjBuilder)
     }
 
+    if (type.oneOfs().size != 0) {
```
Egorand: Replace with `isNotEmpty()`

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -139,6 +148,23 @@ class KotlinGenerator private constructor(
...
+    val result = CodeBlock.Builder()
+    val nameAllocator = nameAllocator(type)
+    oneOfs
+        .filter { oneOf -> oneOf.fields().size >= 2 }
+        .forEach { oneOf ->
+          val fieldNames = oneOf.fields().map { nameAllocator.get(it) }.joinToString(", ")
+          result.beginControlFlow("if (%T.countNonNull(%L) > 1)",
+              Internal::class, fieldNames)
+          result.addStatement("throw IllegalArgumentException(" +
+              "\"at most one of " + "$fieldNames may be non-null\")")
```
Egorand: nit: is that `+` necessary?

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -662,6 +689,12 @@ class KotlinGenerator private constructor(
     else -> baseClass.asNonNull()
   }
 
+  private fun MessageType.withOneOfs(): ImmutableList<Field> {
```
Egorand: How about `fieldsWithOneOfs()`? This method name is confusing.

I see where `ImmutableList` is coming from.. Can we wrap it into a `listOf()` to avoid referencing this type?

wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/OneOfMessage.kt.java.interop
```diff
@@ -0,0 +1,115 @@
...
+import okio.ByteString
+
+data class OneOfMessage(
+    @WireField(tag = 1, adapter = "com.squareup.wire.ProtoAdapter.INT32") val foo: Int? = null,
+    @WireField(tag = 3, adapter = "com.squareup.wire.ProtoAdapter.STRING") val bar: String? = null,
+    @WireField(tag = 4, adapter = "com.squareup.wire.ProtoAdapter.STRING") val baz: String? = null,
+    val unknownFields: ByteString = ByteString.EMPTY
+) : Message<OneOfMessage, OneOfMessage.Builder>(ADAPTER, unknownFields) {
+    init {
+        if (Internal.countNonNull(foo, bar, baz) > 1) {
```
Egorand: Let's generate a `require()`:
```kotlin
require(Internal.countNonNull(foo, bar, baz) <= 1) {
  "At most one of foo, bar and baz may be non-null"
}
```

wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/OneOfMessage.kt.java.interop
```diff
@@ -0,0 +1,115 @@
...
+
+    class Builder : Message.Builder<OneOfMessage, Builder>() {
+        internal var foo: Int? = null
+
+        internal var bar: String? = null
+
+        internal var baz: String? = null
+
+        /**
+         * What foo. */
```
Egorand: Looks like we should add `\n` when generating KDoc..

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
mc-lovin: yes it did, will move it out.

Egorand:
Nice! @mc-lovin do you wanna tackle the sealed classes version as well?

Issue #728, opened 2018-09-30
Kotlin generator todos


mc-lovin:
- [ ]   Options support
- [x]   Add oneof support
- [x]   Add Map support
- [ ]   Add Custom Adapter support
- [x]  replace find with an O(1) operation
- [x] use kotlinpoet for generating lambdas
- [ ] Generate default fields

Egorand:
@mc-lovin I can start working on Map support if you don't mind

Egorand:
@mc-lovin what is "use kotlinpoet for generating lambdas" about?

mc-lovin:
@Egorand sorry missed this comment, This was about generating this block 

```
        val unknownFields = reader.forEachTag { tag ->
                        when (tag) {
                            1 -> number = ProtoAdapter.STRING.decode(reader)
                            2 -> type = PhoneType.ADAPTER.decode(reader)
                            else -> TagHandler.UNKNOWN_TAG
                        }
                    }
```
using updated kotlin poet functions instead of using the generic `addStatment` for generating the lambda

Egorand:
Got it, I can work on this. I'm pretty sure though we're already using the latest APIs, but can make sure.

Egorand:
Yeah, we're good, as we didn't introduce any new APIs in KotlinPoet to support this - it was an implementation change that's present in 1.0.0.

Egorand:
I'm gonna close this issue and open separate issues for each TODO

PR #750, opened 2018-11-16
Add Map support to Kotlin generator

Egorand:
Part of #728 

Missing support for `RuntimeMessageAdapter`, which is not compatible with Kotlin protos as is.


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -286,23 +286,27 @@ class KotlinGenerator private constructor(
...
-      var fieldClass: TypeName = typeName(field.type())
+      var fieldClass: TypeName
       val fieldName = nameAllocator.get(field)
       var defaultValue = CodeBlock.of("null")
 
       when {
-        field.isOptional -> {
-          fieldClass = fieldClass.asNullable()
-        }
+        field.isOptional -> fieldClass = typeName(field.type()).asNullable()
```
mc-lovin: nit can declare a variable for `field.type()`

wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/map/Mappy.kt
```diff
@@ -0,0 +1,60 @@
...
+            override fun encodedSize(value: Mappy): Int = thingsAdapter.encodedSizeWithTag(1, value.things) +
+                value.unknownFields.size
+
+            override fun encode(writer: ProtoWriter, value: Mappy) {
+                thingsAdapter.encodeWithTag(writer, 1, value.things)
+                writer.writeBytes(value.unknownFields)
+            }
+
+            override fun decode(reader: ProtoReader): Mappy {
+                var things = mutableMapOf<String, Thing>()
```
mc-lovin: should this be val?

mc-lovin: if yes should be changed for list declarations as well, can be done in a separate pr

wire-tests/src/test/java/com/squareup/wire/KotlinMapTest.kt
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright 2016 Square Inc.
```
mc-lovin: 2018

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -362,13 +366,28 @@ class KotlinGenerator private constructor(
         .addFunction(encodedSizeFun(type))
         .addFunction(encodeFun(type))
         .addFunction(decodeFun(type))
-        .build()
+
+    for (field in type.fields()) {
+      if (field.type().isMap) {
+        val adapterType = ProtoAdapter::class.asTypeName().parameterizedBy(Map::class.asTypeName()
+            .parameterizedBy(typeName(field.type().keyType()), typeName(field.type().valueType())))
```
mc-lovin: nit can extract `keyType` and `valueType` into variables

wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/map/Mappy.kt
Egorand: Yes, I'll open an issue

PR #754, opened 2018-11-19
Replace manual indents with the %> modifier

Egorand:


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -485,19 +480,17 @@ class KotlinGenerator private constructor(
...
     }
 
     val tagHandlerClass = ClassName("com.squareup.wire", "TagHandler")
 
-    decodeBlock.addStatement("%Lelse -> %T.%L", indentation.repeat(2), tagHandlerClass,
-        "UNKNOWN_TAG")
-    decodeBlock.addStatement("%L}", indentation)
-    decodeBlock.addStatement("}")
+    decodeBlock.addStatement("else -> %T.%L", tagHandlerClass, "UNKNOWN_TAG")
+    decodeBlock.add("%<}\n".repeat(2)) // close the block
```
JakeWharton: Not sure I like the use of `repeat(2)` here. It's easy to miss.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -485,19 +480,17 @@ class KotlinGenerator private constructor(
       }
 
       declarationBody.addStatement("%L", fieldDeclaration)
-      decodeBlock.addStatement(decodeBodyTemplate, indentation.repeat(2), field.tag(), fieldName,
-          adapterName)
-      returnBody.add("%L%L = %L%L,\n", indentation, fieldName, fieldName, throwExceptionBlock)
+      decodeBlock.addStatement(decodeBodyTemplate, field.tag(), fieldName, adapterName)
+      returnBody.addStatement("%L = %L%L,", fieldName, fieldName, throwExceptionBlock)
```
JakeWharton: nit: could use numbers to avoid passing same thing twice

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
Egorand: Fair point, can change

PR #755, opened 2018-11-19
O(1) implementation of WireEnum.fromValue()

Egorand:
Part of #728


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -586,7 +586,15 @@ class KotlinGenerator private constructor(
             .addModifiers(OVERRIDE)
             .addParameter(valueName, Int::class)
             .returns(parentClassName.asNullable())
```
JakeWharton: Why nullable?

Egorand: Shouldn't be, will fix

swankjesse:
Are switches O(1) ?!

PR #775, opened 2018-11-28
Write a new .proto file loader that will work as WireRun prescribes

swankjesse:
It's more flexible in how inputs are provided - fully-qualified paths,
directories, or .zips. And it validates that the parsed paths are in
the expected directories.

The Root class keeps track of a user-provided directory where .proto
files are loaded.

The NewSchemaLoader loads the user's provided sources, and the protos
they import.

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/schema/Root.kt
+++ b/wire-compiler/src/main/java/com/squareup/wire/schema/Root.kt
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2018 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire.schema
+
+import com.google.common.io.Closer
+import java.nio.file.FileSystems
+import java.nio.file.FileVisitResult
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.ProviderNotFoundException
+import java.nio.file.SimpleFileVisitor
+import java.nio.file.attribute.BasicFileAttributes
+
+internal sealed class Root {
+  /** Returns all proto files within this root. */
+  abstract fun allProtoFiles(): Set<LocationAndPath>
+
+  /** Returns the proto file if it's in this root, or null if it isn't. */
+  abstract fun resolve(import: String): LocationAndPath?
+}
+
+internal fun Path.roots(closer: Closer): List<Root> {
+  return when {
+    Files.isDirectory(this) -> listOf(DirectoryRoot(this, this))
+
+    endsWithDotProto() -> listOf(StandaloneFileRoot(this))
+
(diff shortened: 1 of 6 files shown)


wire-compiler/src/main/java/com/squareup/wire/schema/Root.kt
```diff
@@ -0,0 +1,91 @@
...
+        val sourceFs = FileSystems.newFileSystem(this, javaClass.classLoader)
+        closer.register(sourceFs)
+        sourceFs.rootDirectories.map { inZipDirectory -> DirectoryRoot(this, inZipDirectory) }
+      } catch (e: ProviderNotFoundException) {
+        throw IllegalArgumentException(
+            "expected a directory, archive (.zip / .jar / etc.), or .proto: $this")
+      }
+    }
+  }
+}
```
jrodbx: nice

wire-compiler/src/test/java/com/squareup/wire/schema/NewSchemaLoaderTest.kt
```diff
@@ -0,0 +1,121 @@
...
+        |import "squareup/polygons/triangle.proto";
+        |message Circle {
+        |}
+        """.trimMargin(),
+        "squareup/curves/oval.proto" to """
+        |syntax = "proto2";
+        |package squareup.curves;
+        |message Oval {
+        |}
+        """.trimMargin())
```
jrodbx: 👌

wire-compiler/src/test/java/com/squareup/wire/schema/NewSchemaLoaderTest.kt
```diff
@@ -0,0 +1,121 @@
...
+
+    val sourcePath = listOf("colors/src/main/proto")
+    val exception = assertFailsWith<IllegalArgumentException> {
+      NewSchemaLoader(fs, sourcePath).use { it.load() }
+    }
+    assertThat(exception).hasMessage("expected colors/src/main/proto/squareup/shapes/blue.proto " +
+        "to have a path ending with squareup/colors/blue.proto")
+  }
```
jrodbx: good edge case

wire-compiler/src/test/java/com/squareup/wire/schema/RootTest.kt
```diff
@@ -0,0 +1,155 @@
...
+import java.nio.file.FileSystem
+import java.nio.file.Files
+import java.util.zip.ZipEntry
+import java.util.zip.ZipOutputStream
+import kotlin.test.assertFailsWith
+import kotlin.text.Charsets.UTF_8
+
+class RootTest {
+  private val fs = Jimfs.newFileSystem(Configuration.unix())
+  private val closer = Closer.create()
```
jrodbx: nit: explicit setup in @Before?

wire-compiler/src/main/java/com/squareup/wire/schema/Root.kt
```diff
@@ -0,0 +1,91 @@
...
+internal fun Path.roots(closer: Closer): List<Root> {
+  return when {
+    Files.isDirectory(this) -> listOf(DirectoryRoot(this, this))
+
+    endsWithDotProto() -> listOf(StandaloneFileRoot(this))
+
+    // Handle a .zip or .jar file by adding all .proto files within.
+    else -> {
+      try {
+        val sourceFs = FileSystems.newFileSystem(this, javaClass.classLoader)
```
jrodbx: TIL

jrodbx:
LGTM!

wire-compiler/src/test/java/com/squareup/wire/schema/RootTest.kt
swankjesse: I think it’s less code as-is? But I agree it’s asymmetric

Issue #776, opened 2018-11-28
StackoverflowError when model have circular reference


killagu:
# pb
```
//CardModelEntryPB.proto
import "ChildModelEntryPB.proto";
package com_foo_alert;

option java_package = "com.foo.alert";

message CardModelEntryPB {
  optional string cardTypeId = 1;
  optional string cellId = 2;
  optional ChildModelEntryPB childModel = 3;
}
```

```
//ChildModelEntryPB.proto
import "CardModelEntryPB.proto";
package com_foo_alert;

option java_package = "com.foo.alert";

message ChildModelEntryPB {
  repeated string cells = 1;
  repeated CardModelEntryPB children = 2;
  optional int64 serverTimestamp = 3;
}
```

```
// protos.include
ChildModelEntryPB.proto
CardModelEntryPB.proto
```

# cmd
```
java -jar wire-compiler/target/wire-compiler-1.8.0-jar-with-dependencies.jar \
--proto_path=pb \
--java_out=out \
--files=protos.include
```

# Error
```
Exception in thread "main" java.lang.StackOverflowError
	at java.lang.StringCoding$StringEncoder.encode(StringCoding.java:304)
	at java.lang.StringCoding.encode(StringCoding.java:344)
	at java.lang.String.getBytes(String.java:918)
	at java.io.FileInputStream.open0(Native Method)
	at java.io.FileInputStream.open(FileInputStream.java:195)
	at java.io.FileInputStream.<init>(FileInputStream.java:138)
	at java.io.FileInputStream.<init>(FileInputStream.java:93)
	at com.squareup.wire.IO$FileIO.parse(IO.java:41)
	at com.squareup.wire.WireCompiler.loadSymbolsHelper(WireCompiler.java:621)
	at com.squareup.wire.WireCompiler.loadSymbolsHelper(WireCompiler.java:622)
        .......
	at com.squareup.wire.WireCompiler.loadSymbolsHelper(WireCompiler.java:622)
```

# Ref code
```Java
  // WireCompiler.java
  // Call with pass == LOAD_TYPES, then pass == LOAD_FIELDS
  private void loadSymbolsHelper(ProtoFile protoFile, Set<String> loadedDependencies,
      LoadSymbolsPass pass) throws IOException {
    // Load symbols from imports
    for (String dependency : protoFile.dependencies()) {
      if (!loadedDependencies.contains(dependency)) {
        String dep = repoPath + File.separator + dependency;
[...]

PR #777, opened 2018-11-28
fix: fix stackoverflow when model have circular reference

killagu:
Fixes: https://github.com/square/wire/issues/776


oldergod:
I deleted the target branch wire_18. Please reopen the PR targeting master if that's still a problem.

PR #784, opened 2018-12-06
Add one-of support via sealed classes

mc-lovin:


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -144,9 +152,39 @@ class KotlinGenerator private constructor(
...
     return classBuilder.build()
   }
 
+  private fun generateOneOfClass(type: MessageType, oneOf: OneOf): TypeSpec {
+    val nameAllocator = nameAllocator(type)
+    val oneOfClassName = nameAllocator[oneOf].beginWithUpperCase()
+    val oneOfClassType = type.oneOfClass(oneOf)
+    val builder = TypeSpec.classBuilder(oneOfClassName)
+        .addModifiers(SEALED)
+    oneOf.fields().forEach {
```
oldergod: How about naming the parameter when the block is that long?

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -351,6 +389,19 @@ class KotlinGenerator private constructor(
           .build())
     }
 
+    if (!javaInterOp) {
+      message.oneOfs().forEach {
```
oldergod: name parameter please

wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/OneOfMessage.kt
```diff
@@ -0,0 +1,79 @@
...
+            OneOfMessage::class.java
+        ) {
+            override fun encodedSize(value: OneOfMessage): Int = 
+                when (value.choice) {
+                    is Choice.Foo -> ProtoAdapter.INT32.encodedSizeWithTag(1, value.choice.foo)
+                    is Choice.Bar -> ProtoAdapter.STRING.encodedSizeWithTag(3, value.choice.bar)
+                    is Choice.Baz -> ProtoAdapter.STRING.encodedSizeWithTag(4, value.choice.baz)
+                    else -> 0
+                } +
+                value.unknownFields.size
```
oldergod: nit: maybe that'd be easier to read if this line was first?
something like
```kotlin
value.unknownFields.size + when (value.choice) {
    is Choice.Foo -> ProtoAdapter.INT32.encodedSizeWithTag(1, value.choice.foo)
    is Choice.Bar -> ProtoAdapter.STRING.encodedSizeWithTag(3, value.choice.bar)
    is Choice.Baz -> ProtoAdapter.STRING.encodedSizeWithTag(4, value.choice.baz)
    else -> 0
}
```

oldergod: but on the other hand, we want unknowFields at the end

wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/OneOfMessage.kt
```diff
@@ -0,0 +1,79 @@
...
+            override fun decode(reader: ProtoReader): OneOfMessage {
+                var choice: Choice? = null
+                val unknownFields = reader.forEachTag { tag ->
+                    when (tag) {
+                        1 -> choice = Choice.Foo(ProtoAdapter.INT32.decode(reader))
+                        3 -> choice = Choice.Bar(ProtoAdapter.STRING.decode(reader))
+                        4 -> choice = Choice.Baz(ProtoAdapter.STRING.decode(reader))
+                        else -> TagHandler.UNKNOWN_TAG
+                    }
+                }
```
oldergod: the whole block is hard to read to me because of the side-effects mix but I don't have a better alternative right now.

wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/OneOfMessage.kt
Egorand: Plus, this is just an edge case where the proto has a single field. Formatting multiple `oneof` fields in a meaningful way would be interesting.

wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/OneOfMessage.kt
oldergod: Specially I'm confused that the block assigns stuff but also returns a value

mc-lovin: right the way we collect unknown fields here is indeed al little hacky. I will think about this more. 

maybe instead of returning the assigned value or  `TagHandler.UNKNOWN_TAG` we can do a precheck to see if the tag belongs to the list of used tags

PR #787, opened 2018-12-14
Add defaults to WireRun

jrodbx:


jrodbx:
My bad.  Figured it was small/trivial enough to move forward.  Happy to revert if you think it's not the right approach.

PR #788, opened 2018-12-19
Apply java_package option when generating output

jrodbx:
Copied from SchemaLoader with an added test

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/schema/NewSchemaLoader.kt
+++ b/wire-compiler/src/main/java/com/squareup/wire/schema/NewSchemaLoader.kt
@@ -16,6 +16,10 @@
 package com.squareup.wire.schema
 
 import com.google.common.io.Closer
+import com.squareup.wire.schema.SchemaLoader.DESCRIPTOR_PROTO
+import com.squareup.wire.schema.internal.parser.ProtoParser
+import okio.buffer
+import okio.source
 import java.io.Closeable
 import java.io.IOException
 import java.nio.file.FileSystem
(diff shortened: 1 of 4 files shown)


wire-compiler/src/main/java/com/squareup/wire/schema/NewSchemaLoader.kt
```diff
@@ -52,6 +56,18 @@ class NewSchemaLoader(
...
+//    val jarFileSystem = FileSystems.newFileSystem(resourceUri, env)
+//    val resourceRoots =
+//        allRoots(closer, listOf(jarFileSystem.getPath(DESCRIPTOR_PROTO).toString()))
+//    for (resourceRoot in resourceRoots) {
+//      for (locationAndPath in resourceRoot.allProtoFiles()) {
+//        load(locationAndPath)
+//      }
+//    }
+//    loaded[DESCRIPTOR_PROTO] = loadDescriptorProto()
+
```
jrodbx: original attempt that didn't work; plan was to create a new `ResourceRoot`.

wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
```diff
@@ -145,13 +145,16 @@ data class WireRun(
     val schemaLoader = NewSchemaLoader(fs, sourcePath, protoPath)
     val protoFiles = schemaLoader.load()
 
-    // 2. Validate the schema and resolve references
-    val fullSchema = Linker(protoFiles).link()
+    // 2. Load descriptor proto
+    val protoFilesPlusDescriptor = protoFiles.plus(schemaLoader.loadDescriptorProto())
```
jrodbx: could have put this in NewSchemaLoader#load, but felt side-effecty and broke current tests.  Open to moving it back.

wire-schema/src/main/java/com/squareup/wire/schema/SchemaLoader.java
```diff
@@ -49,7 +49,7 @@
  * relative to the root of the archive.
  */
 public final class SchemaLoader {
-  private static final String DESCRIPTOR_PROTO = "google/protobuf/descriptor.proto";
+  static final String DESCRIPTOR_PROTO = "google/protobuf/descriptor.proto";
```
jrodbx: opted to weaken this vs copy into NewSchemaLoader for now.  Open to moving now.

PR #793, opened 2019-01-02
Always emit WireField annotations for Kotlin properties

### LINKED ISSUE #780: Add WireField annotations for non-interop mode
`WireField` annotations are only emitted in `java_interop` mode, do we need them in Kotlin-only mode? Should we put them behind a custom flag? Do we need them at all?

Egorand:
Fixes #780

### DIFF
--- a/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
+++ b/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
@@ -377,12 +377,10 @@ class KotlinGenerator private constructor(
         parameterSpec.defaultValue(defaultValue)
       }
 
-      if (javaInterOp) {
-        parameterSpec.addAnnotation(AnnotationSpec.builder(WireField::class)
-            .addMember("tag = %L", field.tag())
-            .addMember("adapter = %S", getAdapterName(field))
-            .build())
-      }
+      parameterSpec.addAnnotation(AnnotationSpec.builder(WireField::class)
+          .addMember("tag = %L", field.tag())
+          .addMember("adapter = %S", getAdapterName(field))
+          .build())
 
       constructorBuilder.addParameter(parameterSpec.build())
       classBuilder.addProperty(PropertySpec.builder(fieldName, fieldClass)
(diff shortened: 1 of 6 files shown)


wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt
```diff
@@ -21,10 +22,11 @@ import kotlin.jvm.JvmStatic
 import okio.ByteString
 
 data class Person(
-    val name: String,
-    val id: Int,
-    val email: String? = null,
-    val phone: List<PhoneNumber> = emptyList(),
+    @WireField(tag = 1, adapter = "com.squareup.wire.ProtoAdapter.STRING") val name: String,
+    @WireField(tag = 2, adapter = "com.squareup.wire.ProtoAdapter.INT32") val id: Int,
+    @WireField(tag = 3, adapter = "com.squareup.wire.ProtoAdapter.STRING") val email: String? =
```
jrodbx: nit: `val email...` on new line

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -377,12 +377,10 @@ class KotlinGenerator private constructor(
...
-      if (javaInterOp) {
-        parameterSpec.addAnnotation(AnnotationSpec.builder(WireField::class)
-            .addMember("tag = %L", field.tag())
-            .addMember("adapter = %S", getAdapterName(field))
-            .build())
-      }
+      parameterSpec.addAnnotation(AnnotationSpec.builder(WireField::class)
+          .addMember("tag = %L", field.tag())
+          .addMember("adapter = %S", getAdapterName(field))
+          .build())
```
jrodbx: Should the annotation be limited to the field element via [use-site targets](https://kotlinlang.org/docs/reference/annotations.html#annotation-use-site-targets)?

wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt
Egorand: Yeah, that's KotlinPoet

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
Egorand: Yeah, guess that makes sense

PR #804, opened 2019-01-18
Fix Kotlin generated adapter names (issue #802)

### LINKED ISSUE #802: Kotlin generator: enum from another message creates broken code
The following example should reproduce the issue:

```
message TestOne {
  optional string id = 1;
  enum TestEnum {
    ENTRY = 1;
    ANOTHER = 2;
  }
}

message TestTwo {
  optional TestOne.TestEnum e = 2;
}
```

The enum is not fully qualified in the generated TestTwo class.

andrew-arnott:
Also a minor fix in here that happened accidentally while I was doing a little refactoring. Check out the first commit.

Fixes #802


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -67,13 +67,13 @@ class KotlinGenerator private constructor(
   private val emitAndroid: Boolean,
   private val javaInterOp: Boolean
 ) {
-  val nameAllocatorStore = mutableMapOf<Type, NameAllocator>()
+  private val nameAllocatorStore = mutableMapOf<Type, NameAllocator>()
```
Egorand: Yep, this should've been private.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -67,13 +67,13 @@ class KotlinGenerator private constructor(
...
 
-  /** Returns the Kotlin type for [protoType]. */
-  fun typeName(protoType: ProtoType) = nameToKotlinName.getValue(protoType)
+  /** Returns the Kotlin type for [this]. */
+  private fun ProtoType.typeName() = nameToKotlinName.getValue(this)
```
Egorand: Nice, I like this better

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -873,8 +878,10 @@ class KotlinGenerator private constructor(
 
   private fun MessageType.oneOfClass(oneOf: OneOf): ClassName {
     val name = oneOf.name().capitalize()
-    return typeName(type()).nestedClass(name)
+    return type().typeName().nestedClass(name)
   }
 }
 
+
```
Egorand: nit: empty lines

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -812,6 +791,32 @@ class KotlinGenerator private constructor(
...
+  private fun Field.getDefaultValue(): CodeBlock {
+    return when {
+      isRepeated -> CodeBlock.of("emptyList()")
+      type().isMap -> CodeBlock.of("emptyMap()")
+      default != null && isEnum() -> CodeBlock.of("%T.%L", type().typeName(), default)
+      default != null && isRepeated -> {
+        val internalClass = ClassName("com.squareup.wire.internal", "Internal")
+        CodeBlock.of("%T.newMutableList()", internalClass)
+      }
+      default != null ->CodeBlock.of("%L", default)
```
Egorand: nit: space after `->`

wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt
```diff
@@ -26,7 +26,7 @@ data class Person(
     @field:WireField(tag = 2, adapter = "com.squareup.wire.ProtoAdapter.INT32") val id: Int,
     @field:WireField(tag = 3, adapter = "com.squareup.wire.ProtoAdapter.STRING") val email: String?
             = null,
-    @field:WireField(tag = 4, adapter = "PhoneNumber.ADAPTER") val phone: List<PhoneNumber> =
+    @field:WireField(tag = 4, adapter = "Person.PhoneNumber.ADAPTER") val phone: List<PhoneNumber> =
```
Egorand: There's no name clash here, so I don't think we should use FQ name

wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt.java.interop
```diff
@@ -45,7 +45,7 @@ data class Person(
 
         internal var email: String? = null
 
-        internal var phone: List<PhoneNumber> = Internal.newMutableList()
+        internal var phone: List<PhoneNumber> = emptyList()
```
Egorand: Was there a change in the generator that fixes this?

Egorand:
The PR is hard to read because it mixes refactoring with the actual bug fix. I like the refactoring part! Let's split this into a separate PR and that should be an easy merge. It'll also make it easier to understand what the fix was: I don't think it's correct since it shouldn't have affected existing protos, but I'd like to take another look once it's in its own PR.

andrew-arnott:
I rebased this off of #805 so the last commit contains the actual fix to issue #802. See https://github.com/square/wire/pull/804/commits/a7b51a12736f28e3b0127740c26eb1d2177aea24 for the changes.

I don't know Kotlin poet well enough to figure out how to make it smart enough to only qualify the names as much as is necessary. I noticed that the actual enum usages already do that, but the enum's adapter references were no qualified correctly (which is where I made them fully qualified).

andrew-arnott:
Was there any changes I need to make to get this merged?

I'm not sure I can get this working without FQ names right now (until the NameAllocator in KotlinPoet is improved, at least).

wire-tests/src/test/proto/kotlin/same_name_enum.proto
```diff
@@ -0,0 +1,21 @@
+package squareup.protos.kotlin;
```
Egorand: nit: copyright

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -654,7 +654,15 @@ class KotlinGenerator private constructor(
     return when {
       isScalar -> CodeBlock.of("%T.%L", ProtoAdapter::class, simpleName().toUpperCase(Locale.US))
       isMap -> throw IllegalArgumentException("Can't create single adapter for map type $this")
-      else -> CodeBlock.of("%L.ADAPTER", typeName.simpleName)
+      else -> CodeBlock.of("%L.ADAPTER", typeName.getQualifiedName())
```
Egorand: I think using `%T` here instead of `%L` is the right thing to do. This will automatically figure out the shortest unambiguous name to qualify the type.

PR #805, opened 2019-01-19
Refactor KotlinGenerator

andrew-arnott:
I added some more minor changes to the refactoring, but this was the first part of my other PR (#804).

### DIFF
--- a/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/FieldExtensions.kt
+++ b/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/FieldExtensions.kt
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2019 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Extension functions/properties for the Field class
+ */
+package com.squareup.wire.kotlin
+
+import com.squareup.wire.schema.Field
+import com.squareup.wire.schema.ProtoType
+
+val Field.isMap: Boolean
+  get() = type().isMap
+
+val Field.keyType: ProtoType
+  get() = type().keyType()
+
+val Field.valueType: ProtoType
+  get() = type().valueType()
--- a/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/FieldExtensions.kt
+++ b/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/FieldExtensions.kt
@@ -22,11 +22,11 @@ package com.squareup.wire.kotlin
 import com.squareup.wire.schema.Field
 import com.squareup.wire.schema.ProtoType
 
-val Field.isMap: Boolean
+internal val Field.isMap: Boolean
   get() = type().isMap
 
-val Field.keyType: ProtoType
+internal val Field.keyType: ProtoType
   get() = type().keyType()
 
-val Field.valueType: ProtoType
+internal val Field.valueType: ProtoType
   get() = type().valueType()
--- a/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
+++ b/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
@@ -831,14 +831,22 @@ class KotlinGenerator private constructor(
 
   private fun Field.getDefaultValue(): CodeBlock {
     return when {
-      isRepeated && !javaInterOp -> CodeBlock.of("emptyList()")
-      isRepeated && javaInterOp -> {
-        val internalClass = ClassName("com.squareup.wire.internal", "Internal")
-        CodeBlock.of("%T.newMutableList()", internalClass)
+      isRepeated -> {
+        if (javaInterOp) {
+          val internalClass = ClassName("com.squareup.wire.internal", "Internal")
+          CodeBlock.of("%T.newMutableList()", internalClass)
+        } else {
+          CodeBlock.of("emptyList()")
+        }
       }
       isMap -> CodeBlock.of("emptyMap()")
-      default != null && isEnum -> CodeBlock.of("%T.%L", type().typeName, default)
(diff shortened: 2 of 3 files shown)


wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt.java.interop
```diff
@@ -25,7 +25,7 @@ data class Person(
     @field:WireField(tag = 3, adapter = "com.squareup.wire.ProtoAdapter.STRING") val email: String?
             = null,
     @field:WireField(tag = 4, adapter = "PhoneNumber.ADAPTER") val phone: List<PhoneNumber> =
-            emptyList(),
+            Internal.newMutableList(),
```
andrew-arnott: Can you verify if this new output is actually correct?

wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt
```diff
@@ -119,7 +119,7 @@ data class Person(
     data class PhoneNumber(
         @field:WireField(tag = 1, adapter = "com.squareup.wire.ProtoAdapter.STRING") val number:
                 String,
-        @field:WireField(tag = 2, adapter = "PhoneType.ADAPTER") val type: PhoneType =
+        @field:WireField(tag = 2, adapter = "PhoneType.ADAPTER") val type: PhoneType? =
```
andrew-arnott: This type is optional in the proto, but was being generated as non-nullable. My refactoring ended up catching that and fixing it by accident :)

Egorand: Good catch!

wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt.java.interop
Egorand: Sure, `./gradlew clean :wire-kotlin-generator:check :wire-compiler:check`

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/FieldExtensions.kt
```diff
@@ -0,0 +1,25 @@
+/**
```
Egorand: nit: Copyright

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/FieldExtensions.kt
```diff
@@ -0,0 +1,25 @@
...
+ * Extension functions/properties for the Field class
+ */
+package com.squareup.wire.kotlin
+
+import com.squareup.wire.schema.Field
+import com.squareup.wire.schema.ProtoType
+
```
Egorand: Neat! I like the extensions that fold a level of nesting, e.g. `type.valueType() -> valueType`, but I think introducing property extensions just to turn a method into a property is a bit of an overkill. Plus, we might kotlinize the schema module in future and will turn those into properties anyway.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -443,24 +425,22 @@ class KotlinGenerator private constructor(
...
-                .initializer(
-                    "%T.newMapAdapter(%L, %L)",
-                    ProtoAdapter::class,
-                    getAdapterName(keyType),
-                    getAdapterName(valueType)
-                )
-                .build())
-      }
-    }
+    type.fields()
```
Egorand: I think the old version was fine and the new one is not necessarily better: although not really an issue here, functional operators create copies of the collection, so they're more memory-heavy and slower. I'd suggest reverting this piece.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/FieldExtensions.kt
andrew-arnott: I see what you mean about it being a bit overkill. I feel like it can make code a bit more natural to work with, however (properties are data that the object owns, functions do some action). That's at least what I was thinking when I wrote this. But it is extra code to read that doesn't do very much. I can remove them.

What did you think about the other uses of properties where it isn't just replacing a function with the same name?

andrew-arnott:
Okay -- this PR is now just a refactor. The fix to the optional type not being generated correctly will come in a separate PR after this is merged.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/FieldExtensions.kt
Egorand: Yeah, I like those. Let's keep the other ones for consistency. If we decide to rewrite the schema module in Kotlin we can use these extensions. Just one thing - please make these extensions `internal`, we don't want them in the API.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -778,47 +807,54 @@ class KotlinGenerator private constructor(
...
 
   private fun Field.getClass(baseClass: TypeName = nameToKotlinName.getValue(type())) = when {
     isRepeated -> List::class.asClassName().parameterizedBy(baseClass)
     isOptional && default == null -> baseClass.copy(nullable = true)
     else -> baseClass.copy(nullable = false)
   }
 
+  private fun Field.getDefaultValue(): CodeBlock {
+    return when {
+      isRepeated && !javaInterOp -> CodeBlock.of("emptyList()")
```
Egorand: Some of these cases look repetitive, would that work better with nested conditionals?
```kotlin
return when {
  isRepeated -> {
    if (javaInterOp) {
      ...
    } else {
      ...
    }
  }
}
```

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -778,47 +807,54 @@ class KotlinGenerator private constructor(
...
+      isRepeated && javaInterOp -> {
+        val internalClass = ClassName("com.squareup.wire.internal", "Internal")
+        CodeBlock.of("%T.newMutableList()", internalClass)
```
Egorand: I think we should replace it with `mutableListOf()`, but that's a separate PR.

Egorand:
@andrew-arnott I wanna merge this first and ask you to rebase the other PRs on top of the new master - this should eliminate duplicate changes and make them easier to review, plus will make sure that Git history is clean. All of the feedback above is optional, lmk if you're with merging this as is, or if you'd like to implement the changes.

andrew-arnott:
Okay -- made the changes you suggested. Should be good to merge after it builds.

PR #806, opened 2019-01-19
Fix KotlinGenerator issue where optional was not always nullable

andrew-arnott:
Check out the second commit to see the actual changes this makes. The first one is in a separate PR, but this commit depends on that refactoring work.

### DIFF
--- a/wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt.java.interop
+++ b/wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt.java.interop
@@ -45,7 +45,7 @@ data class Person(
 
         internal var email: String? = null
 
-        internal var phone: List<PhoneNumber> = Internal.newMutableList()
+        internal var phone: List<PhoneNumber> = emptyList()
 
         /**
          * The customer's full name.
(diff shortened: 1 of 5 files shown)


Egorand:
I'm looking at the second commit and it's not at all clear what the fix is. Can you please rebase this onto #805, so that this PR only contains the bug fix and necessary test updates?

Egorand:
Please rebase onto the new master so that this whole PR only contains changes relevant to the fix.

wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt.java.interop
```diff
@@ -25,7 +25,7 @@ data class Person(
     @field:WireField(tag = 3, adapter = "com.squareup.wire.ProtoAdapter.STRING") val email: String?
             = null,
     @field:WireField(tag = 4, adapter = "PhoneNumber.ADAPTER") val phone: List<PhoneNumber> =
-            emptyList(),
+            Internal.newMutableList(),
```
Egorand: Guess I know where this is coming from: the default value for a repeated field is calculated as
```kotlin
isRepeated -> {
  if (javaInterOp) {
    val internalClass = ClassName("com.squareup.wire.internal", "Internal")
    CodeBlock.of("%T.newMutableList()", internalClass)
  } else {
    CodeBlock.of("emptyList()")
  }
}
```
I think we should always use `emptyList()`, even in Java interop mode, Wire compiler will still depend on kotlin-stdlib, so that's fine. Could you please change that and regenerate the tests?

andrew-arnott:
Changes are in -- should be good to merge.

PR #808, opened 2019-01-22
Support enclosing types

jrodbx:
Example:
```
message A {
  message B {
  }
  optional B b = 1;
}
```
schema.prune(include("A.B")) -> A == enclosing type


jrodbx:
Agreed that the diffs were a bit confusing.  On the flip side, since the change was localized to the generators, I thought unit testing those classes directly was the cleaner way to surface the problem.

PR #810, opened 2019-01-23
Allow reflection via Java interop

jrodbx:


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -288,12 +287,16 @@ class KotlinGenerator private constructor(
     type.fieldsWithJavaInteropOneOfs().forEach { field ->
       val fieldName = nameAllocator[field]
 
+      val propertyBuilder = PropertySpec.builder(fieldName, field.declarationClass)
+          .mutable(true)
```
Egorand: nit: `true` can be omitted

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -288,12 +287,16 @@ class KotlinGenerator private constructor(
...
+          .initializer(field.getDefaultValue())
+
+      if (javaInterOp) {
+        propertyBuilder.addAnnotation(JvmField::class)
+      }
+
       builder
-          .addProperty(PropertySpec.builder(fieldName, field.declarationClass)
```
Egorand: Pretty sure this code only runs in interop, since I don't see any changes to non-interop tests. Would it make sense to just change the code in place?

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
jrodbx: `.mutable(true)` was there before; i kinda like the explicitness

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
jrodbx: According to this:
```
  private fun MessageType.fieldsWithJavaInteropOneOfs(): List<Field> {
    return if (javaInterOp)
      fieldsAndOneOfFields()
    else
      fields()
  }
```

this code could still run in non-interop mode.

PR #811, opened 2019-01-24
Refactor KotlinGenerator by splitting up functions into classes

andrew-arnott:


andrew-arnott:
No functional changes. This is completely just splitting things into classes and organizing the code.

andrew-arnott:
@Egorand or @swankjesse --thoughts on this PR?

jrodbx:
Thanks for the PR.  I think we prefer the current file organization though; it mirrors JavaGenerator and allows other potential generators to follow suit.  Also, feel free to file issues to start a discussion before introducing structural changes.

swankjesse:
Can you describe the benefits here? It seems like it's just moving things into different files.

andrew-arnott:
@swankjesse Yup -- that's pretty much all that the change is. I intentionally tried to keep it minimal.

The KotlinGenerator class is a ~900 line monstrosity. When I was working with it, I noticed some pretty obvious distinct types that could be isolated, so I pulled them out into their own internal classes.

The obvious benefit is that this change creates more cohesive units. It's not perfect, but I feel it starts putting this code on the right track. Eventually some unit tests could be created against the various internal classes to isolate checking correctness of individual parts of the generated protos.

PR #819, opened 2019-02-04
Delete wire-maven-plugin

Egorand:


jrodbx:
Do we know if anyone in the community still uses this?

swankjesse:
I assume this is used. I'd prefer a deprecation phase. It doesn't need to ever support Kotlin though!

Egorand:
Agree, declining. Let's discuss deprecation plan offline.

PR #823, opened 2019-02-04
Promotes gRPC related code to wire-runtime

oldergod:


wire-runtime/build.gradle
```diff
@@ -23,6 +23,8 @@ animalsniffer {
 
 dependencies {
   api deps.okio
+  api deps.okhttp
+  api deps.kotlin.coroutines.core
```
JakeWharton: Yuck! Why?

oldergod: Trying to see where all these could fit. Maybe we can have a `wire-grpc-runtime`

JakeWharton: This is a _very_ specific implementation of gRPC built on OkHttp. It only supports protos whereas gRPC is agnostic. It only supports coroutines whereas gRPC doesn't care. I would consider not even hosting it in this repo since it's an integration of 3 specific libraries into a new one.

swankjesse: I don’t mind it being specific, I want an _opinionated_ implementation of gRPC.

It’s true that gRPC supports more than protos, but Wire doesn’t care about those; Wire is a protos library! And I think Google is mistaken by treating gRPC and protos as separate projects; you need a `.proto` file to declare a gRPC endpoint! Is there any serious use of gRPC with other payload types?

That said, the dependency on OkHttp is slightly annoying, especially as servers that use Wire will drag this in. Maybe we want `wire-runtime` and `wire-grpc-client` ?

swankjesse: And maybe some kind person will later build for us `wire-grpc-server` that binds to Misk or servlets or something.

JakeWharton: Sure but if it was built two years ago it would have been RxJava-based and now you'd have `wire-runtime` depending on OkHttp and RxJava. If it was three years ago it might have even been RxJava 1.

_Because_ it's opinionated is why it should not live in this runtime.

oldergod: ~I'm down to make an RxJava implementation too when Jesse is on vacation~

wire-grpc-client/src/main/java/com/squareup/wire/GrpcClient.kt
```diff
@@ -44,9 +45,7 @@ class GrpcClient private constructor(
 ) {
   fun <T : Service> create(service: KClass<T>): T {
     val methodToService: Map<Method, GrpcMethod<*, *>> =
-        service.java.methods.associate { method ->
-          method to method.toGrpc<Any, Any>()
-        }
+        service.java.methods.associate { method -> method to method.toGrpc<Any, Any>() }
```
JakeWharton: `associateWith`

oldergod: was looking for this! It appears `associateWith` is on `Iterable`s but `methods` is an array here.

JakeWharton: Ah, that's a shame

PR #825, opened 2019-02-06
Refactoring kotlinFileWriter for better type safety

oldergod:
One side effect is current generated `Type` in kotlin are now using 2 spaces indent.

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/PendingFileSpec.kt
+++ b/wire-compiler/src/main/java/com/squareup/wire/PendingFileSpec.kt
@@ -1,0 +1,8 @@
+package com.squareup.wire
+
+import com.squareup.wire.schema.Service
+import com.squareup.wire.schema.Type
+
+internal sealed class PendingFileSpec
+internal data class PendingTypeFileSpec(val type: Type) : PendingFileSpec()
+internal data class PendingServiceFileSpec(val service: Service) : PendingFileSpec()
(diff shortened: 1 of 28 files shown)


wire-compiler/src/main/java/com/squareup/wire/PendingFileSpec.kt
```diff
@@ -0,0 +1,9 @@
+package com.squareup.wire
+
+import com.squareup.wire.schema.Service
+import com.squareup.wire.schema.Type
+
+internal sealed class PendingFileSpec {
+  data class PendingTypeFileSpec(val type: Type) : PendingFileSpec()
+  data class PendingServiceFileSpec(val service: Service) : PendingFileSpec()
```
JakeWharton: You can put these as top-level instead of nested since the naming scheme sorta wants them to be. Usually when I nest I omit repeating the prefixes and suffixes but in this case you'd wind up with `Type` and `Service` which conflict with the schema

oldergod: You mean
```kotlin
internal sealed class PendingFileSpec
internal data class PendingTypeFileSpec(val type: Type) : PendingFileSpec()
internal data class PendingServiceFileSpec(val service: Service) : PendingFileSpec()
```
?

JakeWharton: Yep

Egorand:
👍

PR #826, opened 2019-02-10
Support for redacted fields in KotlinGenerator

### LINKED ISSUE #813: Support redacted fields in KotlinGenerator
Example:
```proto
message Taco {
  optional string recipe = 1 [(squareup.redacted) = true];
}
```

This should:
1. Add `redacted = true` field to the `WireField` annotation.
2. Replace the value of the field with `██` in `toString()` output.

Egorand:
Closes #813


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -722,6 +757,78 @@ class KotlinGenerator private constructor(
...
+  private fun redactFun(message: MessageType): FunSpec {
+    val className = nameToKotlinName.getValue(message.type())
+    val nameAllocator = nameAllocator(message)
+    val result = FunSpec.builder("redact")
+        .addModifiers(OVERRIDE)
+        .addParameter("value", className)
+        .returns(className.copy(nullable = true))
+
+    val redacted = message.fields().filter { it.isRedacted }
+    val requiredRedacted = redacted.filter { it.isRequired }
```
oldergod: can we rename it so it conveys the object is actually a list?

wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/MessageUsingMultipleEnums.kt
```diff
@@ -69,6 +69,11 @@ data class MessageUsingMultipleEnums(
           unknownFields = unknownFields
         )
       }
+
+      override fun redact(value: MessageUsingMultipleEnums): MessageUsingMultipleEnums? =
+          value.copy(
+        unknownFields = ByteString.EMPTY
+      )
```
oldergod: I don't like this indent, is it expected behavior or could this be a poet bug?

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -444,6 +451,7 @@ class KotlinGenerator private constructor(
           .useSiteTarget(FIELD)
           .addMember("tag = %L", field.tag())
           .addMember("adapter = %S", field.getAdapterName(nameDelimiter = '#'))
+          .apply { if (field.isRedacted) addMember("redacted = true") }
```
oldergod: :sweet-kotlin:

wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/MessageUsingMultipleEnums.kt
Egorand: The indentation for arguments is applied manually. It's a missing KP feature that I'm working on.

Egorand:
@oldergod following up on our offline discussion: I'm gonna merge this as is since it at least follows the approach of the `JavaGenerator`, if we come up with a better solution in the future I'll be happy to submit a followup PR.

PR #836, opened 2019-02-24
Generate @Deprecated annotation for a deprecated enum constant

### LINKED ISSUE #835: Generate Deprecated annotations for deprecated enum constants

Egorand:
Closes #835


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -915,6 +916,11 @@ class KotlinGenerator private constructor(
             if (constant.documentation().isNotBlank()) {
               addKdoc("%L\n", constant.documentation())
             }
+            if ("true" == constant.options().get(ENUM_DEPRECATED)) {
```
oldergod: nit: I don't like Yoda statements
why not `if (constant.options().get(ENUM_DEPRECATED) == "true")`

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -915,6 +916,11 @@ class KotlinGenerator private constructor(
...
+              addAnnotation(AnnotationSpec.builder(Deprecated::class)
+                  .addMember("message = %S", "${constant.name()} is deprecated")
```
oldergod: Not sure it adds relevant information to what the annotation already provides.

wire-tests/src/test/proto/kotlin/deprecated_enum.proto
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+package squareup.protos.kotlin;
+
+option java_package = "com.squareup.wire.protos.kotlin";
+
+enum DeprecatedEnum {
+  DISABLED = 1 [deprecated = true];
+  ENABLED = 2 [deprecated = true];
+  ON = 3;
```
oldergod: Maybe we can add one `[deprecated = false]` or whatever to test the `"true"` check as well?

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
Egorand: Yeah, but `message` is mandatory in `Deprecated`, so you have to pass something. This is consistent with what we pass in for deprecated fields.

wire-tests/src/test/proto/kotlin/deprecated_enum.proto
Egorand: I don't think that `[deprecated = false]` makes sense in practice, but I can add it for the sake of testing.

oldergod: Or then, we could just check that the `deprecated` option is there without checking its value

Egorand: That would be a test for the schema, not for the generator, and I'm pretty sure that test exists. Here we only care about the annotation being generated if the value of `deprecated` is `true`.

PR #838, opened 2019-02-24
Generate enum message KDoc

### LINKED ISSUE #834: Include enum message documentation into the generated code

Egorand:
Closes #834

### DIFF
--- a/wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt
+++ b/wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt
@@ -92,6 +92,9 @@ data class Person(
     }
   }
 
+  /**
+   * Represents the type of the phone number: mobile, home or work.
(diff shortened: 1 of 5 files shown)


wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt
```diff
@@ -92,6 +92,9 @@ data class Person(
     }
   }
 
+  /**
+   * Represents the type of the phone number: mobile, home or work.
+   */
   enum class PhoneType(private val value: Int) : WireEnum {
     MOBILE(0),
```
oldergod: How about we have a comment on `MOBILE` or any enum value? Maybe not worth it?

Egorand: There's a comment on the `WORK` constant, line 106.

PR #840, opened 2019-02-27
Generate KDoc for message fields

Egorand:


build.gradle
```diff
@@ -103,6 +103,9 @@ allprojects {
 
   repositories {
     mavenCentral()
+    maven {
```
oldergod: TODO to remove it when kotlinpoet ships? or is it really a problem to have it always? I don't really know.

JakeWharton: You cannot release with a snapshot dependency, yes

Egorand: Sure, KotlinPoet 1.1 goes live soon

PR #841, opened 2019-02-28
Adding internal access control for Kotlin types (class and enum).

LanderlYoung:
Add `--kotlin_interal` command line argument to make all generated class `internal`, so that an SDK using wire to generate protocol class can hide those classes.


JakeWharton:
Isn't something like ProGuard or jarjar the right option here? I'm not sure every code generation library will want to support this use-case by adding a flag like this.

LanderlYoung:
Thanks for the reply, ProGuard and jarjar can do the work. But it doesn't seems to be perfect. It still exposes classes with a dummy name, while kotlin's "internal" access control can completely hide them.

Still, Java code can access Kotlin's internal types, but this is another topic.

Thanks!

jrodbx:
Thanks for the PR, but we're not looking to add this as an option at this time.

PR #846, opened 2019-03-04
Validate enum namespace in file context are unique

JakeWharton:


wire-compiler/src/test/java/com/squareup/wire/WireCompilerErrorTest.kt
```diff
@@ -119,6 +119,30 @@ class WireCompilerErrorTest {
...
   }
 
+  @Test
+  fun testEnumNamespaceFile() {
+    val e = assertFailsWith<SchemaException> {
+      compile("""
+          |package com.squareup.protos.test;
+          |
+          |enum Bar {
+          |  QUIX = 0;
```
swankjesse: FWIW, `0` is a terrible value for an enum constant with protos

swankjesse:
Protos are bad.

wire-compiler/src/test/java/com/squareup/wire/WireCompilerErrorTest.kt
oldergod: Because of proto3 absence of absence, right? Or even in proto2?

swankjesse: Yeah.

PR #851, opened 2019-03-05
Don't trim leading slash from path; assumes base is never empty

jrodbx:


swankjesse:
```
* What went wrong:
Execution failed for task ':wire-compiler:shadowJar'.
> path may not be null or empty string. path='null'
```

jrodbx:
> path may not be null or empty string. path='null'

Fixed as part of https://github.com/square/wire/pull/850

PR #854, opened 2019-03-05
Move GrpcMethod inside GrpcClient

oldergod:
I also DRY the invoke logic of the methods.


wire-grpc-tests/src/test/java/com/squareup/wire/MockRouteGuideService.kt
```diff
@@ -57,7 +57,7 @@ class MockRouteGuideService : RouteGuideGrpc.RouteGuideImplBase(), TestRule {
 
   suspend fun awaitSuccess() {
     try {
-      withTimeout(2000L) {
+      withTimeout(3000L) {
```
oldergod: Flaky tests seem to timeout here so adding a minute for the lucky strike, don't fire me

wire-grpc-client/src/main/java/com/squareup/wire/GrpcClient.kt
```diff
@@ -165,4 +171,139 @@ class GrpcClient private constructor(
...
+      requestAdapter: ProtoAdapter<S>,
+      responseAdapter: ProtoAdapter<R>
+    ) : GrpcMethod<S, R>(path, requestAdapter, responseAdapter) {
+      fun invoke(
+        continuation: Continuation<Any>,
+        grpcClient: GrpcClient
+      ): Any {
+        val (requestChannel, responseChannel) = super.invoke(continuation, grpcClient)
+
+        return requestChannel to responseChannel
```
swankjesse: why destructure and then undestructure? Just return the super call?

wire-grpc-tests/src/test/java/com/squareup/wire/MockRouteGuideService.kt
oldergod: yeah still don't understand

Issue #742, opened 2018-10-31
Introduce new wire-gradle-plugin


jrodbx:
Dumping some initial design thoughts discussed with @swankjesse and @Egorand on the new and improved Wire Gradle Plugin.

* Analogous to a classpath, the plugin should include a concept of a protopath, providing proto dependencies.
* The plugin should minimizing the number of rounds to parse *.proto files.
* Need a sane way to describe inputs, outputs.  Ideally, reuse inputs for multiple output groups.

```
INPUTS
src/main/proto/squareup/cash/boost/boost.proto
banking.jar!/squareup/cash/banking/banking.proto
banking.jar!/squareup/cash/banking/banking_services.proto

PROTOPATH
all-protos.jar!/squareup/cash/common/money.proto
all-protos.jar!/squareup/cash/common/customers.proto

OUTPUT SET
  language = {java|kotlin}
  directory = target/proto-{java|kt}
  target/proto-{java|kt}/com/squareup/cash/banking/CashCard.{java|kt}
```

Things to consider:
* language (globs etc.) to describe subsets of inputs, excludes, includes
* where do the .proto files live?
* which protos are inputs vs. protopath?

WIP Example Gradle config
```
wire {
  sourceSet += src/main/proto
  protoPath += maven("com.squareup.cash.boost:boost-client:1.0.0")

  javaTarget {
    android = true
    grpcType = rxjava2
    sourceSet += maven("something-extra")
    includes += "squareup.cash.**"
    excludes += "squareup.cash.BankingService"
    prunes += "squareup.cash.Experiment"
    outDirectory = "target/main/wire-java"
  }

  kotlinTarget {
    android = true
    includes = "squareup.cash.BankingService"
    outDirectory = target/main/wire-kotlin
  }
}
```

includes
* include these sources when generating
* if no includes are specified, we include everything

excludes
* applied after includes
* omit from generating
* retain references to these types

prunes
* omit from generating
* strip references to these types

PR #858, opened 2019-03-12
Introduce Wire Gradle Plugin

jrodbx:
Adds configuration to JVM Gradle projects that want to generate protos.

Example DSL:
```
apply plugin: 'com.squareup.wire'

wire {
  // Keeps only 'Dinosaur#name' as the root of the object graph
  roots 'squareup.dinosaurs.Dinosaur#name'

  // Keeps all fields, except 'name', in 'Dinosaur'
  prunes 'squareup.dinosaurs.Dinosaur#name'

  // Both roots and prunes in an external file
  rules 'rules.txt'

  kotlin {
    javaInterop true
    out "${buildDir}/generated/custom"
  }
}
```

Default settings:
* assume JVM java target
* include all protos found in `src/main/proto`
* output all to `build/generated/src/main/java`


wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    if (!kotlin && !java) {
+      throw IllegalArgumentException(
+          "The Wire Gradle plugin requires either the Java or Kotlin plugin to be applied prior to its being applied."
+      )
+    }
+
+    project.afterEvaluate {
+      val ssc = it.property("sourceSets") as SourceSetContainer
```
AlecStrong: can be moved inside if statement

jrodbx: good catch

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -0,0 +1,217 @@
...
+
+    val targets = mutableListOf<Target>()
+    val defaultBuildDirectory = "${project.buildDir}/generated/src/main/java"
+    val javaOutDirs = mutableListOf<String>()
+    val kotlinOutDirs = mutableListOf<String>()
+
+    val kotlinTarget = extension.kotlinTarget
+    val javaTarget = extension.javaTarget ?: if (kotlinTarget != null) null else JavaTarget()
+
+    javaTarget?.let { target ->
```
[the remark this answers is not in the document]
jrodbx: two things:
1) one can have both a java and a kotlin target, to allow for incremental proto migration.
2) i separated the null check, since the first sets up the targets to be passed to the task, and the second links task dependencies.  open to suggestions on how to organize that better.

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -0,0 +1,217 @@
...
+
+  private fun mergeDependencyPaths(
+    project: Project,
+    dependencyPaths: Set<String>,
+    dependencyTrees: Set<SourceDirectorySet> = emptySet()
+  ): List<Any> {
+    val allPaths = mutableListOf<Any>()
+    dependencyTrees.forEach {
+      allPaths += it as Any
+    }
```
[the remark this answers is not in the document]
jrodbx: niiiiice

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -0,0 +1,217 @@
...
+      val ssc = it.property("sourceSets") as SourceSetContainer
+      if (logger.isDebugEnabled) {
+        ssc.forEach {
+          logger.debug("source set: ${it.name}")
+        }
+      }
+
+      when {
+        kotlin || java -> applyJvm(it, extension)
+        else -> throw IllegalStateException("Impossible")
```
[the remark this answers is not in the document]
jrodbx: doh.  fixed.

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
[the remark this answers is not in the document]
jrodbx: niiiiice

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
```diff
@@ -0,0 +1,134 @@
...
+    return protoPaths
+  }
+
+  @InputFiles
+  @Optional
+  fun getProtoTrees(): Set<SourceDirectorySet> {
+    return protoTrees
+  }
+
+  fun protoPath(vararg protoPaths: String) {
```
Egorand: I like KDocs on `sourcePath()`, can we include them on all methods/properties in this class? Fine as a follow-up

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -0,0 +1,210 @@
...
+  private var java = false
+  override fun apply(project: Project) {
+    val logger = project.logger
+
+    val extension = project.extensions.create(
+        "wire", WireExtension::class.java, project, sourceDirectorySetFactory
+    )
+
+    project.plugins.all {
+      logger.debug("plugin: $it")
```
Egorand: Remove this?

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -0,0 +1,210 @@
...
+        ssc.forEach {
+          logger.debug("source set: ${it.name}")
+        }
+      }
+
+      when {
+        kotlin || java -> applyJvm(it, extension)
+        else ->
+          throw IllegalArgumentException(
+              "The Wire Gradle plugin requires either the Java or Kotlin plugin to be applied prior to its being applied."
```
Egorand: nit: Is there a better way to phrase this? specifically `requires ... to be applied prior to its being applied`

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -0,0 +1,210 @@
...
+
+  private fun mergeDependencyPaths(
+    project: Project,
+    dependencyPaths: Set<String>,
+    dependencyTrees: Set<SourceDirectorySet> = emptySet()
+  ): List<Any> {
+    val allPaths = dependencyTrees.toMutableList<Any>()
+
+    dependencyPaths.forEach { path ->
+      val parser = FileOrUriNotationConverter.parser()
```
Egorand: nit: Is it possible to allocate the parser once vs on each iteration?

wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
```diff
@@ -0,0 +1,343 @@
...
+    assertThat(result.output)
+        .contains("Writing com.squareup.dinosaurs.Dinosaur")
+        .contains("Writing com.squareup.geology.Period")
+
+    val generatedProto =
+      File(fixtureRoot, "build/generated/src/main/java/com/squareup/dinosaurs/Dinosaur.kt")
+    assertThat(generatedProto).exists()
+
+    assertThat(generatedProto.readText())
+        .doesNotContain(
```
Egorand: I think this is too brittle, if the structure or the formatting of code emitted by the compiler changes this test will start passing even if the field is present. Would be good to figure out a way to make it more robust.

wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
```diff
@@ -0,0 +1,343 @@
...
```
Egorand: Same here. Generally, `contains()` tests will start failing if generated code changes - it's kinda fine, we'll need to fix them up. `doesNotContain()` tests though are not safe cause they'll turn green if generated code changes.

wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
```diff
@@ -0,0 +1,343 @@
+package com.squareup.wire.gradle
+
+import org.assertj.core.api.Assertions.assertThat
+import org.gradle.testkit.runner.GradleRunner
+import org.junit.Before
+import org.junit.Test
+import java.io.File
+
+class WirePluginTest {
```
Egorand: These tests are awesome!

wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
jrodbx: Def agree, but landed on this after concluding that parsing Kotlin into an AST for assertions didn't seem possible.  Google's compile-testing library only supports Java at the moment.

wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
jrodbx: given that these test fixtures are constant, the generated code will change when we change the generator itself, which is known at the moment.  I assume the generator will stabilize soon + this proto has a small surface area, so I'm less worried.

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
jrodbx: oops, good catch.

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
jrodbx: "Either the Java or Kotlin plugin must be applied before the Wire Gradle plugin."?

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
jrodbx: Was thinking to leave the debug logging, as it's useful to me for...debugging!

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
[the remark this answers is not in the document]
AlecStrong: didnt realize you could have both - makes sense!

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
jrodbx: Added a few more KDocs to the other properties.

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
Egorand: Sounds good!

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
Egorand: Makes sense. Do you have to wrap it with `logger.isDebugEnabled()`?

wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
Egorand: I'd go with just reducing the string to something that's more robust, e.g. just the property name or whatever makes sense.

wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
JakeWharton: Check for less? The absence of `val name` seems pretty safe

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireTask.kt
```diff
@@ -0,0 +1,123 @@
...
+  fun generateWireFiles() {
+    val includes = mutableListOf<String>()
+    val excludes = mutableListOf<String>()
+
+    rules?.let {
+      project.file(it)
+          .forEachLine { line ->
+            when (line.firstOrNull()) {
+              '+' -> includes.add(line.substring(1))
+              '-' -> excludes.add(line.substring(1))
```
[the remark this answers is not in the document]
jrodbx: The configuration can live in the DSL via the `roots` and `prunes` attributes, but for really long lists (a la Cash), I thought it'd be nice to also allow configuration via an external file.

wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
jrodbx: Updated `doesNotContain` tests to only check for `val name`.
Updated `contains` tests to parse field names via regex and check for `val name`.

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
jrodbx: Nope, logger.debug makes the check internally!

wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
jrodbx: cool, updated!

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireTask.kt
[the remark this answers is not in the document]
swankjesse: Can the external file be a .gradle file rather than our custom format?

wire-gradle-plugin/src/test/projects/sourcepath-no-sources/src/main/proto/dummy.md
```diff
@@ -0,0 +1,3 @@
+Added to keep this folder structure for a test
+
+https://stackoverflow.com/questions/115983/how-can-i-add-an-empty-directory-to-a-git-repository/21422128#21422128
```
swankjesse: tricky

JakeWharton: an empty `.gitignore` also works

jrodbx: was going to go that route, but .gitignore wouldn't appear in a naive ls, so chose the dummy route.

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireTask.kt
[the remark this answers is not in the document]
jrodbx: it could...but why would that be desirable?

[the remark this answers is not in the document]
swankjesse: Smaller API. Why offer a custom format when gradles own syntax already works? Especially since gradle format gets gradle’s features.

jrodbx:
Closes https://github.com/square/wire/issues/742

jrodbx:
Closes https://github.com/square/wire/pull/561

PR #859, opened 2019-03-12
Favor lazy configuration of test tasks

jrodbx:
good catch by @autonomousapps!


jrodbx:
> com.squareup.wire.GrpcTest > cancelStreamingResponse PASSED

PR #865, opened 2019-03-15
Konvert ProtoFileElement

jrodbx:


wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ProtoParser.java
```diff
@@ -98,7 +101,6 @@ private Object readDeclaration(String documentation, Context context) {
       if (!context.permitsPackage()) throw reader.unexpected(location, "'package' in " + context);
       if (packageName != null) throw reader.unexpected(location, "too many package names");
       packageName = reader.readName();
-      fileBuilder.packageName(packageName);
```
Egorand: Is it safe to just delete this line? Probably should do `location.copy()` and replace the package name?

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ProtoParser.java
```diff
@@ -538,7 +540,7 @@ private RpcElement readRpc(Location location, String documentation) {
     reader.require(')');
 
     if (reader.peekChar('{')) {
-      ImmutableList.Builder<OptionElement> options = ImmutableList.builder();
+    ImmutableList.Builder<OptionElement> options = ImmutableList.builder();
```
Egorand: Revert?

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ProtoParser.java
jrodbx: See the other changes in this file.

Egorand: Ah, got it, nvm.

PR #868, opened 2019-03-15
Konvert ExtendElement

jrodbx:


wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ExtendElement.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.squareup.wire.schema.internal.Util.appendIndented
+
+data class ExtendElement(
+  val location: Location,
+  val name: String? = null,
+  val documentation: String = "",
+  val fields: List<FieldElement> = emptyList()
+) {
+  fun toSchema(): String {
+    val builder = StringBuilder()
```
JakeWharton: other PRs use `buildString`

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ExtendElement.kt
```diff
@@ -0,0 +1,43 @@
...
+    appendDocumentation(builder, documentation)
+    builder.append("extend ")
+        .append(name)
+        .append(" {")
+    if (!fields.isEmpty()) {
```
Egorand: nit: `fields.isNotEmpty()`

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ExtendElement.kt
jrodbx: fixed

jrodbx:
Also, reverted `ExtendElement#name`'s type back to `String` (auto conversion for some reason, chose `String?`)

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ExtendElement.kt
jrodbx: fixed

PR #869, opened 2019-03-16
Konvert TypeElements

jrodbx:
Depends on https://github.com/square/wire/pull/868


wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/TypeElement.kt
```diff
@@ -13,17 +13,16 @@
...
-  Location location();
-  String name();
-  String documentation();
-  ImmutableList<OptionElement> options();
-  ImmutableList<TypeElement> nestedTypes();
-  String toSchema();
+/** A message type or enum type declaration.  */
+interface TypeElement {
+  val location: Location
+  val name: String?
```
JakeWharton: Names are optional?

jrodbx: reverted.  stupid auto-konverter

PR #870, opened 2019-03-16
 Konvert ExtensionsElement

jrodbx:
Depends on https://github.com/square/wire/pull/869


wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ExtensionsElement.kt
```diff
@@ -0,0 +1,51 @@
...
+
+data class ExtensionsElement(
+  val location: Location,
+  val documentation: String = "",
+  val start: Int,
+  val end: Int
+) {
+  fun toSchema() = buildString {
+    appendDocumentation(this, documentation)
+    append("extensions $start")
```
JakeWharton: I wonder if the compiler is smart enough to realize you're in a `StringBuilder` and not create a new one for interpolation... probably not.

jrodbx: yea, probably not.  Favor concise vs perf in this case.

PR #872, opened 2019-03-16
Konvert ReservedElement

jrodbx:
Depends on https://github.com/square/wire/pull/871


wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ReservedElement.kt
```diff
@@ -0,0 +1,49 @@
...
+package com.squareup.wire.schema.internal.parser
+
+import com.google.common.collect.Range
+import com.squareup.wire.schema.Location
+import com.squareup.wire.schema.internal.Util.appendDocumentation
+
+data class ReservedElement(
+  val location: Location,
+  val documentation: String = "",
+  /** A [String] name or [Integer] or [Range&amp;lt;Integer&gt;][Range] tag.  */
```
JakeWharton: are the HTML escapes needed now that we're markdown? Since you're making a link should it be surrounded with `` ` `` so it's not plaintext but code font?

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ReservedElement.java
```diff
@@ -13,47 +13,37 @@
...
  * limitations under the License.
  */
-package com.squareup.wire.schema.internal.parser;
+package com.squareup.wire.schema.internal.parser
 
-import com.google.auto.value.AutoValue;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Range;
-import com.squareup.wire.schema.Location;
+import com.google.common.collect.Range
```
Egorand: Follow-up: use Kotlin's ranges

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ReservedElement.kt
jrodbx: The escapes are not needed.  Fixed.

PR #873, opened 2019-03-16
Konvert EnumConstantElement

jrodbx:
depends on https://github.com/square/wire/pull/872


wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/EnumConstantElement.kt
```diff
@@ -0,0 +1,40 @@
...
+  val tag: Int = 0,
+  val documentation: String = "",
+  val options: List<OptionElement> = emptyList()
+) {
+
+  fun toSchema() = buildString {
+    appendDocumentation(this, documentation)
+    append("$name = $tag")
+
+    if (!options.isEmpty()) {
```
JakeWharton:
```suggestion
    if (options.isNotEmpty()) {
```

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/EnumConstantElement.kt
```diff
@@ -0,0 +1,40 @@
...
+ */
+package com.squareup.wire.schema.internal.parser
+
+import com.squareup.wire.schema.Location
+import com.squareup.wire.schema.internal.Util
+import com.squareup.wire.schema.internal.Util.appendDocumentation
+
+data class EnumConstantElement(
+  val location: Location,
+  val name: String? = null,
```
JakeWharton: Hmm another null name. From auto convert or actually needed?

wire-schema/src/test/java/com/squareup/wire/schema/internal/parser/EnumElementTest.kt
```diff
@@ -142,40 +131,37 @@ class EnumElementTest {
...
-        .build()
+    val value = EnumConstantElement(
+        location = location,
+        name = "NAME",
+        tag = 1,
+        documentation = "Hello"
+    )
     val expected = """
-        |// Hello
+      |// Hello
```
JakeWharton: 👀

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/EnumConstantElement.kt
jrodbx: Auto convert.  Reverted.

PR #875, opened 2019-03-16
 Konvert OptionElement and OptionReader

jrodbx:
Depends on https://github.com/square/wire/pull/874


wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/OptionElement.kt
```diff
@@ -0,0 +1,115 @@
...
+    when (kind) {
+      Kind.STRING -> return """${formatName()} = "$value""""
+      Kind.BOOLEAN,
+      Kind.NUMBER,
+      Kind.ENUM -> return "${formatName()} = $value"
+      Kind.OPTION -> {
+        // Treat nested options as non-parenthesized always, prevents double parentheses.
+        val optionValue = (value as OptionElement).run {
+          OptionElement.create(name, kind, value)
+        }
```
JakeWharton: I think this is just `OptionElement.create(name, kind, value as OptionElement)`

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/OptionElement.kt
```diff
@@ -0,0 +1,115 @@
...
+      value: Any,
+      isParenthesized: Boolean = false
+    ) = OptionElement(name, kind, value, isParenthesized)
+
+    private fun formatOptionMap(
+      builder: StringBuilder,
+      valueMap: Map<String, *>
+    ) {
+      valueMap.entries.forEach { entry ->
+        val endl = if (entry != valueMap.entries.last()) "," else ""
```
JakeWharton: switch to `forEachIndexed` and use a prefix comma for `i > 0`

JakeWharton: Maybe you can't because of `appendIndented` actually. I'd feel better if this was index based somehow.

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/OptionElement.kt
```diff
@@ -0,0 +1,115 @@
...
+          append('}')
+        }
+        is List<*> -> {
+          append("[\n")
+          var i = 0
+          val count = value.size
+          while (i < count) {
+            val endl = if (i < count - 1) "," else ""
+            appendIndented(this, formatOptionMapValue(value[i]!!) + endl)
+            i++
```
JakeWharton: This one is index based but the above one isn't. I think you switch this to `forEachIndexed`, change `count` to `lastIndex = value.size - 1`. For both.

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/OptionElement.kt
```diff
@@ -0,0 +1,115 @@
...
+          }
+          append("]")
+        }
+        else -> {
+          return value.toString()
```
JakeWharton: nit: `append(value)` (and for `String`). otherwise move the `buildString` into the `when` arms. Mixing `buildString` and `return` throws me off... plus you've already allocated the `StringBuilder` at this point, might as well use it.

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/OptionReader.kt
```diff
@@ -0,0 +1,183 @@
...
+    }
+  }
+
+  /** Adds an object or objects to a List.  */
+  private fun addToList(
+    list: MutableList<Any>,
+    value: Any
+  ) {
+    if (value is List<*>) {
+      list.addAll(value as Collection<Any>)
```
JakeWharton: Do you still need a cast here? I guess because of the generic...

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/OptionReader.kt
```diff
@@ -0,0 +1,183 @@
...
+    }
+  }
+
+  /**
+   * Returns a list of values. This is similar to JSON with '[' and ']' surrounding the list and ','
+   * separating values.
+   */
+  private fun readList(): List<Any> {
+    reader.require('[')
+    val result = ArrayList<Any>()
```
JakeWharton:
```suggestion
    val result = mutableListOf<Any>()
```

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/OptionReader.kt
```diff
@@ -0,0 +1,183 @@
...
+   * Returns a map of string keys and values. This is similar to a JSON object, with '{' and '}'
+   * surrounding the map, ':' separating keys from values, and ',' separating entries.
+   */
+  private fun readMap(
+    openBrace: Char,
+    closeBrace: Char,
+    keyValueSeparator: Char
+  ): Map<String, Any> {
+    if (reader.readChar() != openBrace) throw AssertionError()
+    val result = LinkedHashMap<String, Any>()
```
JakeWharton:
```suggestion
    val result = mutableMapOf<String, Any>()
```

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/OptionElement.java
```diff
@@ -36,99 +34,82 @@
...
-      case BOOLEAN:
-      case NUMBER:
-      case ENUM:
-        return formatName() + " = " + value;
-      case OPTION: {
-        StringBuilder builder = new StringBuilder();
-        OptionElement optionValue = (OptionElement) value;
+  fun toSchema(): String = buildString {
+    when (kind) {
+      Kind.STRING -> return """${formatName()} = "$value""""
```
Egorand: In this case we end up throwing away the `StringBuilder` instance, maybe call `buildString` only for cases where it's needed?

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/OptionElement.java
```diff
@@ -36,99 +34,82 @@
...
-  static void formatOptionMap(StringBuilder builder, Map<String, ?> valueMap) {
-    List<? extends Map.Entry<String, ?>> entries = new ArrayList<>(valueMap.entrySet());
-    for (int i = 0, count = entries.size(); i < count; i++) {
-      Map.Entry<String, ?> entry = entries.get(i);
-      String endl = (i < count - 1) ? "," : "";
-      appendIndented(builder,
-          entry.getKey() + ": " + formatOptionMapValue(entry.getValue()) + endl);
-    }
-  }
+  private fun formatName() = if (isParenthesized) "($name)" else name
```
Egorand: `val`?

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/OptionElement.java
```diff
@@ -36,99 +34,82 @@
...
     }
-    return value.toString();
-  }
 
-  private String formatName() {
-    return isParenthesized() ? '(' + name() + ')' : name();
+    private fun formatOptionMapValue(value: Any): String = buildString {
+      when (value) {
+        is String -> {
+          return """"$value""""
```
Egorand: Same here, `buildString` feels wasteful

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/OptionElement.java
```diff
@@ -36,99 +34,82 @@
...
+        is Map<*, *> -> {
+          append("{\n")
+          formatOptionMap(this, value as Map<String, *>)
+          append('}')
+        }
+        is List<*> -> {
+          append("[\n")
+          var i = 0
+          val count = value.size
+          while (i < count) {
```
Egorand: `while` -> `for`?

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/OptionElement.kt
jrodbx: it can be done!

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/OptionElement.kt
jrodbx: It's more like `val optionValue = (value as OptionElement).copy()`.  Updated.

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/OptionReader.kt
jrodbx: yep

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/OptionElement.java
jrodbx: changed to `append`

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/OptionElement.java
jrodbx: changed to `append`

PR #878, opened 2019-03-17
 Konvert FieldElement

jrodbx:
Depends on https://github.com/square/wire/pull/875


wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ProtoParser.java
```diff
@@ -341,23 +341,22 @@ private FieldElement readField(
...
-        .build();
+    return new FieldElement(
+        location,
+        label,
+        type,
+        name,
+        defaultValue,
+        tag,
+        documentation,
+        ImmutableList.copyOf(options));
```
JakeWharton: Guava 🚫 Not sure if you're doing now or follow-up

jrodbx: replaced with `new ArrayList()`.  Eventually, will be nuking the other Guava bits.

PR #883, opened 2019-03-26
Support kotlin projects using java protos

jrodbx:
Currently, kotlin projects using java protos fail because of:
1) the way we currently link task dependencies
2) the way the kotlin plugin consumes sourceSets in an afterEvaluate block, requiring the Wire plugin to be applied before it: https://youtrack.jetbrains.com/issue/KT-12715

Re: 2), there might be a way to use lazy properties to avoid using `project.afterEvaluate` in the Wire plugin, but that's for a later PR.

### DIFF
--- a/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
+++ b/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
@@ -19,14 +19,20 @@ import com.squareup.wire.gradle.WireExtension.JavaTarget
 import com.squareup.wire.schema.Target
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.UnknownDomainObjectException
 import org.gradle.api.file.SourceDirectorySet
+import org.gradle.api.internal.HasConvention
 import org.gradle.api.internal.file.FileOrUriNotationConverter
 import org.gradle.api.internal.file.SourceDirectorySetFactory
 import org.gradle.api.plugins.JavaBasePlugin
+import org.gradle.api.tasks.SourceSet
 import org.gradle.api.tasks.SourceSetContainer
 import org.gradle.api.tasks.TaskProvider
 import org.gradle.api.tasks.compile.JavaCompile
+import org.jetbrains.kotlin.gradle.plugin.KOTLIN_DSL_NAME
+import org.jetbrains.kotlin.gradle.plugin.KOTLIN_JS_DSL_NAME
 import org.jetbrains.kotlin.gradle.plugin.KotlinBasePluginWrapper
+import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 import java.io.File
 import java.net.URI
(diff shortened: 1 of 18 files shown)


wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -146,18 +154,32 @@ class WirePlugin @Inject constructor(
...
-      compileTask.configure {
-        it.source(kotlinOutDirs)
-        it.dependsOn(wireTask)
+      try {
+        val compileKotlinTask = project.tasks.named("compileKotlin") as TaskProvider<KotlinCompile>
+        compileKotlinTask.configure {
+          it.source(kotlinOutDirs)
+          it.dependsOn(wireTask)
+        }
+      } catch (e: UnknownDomainObjectException) {
```
JakeWharton: Is there a way to query this that doesn't involve catching an exception?

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -146,18 +154,32 @@ class WirePlugin @Inject constructor(
...
+        throw IllegalStateException("To generate Kotlin protos, please apply a Kotlin plugin.", e)
```
AlecStrong: nice

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
jrodbx: not in TaskProvider world, the others would `realize` and return the actual Task.

JakeWharton: That seems like a really weird ergonomic issue. Should we file a Gradle bug asking for a `configureOrElse` or something?

autonomousapps: You're referring to https://github.com/gradle/gradle/issues/8057, and it's a whole can of worms. Your options are either try/catch or `findByName`, which will realize the task.

jrodbx: Thanks @autonomousapps, maybe I'll employ some of the suggestions in gradle/gradle#8057.

JakeWharton: I guess it's not clear why try/catch, a synchronous construct, is able to work where a query or some other mechanism wouldn't.

autonomousapps: I don't disagree with you, and I welcome any input you might have on the issue.

jrodbx: I wonder if returning a NamedDomainObjectCollection would be better, since then we could configureEach which no-ops if the task doesn't exist?

autonomousapps: `tasks.withType<KotlinCompile>().configureEach { ... }` would lazily configure all KotlinCompile tasks (and not failing if one weren't found).

PR #885, opened 2019-03-27
 Allow multiple roots and prunes

jrodbx:


wire-gradle-plugin/src/test/projects/field-rule-root/build.gradle
```diff
@@ -5,7 +5,7 @@ plugins {
 }
 
 wire {
-  kotlin {
-    rules 'rules.txt'
-  }
+  rules 'rules.txt'
```
JakeWharton: Because Groovy, that's how

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
```diff
@@ -37,20 +37,30 @@ open class WireExtension(
...
+  fun roots() = roots.toSet()
 
   /**
    * See [com.squareup.wire.schema.WireRun.treeShakingRoots]
    */
-  @get:Input
-  @get:Optional
-  var roots: Array<String>? = null
+  fun root(vararg roots: String) {
+    this.roots.addAll(Arrays.asList(*roots))
```
JakeWharton: You shouldn't need to create a list. There's an overload for `addAll` that takes an array: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/add-all.html
```suggestion
    this.roots.addAll(roots)
```

jrodbx: good catch.  that was copy-pasta from another method.

jrodbx: actually, i'll address all the methods that do this in a follow-up PR.

PR #890, opened 2019-03-28
Query Gson adapter once to avoid invoking gson.fromJson inside loops

Egorand:


wire-gson-support/src/main/java/com/squareup/wire/MessageTypeAdapter.kt
```diff
@@ -124,11 +125,12 @@ internal class MessageTypeAdapter<M : Message<M, B>, B : Message.Builder<M, B>>(
 
       val keyType = fieldBinding.keyAdapter().javaType
       val valueType = fieldBinding.singleAdapter().javaType
+      val valueAdapter = gson.getAdapter(valueType)
 
       val jsonObject = element.asJsonObject
       return jsonObject.entrySet().associateBy(
           { gson.fromJson(it.key, keyType) },
```
Egorand: Tried to do the same for the key, but its type is inferred as String, and `keyAdapter.fromJson(String)` fails tests. Is there a correct way to do this?

JakeWharton: What's the failure look like?

Egorand: `com.google.gson.JsonSyntaxException: com.google.gson.stream.MalformedJsonException: Use JsonReader.setLenient(true) to accept malformed JSON at line 1 column 1 path $`

JakeWharton: Gson might not support top-level strings despite being valid JSON. I can't remember. You can do what it says and create your own `JsonReader` and set lenient to true, but probably not worth the effort.

Egorand: Ah, makes sense. Will leave as is for now.

PR #893, opened 2019-04-01
Konvert Reserved

jrodbx:


wire-schema/src/main/java/com/squareup/wire/schema/Reserved.java
```diff
@@ -13,74 +13,29 @@
...
-
-  public Location location() {
-    return location;
+class Reserved(
+  val location: Location,
+  val documentation: String,
+  val values: List<Any>
+) {
+  fun matchesTag(tag: Int) = values.any {
+    it is Int && tag == it || (it as? Range<Int>)?.contains(tag) == true
```
Egorand: I wonder if `tag in (it as? Range<Int>)` will compile without safe-calls

jrodbx: Screen Shot 2019-04-01 at 11 45 12 PM

Nope.

PR #894, opened 2019-04-01
 Konvert EnclosingType

jrodbx:


wire-schema/src/main/java/com/squareup/wire/schema/EnclosingType.java
```diff
@@ -13,94 +13,40 @@
...
+  internal override fun retainAll(
+    schema: Schema,
+    markSet: MarkSet
+  ): Type? {
+    val retainedNestedTypes = nestedTypes.mapNotNull { it.retainAll(schema, markSet) }
+    return if (retainedNestedTypes.isEmpty()) null
+    else EnclosingType(location, type, documentation, retainedNestedTypes)
+  }
+
+  fun toElement() = MessageElement(
```
Egorand: `internal`

jrodbx: can't, until `Type.java` is konverted.

PR #896, opened 2019-04-02
 Konvert EnumConstant

jrodbx:


wire-schema/src/main/java/com/squareup/wire/schema/EnumConstant.java
```diff
@@ -13,78 +13,35 @@
...
+  val location: Location,
+  val name: String,
+  val tag: Int,
+  val documentation: String,
+  val options: Options
+) {
+  internal fun toElement() =
+    EnumConstantElement(location, name, tag, documentation, options.toElements())
+
+  internal fun linkOptions(linker: Linker) = options.link(linker)
```
Egorand: nit: space below the line?

wire-schema/src/main/java/com/squareup/wire/schema/EnumConstant.java
```diff
@@ -13,78 +13,35 @@
...
+  internal fun retainAll(
```
Egorand: We got rid of this kind of methods in KotlinPoet in favor of data classes and `copy()`, that's a more idiomatic approach to API. We should into doing the same everywhere in Wire.

jrodbx: Good follow-up

PR #897, opened 2019-04-02
 Convert files in wire-moshi-adapter to Kotlin

Egorand:


wire-moshi-adapter/src/main/java/com/squareup/wire/WireJsonAdapterFactory.kt
```diff
@@ -0,0 +1,146 @@
...
+        Types.nextAnnotations(annotations, Uint64::class.java) != null) {
+      return UINT64_JSON_ADAPTER
+    }
+    val rawType = Types.getRawType(type)
+    if (rawType == List::class.java &&
+        (type as ParameterizedType).actualTypeArguments[0] === Long::class.javaObjectType &&
+        Types.nextAnnotations(annotations, Uint64::class.java) != null) {
+      return LIST_OF_UINT64_JSON_ADAPTER
+    }
+    if (!annotations.isEmpty()) {
```
oldergod: `isNotEmpty()`? Does it exist?

Egorand: Sure! Fixed.

PR #899, opened 2019-04-14
Gradle 5.3.1

Egorand:


Egorand:
@jrodbx can you please look into the build failure?

Egorand:
Here's what I'm seeing when I build with 4.10.2:
```
> Task :wire-gradle-plugin:test
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.codehaus.groovy.reflection.CachedClass (file:/Users/Egor/.gradle/wrapper/dists/gradle-4.10.2-all/9fahxiiecdb76a5g3aw9oi8rv/gradle-4.10.2/lib/groovy-all-2.4.15.jar) to method java.lang.Object.finalize()
WARNING: Please consider reporting this to the maintainers of org.codehaus.groovy.reflection.CachedClass
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

JakeWharton:
Those are just warnings. Presumably you're running with JDK 9 or newer?

Egorand:
Yep, I was running it with JDK 11. These are warnings in Gradle 4.10.2, but seems like they fail the build in Gradle 5.3.1. Travis runs with JDK 8 and the failures look similar to what I was seeing locally. On a side note, I guess we should run with `-s` on Travis, there's not a lot of output right now to debug.

JakeWharton:
Gradle 5 supports JDK 11 or newer and JDK 8 doesn't have the module system
where that reflection is prohibited so it would be surprising to see it
caused by the same error.

Egorand:
Actually the gist of the build failure is the following:
```
org.gradle.api.InvalidUserDataException: Project directory '/Users/egor/Development/square-workspace/wire/wire-gradle-plugin/src/test/projects/different-kotlin-out' is not part of the build defined by settings file '/Users/egor/Development/square-workspace/wire/settings.gradle'. If this is an unrelated build, it must have its own settings file.
```
Adding `settings.gradle` files to all `wire-gradle` test projects should do the job.

JakeWharton:
Yep. In SQL Delight the fixture runner writes an empty file if one doesn't
exist.

Egorand:
Thanks, I'll steal it!

Egorand:
Still failing cause fixtures don't declare a dependency on the stdlib and generated code fails to compile. Once runtime is in Kotlin it'll be exposing the `api` dependency anyway, which will fix the build. I'll submit a small PR adding that dependency to runtime, just to unblock merging this one.

jrodbx:
Looks like you're now unblocked on your previous build failures.  Ping me again if anything new comes up.

PR #903, opened 2019-04-14
 Convert WireField to Kotlin

Egorand:


wire-runtime/src/main/java/com/squareup/wire/WireField.kt
```diff
@@ -0,0 +1,66 @@
...
+  /** A protocol buffer label.  */
+  enum class Label {
+    REQUIRED,
+    OPTIONAL,
+    REPEATED,
+    ONE_OF,
+    /** Implies [REPEATED]. */
+    PACKED;
+
+    val isRepeated: Boolean
```
JakeWharton: Internal for these?

Egorand: They're accessed from `wire-gson-support`. Can also reimplement them in `wire-gson-support` since they're simple.

JakeWharton: They don't do too much harm being public either though

PR #905, opened 2019-04-14
 Convert WireEnum.java to Kotlin

Egorand:


Egorand:
@JakeWharton this required patching up codegen, PTAL: e68c950

Egorand:
This change fails Gradle plugin tests as it's source-incompatible with the published version of runtime that the tests rely on. I think fixtures should rely on the local version of the runtime to avoid this kind of problems.

JakeWharton:
Indeed

PR #908, opened 2019-04-14
 Convert RuntimeEnumAdapter.java to Kotlin

Egorand:


wire-runtime/src/main/java/com/squareup/wire/RuntimeEnumAdapter.kt
```diff
@@ -0,0 +1,41 @@
...
+ */
+internal class RuntimeEnumAdapter<E : WireEnum>(private val type: Class<E>) : EnumAdapter<E>(type) {
+  private var fromValueMethod: Method? = null // Lazy to avoid reflection during class loading.
+
+  private fun getFromValueMethod(): Method {
+    val method = this.fromValueMethod
+    if (method != null) {
+      return method
+    }
+    return type.getMethod("fromValue", Int::class.javaPrimitiveType).also {
```
JakeWharton:
```kotlin
return fromValueMethod ?: type.get....
```
and delete first 4 lines

PR #909, opened 2019-04-14
 Convert MessageSerializedForm.java to Kotlin

Egorand:


wire-runtime/src/main/java/com/squareup/wire/MessageSerializedForm.kt
```diff
@@ -0,0 +1,42 @@
...
+    try {
+      // Extensions will be decoded as unknown values.
+      return adapter.decode(bytes)
+    } catch (e: IOException) {
+      throw StreamCorruptedException(e.message)
+    }
+  }
+
+  companion object {
+    private const val serialVersionUID = 0L
```
JakeWharton: Does this correctly show up on the enclosing class as a private static field? It needs to be there, and not on the companion class.

Egorand: It does, verified by decompiler.

PR #910, opened 2019-04-14
 Convert Message.java to Kotlin

Egorand:


wire-runtime/src/main/java/com/squareup/wire/Message.kt
```diff
@@ -0,0 +1,156 @@
...
+        unknownFieldsWriter = ProtoWriter(unknownFieldsBuffer!!)
+        // Writes the cached unknown fields to the buffer.
+        unknownFieldsWriter!!.writeBytes(unknownFieldsByteString)
+        unknownFieldsByteString = ByteString.EMPTY
+      }
+    }
+  }
+
+  companion object {
+    private const val serialVersionUID = 0L
```
JakeWharton: Same question as the other PR. This needs to end up as a static field on the `Message` class. Does it?

Egorand: Yup
image

PR #913, opened 2019-04-15
 Convert ProtoReader.java to Kotlin

Egorand:


wire-runtime/src/main/java/com/squareup/wire/ProtoReader.kt
```diff
@@ -0,0 +1,440 @@
...
+import okio.BufferedSource
+import okio.ByteString
+import java.io.EOFException
+import java.io.IOException
+import java.net.ProtocolException
+
+/**
+ * Reads and decodes protocol message fields.
+ */
+class ProtoReader(@JvmField private val source: BufferedSource) {
```
JakeWharton: annotation needed?

Egorand: Nope, removed

PR #918, opened 2019-04-16
 Convert RuntimeMessageAdapter.java to Kotlin

Egorand:


wire-runtime/build.gradle
```diff
@@ -53,6 +53,10 @@ task japicmp(type: JapicmpTask, dependsOn: 'jar') {
   packageExcludes = [
       'com.squareup.wire.internal',
   ]
+  classExcludes = [
+      // package protected -> public    
+      'com.squareup.wire.RuntimeMessageAdapter',
```
Egorand: Gets accessed from `wire-gson-support` so has to be public

wire-runtime/src/main/java/com/squareup/wire/FieldBinding.java
```diff
@@ -25,7 +25,7 @@
  * Read, write, and describe a tag within a message. This class knows how to assign fields to a
  * builder object, and how to extract values from a message object.
  */
-final class FieldBinding<M extends Message<M, B>, B extends Message.Builder<M, B>> {
+public final class FieldBinding<M extends Message<M, B>, B extends Message.Builder<M, B>> {
```
Egorand: Is exposed through `RuntimeMessageAdapter` hence should also become public

wire-runtime/build.gradle
JakeWharton: Is this still needed?

wire-runtime/src/main/java/com/squareup/wire/FieldBinding.java
JakeWharton: Should also be in 'internal'

Egorand: That's in a different PR

wire-runtime/build.gradle
Egorand: Thought it would be since the class is missing. Japicmp doesn't complain though, so I removed it.

PR #928, opened 2019-04-23
Run wire-gradle-plugin tests against a local version of wire-runtime

Egorand:


wire-gradle-plugin/src/test/projects/java-project-java-protos/build.gradle
```diff
@@ -6,10 +6,10 @@ plugins {
 mainClassName = 'com.squareup.dinosaurs.Sample'
 
 repositories {
+  mavenLocal()
   mavenCentral()
 }
 
 dependencies {
-  implementation 'com.squareup.wire:wire-runtime:+'
```
Egorand: Somehow `+` still resolves to `3.0.0-alpha01`, is there a way to make it resolve to the latest SNAPSHOT?

wire-gradle-plugin/src/test/projects/java-project-java-protos/build.gradle
```diff
@@ -6,10 +6,10 @@ plugins {
...
-  implementation 'com.squareup.okio:okio:+'
+  implementation 'com.squareup.wire:wire-runtime:3.0.0-SNAPSHOT'
```
JakeWharton: Why not retain `+`? Otherwise you'll want to inject the version as a system property from the enclosing build. We do this in SQL Delight.

Egorand: See my comment above

JakeWharton: I would inject the version into the tests as a system property and then inject it into the Gradle project by writing it to `gradle.properties` or something.

Egorand: Yeah, it might already be injectable since it's in `gradle.properties`. Guess the problem should also disappear once we publish 3.0, `+` should start resolving to `3.1.0-SNAPSHOT` properly.

JakeWharton: It's in this projects `gradle.properties` but not the test project's

Egorand: PTAL 6996779

PR #931, opened 2019-04-24
Make return type of ProtoAdapter.redact() non-nullable

Egorand:


wire-runtime/src/main/java/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -46,7 +46,7 @@ abstract class ProtoAdapter<E>(
   internal var repeatedAdapter: ProtoAdapter<List<E>>? = null
 
   /** Returns the redacted form of `value`. */
-  open fun redact(value: E): E? = null
+  open fun redact(value: E): E = throw UnsupportedOperationException()
```
JakeWharton: Do you just want to make this `abstract`?

Egorand: Yep, good point

Egorand: That breaks binary compat though.. I'm gonna merge as is, can revisit later

JakeWharton: It's a major version bump though. Generated code is only guaranteed to work with the same compiler version.

Egorand: Well yeah, guess that's fine.

PR #937, opened 2019-04-29
Allow jar inputs

### LINKED ISSUE #898: Different proto files loaded between SchemaLoader and NewSchemaLoader
This might not be a bug and just user error :) To start I just wanted to document my findings, in case others run into similar issues.

When using the `WireCompiler`, which uses the `SchemaLoader`, only the explicit proto files I have listed in the arguments are loaded. For example:

```
java WireCompiler --proto-path=large-jar-with-protos.jar my_proto1.proto my_proto2.proto
```

Even though `large-jar-with-protos.jar` has 1000s of proto files bundled in it, only `my_proto1.proto` and `my_proto2.proto` (and their dependencies) are loaded.

However, it appears when using `WireRun`, which uses the `NewSchemaLoader`, it loads the 1000s of proto files in `large-jar-with-protos.jar`. We are using the new wire gradle plugin, which uses `WireRun`.

```
wire {
  sourcePath large-jar-with-protos.jar
}
```

This is causing issues for my org, since the large proto file has colliding `FieldOptions`, which now result in fully qualified name methods like `fully_qualified_required(...)` instead of `required()`. This is now not backwards compatible with previously generated Wire code.

jrodbx:
Closes #898


wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
```diff
@@ -133,11 +140,38 @@ open class WireExtension(
...
+    sourceJars: MutableSet<ProtoRootSet>,
+    name: String
+  ) {
+    val protoRootSet = objectFactory.newInstance(ProtoRootSet::class.java)
+    action.execute(protoRootSet!!)
+
+    val hasSrcDirs = protoRootSet.srcDirs.isNotEmpty()
+    val hasSrcJar = protoRootSet.srcJar != null
+
+    if (hasSrcDirs && hasSrcJar) {
```
swankjesse: `check()`

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -183,43 +187,66 @@ class WirePlugin @Inject constructor(
...
+    path: String,
+    project: Project,
+    jarIncludes: List<String> = emptyList()
+  ): Any {
+    val converted = parser.parseNotation(path)
+
+    if (converted is File) {
+      val file =
+        if (!converted.isAbsolute) File(project.projectDir, converted.path) else converted
+      if (!file.exists()) {
```
swankjesse: check

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -183,43 +187,66 @@ class WirePlugin @Inject constructor(
...
         throw IllegalArgumentException(
-            "Invalid path string: \"$path\". URL dependencies are not allowed."
+            "Invalid path string: \"$path\". Path does not exist."
+        )
+      }
+      return when {
+        file.isDirectory -> project.files(path)
+        file.isJar -> {
+          if (jarIncludes.isNotEmpty()) {
+            jarToIncludes[file.path] = jarIncludes
```
swankjesse: I find this side-effect surprising

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -183,43 +187,66 @@ class WirePlugin @Inject constructor(
...
+          }
+          project.files(file)
+        }
+        else -> throw IllegalArgumentException(
+            "Invalid path string: \"$path\". For individual files, use the closure syntax."
```
swankjesse: instead of saying "use the closure syntax" wanna just provide an example?

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -229,4 +256,7 @@ class WirePlugin @Inject constructor(
     } catch (e: Exception) {
       false
     }
-}
\ No newline at end of file
+}
+
+private val File.isJar
```
swankjesse: I don’t think this is going to be a good approach long term. The extension might be absent depending on where the file came from.

swankjesse: Could you instead optimistically treat it as a jar, and throw a nice exception if that fails?

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
swankjesse: Not sure if I see a better way though.

Still I don’t like how this method returns a value and changes local state, and the 2nd part is kind of invisible.

wire-gradle-plugin/src/test/projects/sourcejar-remote-many-files/build.gradle
```diff
@@ -0,0 +1,17 @@
+plugins {
+  id 'application'
+  id 'com.squareup.wire'
+}
+
+repositories {
+  maven {
+    url "file://${projectDir.absolutePath}/../../../../../build/localMaven"
```
swankjesse: comment this?

swankjesse:
Very cool

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
Egorand: Return a Pair/custom class and let the caller process it as needed?

jrodbx: Agreed on all points.  Can revisit down the road, but this unblocks in the short term.

wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
```diff
@@ -159,6 +159,75 @@ class WirePluginTest {
...
+    val fixtureRoot = File("src/test/projects/sourcejar-remote-protopath")
+
+    val result = gradleRunner
+        .withProjectDir(fixtureRoot)
+        .build()
+
+    assertThat(result.task(":generateProtos")).isNotNull
+    assertThat(result.output)
+        .contains("Writing com.squareup.dinosaurs.Dinosaur")
+        .contains("Writing com.squareup.geology.Period")
```
Egorand: `period.proto` is on the `protoPath`, doesn't it mean that it shouldn't get compiled, as it's a dependency?

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
jrodbx: Hmm, but we want to avoid individual proto file paths, like this:
```
wire {
  sourcePath 'src/main/proto/squareup/geology/period.proto'
}
```
since it makes the Location.get(base, path) split logic tricky.  

We previously accomplished this by disallowing all non-directory files.  The isJar check weakens it enough since a jar is essentially a directory root.

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
jrodbx: I'll address in a follow-up, as this existed before.

wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
jrodbx: great catch.  turns out the test setup isn't quite right.  i'll fix that.

jrodbx: https://github.com/square/wire/pull/940

jrodbx: fixed

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
swankjesse: Ship it as-is. Later we should differentiate zip vs. non-zip by attempting to open the file as a .zip, not by its extension.

wire-gradle-plugin/src/test/projects/sourcejar-remote-many-files/build.gradle
jrodbx: done

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
jrodbx: @Egorand here's one way that can look: 
https://github.com/square/wire/commit/4e4351306416c7df90db119c6e6eb4d2a01aae1c

not sure i like it better though.

jrodbx: spoke with @Egorand offline who gave some suggestions using sealed classes.  i'll merge this and follow up.

PR #942, opened 2019-04-30
Move jarIncludes up to callers

jrodbx:
Potential cleanup for https://github.com/square/wire/pull/937


wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -261,6 +263,23 @@ class WirePlugin @Inject constructor(
     } catch (e: Exception) {
       false
     }
+
+  sealed class DependencyType(val project: Project, open val path: String) {
```
Egorand: Make path `abstract` instead of `open`?

jrodbx: done

PR #948, opened 2019-05-03
Add Kotlin Multiplatform dependencies

Egorand:


build.gradle
```diff
@@ -50,11 +50,16 @@ buildscript {
...
+              'common': "org.jetbrains.kotlin:kotlin-stdlib-common",
               'jdk6': "org.jetbrains.kotlin:kotlin-stdlib",
               'jdk8': "org.jetbrains.kotlin:kotlin-stdlib-jdk8",
+              'js': "org.jetbrains.kotlin:kotlin-stdlib-js",
           ],
           'test': [
+              'common': "org.jetbrains.kotlin:kotlin-test-common",
+              'annotations': "org.jetbrains.kotlin:kotlin-test-annotations-common",    
               'junit': "org.jetbrains.kotlin:kotlin-test-junit",
+              'js': "org.jetbrains.kotlin:kotlin-test-js",
```
jrodbx: should we omit js for now?

Egorand: As a platform, or just omit its dependencies? This PR adds a bunch of stuff we'll need in a sec

JakeWharton: There should be little reason to omit JS. Everything kinda just works on it. Native is the problematic one.

PR #949, opened 2019-05-03
Build wire-runtime with Kotlin MPP

Egorand:


wire-maven-plugin/build.gradle
```diff
@@ -1,4 +1,5 @@
 apply plugin: 'com.liferay.maven.plugin.builder'
+apply plugin: 'org.jetbrains.kotlin.jvm'
```
Egorand: Modules that depend on runtime module have to get built by a Kotlin plugin, even though they're Java libraries - this is needed so that they can resolve the correct configuration of a multiplatform project

wire-runtime/build.gradle
```diff
@@ -1,12 +1,98 @@
...
-  testImplementation deps.junit
-  testImplementation deps.assertj
+apply plugin: 'org.jetbrains.kotlin.multiplatform'
+
+kotlin {
+  jvm()
+  js() {
+    configure([compilations.main, compilations.test]) {
+      tasks.getByName(compileKotlinTaskName).kotlinOptions {
+        moduleKind = "commonjs"
```
JakeWharton: umd. commonjs is basically useless.

wire-gradle-plugin/src/test/projects/kotlin-project-kotlin-protos/build.gradle
```diff
@@ -7,12 +7,14 @@ plugins {
...
-  mavenLocal()
+  maven {
+    url "file://${projectDir.absolutePath}/../../../../../build/localMaven"
+  }
   mavenCentral()
 }
 
 dependencies {
-  implementation "com.squareup.wire:wire-runtime:$VERSION_NAME"
+  implementation "com.squareup.wire:wire-runtime-jvm:$VERSION_NAME"
```
Egorand: Can alternatively depend on `wire-runtime` and build fixtures with Kotlin plugin

wire-runtime/build.gradle
```diff
@@ -1,12 +1,98 @@
...
+        api deps.kotlin.stdlib.js
+      }
+    }
+    jsTest {
+      dependencies {
+        implementation deps.kotlin.test.js
+      }
+    }
+    nativeMain {
+      dependsOn commonMain
```
JakeWharton: These are needed?

wire-maven-plugin/build.gradle
Egorand: Or I can actually point them to correct configuration manually, similar to `:wire-runtime:japicmp`.

wire-runtime/build.gradle
Egorand: Need to fix in Okio

wire-runtime/build.gradle
Egorand: Copied from Okio, I believe @bnorm added those to fix IDE issues

bnorm: https://kotlinlang.org/docs/reference/building-mpp-with-gradle.html#connecting-source-sets

> Custom source sets created in addition to the default ones should be explicitly included into the dependencies hierarchy to be able to use declarations from other source sets and, most importantly, to take part in compilations. Most often, they need a dependsOn(commonMain) or dependsOn(commonTest) statement, and some of the default platform-specific source sets should depend on the custom ones, directly or indirectly:

This is a weird intersection where IntelliJ is unhappy with all the code in `nativeMain` without the `dependsOn` because it is correctly interpreting the rules but Gradle is happy without it because it never actually compiles `nativeMain`, it compiles `linuxX64Main` which is a default source set so it automatically includes `commonMain` and we're adding `nativeMain` to that so everything resolves.

Is this an IntelliJ bug because it displays code red when it actually compiles? Is this a compiler bug because `nativeMain` doesn't actually depend on `commonMain`? Is it not a bug at all because the documentation clear states that you need this `dependsOn`?

PR #955, opened 2019-05-10
Move MutableOnWriteList to common

Egorand:


wire-runtime/src/jsMain/kotlin/com/squareup/wire/internal/-Platform.kt
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+package com.squareup.wire.internal
+
+import okio.IOException
+import kotlin.reflect.KClass
+
+actual interface Serializable
+
+actual annotation class Throws(actual vararg val exceptionClasses: KClass<out Throwable>)
```
JakeWharton: Use https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-optional-expectation/index.html to avoid actuals in other platforms

wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/-Platform.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import java.util.Collections
+
+actual typealias Serializable = java.io.Serializable
+
+actual typealias Throws = kotlin.jvm.Throws
+
+actual typealias ObjectStreamException = java.io.ObjectStreamException
+
+actual fun <T> MutableList<T>.toUnmodifiableList(): List<T> = Collections.unmodifiableList(this)
```
JakeWharton: inline?

wire-runtime/src/nativeMain/kotlin/com/squareup/wire/internal/-Platform.kt
```diff
@@ -0,0 +1,27 @@
...
+import okio.IOException
+import kotlin.reflect.KClass
+
+actual interface Serializable
+
+actual annotation class Throws(actual vararg val exceptionClasses: KClass<out Throwable>)
+
+actual abstract class ObjectStreamException : IOException()
+
+actual fun <T> MutableList<T>.toUnmodifiableList(): List<T> = toList()
```
JakeWharton: Can just return `this` and be `inline`

wire-runtime/src/jsMain/kotlin/com/squareup/wire/internal/-Platform.kt
```diff
@@ -0,0 +1,27 @@
...
+
+actual abstract class ObjectStreamException : IOException()
+
+actual fun <T> MutableList<T>.toUnmodifiableList(): List<T> = toList()
```
JakeWharton: Can just return `this` and be `inline`

PR #959, opened 2019-05-11
Move Internal to common

Egorand:


wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire
+
+expect abstract class ProtoAdapter<E> {
```
Egorand: Temporary `expect` class, `ProtoAdapter` will eventually completely move into common

PR #961, opened 2019-05-12
Change ProtoAdapter to use KClass instead of java.lang.Class

Egorand:


swankjesse:
Impact on Java callers?

Egorand:
`ProtoAdapter` and `FieldEncoding` constructors accepting `java.lang.Class` are gone, so Java callers will have to pass in a `KClass`, which can be obtained using `JvmMappingsKt.getKotlinClass(java.lang.Class)` - see regenerated tests for an example. Since runtime depends on the stdlib anyway, it should just be a minor annoyance.

wire-gson-support/src/main/java/com/squareup/wire/MessageTypeAdapter.kt
```diff
@@ -115,7 +115,7 @@ internal class MessageTypeAdapter<M : Message<M, B>, B : Message.Builder<M, B>>(
       if (element.isJsonNull) {
         return emptyList<Any>()
       }
-      val itemType = fieldBinding.singleAdapter().javaType
+      val itemType = fieldBinding.singleAdapter().type?.javaObjectType
```
JakeWharton: Why are these all using safe-call?

wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/RuntimeMessageAdapter.kt
```diff
@@ -62,9 +63,11 @@ class RuntimeMessageAdapter<M : Message<M, B>, B : Builder<M, B>>(
     for (fieldBinding in fieldBindings.values) {
       if (fieldBinding.redacted && fieldBinding.label == WireField.Label.REQUIRED) {
         throw UnsupportedOperationException(
-            "Field '${fieldBinding.name}' in ${javaType?.name} is required and cannot be redacted.")
+            "Field '${fieldBinding.name}' in ${type?.javaObjectType?.name} is required and " +
+                "cannot be redacted.")
       }
-      val isMessage = Message::class.java.isAssignableFrom(fieldBinding.singleAdapter().javaType)
+      val isMessage = Message::class.java
+          .isAssignableFrom(fieldBinding.singleAdapter().type?.javaObjectType)
```
JakeWharton: Should this be `!!`

PR #965, opened 2019-05-14
 Move ProtoAdapter, ProtoReader, ProtoWriter and FieldEncoding into common

Egorand:


wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -15,6 +15,527 @@
...
+      override fun decode(reader: ProtoReader): List<E> = listOf(this@ProtoAdapter.decode(reader))
+
+      override fun redact(value: List<E>): List<E> = emptyList()
+    }
+  }
+
+  class EnumConstantNotFoundException(
+    @JvmField val value: Int,
+    type: KClass<*>?
+      // TODO(egorand): Error message should print out type.qualifiedName, but this currently fails
```
swankjesse: simpleName is good enough

wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/-Platform.kt
```diff
@@ -24,6 +24,14 @@ expect annotation class Throws(vararg val exceptionClasses: KClass<out Throwable
...
+
 expect fun <T> MutableList<T>.toUnmodifiableList(): List<T>
 
 expect fun <K, V> MutableMap<K, V>.toUnmodifiableMap(): Map<K, V>
+
+internal expect fun String.format(vararg args: Any?): String
+
+internal expect fun Int.intBitsToFloat(): Float
+
+internal expect fun Long.longBitsToDouble(): Double
```
JakeWharton: Both of these are built-in to Kotlin already as `toBits()`

Egorand: I used `toBits()` for `Float` to `Int` and `Double` to `Long` conversion, but there are no `Int` to `Float` and `Long` to `Double` bits converters. Not sure if `Int.toFloat()` and `Long.toDouble()` are safe to use.

JakeWharton: There's `fromBits`: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/from-bits.html

Funny enough they're static factories instead of extensions which is funny given my strong argument for extensions today.

Egorand: Since I voted for factories I feel defeated by my own weapon. Rip.

PR #968, opened 2019-05-15
Move Message to common

Egorand:


Egorand:
Currently won't compile due to a missing Okio method: https://github.com/square/okio/pull/612

wire-runtime/src/commonMain/kotlin/com/squareup/wire/MessageSerializedForm.kt
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+package com.squareup.wire
+
+import com.squareup.wire.internal.ObjectStreamException
+import com.squareup.wire.internal.Serializable
+import com.squareup.wire.internal.Throws
+import kotlin.reflect.KClass
+
+internal expect class MessageSerializedForm<M : Message<M, B>, B : Message.Builder<M, B>>(
```
JakeWharton: This really shouldn't be in commonMain. I realize it requires expect/actual-ing Message, but I'm not sure dragging in this an serializable and object stream exception is any better.

Egorand: Agree, looks like I won't be able to avoid expect/actual without hacks. I'll experiment with slimming down `Builder` since generated Kotlin code doesn't need it, this might help make JS and Native actuals pretty small.

PR #969, opened 2019-05-15
WireRun generates service

oldergod:
`WireCompiler` supported; now `WireRun` does too!

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
+++ b/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
@@ -35,7 +35,11 @@ sealed class Target {
    */
   abstract val elements: List<String>
 
-  internal abstract fun newHandler(schema: Schema, fs: FileSystem, logger: WireLogger): TypeHandler
+  internal abstract fun newHandler(
+    schema: Schema,
+    fs: FileSystem,
+    logger: WireLogger
+  ): SchemaHandler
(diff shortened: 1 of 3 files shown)


wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -135,24 +143,51 @@ sealed class Target {
...
+
+        override fun handle(service: Service) {
+          logger.artifactSkipped(service.type())
+        }
       }
     }
   }
 
-  interface TypeHandler {
+  interface SchemaHandler {
```
oldergod: Tell me if there's a better word

wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -90,6 +94,10 @@ sealed class Target {
                 "to $outDirectory", e)
           }
         }
+
+        override fun handle(service: Service) {
+          throw IllegalStateException("Service handling isn't supporting in Java.")
```
swankjesse: silently fail instead please?

wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
swankjesse: Works for me.

PR #983, opened 2019-05-24
Adds singleMethod plugin configuration for Services

oldergod:
```groovy
wire {
  kotlin {
    blockingServices true
    singleMethodServices true
  }
}
```

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
+++ b/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
@@ -116,7 +116,7 @@ sealed class Target {
     val javaInterop: Boolean = false,
 
     /** True for emitted services to implement one interface per RPC. */
-    val servicesAsSingleMethod: Boolean = false
+    val singleMethodServices: Boolean = false
   ) : Target() {
     override fun newHandler(schema: Schema, fs: FileSystem, logger: WireLogger): SchemaHandler {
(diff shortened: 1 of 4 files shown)


wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -257,12 +259,12 @@ class WirePlugin @Inject constructor(
...
-      true
-    } catch (e: Exception) {
-      false
-    }
+      try {
+        uri.toURL()
+        true
+      } catch (e: Exception) {
+        false
+      }
```
jrodbx: can we move formatting changes to a separate PR?

wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -115,8 +115,12 @@ sealed class Target {
     /** True for emitted types to implement APIs for easier migration from the Java target. */
     val javaInterop: Boolean = false,
 
+    /** True for emitted services to provide blocking APIs only. */
+    // TODO(benoit) use it
+    val blockingServices: Boolean = false,
```
jrodbx: can we move the introduction of this parameter to the PR that provides its implementation?

Issue #993, opened 2019-05-30
Bad value for @WireField#adapter with Kotlin and nested classes


oldergod:
```proto
message FormBlocker {
  message Element {}
}
```

if generated in Java, we get
```java
  @WireField(
      tag = 1,
      adapter = "com.squareup.protos.franklin.api.FormBlocker$Element#ADAPTER",
      label = WireField.Label.REPEATED
  )
  public final List<Element> elements;
```
but if generated in Kotlin, we get
```kotlin
  @WireField(
      tag = 1,
      adapter = "com.squareup.protos.franklin.api.FormBlocker.Element#ADAPTER"
   )
   @JvmField
   @NotNull
   public final List elements;
```

The value should have a `$` before the nested class.

PR #994, opened 2019-05-30
Tests for nested message in Kotlin

oldergod:
So we wanna make those pass
#993

### DIFF
--- a/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
+++ b/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
@@ -91,6 +91,7 @@ class KotlinGenerator private constructor(
 
   /** Returns the full name of the class generated for [type].  */
   fun generatedTypeName(type: Type) = type.typeName
+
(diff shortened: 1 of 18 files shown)


oldergod:
oh you're right, even the Person generated Kotlin is wrong

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -881,7 +882,17 @@ class KotlinGenerator private constructor(
...
+    }
+  }
+
+  private fun ProtoType.reflectionName(): String {
+    val className = nameToKotlinName.getValue(this)
+
+    return if (className.packageName.isEmpty()) {
+      className.simpleNames.joinToString(separator = "\$")
+    } else {
+      "${className.packageName}.${className.simpleNames.joinToString(separator = "\$")}"
```
oldergod: inspired by https://github.com/square/wire/blob/master/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java#L1219

oldergod:
Fixed by @Egorand

PR #995, opened 2019-05-30
Improve regex in Gradle plugin tests to ignore formatting when matching fields

Egorand:
`.` doesn't match line separators by default, `DOT_MATCHES_ALL` changes this behavior. But since this makes the regex less restrictive, and `*` quantifier is greedy by default, it will now match the last `val` it can find. Switched to using reluctant quantifiers to make sure the closest `val` to `@field:WireField` is matched.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
       'gson': '2.8.1',
       'guava': '20.0',
       'javapoet': '1.10.0',
-      'kotlinpoet': '1.1.0',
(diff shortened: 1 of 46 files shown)


Egorand:
@jrodbx sorry, forgot that KotlinPoet 1.3.0 branch target this one and merged. See this commit only: https://github.com/square/wire/pull/995/commits/e7d4e5fac466eb322067f1706282821b0f0766b7

jrodbx:
👍

PR #1002, opened 2019-06-05
Remove sealed classes-based oneof implementation

### LINKED ISSUE #998: Drop oneof with sealed classes
We won't be going forward with this design, so the plan is to go back to the strategy we use in Java codegen. Since we already use it in interop mode, we should just make it default and remove the sealed classes approach.

Egorand:
Closes #998

### DIFF
--- a/wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/OneOfMessage.kt
+++ b/wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/OneOfMessage.kt
@@ -8,7 +8,8 @@ import com.squareup.wire.ProtoAdapter
 import com.squareup.wire.ProtoReader
 import com.squareup.wire.ProtoWriter
 import com.squareup.wire.TagHandler
-import java.lang.UnsupportedOperationException
+import com.squareup.wire.WireField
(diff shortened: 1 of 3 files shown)


wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/OneOfMessage.kt
```diff
@@ -19,8 +20,29 @@ import okio.ByteString
 /**
  * It's a one of message.
  */
-data class OneOfMessage(val choice: Choice? = null, val unknownFields: ByteString =
-    ByteString.EMPTY) : Message<OneOfMessage, OneOfMessage.Builder>(ADAPTER, unknownFields) {
+data class OneOfMessage(
+  /**
+   * What foo.
+   */
+  @field:WireField(tag = 1, adapter = "com.squareup.wire.ProtoAdapter#INT32") val foo: Int? = null,
```
oldergod: I thought latest kotlin poet would pretify those for us, no?

Egorand: It hasn't been merged yet

PR #1004, opened 2019-06-07
Wrap oneof error message properly

### LINKED ISSUE #991: Error message wrapped incorrectly in a oneof class in Java interop mode
image

The error message used in `oneOf`s is wrapped without making the whole thing one string.

Egorand:
Fixes #991


wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/Form.kt
```diff
@@ -0,0 +1,765 @@
...
+    adapter = "com.squareup.wire.protos.kotlin.Form${'$'}CurrencyConversionFlagsElement#ADAPTER"
+  )
+  val currency_conversion_flags_element: CurrencyConversionFlagsElement? = null,
+  val unknownFields: ByteString = ByteString.EMPTY
+) : Message<Form, Form.Builder>(ADAPTER, unknownFields) {
+  init {
+    require(countNonNull(button_element, local_image_element, remote_image_element, money_element,
+        spacer_element, text_element, customized_card_element, address_element, text_input_element,
+        option_picker_element, detail_row_element, currency_conversion_flags_element) <= 1) {
+      "At most one of button_element, local_image_element, remote_image_element, money_element, spacer_element, text_element, customized_card_element, address_element, text_input_element, option_picker_element, detail_row_element, currency_conversion_flags_element may be non-null"
```
oldergod: hmmm, there's no wrapping though? Shouldn't we break lines?

Egorand: Nope, strings don't wrap.

PR #1005, opened 2019-06-07
Use non-nullable types for fields that have default values

### LINKED ISSUE #997: Kotlin generator assigns nullable refs to non-nullable variables in Java interop mode
We assign a [nullable](https://github.com/square/wire/blob/d9e9eb9c9f35aaceabc66a298ccf9131842072f5/wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt.java.interop#L237) to a [non-nullable](https://github.com/square/wire/blob/d9e9eb9c9f35aaceabc66a298ccf9131842072f5/wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt.java.interop#L253) here https://github.com/square/wire/blob/d9e9eb9c9f35aaceabc66a298ccf9131842072f5/wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt.java.interop#L243

Egorand:
Fixes #997


Egorand:
This solution is wrong, see #1006

PR #1008, opened 2019-06-09
Fix dependencies

Egorand:


proto-pruner/build.gradle
```diff
@@ -16,6 +16,7 @@ jar {
 dependencies {
   implementation project(':wire-schema')
   implementation deps.kotlin.stdlib.jdk8
+  implementation deps.guava
```
jrodbx: is this dependency being pulled in from wire-schema?  if so, should it remain as an api dependency there?

Egorand: That's how it was before the change, but in fact both use Guava as implementation.

PR #1010, opened 2019-06-13
Move Message to common

Egorand:


wire-runtime/src/commonMain/kotlin/com/squareup/wire/Message.kt
```diff
@@ -0,0 +1,41 @@
...
+/** A protocol buffer message. */
+expect abstract class Message<M : Message<M, B>, B : Message.Builder<M, B>> {
+  /** If non-zero, the hash code of this message. Accessed by generated code. */
+  protected var hashCode: Int
+
+  /**
+   * Returns a byte string containing the proto encoding of this message's unknown fields. Returns
+   * an empty byte string if this message has no unknown fields.
+   */
+  fun unknownFields(): ByteString
```
JakeWharton: Unrelated: could this be a val?

wire-runtime/src/commonMain/kotlin/com/squareup/wire/Message.kt
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * Returns a new builder initialized with the data in this message.
+   */
+  // TODO(egorand): Remove from common and not generate in Kotlin.
```
Egorand: Hmm, doesn't seem feasible, `RuntimeMessageAdapter` still needs to reference it in a generic way.

wire-runtime/src/commonMain/kotlin/com/squareup/wire/Message.kt
Egorand: Totally!

Egorand: Making this a val is harder than I thought, cause there's already a primary constructor val with the same name. Can probably rename the constructor val, but I couldn't come up with a sane name. Otherwise, turning it into a constructor param doesn't work with data classes. Lmk if you have any ideas.

JakeWharton: If the primary constructor already has a val then what is the function for?

Egorand: Java still needs it. I can remove it from `Message` and generate it in Java without `@Override`, but then we lose the polymorphic call.

PR #1011, opened 2019-06-13
Make KotlinGenerator depend on common runtime artifact

Egorand:


wire-kotlin-generator/build.gradle
```diff
@@ -12,9 +12,7 @@ jar {
 
 dependencies {
   api project(':wire-schema')
-  // TODO(egorand): Ideally KotlinGenerator should generate platform-independent code, once runtime
-  // is 100% multiplatform find the right configuration to enforce this
-  implementation project(path: ':wire-runtime', configuration: 'jvmRuntimeElements')
+  implementation project(path: ':wire-runtime', configuration: 'runtimeElements')
```
JakeWharton: Is the configuration needed? It should just do the right thing with a normal project dep.

Egorand: Hmm, I remember it being unhappy with the normal dep. Lemme try again.

JakeWharton: We do this in SQLDelight for the JDBC sqlite driver depending on the multiplatform runtime

PR #1012, opened 2019-06-13
Remove usages of path and configuration for dependency resolution

Egorand:
Bringing this in line for all modules


Egorand:
Tests fail cause tasks in `gen-tests.gradle` depend on the compiler through a custom configuration, and it doesn't seem to properly resolve the variants of transitive multiplatform dependencies. We've faced a similar issue in Cash and I've yet to find a solution. Will continue digging.

PR #1013, opened 2019-06-13
Don't use JvmField and JvmStatic in non-interop mode

Egorand:


wire-grpc-tests/src/test/proto-grpc/routeguide/Feature.kt
```diff
@@ -51,7 +50,6 @@ data class Feature(
   }
 
   companion object {
-    @JvmField
     val ADAPTER: ProtoAdapter<Feature> = object : ProtoAdapter<Feature>(
```
JakeWharton: This is still looked up via reflection in ProtoAdapterJvm

wire-moshi-adapter/src/test/java/com/squareup/wire/protos/person/kotlin/Person.kt
```diff
@@ -140,14 +137,12 @@ data class Person(
...
 
     companion object {
-      @JvmField
       val ADAPTER: ProtoAdapter<PhoneType> = object : EnumAdapter<PhoneType>(
         PhoneType::class
       ) {
         override fun fromValue(value: Int): PhoneType = PhoneType.fromValue(value)
       }
 
-      @JvmStatic
```
JakeWharton: This is required for RuntimeEnumAdapter

Egorand:
I'll fix reflection to look for both static fields and companions to support both modes.

JakeWharton:
Is there a motivation to remove these? They don't prevent the generated code from being multiplatform.

Egorand:
IDE complained about them when I tried compiling against JS

JakeWharton:
Are you using kotlin2js or the multiplatform plugin? We have these in Okio's common code, for example, which compiles to JS just fine.

Egorand:
As discussed offline, will punt this for now, JVM annotations should work fine if generated code is used by a multiplatform module. In future we might enable the compiler to produce JS or Native-specific code.

PR #1014, opened 2019-06-13
Add JsName to Message#hashCode to avoid name clash on JS

Egorand:


Egorand:
Good point, this feels like a less invasive solution, and it won't require regenerating Java code. Besides, JS doesn't reference `cachedHashCode`.

PR #1017, opened 2019-06-19
Store unknown fields in ProtoReader instead of Builder

Egorand:


wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoReader.kt
```diff
@@ -90,13 +93,31 @@ class ProtoReader(private val source: BufferedSource) {
...
+  }
+
+  /**
+   * End a length-delimited nested message. Calls to this method must be symmetric with calls to
+   * [beginMessage].
+   *
+   * @param token value returned from the corresponding call to [beginMessage].
+   */
+  @Throws(IOException::class)
+  @Deprecated(level = DeprecationLevel.HIDDEN, message = "prefer endMessageAndGetUnknownFields()")
```
swankjesse: Wanna do WARNING deprecation? Code that has a compiler/runtime mismatch should still build?

wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoReader.kt
```diff
@@ -391,29 +412,22 @@ class ProtoReader(private val source: BufferedSource) {
...
-      @Suppress("UNCHECKED_CAST") // We encode and decode the same types.
-      (protoAdapter as ProtoAdapter<Any>).encodeWithTag(unknownFieldsWriter!!, tag, value)
+      tagHandler(tag)
     }
-    endMessage(token)
+    return endMessageAndGetUnknownFields(token)
+  }
 
-    return unknownFieldsBuffer?.readByteString() ?: ByteString.EMPTY
+  fun readUnknownField(tag: Int) {
```
swankjesse: Document that the unknown field will be included in endMessageAndGetUnknownFields’s return value?

swankjesse:
Worth adding a test where the enclosing and enclosed messages both have unknown fields? As is I think we could incorrectly share buffers and our tests would still pass.

wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoReader.kt
Egorand: Good catch, code generated by previous versions of compiler will fail to compile after this change.

Egorand:
@swankjesse PTAL https://github.com/square/wire/pull/1017/commits/dc36207d177a0e61c9847e6eeafe8282c3f5d108

PR #1018, opened 2019-06-19
Remove TagHandler and ProtoReader.forEachTag(TagHandler)

Egorand:


wire-runtime/japicmp/build.gradle
```diff
@@ -39,6 +39,9 @@ task japicmp(type: JapicmpTask, dependsOn: 'jar') {
   packageExcludes = [
       'com.squareup.wire.internal',
   ]
+  classExcludes = [
+      'com.squareup.wire.TagHandler',
```
Egorand: Turns out it was used before, I'll look into history to make sure we won't be breaking things if we just delete it

Egorand: `TagHandler` was introduced for `KotlinGenerator`, so it's safe to delete

PR #1020, opened 2019-06-20
expect/actual ProtoAdapter

### LINKED ISSUE #1019: expect/actual ProtoAdapter to preserve backwards compatibility
Use `expect`/`actual` to move methods from `ProtoAdapterJvm` back into `ProtoAdapter` and keep backwards compat. Breaking backwards compatibility in the runtime module will make it hard to gradually migrate code from Wire 2 to Wire 3 and will require changing the package name.

Egorand:
Closes #1019


Egorand:
Moved `InternalTest.kt` from `common` back to `jvm`, it's also affected by https://youtrack.jetbrains.com/issue/KT-31473 and somehow it surfaced after the changes I've made in this PR. Anyway, target version for that fix is 1.3.50, so will hopefully move it back into `common` soon.

PR #1021, opened 2019-06-20
Adding some gRPC tests around cancellation

oldergod:


wire-grpc-tests/src/test/java/com/squareup/wire/GrpcTest.kt
```diff
@@ -112,7 +110,7 @@ class GrpcTest {
     }
   }
 
-  @Test
+  @ObsoleteCoroutinesApi @Test
```
oldergod: I was sick with the warnings

JakeWharton: Put it on the class so you only do it once

wire-grpc-tests/src/test/java/com/squareup/wire/GrpcTest.kt
```diff
@@ -254,20 +250,100 @@ class GrpcTest {
...
+    mockService.enqueue(
+        ReceiveMessage(RouteGuideProto.RouteNote.newBuilder().setMessage("marco").build()))
+    mockService.enqueue(ReceiveComplete)
+
+    val (requestChannel, responseChannel) = routeGuideService.RouteChat()
+    runBlocking {
+      requestChannel.send(RouteNote(message = "marco"))
+      requestChannel.close()
+      // We wait for mockService to process our actions.
+      delay(500)
```
Egorand: I wonder if we can avoid this kind of delays and make the mock service suspend until it's processed the actions, is that possible?

wire-grpc-tests/src/test/java/com/squareup/wire/GrpcTest.kt
```diff
@@ -49,6 +47,8 @@ import java.time.Duration
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicReference
 
+@ExperimentalCoroutinesApi
+@ObsoleteCoroutinesApi
```
swankjesse: this is out of hand, JetBrains has gone too far

wire-grpc-tests/src/test/java/com/squareup/wire/GrpcTest.kt
swankjesse: possibly? I find interleaving different parts of the code pretty difficult to do neatly

swankjesse:
Good stuff.

wire-grpc-tests/src/test/java/com/squareup/wire/GrpcTest.kt
oldergod: And the thing supposed to replace it isn't released yet...

PR #1022, opened 2019-06-20
Kotlin 1.3.40

Egorand:


Egorand:
Build fails due to incompatibility with Okio. Okio has been updated to 1.3.40 and fresh snapshots published. Should wait for it to sync properly and it should go through.

PR #1023, opened 2019-06-21
Generate constants for default value options in Kotlin

### LINKED ISSUE #1006: Incorrect semantics of default field values
Currently, for the following proto:
```proto
message PhoneNumber {
  optional PhoneType type = 2 [default = HOME];
}
```
we'll generate the following Kotlin class:
```kotlin
data class PhoneNumber(
  @WireField(...)
  val type: PhoneType? = PhoneType.HOME,
  val unknownFields: ByteString
)
```
The type's `ProtoAdapter` decodes the bytes in the following way:
```kotlin
override fun decode(reader: ProtoReader): PhoneNumber {
  var type: PhoneType = PhoneType.HOME
    when (tag) {
      2 -> type = PhoneType.ADAPTER.decode(reader)
      else -> TagHandler.UNKNOWN_TAG
    }
  }
  return PhoneNumber(
    type = type,
    unknownFields = unknownFields
  )
}
```
This means that:

- If tag `2` is missing on the wire - the value of the field will be `PhoneType.HOME`
- If tag `2` is present on the wire - the field will get the decoded value 

What's missing here is the ability to distinguish the case when the tag was missing in the payload from the case when the data included the tag and the value of it was `PhoneType.HOME`. In contrast, Java represents default field values as constants, making the following logic possible:
```kotlin
val phoneType = phoneNumber.type ?: run {
  Timber.e("Hey server, you should be sending us the type of that phone number!")
  return@run PhoneNumber.DEFAULT_PHONE_TYPE
}
```
In order to retain proto2 semantics, we should revert back to Java's approach to default values. However, since current implementation adheres to proto3 semantics (where lack of value is indistinguishable from the default value), we can potentially generate different code based on the version of the schema.

Egorand:
Fixes #1006


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -524,6 +528,113 @@ class KotlinGenerator private constructor(
...
+    messageType: MessageType,
+    protoType: ProtoType,
+    nameAllocator: NameAllocator,
+    value: Any?
+  ): CodeBlock {
+    val typeName = protoType.typeName
+    return when {
+      value is List<*> -> buildCodeBlock {
+        add("listOf(")
+        var first = true
```
oldergod: Can't you do a map of each line with a `joinAsString` kind of thing?

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -524,6 +528,113 @@ class KotlinGenerator private constructor(
...
+
+  private fun Any?.toIntFieldInitializer(): CodeBlock = when (val int = valueToInt()) {
+    Int.MIN_VALUE -> CodeBlock.of("%T.MIN_VALUE", INT)
+    Int.MAX_VALUE -> CodeBlock.of("%T.MAX_VALUE", INT)
+    else -> CodeBlock.of("%L", int)
+  }
+
+  private fun Any?.valueToInt(): Int {
+    if (this == null) return 0
+    val string = toString()
```
oldergod: Doesn't work with `when (val string = toString())`?

wire-tests/src/test/proto-kotlin/com/squareup/wire/protos/kotlin/person/Person.kt.android
```diff
@@ -208,7 +211,7 @@ data class Person(
 
         override fun decode(reader: ProtoReader): PhoneNumber {
           var number: String? = null
-          var type: PhoneType = PhoneType.HOME
+          var type: PhoneType? = null
```
oldergod: niiice

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
Egorand: Good point, I'll try

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
Egorand: Should work!

Egorand: Wait, no, `when (val string = toString())` would require all branch conditions to resolve to String, whereas what I need is Boolean

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
Egorand: I've actually removed that block of code and simplified the method: https://github.com/square/wire/pull/1023/commits/638226470d90a5feb12215f317e49c67950856f9

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
JakeWharton: That's not how that works. It just scopes the local to live inside the when. Doesn't affect what you do in the branches.

Egorand: I must be missing something, cause when I follow Benoit's suggestion I get compilation errors on branch conditions saying `Incompatible types: Boolean and String`.

oldergod: I think you're both talking about something different.

PR #1034, opened 2019-06-23
Declare Guava as api dependency in wire-schema

### LINKED ISSUE #1030: In wire-schema, change dependency on Guava back to "api"
[This PR](https://github.com/square/wire/pull/1008) changes dependency to `implementation`. In fact, `wire-schema` does have public APIs returning Guava types. Should change back to `api` so that callers are not required to depend on Guava.

Egorand:
Fixes #1030

### DIFF
--- a/wire-schema/build.gradle
+++ b/wire-schema/build.gradle
@@ -14,7 +14,7 @@ dependencies {
   api project(':wire-runtime')
   api deps.kotlin.stdlib.jdk8
   implementation deps.okio.jvm
-  implementation deps.guava
+  api deps.guava
   testImplementation project(':wire-test-utils')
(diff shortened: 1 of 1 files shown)


wire-schema/build.gradle
```diff
@@ -14,7 +14,7 @@ dependencies {
   api project(':wire-runtime')
   api deps.kotlin.stdlib.jdk8
   implementation deps.okio.jvm
-  implementation deps.guava
```
jrodbx: how many Guava usages remain?  shouldn't be many after konverting a bunch of classes to use Kotlin lists...

Egorand: Definitely not that many

PR #1047, opened 2019-07-05
Support includes on coordinate dependencies.

swankjesse:
Previously if you included a dependency by its Maven coordinates you
wouldn't benefit from includes filtering.

### DIFF
--- a/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
+++ b/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
@@ -23,6 +23,7 @@ import com.squareup.wire.gradle.WirePlugin.DependencyType.Path
 import com.squareup.wire.schema.Target
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.artifacts.Dependency
 import org.gradle.api.file.SourceDirectorySet
 import org.gradle.api.internal.file.FileOrUriNotationConverter
 import org.gradle.api.internal.file.SourceDirectorySetFactory
@@ -43,7 +44,7 @@ class WirePlugin @Inject constructor(
 ) : Plugin<Project> {
   private var kotlin = false
   private var java = false
-  private val jarToIncludes = mutableMapOf<String, List<String>>()
+  private val dependencyToIncludes = mutableMapOf<Dependency, List<String>>()
 
   private lateinit var sourceSetContainer: SourceSetContainer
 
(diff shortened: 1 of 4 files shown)


wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireTask.kt
```diff
@@ -112,20 +114,27 @@ open class WireTask : SourceTask() {
...
+          base = srcDir.path,
+          path = path.substring(srcDir.path.length + 1)
+      ))
+    }
+
+    val includes = dependencyToIncludes[dependency]
+        ?: return listOf(Location.get(path))
+
+    return includes.map {
+      include -> Location.get(base = path, path = include)
```
oldergod: super nit but that feels odd for the parameter to be declared here.

swankjesse: I wanna leave it ’cause it’s weird that the parameter named path isn’t the path local variable.

oldergod: I meant I'm used to
```kotlin
return includes.map { include -> 
      Location.get(base = path, path = include)
}
```

swankjesse: ahh, yeah that’s just a mistake.

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -192,32 +189,36 @@ class WirePlugin @Inject constructor(
     }
   }
 
+  /** Returns a list of dependencies for the build's paths, trees, and .jars. */
   private fun mergeDependencyPaths(
```
jrodbx: nit: method name should be updated as it now merges dependencies vs its paths

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -192,32 +189,36 @@ class WirePlugin @Inject constructor(
...
     val parser = FileOrUriNotationConverter.parser()
 
     dependencyJars.forEach { depJar ->
       depJar.srcJar?.let { path ->
         val depPath = mapDependencyPath(parser, path, project)
-        if (depPath is Jar && depJar.includes.isNotEmpty()) {
-          jarToIncludes[depPath.path] = depJar.includes
-        }
-        allPaths += depPath.dependency
+        val dependency = project.dependencies.create(depPath.dependency)
```
jrodbx: nice cleanup

jrodbx:
👍

Issue #1048, opened 2019-07-05
[wire-runtime] IllegalFormatConversionException when decoding an invalid Boolean


pavlospt:
### Problem
Try to decode an invalid value returned with `ProtoReader#readVarInt32()` and in case it fails to match either `1 or 0` it fallbacks to throwing an exception:

```kotlin
throw IOException("Invalid boolean value 0x%02x".format(value))
```
Here value is passed as a `vararg` but it is not exploded in `wire-runtime/jvmMain/internal/-Platform (L37)`

```kotlin
String.format(this, args)
```

causing an `java.util.IllegalFormatConversionException: x != [Ljava.lang.Object;` exception.

PR #1049, opened 2019-07-05
[wire-runtime] Explode vararg `args` before passing them to String.format

pavlospt:
Resolves https://github.com/square/wire/issues/1048

### DIFF
--- a/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/-Platform.kt
+++ b/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/-Platform.kt
@@ -34,4 +34,4 @@ actual inline fun <K, V> MutableMap<K, V>.toUnmodifiableMap(): Map<K, V> =
     Collections.unmodifiableMap(this)
 
 @Suppress("NOTHING_TO_INLINE") // Syntactic sugar.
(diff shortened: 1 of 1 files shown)


oldergod:
Thanks! How hard would it be to write a test case for that?

Egorand:
I don't think it warrants a test, it's a simple omission and the fix is straightforward.

Egorand:
@pavlospt could you please sign our [CLA](https://spreadsheets.google.com/spreadsheet/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1) so that I can merge the PR?

pavlospt:
@oldergod I could write test for it but i consider it an overkill since this kind of extensions are rarely changing. 
@Egorand I signed the CLA!

JakeWharton:
It's not the function that needs a test–it's the codepath which invokes it.

Egorand:
I can look into it when I have time. `String.format()` is only used to format an exception message though, so I consider it low priority.

JakeWharton:
The fact that no test invokes the exceptional path is the problem and would have uncovered the problem. We don't actually need to care that much about the actual message.

pavlospt:
@Egorand if you add the bytes & message from this issue: https://github.com/square/wire/issues/1056 i think you have an almost ready test case.

PR #1051, opened 2019-07-07
Kotlin 1.3.41

Egorand:


Egorand:
Filed https://youtrack.jetbrains.com/issue/KT-32457

Egorand:
The target version for the issue above is 1.3.50, so closing this in favor of #1100.

PR #1055, opened 2019-07-07
Allow using kotlin.Experimental in wire-runtime

Egorand:
Using `@OptionalExpectation` (required for `@Throws` that we
declare in `-Platform.kt`) requires opt in. Hopefully this will
also indicate to wire-runtime consumers that it's in experimental
state, since it's a multiplatform library.

### DIFF
--- a/wire-runtime/build.gradle
+++ b/wire-runtime/build.gradle
@@ -28,6 +28,11 @@ kotlin {
   macosX64()
   mingwX64('winX64')
   sourceSets {
+    all {
+      languageSettings {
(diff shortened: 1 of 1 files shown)


JakeWharton:
Is there a YouTrack for making the built-in `@Throws` an optional expectation annotation in common? I couldn't find one.

Egorand:
Yep, there's a link in the code: https://youtrack.jetbrains.com/issue/KT-26283

Egorand:
Weird, both builds passed but the status didn't get updated in the PR. Merging.

PR #1061, opened 2019-07-10
[wire-kotlin-generator] Use NameAllocator to avoid param name clash in ADAPTER#decode

### LINKED ISSUE #1060: Lambda variable name may conflict with field name.
```
override fun decode(reader: ProtoReader): CreateHashTagResponse {
	var base_resp: BaseResponse? = null
	var tag: HashTag? = null
	when (tag) {
		1 -> base_resp = BaseResponse.ADAPTER.decode(reader)
		2 -> tag = HashTag.ADAPTER.decode(reader)
		else -> reader.readUnknownField(tag)
	}
}
	return CreateHashTagResponse(
		base_resp = base_resp,
		tag = tag,
		unknownFields = unknownFields
	)
}
```
`reader.forEachTag { tag ->` conflict with `var tag: HashTag? = null`.
Maybe change to `reader.forEachTag { it ->` is better.

pavlospt:
Resolves #1060

### DIFF
--- a/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
+++ b/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
@@ -193,7 +193,6 @@ class KotlinGenerator private constructor(
             newName("reader", "reader")
             newName("Builder", "Builder")
             newName("builder", "builder")
-            newName("readerTag", "readerTag")
 
             if (emitAndroid) {
               newName("CREATOR", "CREATOR")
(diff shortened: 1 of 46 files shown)


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -193,6 +193,7 @@ class KotlinGenerator private constructor(
             newName("reader", "reader")
             newName("Builder", "Builder")
             newName("builder", "builder")
+            newName("readerTag", "readerTag")
```
JakeWharton: This name is not used as part of public API so it doesn't need to be part of these name allocators which track public API and change properties (as shown by the test).

Since the name is local to a method, you can clone the allocator inside `decodeFun` and call `newName` on the cloned allocator. This will avoid collision with the API names while not leaking the local name to the global name allocator.

JakeWharton:
Tweaked, squashed, and merged as b99055af. I reverted to the original name to prevent the churn on the generated code. Thanks!

Issue #957, opened 2019-05-10
gRPC Server API codegen


swankjesse:
I’d like one interface per method. No suspending initially, but perhaps soon.

```kotlin
interface RouteGuideGetFeature : Service {
  @WireRpc(
      path = "/routeguide.RouteGuide/GetFeature",
      requestAdapter = "routeguide.Point#ADAPTER",
      responseAdapter = "routeguide.Feature#ADAPTER"
  )
  fun GetFeature(request: Point): Feature
}
```

What does the Gradle config look like?
```groovy
wire {
  kotlin {
    services {
      // This prevents us from adding the 'suspend' modifier
      blocking: true

      // This emits a single interface for each service method
      singleMethod: true
    }
  }
}
```

Tasks

 * [x] Add new configuration to KotlinTarget for blocking, singleMethod.
 * [x] Implement that configuration in KotlinGenerator.
 * [x] Split `wire-grpc-client` into `wire-grpc`, `wire-grpc-client`.
 * [x] Create `wire-grpc-service` in Wire or Misk that depends on Jetty. Maybe a Servlet?
 * [x] Promote GrpcReader, GrpcWriter to public APIs.
 * [x] GrpcWriter to support gzip
 * [ ] GrpcWriter gzip tests

oldergod:
Need to promote GrpcReader, GrpcWriter to public APIs. Might mean changing the way we hook up GrpcEncoding so it can stay internal.

oldergod:
@swankjesse do we have to migrate the grpc channels from misk to wire to be used in server code generation?

swankjesse:
I think we need *FOUR DIFFERENT SIGNATURES* for our services. This depends on whether you’re blocking or suspending, client or server…

```
service TacoApi {
  rpc SubscribeToTacos(stream Money) returns (stream Taco) {}
}
```

Here’s my proposal for naming conventions:

*Blocking client*
```
interface TacoApiBlockingClient : Service {
  fun SubscribeToTacos() : Pair<MessageSink<Money>, MessageSource<Taco>>
}
```

*Blocking server*
```
interface TacoApiBlockingServer : Service {
  fun SubscribeToTacos(request: MessageSource<Money>, response: MessageSink<Taco>>)
}
```

*Suspending client*
```
interface TacoApiClient : Service {
  fun SubscribeToTacos() : Pair<SendChannel<Money>, ReceiveChannel<Taco>>
}
```

*Suspending server* (not implemented)
```
[...]

PR #1062, opened 2019-07-10
Creates wire-grpc module

oldergod:
To be shared by `grpc-client` and `grpc-server`.

Moved `Service` and `WireRpc` in it for now. Will follow up with `GrpcReader` and `GrpcWriter`.


swankjesse:
Yass!

oldergod:
related to #957

PR #1063, opened 2019-07-11
Promote GrpcReader and GrpcWriter to public API

oldergod:
related to #957

### DIFF
--- a/wire-grpc/src/main/java/com/squareup/wire/GrpcEncoding.kt
+++ b/wire-grpc/src/main/java/com/squareup/wire/GrpcEncoding.kt
@@ -31,16 +31,14 @@ internal sealed class GrpcEncoding(val name: String) {
   internal object GzipGrpcEncoding : GrpcEncoding("gzip") {
     override fun decode(source: BufferedSource) = GzipSource(source)
   }
+}
 
-  companion object {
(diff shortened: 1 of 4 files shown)


swankjesse:
Only request: we should get the writer to support Gzip before we ship

wire-grpc/src/main/java/com/squareup/wire/GrpcEncoding.kt
```diff
@@ -31,16 +31,14 @@ internal sealed class GrpcEncoding(val name: String) {
...
-        else -> throw ProtocolException("unsupported grpc-encoding: $this")
-      }
-    }
+internal fun String.toGrpcEncoding(): GrpcEncoding {
+  return when (this) {
+    "identity" -> GrpcEncoding.IdentityGrpcEncoding
+    "gzip" -> GrpcEncoding.GzipGrpcEncoding
+    "deflate" -> throw ProtocolException("deflate not yet supported")
+    "snappy" -> throw ProtocolException("snappy not yet supported")
+    else -> throw ProtocolException("unsupported grpc-encoding: $this")
```
oldergod: It looks like the spec allows custom value here but doesn't say what to do with it.
https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests

PR #1064, opened 2019-07-11
GrpcWriter supports gzip

oldergod:


wire-grpc/src/main/java/com/squareup/wire/GrpcDecoder.kt
```diff
@@ -20,23 +20,23 @@ import okio.GzipSource
 import okio.Source
 import java.net.ProtocolException
 
-internal sealed class GrpcEncoding(val name: String) {
+internal sealed class GrpcDecoder(val name: String) {
```
oldergod: Awkward splitting between the reading and the writing

wire-grpc/src/main/java/com/squareup/wire/GrpcWriter.kt
```diff
@@ -31,21 +33,24 @@ class GrpcWriter<T> private constructor(
...
+      messageAdapter: ProtoAdapter<T>,
+      grpcEncoding: String = "identity"
+    ) = GrpcWriter(sink, messageAdapter, grpcEncoding)
   }
 
   fun writeMessage(message: T) {
-    // TODO: support writing nontrivial encodings
-    val compressedFlag = 0
+    val compressedFlag = if (grpcEncoding == "identity") 0 else 1
     sink.writeByte(compressedFlag)
```
oldergod: Should we use the gzip sink here too?

PR #1066, opened 2019-07-11
Workaround for generated annotation on java9+

### LINKED ISSUE #1003: Generated gRPC code doesn't compile with Java 12
Looks like `javax.annotation.Generated` has been removed from the JDK.

oldergod:
kind of fixes #1003

It doesn't look like the proto team fixed it. We are using it only in tests so I think that's enough for now.

### DIFF
--- a/wire-grpc-tests/build.gradle
+++ b/wire-grpc-tests/build.gradle
@@ -58,6 +58,11 @@ dependencies {
   implementation project(':wire-grpc')
   implementation project(':wire-grpc-client')
   implementation deps.okio.jvm
+  if (JavaVersion.current().isJava9Compatible()) {
+    // Workaround for @javax.annotation.Generated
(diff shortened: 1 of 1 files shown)


wire-grpc-client/build.gradle
```diff
@@ -19,4 +19,9 @@ dependencies {
   api deps.okio.jvm
   api deps.okhttp
   api deps.kotlin.coroutines.core
+  if (JavaVersion.current().isJava9Compatible()) {
+    // Workaround for @javax.annotation.Generated
+    // see: https://github.com/grpc/grpc-java/issues/3633
+    api 'javax.annotation:javax.annotation-api:1.3.1'
```
JakeWharton: I thought you said this only affects tests. Why is it being added as a transitive dependency of this artifact?

oldergod: Because I totally pooped on myself. It's updated.

PR #1069, opened 2019-07-11
Fix deserializing null values in Gson adapter

Egorand:
This is a regression introduced during Kotlin rewrite


Egorand:
Yeah, I ran a similar test on Moshi and it passed.

oldergod:
Maybe you should have leaved it in there!

PR #1072, opened 2019-07-13
Creates wire-grpc-server

oldergod:
Also introduces GrpcChannel to be used by generated server code.


wire-grpc-server/src/main/java/com/squareup/wire/GrpcChannel.kt
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire
+
+import java.io.Closeable
+
+interface GrpcSendChannel<T> : Closeable {
```
swankjesse: Should these go in grpc-common to give us the option for blocking clients?

oldergod: Why not? We'd have to support it in the runtime as well. Our `GrpcClient` assumes coroutines to create the service.

You'd add something like in the plugin config?
```kotlin
    var blockingServices: Boolean
    var singleMethodServices: Boolean
/!\ var kotlinCoroutine: Boolean /!\
```

swankjesse: I think it's just blockingServices=true to use these.

Egorand: nit: `<in T>`?

wire-grpc-server/src/main/java/com/squareup/wire/GrpcChannel.kt
```diff
@@ -0,0 +1,35 @@
...
+  fun send(message: T)
+}
+
+interface GrpcReceiveChannel<T> : Closeable {
```
Egorand: nit: `<out T>`?

wire-grpc-server/src/main/java/com/squareup/wire/GrpcChannel.kt
```diff
@@ -0,0 +1,35 @@
...
+  fun receiveOrNull(): T?
+}
+
+fun <T> GrpcReceiveChannel<T>.consumeEachAndClose(block: (T) -> Unit) {
```
Egorand: nit: `inline`

wire-grpc-server/src/main/java/com/squareup/wire/GrpcChannel.kt
oldergod: `blockingServices` makes it so we don't use `suspend` for a single request/response API. Others API are non-blocking by default.
Why would `blockingServices=true` affect the non-blocking APIs signature?

swankjesse: I think there's two independent features:

 * blocking/suspending
 * single interface/many interfaces

We could build blocking clients and suspending servers. Just don't need those yet.

oldergod: Ok, then maybe we don't need the `wire-grpc-server` for now. I'll move the GrpcChannels into our `wire-grpc` module.

PR #1075, opened 2019-07-14
Add UseExperimental annotations to GrpcClient

Egorand:
All of these are in `internal` code, so the intent is to just silence the warnings.

### DIFF
--- a/wire-grpc-client/build.gradle
+++ b/wire-grpc-client/build.gradle
@@ -13,6 +13,12 @@ animalsniffer {
   ignore 'com.squareup.wire.internal'
 }
 
+compileKotlin {
+  kotlinOptions{
+    freeCompilerArgs += "-Xuse-experimental=kotlin.Experimental"
+  }
+}
+
 dependencies {
   implementation project(':wire-runtime')
   api project(':wire-grpc')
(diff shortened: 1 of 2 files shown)


JakeWharton:
I don't think this is the right approach without defining your own experimental annotation and requiring its use. If someone updates the coroutine library on an old version of Wire you'll receive NoClassDefFoundError or NoSuchMethodError because of the use of these unstable functions.

Egorand:
I see, so we should actually be propagating the experimental annotations. What's the value of declaring our own annotation vs propagating `@ExperimentalCoroutinesApi` and `@ObsoleteCoroutinesApi`? If consuming code has already opted into the standard ones it'll automatically opt into using `wire-grpc`.

JakeWharton:
I would say that the advantage of having your own is that you don't need to propagate multiple of the stdlib ones. If someone is suppressing the built-in ones already they might not realize that wire is also relying on these unstable APIs.

That being said, I don't feel too strongly about it. I just don't think suppression is the right thing here since it's prone to potential breakage in a non-obvious way.

Egorand:
Yeah, agree. I'll go with a custom annotation, thinking more about it, a consumer might be fine using experimental API in their own code but not in 3rd party code, in which case they can choose to not opt into using experimental parts of Wire.

PR #1078, opened 2019-07-16
Don't use KClass.simpleName to avoid needing kotlin-reflect dependency

### LINKED ISSUE #1077: Don't use KClass.simpleName in EnumConstantNotFoundException
This requires usage of `kotlin-reflect`, which is too big for Android apps.

Egorand:
Fixes #1077


JakeWharton:
Kinda bullshit this isn't a compile-time failure

Egorand:
☝️ 👈 this here

PR #1080, opened 2019-07-17
Merging wire-grpc to wire-runtime/wire-grpc-client

oldergod:
I don't think we want to leave `Service` and `WireRpc` into `commonMain`, let me know otherwise.


JakeWharton:
What's the motivation for this?

oldergod:
Wire wouldn't be about messages only ; we want Wire to deal with stream as well, from its core.

JakeWharton:
I didn't even want this gRPC stuff in the wire project from the get go, so I'm definitely opposed to this even more so than that.

Protocol buffers are an implementation detail of the gRPC protocol which is agnostic to the serialization format of the actual transported body. There's no reason or benefit to intrinsically tying these two together. Nothing in the proto spec deals with streaming, and when it is mentioned it talks about prefixed, length-delimited messages.

This whole effort feels like it should live in OkHttp or a separate project that combines the two (similar to how gRPC itself does this with Google protos and various HTTP clients).

swankjesse:
> Nothing in the proto spec deals with streaming, and when it is mentioned it talks about prefixed, length-delimited messages.

From [the spec](https://developers.google.com/protocol-buffers/docs/reference/proto3-spec#service_definition):

```
service = "service" serviceName "{" { option | rpc | emptyStatement } "}"
rpc = "rpc" rpcName "(" [ "stream" ] messageType ")" "returns" "(" [ "stream" ]
messageType ")" (( "{" {option | emptyStatement } "}" ) | ";")
```

JakeWharton:
Even ignoring the argument that you can use other encodings for bodies with gRPC, there's still no reason to merge the two artifacts. Just because one thing requires another does not mean they need to be the same library. It's like having every JSON library also include facilities for JSON-RPC. If it must live in wire, it gRPC fundamentally has nothing to do with the protocol buffer specification and should continue to live in a separate artifact. If RSocket was built on protocol buffers would you embed an RSocketReader and RSocketWriter into wire-runtime? What about if stubby was open sourced? Or any other transport protocol that happens to build with protocol buffers.

JakeWharton:
Notably, that spec piece says nothing about gRPC and does not require the use of gRPC, as you well know seeing as the thing you work on uses regular HTTP to implement those service defintiions.

oldergod:
I updated the PR.

*wire-runtime*
- `@WireRpc`, `Service` (marker interface)
- `MessageSink<T>`, `MessageSource<T>` interfaces

*wire-grpc-client*
- `GrpcMessageSink<T>`, `GrpcMessageSource<T>` implementations (Previously GrpcWriter and GrpcReader)

oldergod:
Should probably add some comments on mesageSource and sink...

wire-grpc-client/src/main/java/com/squareup/wire/GrpcMessageSink.kt
```diff
@@ -35,7 +34,12 @@ class GrpcWriter<T> private constructor(
...
-    ) = GrpcWriter(sink, messageAdapter, grpcEncoding)
+    ) = GrpcMessageSink(sink, messageAdapter, grpcEncoding)
+  }
+
+  override fun send(message: T) {
+    writeMessage(message)
+    flush()
   }
 
   fun writeMessage(message: T) {
```
JakeWharton: Is this needed as a separate concept from `send`?  If no, private.

wire-runtime/src/commonMain/kotlin/com/squareup/wire/MessageSink.kt
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire
+
+expect interface MessageSink<in T> {
+  fun send(message: T)
```
JakeWharton: I'm tempted to have this called `write` and the source one called `readOrNull` for some symmetry with Okio's sink/source. @swankjesse ?

wire-grpc-client/src/main/java/com/squareup/wire/GrpcMessageSink.kt
```diff
@@ -17,14 +17,13 @@ package com.squareup.wire
...
-import java.io.Closeable
 
 /** Writes a sequence of GRPC messages as an HTTP/2 stream. */
-class GrpcWriter<T> private constructor(
+internal class GrpcMessageSink<T> private constructor(
   private val sink: BufferedSink,
   private val messageAdapter: ProtoAdapter<T>,
   private val grpcEncoding: String
-) : Closeable by sink {
+) : MessageSink<T> {
```
JakeWharton: You should still be able to do `MessageSink<T>, Closeable by sink` and omit the explicit function below. Same for reader

wire-runtime/src/commonMain/kotlin/com/squareup/wire/MessageSource.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+package com.squareup.wire
+
+expect interface MessageSource<out T> {
+  fun receiveOrNull(): T?
+
+  fun close()
+}
+
+expect inline fun <T> MessageSource<T>.consumeEachAndClose(block: (T) -> Unit)
```
JakeWharton: Hmm I'm actually tempted to leave this as an internal function of the client for now. I don't think we expose anything this DSL-like for anything else, do we?

wire-runtime/src/commonMain/kotlin/com/squareup/wire/MessageSource.kt
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire
+
+expect interface MessageSource<out T> {
+  fun receiveOrNull(): T?
```
JakeWharton: Definitely needs docs

wire-runtime/src/commonMain/kotlin/com/squareup/wire/MessageSink.kt
JakeWharton: Mostly because send/receive seems to indicate network where this might be reading from memory or filesystem or somewhere else.

swankjesse: I like `write()`.

It also makes it more clear that this operates on a value not a reference. Ie. if you write something and then mutate it, the sink doesn't observe the mutation.

wire-runtime/src/commonMain/kotlin/com/squareup/wire/MessageSource.kt
swankjesse: Should this just be `read()`? The OrNull suffix is most useful in Java where it's otherwise unclear that null is a possibility.

swankjesse:
I’m going to merge @oldergod’s PR as is and address follow-up in a separate PR. (He’s out for 2 weeks and invited me to do this!)

swankjesse:
https://github.com/square/wire/pull/1083/files

wire-runtime/src/commonMain/kotlin/com/squareup/wire/MessageSource.kt
zach-klippenstein: Late review, but should `T` have an upper bound of `Any`? Otherwise `null` is ambiguous if `T` is nullable. And if `T` _should_ be nullable, then this API is inherently problematic. I'm assuming that `null` here is intended to indicate the source is closed or something. Channels are (eventually) getting [`receiveOrClosed`](https://github.com/Kotlin/kotlinx.coroutines/commit/a8904e2b6fb774c251e491c05ffcc1ee2a2cfe2c#diff-bb85d2a203a08f1a61323b053fa20e87R268) to solve this same problem with their API.

swankjesse: https://github.com/square/wire/pull/1088

wire-grpc-client/src/main/java/com/squareup/wire/GrpcClient.kt
```diff
@@ -128,7 +128,7 @@ class GrpcClient private constructor(
     // 2. write it to the stream (blocking)
     // 3. repeat. We also have to wait for all 2s to end before closing the writer
     CoroutineScope(Dispatchers.IO).launch {
-      val requestWriter = GrpcWriter.get(
+      val requestWriter = GrpcMessageSink.get(
```
Egorand: I was wondering why we create `GrpcMessageSource` using a constructor call and `GrpcMessageSink` using a factory method?

swankjesse: Not any more!

PR #1082, opened 2019-07-17
Optimize decode() code for protos with no fields

Egorand:


Egorand:
We've got quite a few request/response protos that don't have any fields, don't know how useful that is. We actually had a couple unexpected crashes caused by R8 pruning those types as they're only used as generic parameters to Retrofit's return types.

PR #1087, opened 2019-07-19
Follow symlinks when building

swankjesse:


wire-compiler/src/main/java/com/squareup/wire/schema/Root.kt
```diff
@@ -134,15 +135,16 @@ internal class DirectoryRoot(
...
-    Files.walkFileTree(rootDirectory, object : SimpleFileVisitor<Path>() {
-      override fun visitFile(descendant: Path, attrs: BasicFileAttributes): FileVisitResult {
-        if (descendant.endsWithDotProto()) {
-          val location = Location.get(base, "${rootDirectory.relativize(descendant)}")
-          result.add(ProtoFilePath(location, descendant))
-        }
-        return FileVisitResult.CONTINUE
-      }
-    })
+    Files.walkFileTree(rootDirectory, setOf(FileVisitOption.FOLLOW_LINKS), Integer.MAX_VALUE,
```
Egorand: meganit: `Int.MAX_VALUE`

PR #1088, opened 2019-07-19
Fix the upper bounds of MessageSource and MessageSink

swankjesse:


swankjesse:
Yep, that turned out to have a pretty big blast radius.

JakeWharton:
Yeah but now you can relax and bask in the warmth of that explosion of correctness.

wire-grpc-client/src/main/java/com/squareup/wire/GrpcClient.kt
```diff
@@ -260,13 +260,13 @@ class GrpcClient private constructor(
         return unconfinedCoroutineScope.produce<Any> {
           requestChannel.send(parameter as S)
           requestChannel.close()
-          (responseChannel as Channel<Any>).toChannel(channel)
+          responseChannel.toChannel(channel)
```
Egorand: This is a very nice effect

PR #1094, opened 2019-07-25
Honor Java package names in Wire gRPC services

swankjesse:
Also add support for non-exclusive targets. This can be used to
emit a service both as a client and as a server in a single run


swankjesse:
```
BUILD SUCCESSFUL in 3m 28s
136 actionable tasks: 118 executed, 18 up-to-date
```

PR #1098, opened 2019-07-27
Change wire-runtime artifact names to preserve 2.x compat

### LINKED ISSUE #1093: Artifact ID should stay the same for Wire 3
Here’s what we did in Okio:

```
  // Use default artifact name for the JVM target
  publications {
    kotlinMultiplatform {
      artifactId = 'okio-multiplatform'
    }
    jvm {
      artifactId = 'okio'
    }
  }
```

Egorand:
Fixes #1093

### DIFF
--- a/gradle/gradle-mvn-mpp-push.gradle
+++ b/gradle/gradle-mvn-mpp-push.gradle
@@ -113,6 +113,18 @@ publishing {
     }
   }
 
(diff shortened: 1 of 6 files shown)


Egorand:
Yep, should be possible. Just copied over Okio's config.

PR #1100, opened 2019-07-28
[WIP] Kotlin 1.3.50

Egorand:
Had to attach a locally published Okio build that uses the same Kotlin version, due to Native compiler incompatibility.


localMaven/com/squareup/okio/okio-iosarm64/2.3.0-SNAPSHOT/maven-metadata.xml
```diff
@@ -0,0 +1,69 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
swankjesse: Yuck! Which tool requires this?

Egorand:
No, I intend to remove locally published Okio once there's at least a snapshot of it built with 1.3.50.

build.gradle
```diff
@@ -9,8 +9,8 @@ buildscript {
       'javapoet': '1.11.1',
       'kotlinpoet': '1.3.0',
       'jsr305': '3.0.2',
-      'kotlin': '1.3.40',
-      'okio': '2.3.0',
+      'kotlin': '1.3.50',
+      'okio': '2.4.0-SNAPSHOT',
```
jrodbx: does this need to rely on the snapshot?

wire-runtime/build.gradle
```diff
@@ -12,6 +12,7 @@ kotlin {
         metaInfo = true
       }
     }
+    nodejs()
```
jrodbx: curious: what does this do?

build.gradle
Egorand: Yes. 2.3.0 was built with 1.3.40 which includes an older, incompatible version of the Native compiler.

wire-runtime/build.gradle
Egorand: If I understand it correctly now it's mandatory to declare a real JS target for the tests to run.

JakeWharton: We probably also want `browser()`

build.gradle
JakeWharton: But we're not going to land until there's a stable version, right?

Egorand: Yeah, we'll need another Okio release to merge this.

wire-runtime/build.gradle
Egorand: Good call.

PR #1103, opened 2019-07-28
Make Message.unknownFields a val

Egorand:
Unfortunately this will require regenerating Kotlin code

### DIFF
--- a/wire-runtime/src/commonMain/kotlin/com/squareup/wire/Message.kt
+++ b/wire-runtime/src/commonMain/kotlin/com/squareup/wire/Message.kt
@@ -21,7 +21,7 @@ import okio.ByteString
 /** A protocol buffer message. */
 expect abstract class Message<M : Message<M, B>, B : Message.Builder<M, B>> protected constructor(
   adapter: ProtoAdapter<M>,
-  unknownFields: ByteString?
+  unknownFields: ByteString
 ) {
   /** If non-zero, the hash code of this message. Accessed by generated code. */
   protected var hashCode: Int
@@ -30,7 +30,7 @@ expect abstract class Message<M : Message<M, B>, B : Message.Builder<M, B>> prot
    * Returns a byte string containing the proto encoding of this message's unknown fields. Returns
    * an empty byte string if this message has no unknown fields.
    */
-  fun unknownFields(): ByteString
+  open val unknownFields: ByteString
 
   /**
    * Returns a new builder initialized with the data in this message.
--- a/wire-grpc-tests/src/test/proto-grpc/routeguide/Feature.kt
+++ b/wire-grpc-tests/src/test/proto-grpc/routeguide/Feature.kt
@@ -39,7 +39,7 @@ data class Feature(
     adapter = "routeguide.Point#ADAPTER"
   )
   val location: Point? = null,
-  val unknownFields: ByteString = ByteString.EMPTY
+  override val unknownFields: ByteString = ByteString.EMPTY
 ) : Message<Feature, Nothing>(ADAPTER, unknownFields) {
   @Deprecated(
     message = "Shouldn't be used in Kotlin",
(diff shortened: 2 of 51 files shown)


wire-runtime/src/commonMain/kotlin/com/squareup/wire/Message.kt
```diff
@@ -30,7 +30,7 @@ expect abstract class Message<M : Message<M, B>, B : Message.Builder<M, B>> prot
    * Returns a byte string containing the proto encoding of this message's unknown fields. Returns
    * an empty byte string if this message has no unknown fields.
    */
-  fun unknownFields(): ByteString
+  open val unknownFields: ByteString
```
JakeWharton: Combine with ctor param?

JakeWharton: Why open? Can't subtypes just pass their value to this supertype?

Egorand: 1. `Expected class constructor can't have a property parameter`.
2. `data` classes can't have non-property parameters in the primary constructor, so they have to declare a `val unknownFields`. The only way to avoid it clashing with the supertype's `val` is to `override`, which requires `open` or `abstract`. Kotlin doesn't allow declaring `abstract` primary constructor properties, so `open` is the only thing that worked for me. Agree that it sucks, is there another way to satisfy these constraints that I'm missing?

JakeWharton: Makes sense. Seems like we should nuke the base class constructor param then and just declare it abstract.

Egorand: That could work, but we'll need to add a new constructor to preserve binary compat. I'll give it a try.

Egorand: Nah, making the property abstract becomes a PITA for Java backwards compat. Happy to either merge as is, or punt until post-3.0.

JakeWharton: The problem is that right now you store the value in two fields: one in the supertype and one in the subtype.

Egorand: Yeah, good point. I can't think of different solution that would not require regenerating Java code.

JakeWharton: Me neither.

You could remove the `data` modifier and generate your own `equals`/`hashCode`/`copy`. You'd wind up making the classes smaller since the `componentN` functions probably aren't super useful since proto allows field re-ordering (provided the tag stays the same).

JakeWharton:
Let's land the null fix in the base class separately first. It's a trivial, obvious change.

Otherwise, I think there's value in pursuing the `val`:

First, let's agree that if we could _trivially_ make this a `val` in the same way that we made APIs in Okio and OkHttp a `val` we would. Is that safe to assume?

Next, the reason we can't do this is because `data class` requires all properties defined in the primary constructor. Before we get to this, let's step back and talk about whether `data class` is even the right choice for protocol buffers.

First, being code generation, saving characters is not a super important thing. Generating `equals`, `hashCode`, and `copy` is a single effort that gets generated making the cost of doing this fixed, not variable. So there's not much incentive there.

Second, This implicitly enables destructuring which is not a safe operation on protos. Protocol buffers are unordered maps from tag to field name. The order of the fields in a proto schema and thus the generated type are free to change provided the tag-to-name mapping remains constant. Currently you are allowed to write `val (foo, bar) = foobarbaz` where the proto schema might change to baz, bar, foo breaking the semantics of your code. If it were some kind of name based operation it would make sense, but without an ordering guarantee you're setting yourself up for future problems. If you want destructuring, it'd be far better to define them as extension functions where you're forced to determine the order-to-name mapping which will then remain fixed, even as the proto ordering changes.

Given that there's an arguable bug in using `data` and the benefits of boilerplate reduction do not apply to generated code, it should be removed. This change can be done after landing the base class null fix.

With the `data` modifier gone, the function can be trivially changed to a `val` as the primary constructor restriction no longer applies.

swankjesse:
Let switch from data classes to regular classes. Then the costs of this improvement is low & I’m on board.

Egorand:
Yep, agree with what Jake said. I'll work on landing both fixes.

PR #1106, opened 2019-07-30
Docs on the Wire compiler and Gradle plugin

swankjesse:


swankjesse:
Rendered: https://github.com/square/wire/blob/7876a6b25cec0012b88ceedd65ce3d592ccf9cd7/docs/wire_compiler.md

PR #1109, opened 2019-07-30
Make PipeDuplexRequestBody internal

swankjesse:
Also promote docs from the JVM project to common where appropriate.

### DIFF
--- a/wire-runtime/src/commonMain/kotlin/com/squareup/wire/MessageSink.kt
+++ b/wire-runtime/src/commonMain/kotlin/com/squareup/wire/MessageSink.kt
@@ -18,7 +18,33 @@ package com.squareup.wire
 import com.squareup.wire.internal.Throws
 import okio.IOException
 
+/**
+ * A writable stream of messages.
+ *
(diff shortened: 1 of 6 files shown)


wire-runtime/src/commonMain/kotlin/com/squareup/wire/MessageSink.kt
```diff
@@ -18,7 +18,33 @@ package com.squareup.wire
...
+ *
+ * Typical implementations will immediately encode messages and enqueue them for transmission, such
+ * as for client-to-server or server-to-client networking. But this interface is not limited to 1-1
+ * networking use cases and implementations may persist, broadcast, validate, or take any other
+ * action with the messages.
+ *
+ * There is no flushing mechanism. Messages are flushed one-by-one as they are written. This
+ * minimizes latency at a potential cost of throughput.
+ *
+ * On its own this offers no guarantees that messages are delivered. For example, a message may
```
Egorand: nit: remove `may`?

PR #1111, opened 2019-07-31
Generate equals() method for Kotlin protos

Egorand:


wire-tests/src/jvmTest/proto-kotlin/com/squareup/wire/protos/kotlin/Form.kt
```diff
@@ -215,6 +235,8 @@ data class Form(
       throw AssertionError()
     }
 
+    override fun equals(other: Any?): Boolean = other is ButtonElement
```
Egorand: Hmm, is this correct? What if the `unknownFields` are different? Implementation of `equals()` is copied from `JavaGenerator`.

JakeWharton: Looks broken!

JakeWharton: And probably should be fixed separately since it affects both generators

PR #1112, opened 2019-07-31
Generate hashCode() method for Kotlin protos

Egorand:


wire-grpc-tests/src/test/proto-grpc/routeguide/Feature.kt
```diff
@@ -59,6 +59,16 @@ data class Feature(
         && location == other.location
   }
 
+  override fun hashCode(): Int {
+    var result = super.hashCode
+    if (result == 0) {
+      result = result * 37 + (name?.hashCode() ?: 0)
```
JakeWharton: Since 1.3 there's a nullable-receiver variant of `hashCode`: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/hash-code.html

PR #1114, opened 2019-07-31
Generate toString() method for all Kotlin protos

Egorand:
Also, make the implementation more consistent with Java.


wire-grpc-tests/src/test/proto-grpc/routeguide/Feature.kt
```diff
@@ -70,6 +70,13 @@ data class Feature(
     return result
   }
 
+  override fun toString(): String = buildString {
+    append("Feature{")
+    if (name != null) append("""name=$name""")
+    if (location != null) append(""", location=$location""")
```
JakeWharton: if `name` is null and `location` is not the output is
```
Feature{, location=...}
```

Here's a workaround I've used in the past:
```kotlin
append("Featur") // <-- should be name minus one character
if (name != null) append(""", name=$name""")
if (location != null) append(""", location=$location""")
set(6, 'e') // these values change based on the length of the name
set(7, '{')
append('}')
```
Since the shortest name is of size 1, this always works. (And in that case you'd have no initial `append`)

JakeWharton: Oh dear it looks like mutating functions on StringBuilder are non-existent in common code 😭

Egorand: Ugh.. Will fix, good catch.

oldergod: could do a `listOf().joinAsString` as well right?

JakeWharton: You would still need to filter nulls and now you're talking about multiple collection allocations, the array allocation to the varargs, their internal expansion when more than 10 properties are defined, etc.

Not great, but might be fine to start with.

Egorand:
@JakeWharton @oldergod PTAL

pavlospt:
@Egorand Has the new solution with the List, removed the `,` between the property name-value pairs?

Egorand:
@pavlospt no, it uses the default separator, which is `, `. I agree that it's a good idea to specify a separator explicitly, I'll add it in a sec.

JakeWharton:
https://youtrack.jetbrains.com/issue/KT-33069

PR #1117, opened 2019-08-01
Make Message.unknownFields a val

Egorand:


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -635,9 +635,6 @@ class KotlinGenerator private constructor(
         ParameterSpec.builder(unknownFields, byteClass)
             .defaultValue("%T.EMPTY", byteClass)
             .build())
-    classBuilder.addProperty(PropertySpec.builder(unknownFields, byteClass)
```
Egorand: KotlinPoet design decisions in action!

PR #1124, opened 2019-08-05
Expose Schema linking as static method

jrodbx:
`SchemaLoader.load()` returned a linked Schema by indirectly calling `Linker.link()`, whereas `NewSchemaLoader.load()` returns a `List<ProtoFile>` pulling the linking behavior up into `WireRun.execute()`.  This change allows tooling (graph visualization, keep rule generator) built on top of the old compiler to migrate to the new Wire approach, albeit temporarily.

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
+++ b/wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
@@ -151,7 +151,7 @@ data class WireRun(
     val protoFilesPlusDescriptor = protoFiles.plus(schemaLoader.loadDescriptorProto())
 
     // 3. Validate the schema and resolve references
-    val fullSchema = Linker(protoFilesPlusDescriptor).link()
+    val fullSchema = Schema.fromFiles(protoFilesPlusDescriptor)
 
     // 4. Optionally prune the schema.
     val schema = treeShake(fullSchema, logger)
--- a/wire-schema/src/main/java/com/squareup/wire/schema/Schema.java
+++ b/wire-schema/src/main/java/com/squareup/wire/schema/Schema.java
@@ -109,6 +109,10 @@ public final class Schema {
     return field;
   }
 
+  public static Schema fromFiles(Iterable<ProtoFile> files) {
+    return new Linker(files).link();
+  }
+
   private static ImmutableMap<String, Type> buildTypesIndex(Iterable<ProtoFile> protoFiles) {
     Map<String, Type> result = new LinkedHashMap<>();
     for (ProtoFile protoFile : protoFiles) {


wire-schema/src/main/java/com/squareup/wire/schema/Linker.java
```diff
@@ -27,14 +27,14 @@
...
-final class Linker {
+public final class Linker {
   private final ImmutableList<ProtoFile> protoFiles;
   private final Map<String, Type> protoTypeNames;
   private final Multimap<String, String> imports;
   private final List<String> errors;
   private final List<Object> contextStack;
 
-  Linker(Iterable<ProtoFile> protoFiles) {
+  public Linker(Iterable<ProtoFile> protoFiles) {
```
JakeWharton: If we're going to expose this type I would move creation into a static factory method just in case.

jrodbx: works for me

jrodbx: nvm, added a top-level extension function instead to return a linked Schema from a list of ProtoFile, mimicking previous behavior without weakening access.

wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
```diff
@@ -225,3 +225,7 @@ data class WireRun(
     return result
   }
 }
+
+fun List<ProtoFile>.link(): Schema {
```
JakeWharton: Do we care about Java callers invoking this? Right now they would call `WireRunKt.link(foo)` which isn't great.

Also I'm not sure whether this should have a more descriptive name like `toSchema` or `linkToSchema`.

JakeWharton: The closest example in the stdlib is probably `Iterable<Pair<K, V>>.toMap` so I think I've vote for `toSchema` for Kotlin callers.

But for Java I'd probably want `Schema.fromFiles(files)`.

Egorand: Agree with Jake's comments, but it also confuses me why we want this method to live in `WireRun.kt`. It should either be a part of `Schema.java`, or maybe `NewSchemaLoader`.

jrodbx: moved to Schema as a public static method

wire-schema/src/main/java/com/squareup/wire/schema/Schema.java
```diff
@@ -109,6 +110,10 @@ public Field getField(ProtoMember protoMember) {
     return field;
   }
 
+  public static Schema fromFiles(List<ProtoFile> files) {
```
JakeWharton: nit: Iterable

jrodbx:
```
* What went wrong:
Execution failed for task ':wire-compiler:dokka'.
> Guice provision errors:
  
  1) Error injecting constructor, java.lang.RuntimeException: Exception while loading package-list from ExternalDocumentationLinkImpl(url=http://docs.oracle.com/javase/6/docs/api/, packageListUrl=http://docs.oracle.com/javase/6/docs/api/package-list)
    at org.jetbrains.dokka.ExternalDocumentationLinkResolver.<init>(ExternalDocumentationLinkResolver.kt:28)
    at org.jetbrains.dokka.ExternalDocumentationLinkResolver.class(ExternalDocumentationLinkResolver.kt:28)
    while locating org.jetbrains.dokka.ExternalDocumentationLinkResolver
      for parameter 4 at org.jetbrains.dokka.DeclarationLinkResolver.<init>(DeclarationLinkResolver.kt:13)
    while locating org.jetbrains.dokka.DeclarationLinkResolver
      for parameter 0 at org.jetbrains.dokka.PackageDocs.<init>(PackageDocs.kt:13)
    at org.jetbrains.dokka.PackageDocs.class(PackageDocs.kt:12)
    while locating org.jetbrains.dokka.PackageDocs
  
  1 error
```

known issue?  ok to merge on red master?

Egorand:
I believe I've seen that Dokka issue once or twice, it recovers on the next run. Might be worth filing a bug report if it's not reported yet.

PR #1131, opened 2019-08-10
Improve generation output path

### LINKED ISSUE #1130: gradle plugin generating code at the wrong folder
In my `app/build.gradle`
```
wire {
  kotlin {
    out "src/main/java"
  }
}
```
The generated code is going to `/src/main/java` instead of `/app/src/main/java`. I believe `WireCompiler` was doing it as expected.

- [ ] write tests.
- [ ] fix.

jrodbx:
Closes #1130

### DIFF
--- a/docs/wire_compiler.md
+++ b/docs/wire_compiler.md
@@ -63,7 +63,7 @@ code. It uses [KotlinPoet][kotlinpoet] internally to generate compact and legibl
 
 ![Basic](images/gradle_basic@2x.png)
 
-Wire will generate files in `build/generated/src/main/java`. It'll also register this directory as a
+Wire will generate files in `build/generated/source/wire`. It'll also register this directory as a
 source directory for the project so the generated sources are compiled by the Kotlin compiler.
 
 Replace `kotlin` with `java` to generate Java sources instead.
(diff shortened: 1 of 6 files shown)


docs/wire_compiler.md
```diff
@@ -63,7 +63,7 @@ code. It uses [KotlinPoet][kotlinpoet] internally to generate compact and legibl
 
 ![Basic](images/gradle_basic@2x.png)
 
-Wire will generate files in `build/generated/src/main/java`. It'll also register this directory as a
+Wire will generate files in `build/generated/source/wire`. It'll also register this directory as a
```
jrodbx: kapt, apt use `build/generated/source`.
sqldelight use `build/sqldelight`.
the former seemed more conventional.

oldergod: I like `build/wire` more; isn't just beautifully simple?

jrodbx: It's shorter, for sure. 
On the other hand, buildd/generated/source/wire vs build/generated/source/apt vs build/generated/source/kapt has a nice symmetric structure IMO.  Do you feel strongly either way?

Egorand: My personal preference is for `build/generated/source/wire`, but both are fine IMO.

PR #1134, opened 2019-08-11
Multiple deploy_website.sh fixes

Egorand:
- Make OSX sed command the default
- Undo the results of sed to jump branches
- --set-upstream before pushing to gh-pages

### DIFF
--- a/deploy_website.sh
+++ b/deploy_website.sh
@@ -46,8 +46,7 @@ git checkout -- .
 # Restore Javadocs from 2.x
 git checkout gh-pages
 git cherry-pick 0e1192aaa1d074c9703748fae100daef707218d4
-git push --set-upstream origin gh-pages
-git push
+git push --set-upstream origin gh-pages && git push
 
 # Delete our temp folder
 cd ..
--- a/deploy_website.sh
+++ b/deploy_website.sh
@@ -28,9 +28,9 @@ cd $DIR
 
 # Fix *.md links to point to where the docs live under Mkdocs.
 # Linux
-sed -i 's/docs\/wire_compiler.md/wire_compiler/' README.md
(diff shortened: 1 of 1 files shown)


deploy_website.sh
```diff
@@ -40,9 +40,13 @@ cp CONTRIBUTING.md docs/contributing.md
 # Build the site and push the new files up to GitHub
 mkdocs gh-deploy
 
+# Undo any file changes to be able to jump branches.
+git checkout -- .
```
jrodbx: would `git reset --soft` suffice?

deploy_website.sh
```diff
@@ -40,9 +40,13 @@ cp CONTRIBUTING.md docs/contributing.md
...
+
 # Restore Javadocs from 2.x
 git checkout gh-pages
 git cherry-pick 0e1192aaa1d074c9703748fae100daef707218d4
+git push --set-upstream origin gh-pages
 git push
```
jrodbx: combine these two lines?

deploy_website.sh
Egorand: No, gotta be `--hard`. What's the difference anyway?

deploy_website.sh
pavlospt: There is also a shorter version with `git push -u origin gh-pages` if you want!

deploy_website.sh
pavlospt: `reset --soft` returns to the desired hash while retaining the changes from all the commits between the `HEAD` and desired hash.

Egorand: I'm trying to undo the changes introduced by the previous commands in this script, and `git reset --soft` doesn't seem to cut it. `git checkout -- .` works, so I'll keep it.

pavlospt: Yes, you need a hard reset here, or a stash!

PR #1135, opened 2019-08-11
Rename Util classes in wire-runtime and wire-grpc-client

### LINKED ISSUE #1133: (Android) grpc-client and runtime both define 'Util.kt'
AFAIK Android support for 3.0.0 is not yet ready so maybe this is known, but I tried crunching through to see where I'd get stuck anyway :)

Because both modules define `com.squareup.wire.internal.Util`, on the latest SNAPSHOT builds (and alpha) Android builds fail with:

```
Duplicate class com.squareup.wire.internal.UtilKt found in modules wire-grpc-client-3.0.0-SNAPSHOT.jar (com.squareup.wire:wire-grpc-client:3.0.0-SNAPSHOT:20190809.205059-66) and wire-runtime-3.0.0-SNAPSHOT.jar (com.squareup.wire:wire-runtime:3.0.0-SNAPSHOT)
```

*Edit:* this was the only issue blocking Android from making a successful gRPC call with grpc-client. I managed to make a successful call after changing the namespace to something more specific.

Egorand:
Fixes #1133

### DIFF
--- a/wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/Util.kt
+++ b/wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/Util.kt
@@ -13,7 +13,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:JvmName("RuntimeUtils")
+
 package com.squareup.wire.internal
+
+import kotlin.jvm.JvmName
 
 @Suppress("NOTHING_TO_INLINE") // Syntactic sugar.
 internal inline infix fun Byte.and(other: Int): Int = toInt() and other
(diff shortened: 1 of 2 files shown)


wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/Util.kt
```diff
@@ -13,8 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:JvmName("RuntimeUtils")
+
 package com.squareup.wire.internal
 
+import kotlin.jvm.JvmName
```
jrodbx: why wasn't this necessary in the other file edit?

Egorand: I guess that's because it's common code, probably `kotlin-stdlib-jvm` has this file in `kotlin.*`, so it doesn't need to be imported, whereas `kotlin-stdlib-common` has it in `kotlin.jvm.*`, hence requires an import.

PR #1136, opened 2019-08-11
Depend on OkHttp 4.0.1

Egorand:


swankjesse:
Did IntelliJ do it all?

Egorand:
Yep!

PR #1139, opened 2019-08-14
Provide explicit option to generate protos on build; default to true

jrodbx:


wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
```diff
@@ -131,6 +131,13 @@ open class WireExtension(
     populateRootSets(action, protoTrees, protoJars, "proto-tree")
   }
 
+  /**
+   * A flag indicating whether to generate protos before every compilation run
+   */
+  @get:Input
+  @get:Optional
+  var generateOnBuild: Boolean = false
```
JakeWharton: Extremely surprising default. Would invert.

AlecStrong:
> I remain, in general, opposed to such a flag. If the task is properly incremental, making your inputs fixed is sufficient to produce this behavior. The `JavaCompile` task does not have a flag to prevent generating class files by default. The `KotlinCompile` task does not have a flag to prevent generating class files by default. SQL Delight does not have a flag to prevent performing code gen by default. Every compiler integration in Gradle does not have a flag to prevent performing its transformation by default.

I think it's different as a client consumer of protos - we're not in charge of the schema like we are with sql or java and generating the entire proto graph introduces breaking changes if a field is removed upstream, so theres no way to selectively update specific protos anymore

JakeWharton:
I understand that, but it's not the Wire compiler or the Wire Gradle plugin's responsibility to support that case. You have the ability to fix your inputs such that from this plugin's point of view the protos are static allowing Gradle to determine that the task is up-to-date and not re-run the compiler.

Imagine the case where I have my server and client in the same repo and are using Wire to generate code for both. Upon adding a `.proto` and adding the Wire plugin, nothing would happen. I would have to manually run the task in order for it to generate files that I could compile against. Commit the proto and your server and client code and send it off to code review. CI fails because it now also needs to explicitly run the task before the server or client can compile. After all, the generated files are put in the `build/` directory by default which is not committed to version control.

I know you commit the generated code and that's fine. That is not the way in which the majority of people will be using this plugin and your setup should not leak into the plugin, certainly not in this way where it's the default behavior.

oldergod:
> so theres no way to selectively update specific protos anymore

We could control that by checking in the proto files.
I think putting a version to the proto we are depending on, and checking in the proto instead of the generated code, is our sweet spot. It looks to me it would fix all solutions without breaking existing ones.

Egorand:
Spoke with Alec and JRod offline about this, there's definitely room for improving the setup to be able to fully rely on the inputs, but right now having a flag is the easiest way to unblock Cash. I suggest keeping the flag undocumented, or explicitly stating that it's not part of the API, so that we can remove it once our proto tooling is in a better place.

JakeWharton:
Can it be deprecated and issue a Gradle warning when set to false that's it's an unsupported configuration? Or maybe it could be set using an undocumented property that you put in `gradle.properties` similar to those AGP uses.

jrodbx:
Figured out a workaround to decouple proto generation from module compilation and unblock Cash:

```
// Don't link proto generation to this module's compilation
afterEvaluate {
  tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
    dependsOn.remove(tasks.named("generateProtos"))
  }
}
```

Globegitter:
@jrodbx Now with the latest gradle (6.6.1) I get the following error: `Removing a task dependency from a task instance is not supported.` did you find any workaround for that?

PR #1140, opened 2019-08-14
Fail if sourcePath.srcDir doesn't exist

### LINKED ISSUE #1132: Gradle plugin fails silently if srcDir does not exist
**Affects version:** 3.0.0-rc01

I accidentally configured the `sourcePath` -> `srcDir` as follows:

```
wire {
    sourcePath {
        srcDir '../non-existent-relative-path'
    }
    ...
```

Since I was a first time user, it took me a while before I eventually ran `gradle generateProtos --info` and saw that I had incorrectly specified the relative path to my source protos.

I imagine this should ideally fail with an error?

jrodbx:
Closes #1132

### DIFF
--- a/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt
+++ b/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt
@@ -60,6 +60,12 @@ internal class WireInput(
 
   fun addTrees(trees: Set<SourceDirectorySet>) {
     for (tree in trees) {
+      // TODO: this eagerly resolves dependencies; fix this!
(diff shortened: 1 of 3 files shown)


wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt
```diff
@@ -60,6 +60,12 @@ internal class WireInput(
 
   fun addTrees(trees: Set<SourceDirectorySet>) {
     for (tree in trees) {
+      // TODO: this eagerly resolves dependencies; fix this!
```
swankjesse: This comment suggests there’s a better place to do this?

PR #1144, opened 2019-08-15
Tighten up the test for gRPC

swankjesse:
I'm about to copy-paste this into the blocking test.


wire-grpc-tests/src/test/java/com/squareup/wire/MockRouteGuideService.kt
```diff
@@ -55,6 +56,44 @@ class MockRouteGuideService : RouteGuideGrpc.RouteGuideImplBase(), TestRule {
     script.add(action)
   }
 
+  fun enqueueReceiveNote(message: String) {
+    enqueue(Action.ReceiveMessage(RouteNote.newBuilder()
```
Egorand: Doesn't `RouteNote(message = message)` work? You shouldn't need to use `newBuilder()`.

swankjesse: This is a Google protobuf!

PR #1146, opened 2019-08-16
Implement the runtime for the blocking gRPC client

swankjesse:
The test cases is a nearly 1-1 port of the coroutines test.

There's quite a few things going on here:

1. Move GrpcMethod into its own file
2. Decompose the call function into reusable parts that don't
   need coroutines. These are mostly in grpc.kt
3. BlockingMessageSource works like a channel, leaning on LinkedBlockingQueue
   to do most of the heavy lifting
4. Use polymorphism instead of 'when' in the dynamic proxy


swankjesse:
Also there’s one major gap in this design. We always use a dispatcher thread while reading the response messages. In a follow up we should change it to only use the dispatcher when strictly necessary - that’s cases where the request is streaming.

wire-grpc-client/src/main/java/com/squareup/wire/internal/grpc.kt
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * 1. read a message (non blocking, suspending code)
+ * 2. write it to the stream (blocking)
+ * 3. repeat. We also have to wait for all 2s to end before closing the writer
+ */
+internal fun <S : Any> ReceiveChannel<S>.writeToRequestBody(
+  requestBody: PipeDuplexRequestBody,
+  requestAdapter: ProtoAdapter<S>
+) {
+  CoroutineScope(Dispatchers.IO).launch {
```
JakeWharton: I realize you copied this code, but this seems like a leak. This should be a suspend function and the caller should do something like
```kotlin
runBlocking {
  launch(IO) {
    channel.writeToRequestBody(.., ..)
  }
  somethingElse()
}
```

wire-grpc-client/src/main/java/com/squareup/wire/internal/BlockingMessageSource.kt
```diff
@@ -0,0 +1,90 @@
...
+
+  override fun close() {
+    call.cancel() // Short-circuit the request stream if it's still active.
+  }
+
+  /** Read messages from the response body and write them to the deque. */
+  fun readFromResponseBodyCallback(): Callback {
+    return object : Callback {
+      override fun onFailure(call: Call, e: IOException) {
+        queue.put(e)
```
JakeWharton:
```suggestion
        queue.put(Failure(e))
```

wire-grpc-client/src/main/java/com/squareup/wire/GrpcMethod.kt
```diff
@@ -0,0 +1,276 @@
...
+              && responseType == MessageSource::class.java) {
+            return BlockingFullDuplex(wireRpc.path, requestAdapter, responseAdapter)
+          }
+        }
+      }
+
+      error("unexpected gRPC method: $this")
+    }
+  }
+}
```
JakeWharton: Check that IDE box!

wire-grpc-client/src/main/java/com/squareup/wire/GrpcMethod.kt
```diff
@@ -0,0 +1,276 @@
...
+      grpcClient: GrpcClient,
+      args: Array<Any>
+    ): Pair<SendChannel<S>, Deferred<R>> {
+      val (requestChannel, responseChannel) = super.callWithChannels(grpcClient)
+
+      return Pair(
+          requestChannel,
+          // We use ATOMIC to guarantee the coroutine doesn't start after potential cancellations
+          // happen.
+          unconfinedCoroutineScope.async(start = CoroutineStart.ATOMIC) {
```
JakeWharton: Need to wrap my head around this more, and this is also copied from existing code, but this feels like another case of scope misuse.

wire-grpc-tests/src/test/java/com/squareup/wire/GrpcBlockingClientTest.kt
```diff
@@ -0,0 +1,284 @@
...
+    val (requestChannel, responseChannel) = routeGuideService.RouteChat()
+    requestChannel.write(RouteNote(message = "marco"))
+    assertThat(responseChannel.read()).isEqualTo(RouteNote(message = "polo"))
+    requestChannel.write(RouteNote(message = "rené"))
+    assertThat(responseChannel.read()).isEqualTo(RouteNote(message = "lacoste"))
+    requestChannel.close()
+    assertThat(responseChannel.read()).isNull()
+
+    // We wait for mockService to process our actions.
+    Thread.sleep(500)
```
JakeWharton: Lots of sleeps :(

wire-grpc-client/src/main/java/com/squareup/wire/internal/BlockingMessageSource.kt
swankjesse: oooh good catch. Test coverage is not there.

wire-grpc-client/src/main/java/com/squareup/wire/GrpcMethod.kt
swankjesse: Yeah, we were learning as we went on this.

wire-grpc-tests/src/test/java/com/squareup/wire/GrpcBlockingClientTest.kt
swankjesse: Yeah. Maybe we want to be able to give the mock server a CountDownLatch or something.

wire-grpc-client/src/main/java/com/squareup/wire/GrpcMethod.kt
swankjesse: Screen Shot 2019-08-16 at 11 04 33 AM

swankjesse: Added.

swankjesse: Also, invisible whitespace is dumb and I’m annoyed that this is even a warning that exists.

wire-grpc-tests/src/test/java/com/squareup/wire/GrpcBlockingClientTest.kt
```diff
@@ -0,0 +1,286 @@
...
+    requestChannel.write(RouteNote(message = "marco"))
+    assertThat(responseChannel.read()).isEqualTo(RouteNote(message = "polo"))
+    requestChannel.write(RouteNote(message = "rené"))
+    assertThat(responseChannel.read()).isEqualTo(RouteNote(message = "lacoste"))
+    assertThat(responseChannel.read()).isNull()
+    requestChannel.close()
+  }
+
+  @Test
+  @Ignore // TODO(jwilson): rework this test for blocking APIs.
```
swankjesse: needed to suppress this too

wire-grpc-client/src/main/java/com/squareup/wire/GrpcMethod.kt
JakeWharton: I don't disagree. Trailing newlines are nice because when you add new code to the end of a file you see only added lines instead of a changed line and then added lines. Not sure why it's still opt-in.

PR #1147, opened 2019-08-16
Support blocking clients in KotlinGenerator

swankjesse:
Still needs to be hooked up in WireRun and the Gradle plugin


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -119,18 +134,17 @@ class KotlinGenerator private constructor(
...
-        if (rpc == null) generatedServiceName(service) to service.rpcs()
-        else generatedServiceRpcName(service, rpc) to listOf(rpc)
-
-    val superinterface = com.squareup.wire.Service::class.java
+    val serviceName = generatedServiceName(service, rpc)
+    val builder = TypeSpec.interfaceBuilder(serviceName)
+        .addSuperinterface(com.squareup.wire.Service::class.java)
+
+    val rpcs = if (rpc == null) service.rpcs() else listOf(rpc)
+    for (it in rpcs) {
```
JakeWharton: skeptical of `it` as an explicit for loop variable name

oldergod: `rpcs.forEach {`?

PR #1150, opened 2019-08-17
Support configuring gRPC options in the Gradle plugin

swankjesse:


wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -60,186 +60,192 @@ sealed class Target {
     logger: WireLogger
   ): SchemaHandler
 
-  /** Generate `.java` sources. */
-  data class JavaTarget(
```
swankjesse: I promoted this to be top-level

wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -60,186 +60,192 @@ sealed class Target {
...
-      }
-    }
+  interface SchemaHandler {
+    fun handle(type: Type)
+    fun handle(service: Service)
   }
+}
 
-  /** Generate `.kt` sources. */
-  data class KotlinTarget(
```
swankjesse: I promoted this to be top-level

wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -60,186 +60,192 @@ sealed class Target {
...
+          throw IOException("Error emitting " +
+              "${kotlinFile.packageName}.${service.type()} to $outDirectory", e)
         }
       }
     }
   }
+}
 
-  /** Omit code generation for these sources. Use this for a dry-run. */
-  data class NullTarget(
```
swankjesse: I promoted this to be top-level

docs/wire_compiler.md
```diff
@@ -318,8 +319,13 @@ wire {
     // target.
     javaInterop = true
 
-    // True for emitted services to provide blocking APIs only.
-    blockingServices = true
+    // SUSPENDING to generate coroutines APIs that require a Kotlin coroutines context.
+    // BLOCKING to generate blocking APIs callable by Java and Kotlin.
+    rpcCallStyle = com.squareup.wire.kotlin.RpcCallStyle.BLOCKING
```
Egorand: With no auto-complete in Groovy I expect these to be painful to use.. Should we introduce overloads that accept Strings?
```groovy
wire {
  kotlin {
    rpcCallStyle = 'blocking'
    rpcRole = 'server'
  }
}
```

swankjesse: What’s idiomatic?

swankjesse: I’m gonna land as-is and add more later.

Egorand: Yeah, not sure what the idiomatic way is.

JakeWharton: definitely a string. you can do `values().singleOrNull { it.equals(whatever, ignoreCase=true) } ?: throw IllegalArgumentException("Unknown whatever $whatever. Valid values: ${values().contentToString()}")`

PR #1154, opened 2019-08-20
Send stream cancels from clients

swankjesse:
This just leans on Call.cancel which cancels the underlying HTTP/2 stream.


wire-runtime/src/commonMain/kotlin/com/squareup/wire/MessageSink.kt
```diff
@@ -44,10 +44,27 @@ expect interface MessageSink<in T : Any> {
...
+  /**
+   * Truncate this stream abnormally. This attempts to signal to readers of this data that it is
+   * incomplete. Note that unlike some cancel methods this is not safe for concurrent use.
+   *
+   * A stream must be closed after it is canceled.
+   *
+   * @throws IllegalStateException if this stream is closed.
+   */
+  @Throws(IOException::class)
+  fun cancel()
```
JakeWharton: Not super enthusiastic about this as an API surface (as mentioned elsewhere) since these types do not intrinsically imply or represent something that could be canceled.

Cancel seems tied to the notion of gRPC and also may not be tied to the notion of streaming. What would it look like on an endpoint that had streaming requests but a single response?

JakeWharton: Additionally, this only exists on sink creating an asymmetry in the API with source. Why can't reading from a source of messages be canceled but a sink can? Aside from cancel being a weird concept to model on these types, that feels like it's leaking the requirements of a specific implementation.

This doesn't feel like the right API.

swankjesse: The sink indicates a writer-initiated failure with cancel(), the source indicates it with an exception.

JakeWharton: That sounds like something to do with your implementation and not general concepts to be modeled in this API.

If I'm modeling a stream of incoming messages from SSE or a WebSocket with MessageSink how do I cancel it?

swankjesse: Discussed offline. As is is the least bad of a few options.

PR #1155, opened 2019-08-20
KotlinGeneratorTest: invoke Kotlin compiler

sullis:


Egorand:
We generate test code into the `wire-tests` module, which now has a number of compilation units that cover all flavors of generated code. That code is getting built and exercised by the test cases inside the same module. Therefore, I think that running compiler against the code generated by `KotlinGeneratorTest` is redundant. I actually think that that test case is redundant as well, so we might remove it in the future. Anyway, thanks for your effort!

PR #1158, opened 2019-08-22
Use routeguide corresponding RPCs for tests

oldergod:
Using derivatives of GetFeature everywhere is confusing


swankjesse:
Much better.

PR #1159, opened 2019-08-23
Make unknownFields nullable to reflection

swankjesse:
This is particularly unfortunate but the least worst of several
bad options. Our backends incorrectly use Gson to reflectively
create Wire objects, those Wire objects don't include a good
representation of unknownFields.


wire-grpc-tests/src/test/proto-grpc/routeguide/Feature.kt
```diff
@@ -55,7 +55,7 @@ class Feature(
   override fun equals(other: Any?): Boolean {
     if (other === this) return true
     if (other !is Feature) return false
-    return unknownFields == other.unknownFields
+    return unknownFields() == other.unknownFields()
```
swankjesse: no longer a val. I couldn’t figure out how to have a nullable JVM field named `unknownFields` and also a non-nullable val with the same name.

wire-gson-support/src/test/java/com/squareup/dinosaurs/Dinosaur.java
```diff
@@ -0,0 +1,252 @@
+// Code generated by Wire protocol buffer compiler, do not edit.
```
swankjesse: Just for the regression test.

wire-runtime/src/commonMain/kotlin/com/squareup/wire/Message.kt
```diff
@@ -30,7 +30,7 @@ expect abstract class Message<M : Message<M, B>, B : Message.Builder<M, B>> prot
    * Returns a byte string containing the proto encoding of this message's unknown fields. Returns
    * an empty byte string if this message has no unknown fields.
    */
-  val unknownFields: ByteString
+  fun unknownFields(): ByteString
```
JakeWharton: This shouldn't have to change to support the behavior.

wire-grpc-tests/src/test/proto-grpc/routeguide/Feature.kt
JakeWharton: You probably need to do something like
```kotlin
@JvmName("unknownFields")
private val _unknownFields: ByteString?

val unknownFields get() = _unknownFields ?: EMPTY
```

swankjesse: I tried it and IntelliJ complained. Lemme try again.

swankjesse: Yeah, `@JvmName` can’t target field names.

```
/**
 * Specifies the name for the Java class or method which is generated from this element.
 *
 * See the [Kotlin language documentation](https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#handling-signature-clashes-with-jvmname)
 * for more information.
 * @property name the name of the element.
 */
@Target(AnnotationTarget.FUNCTION, AnnotationTarget.PROPERTY_GETTER, AnnotationTarget.PROPERTY_SETTER, AnnotationTarget.FILE)
@Retention(AnnotationRetention.BINARY)
@MustBeDocumented
public actual annotation class JvmName(actual val name: String)
```

wire-runtime/src/commonMain/kotlin/com/squareup/wire/Message.kt
swankjesse: Lets try something completely different:
https://github.com/square/wire/pull/1160/files

PR #1160, opened 2019-08-23
Defend against Message.unknownFields being impossibly null

swankjesse:


wire-runtime/src/jvmMain/kotlin/com/squareup/wire/Message.kt
```diff
@@ -28,12 +28,22 @@ actual abstract class Message<M : Message<M, B>, B : Message.Builder<M, B>>
...
    */
-  @field:Transient @get:JvmName("unknownFields") actual val unknownFields: ByteString
-) : Serializable {
+  @field:Transient @get:JvmName("unknownFields")
+  actual val unknownFields: ByteString = unknownFields
+    get() {
+      // Some naughty libraries construct Messages by reflection which causes this non-null field
+      // to have a null value. We defend against this with an otherwise-redundant null check.
+      @Suppress("SENSELESS_COMPARISON")
+      if (field == null) return ByteString.EMPTY
```
Egorand: I don't think it'll work. You'll get an NPE when you try to construct a `Message` object and pass a null for `unknownFields`.

JakeWharton: That's what we want to happen though. This still allows Gson to bypass the constructor which will put the backing field in a default, `null` state.

Egorand: Ah, got it. Nice!

wire-gson-support/src/test/java/com/squareup/dinosaurs/Dinosaur.java
```diff
@@ -0,0 +1,252 @@
+// Code generated by Wire protocol buffer compiler, do not edit.
```
oldergod: did you actually copy paste those?

swankjesse: Yes I did! We could generate them instead but it's less urgent.

PR #1162, opened 2019-08-25
Move KotlinEnumTest to common

Egorand:


wire-tests/src/commonTest/kotlin/com/squareup/wire/KotlinEnumTest.kt
```diff
@@ -0,0 +1,15 @@
+package com.squareup.wire
```
swankjesse: Nit: copyright

PR #1165, opened 2019-08-26
Don't emit proto2 sources for proto3 syntax

swankjesse:
We don't currently implement proto3 syntax. Detect it and skip rather than emitting
an invalid message.


pavlospt:
Why is this needed? Up until now proto3 works fine (or at least for us) and this is with our own responsibility that we use Wire with the unsupported syntax! Will this PR essentially remove that ability of generating sources for proto3 files??

pavlospt:
That's what I am not understanding. Why disable something that is working in a way it works right now and implement the proto3 behaviour on the side? Ideally have it as a Gradle plugin/CLI flag for opt-in, in order to let anyone who wants to still use that possibly error-prone generation of sources for proto3...!

PR #1176, opened 2019-09-06
Changes on printing options and enum

oldergod:
- No empty lines between options and fields for enums (same as types)
- Print options on new lines only when more than one.

Here's the expected output of this test https://github.com/square/wire/blob/47d004cd3271295aca859a2aa7d1d4d2482a5657/proto-pruner/src/test/expected/simple_message.proto

### DIFF
--- a/wire-schema/src/main/java/com/squareup/wire/schema/internal/Util.java
+++ b/wire-schema/src/main/java/com/squareup/wire/schema/internal/Util.java
@@ -35,8 +35,14 @@ public final class Util {
   }
 
   public static void appendOptions(StringBuilder builder, List<OptionElement> options) {
+    int count = options.size();
+    if (count == 1) {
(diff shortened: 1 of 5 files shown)


wire-schema/src/main/java/com/squareup/wire/schema/internal/Util.java
```diff
@@ -35,8 +35,14 @@ public static void appendDocumentation(StringBuilder builder, String documentati
   }
 
   public static void appendOptions(StringBuilder builder, List<OptionElement> options) {
+    int count = options.size();
+    if (count == 1) {
+      builder.append("[").append(options.get(0).toSchema()).append(']');
```
JakeWharton: suuuuuuuper nit
```suggestion
      builder.append('[').append(options.get(0).toSchema()).append(']');
```

PR #1182, opened 2019-09-11
Create separate compilation units for Java test flavors

Egorand:


wire-compiler/src/test/java/com/squareup/wire/WireCompilerTest.kt
```diff
@@ -642,6 +642,19 @@ class WireCompilerTest {
         return expectedFile.also {
           println("Comparing against expected output $name")
         }
+      } else if (protoFolderSuffix() == "java") {
```
swankjesse: Nice

PR #1183, opened 2019-09-11
Move WireKotlinTest into jvmJavaTest

Egorand:


Egorand:
Gonna self-merge cause it's a simple file move, reviews welcome.

PR #1187, opened 2019-09-11
Fix handling of list values in FieldBinding

Egorand:


gen-tests.gradle
```diff
@@ -202,7 +202,8 @@ task generateKotlinJavaInteropTests(type: JavaExec) {
       'percents_in_kdoc.proto',
       'same_name_enum.proto',
       'redacted_one_of.proto',
-      'map.proto'
+      'map.proto',
+      'repeated.proto'
```
JakeWharton:
```suggestion
      'repeated.proto',
```
Trailing comma means future additions won't change two lines

wire-tests/src/jvmKotlinInteropTest/kotlin/com/squareup/wire/KotlinRepeatedTest.kt
```diff
@@ -0,0 +1,54 @@
...
+import okio.ByteString.Companion.decodeHex
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+
+class KotlinRepeatedTest {
+  private val adapter = RuntimeMessageAdapter.create(Repeated::class.java)
+
+  @Test fun serialize() {
+    assertEquals(BYTES, ByteString.of(*adapter.encode(THREE)))
```
JakeWharton:
```suggestion
    assertEquals(BYTES, adapter.encode(THREE).toByteString())
```

gen-tests.gradle
Egorand: Yeah, good call.

PR #1189, opened 2019-09-11
Use by lazy to create packed and repeated adapters

Egorand:
Having the fields as vars and lazily assigning doesn't work in
Kotlin Native, cause it's considered a modification of a frozen
object. Lazy delegates are properly implemented to handle this use
case, so we should rely on them.

### DIFF
--- a/wire-runtime/src/nativeMain/kotlin/com/squareup/wire/ProtoAdapter.kt
+++ b/wire-runtime/src/nativeMain/kotlin/com/squareup/wire/ProtoAdapter.kt
@@ -24,8 +24,13 @@ actual abstract class ProtoAdapter<E> actual constructor(
   internal actual val fieldEncoding: FieldEncoding,
   actual val type: KClass<*>?
 ) {
-  internal actual var packedAdapter: ProtoAdapter<List<E>>? = null
-  internal actual var repeatedAdapter: ProtoAdapter<List<E>>? = null
+  // Using by lazy doesn't work on Native since it creates cyclic refs between objects
+  // (commonCreateXxx() is an extension function on ProtoAdapter), so we can't cache adapters here
+  // and will have to recreate them every time.
+  internal actual val packedAdapter: ProtoAdapter<List<E>>
+    get() = commonCreatePacked()
+  internal actual val repeatedAdapter: ProtoAdapter<List<E>>
+    get() = commonCreateRepeated()
 
   /** Returns the redacted form of `value`. */
   actual abstract fun redact(value: E): E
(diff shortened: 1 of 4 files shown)


wire-runtime/src/nativeMain/kotlin/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -24,8 +24,13 @@ actual abstract class ProtoAdapter<E> actual constructor(
   internal actual val fieldEncoding: FieldEncoding,
   actual val type: KClass<*>?
 ) {
-  internal actual var packedAdapter: ProtoAdapter<List<E>>? = null
-  internal actual var repeatedAdapter: ProtoAdapter<List<E>>? = null
+  // Using by lazy doesn't work on Native since it creates cyclic refs between objects
+  // (commonCreateXxx() is an extension function on ProtoAdapter), so we can't cache adapters here
+  // and will have to recreate them every time.
```
jrodbx: is this expensive?

Egorand: Encoding and decoding will both be creating new instances with this approach, plus repeated and packed adapters for common field types will be recreated, so if your proto has 10 `repeated int32` fields there will be 10 allocations during encoding and 10 during decoding. This sucks. Alternatively we can eagerly create those adapters on Native.

Egorand:
Makes sense.

wire-runtime/src/nativeMain/kotlin/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -24,13 +24,8 @@ actual abstract class ProtoAdapter<E> actual constructor(
...
 ) {
-  // Using by lazy doesn't work on Native since it creates cyclic refs between objects
-  // (commonCreateXxx() is an extension function on ProtoAdapter), so we can't cache adapters here
-  // and will have to recreate them every time.
-  internal actual val packedAdapter: ProtoAdapter<List<E>>
-    get() = commonCreatePacked()
-  internal actual val repeatedAdapter: ProtoAdapter<List<E>>
-    get() = commonCreateRepeated()
+  internal actual val packedAdapter: ProtoAdapter<List<E>> by lazy { commonCreatePacked() }
+  internal actual val repeatedAdapter: ProtoAdapter<List<E>> by lazy { commonCreateRepeated() }
```
oldergod: The previous code wasn't guarding against concurrency so could we `lazy(NONE)`?

PR #1197, opened 2019-09-13
object methods still work for our gRPC services

### LINKED ISSUE #1180: gRPC: handle `routeGuideService.toString()` without crashing

oldergod:
fixes #1180

### DIFF
--- a/wire-grpc-client/src/main/java/com/squareup/wire/GrpcClient.kt
+++ b/wire-grpc-client/src/main/java/com/squareup/wire/GrpcClient.kt
@@ -23,6 +23,7 @@ import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
 import okhttp3.OkHttpClient
 import okhttp3.Request
 import okhttp3.RequestBody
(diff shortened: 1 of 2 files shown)


wire-grpc-client/src/main/java/com/squareup/wire/GrpcClient.kt
```diff
@@ -37,11 +38,24 @@ class GrpcClient private constructor(
...
+            if (method.declaringClass == Object::class.java) {
+              return if (args == null) {
+                method.invoke(self)
+              } else {
+                method.invoke(self, *args)
+              }
+            }
+
+            val grpcMethod = methodToService[method] as GrpcMethod<*, *>
+            return grpcMethod.invoke(this@GrpcClient, args ?: emptyArray())
```
JakeWharton: How about moving this `args ?: emptyArray` before the `if` to avoid doing it in both branches?

oldergod: why not, done

PR #1198, opened 2019-09-14
Adds isCancelled, isExecuted, and clone to grpc interfaces

oldergod:


wire-grpc-tests/src/test/java/com/squareup/wire/GrpcClientTest.kt
```diff
@@ -576,6 +585,127 @@ class GrpcClientTest {
     }
   }
 
+  @Test
+  fun grpcCallIsCancelledWhenItShouldBe() {
```
swankjesse: Nit: canceled with one L

wire-grpc-tests/src/test/java/com/squareup/wire/GrpcClientTest.kt
```diff
@@ -576,6 +585,127 @@ class GrpcClientTest {
...
+    assertThat(grpcStreamingCall.isExecuted()).isTrue()
+
+    runBlocking {
+      assertThat(responseChannel.receive()).isEqualTo(RouteNote(message = "welcome"))
+      requestChannel.close()
+      assertThat(responseChannel.receiveOrNull()).isNull()
+    }
+  }
+
+  @Test @Ignore("Timeing out")
```
swankjesse: Nit: timing out

wire-grpc-tests/src/test/java/com/squareup/wire/GrpcClientTest.kt
oldergod: > Both spellings are correct; Americans favor canceled (one L), while cancelled (two Ls) is preferred in British English and other dialects. However, there is only one correct spelling of the word cancellation, no matter where you are.

swankjesse: I wish canceled took 2 Ls and it's was the possessive of it.

PR #1202, opened 2019-09-15
Remove withoutUnknownFields() from Kotlin

Egorand:


swankjesse:
Can we keep it, but implement it without the builder?

swankjesse:
... or do we already get that with `copy(unknownFields = null)` ?

Egorand:
Yeah, it's easy to do with `copy(unknownFields = ByteString.EMPTY)`, so I don't think this method is useful. We can't really implement it without the builder in `Message` cause `copy()` is not a polymorphic method. If you think we should keep it we can make it `open` and generate an override in Kotlin that calls into `copy()`.

PR #1209, opened 2019-09-29
Copy Builder-specific unknown fields tests to jvmKotlinInteropTest

Egorand:


Egorand:
CI failure is gRPC flakiness, safe to merge.

Egorand:
We should keep a battery of tests running against Java code. Theoretically, we can combine Java and Kotlin interop tests into a set of parameterized tests, but I'm not sure it's worth it.

PR #1219, opened 2019-10-02
Always attempt to read all messages even if only one is expected.

swankjesse:
We need to exhaust the inbound stream until it EOFs otherwise when we
close the stream we inadvertently cancel it.

### DIFF
--- a/wire-grpc-client/src/main/java/com/squareup/wire/internal/GrpcMessageSource.kt
+++ b/wire-grpc-client/src/main/java/com/squareup/wire/internal/GrpcMessageSource.kt
@@ -60,4 +60,13 @@ internal class GrpcMessageSource<T : Any>(
 
     return messageAdapter.decode(messageDecoding.decode(encodedMessage).buffer())
   }
+
+  fun readExactlyOneAndClose(): T {
+    use { reader ->
+      val result = reader.read() ?: throw ProtocolException("expected 1 message but got none")
+      val end = reader.read()
+      if (end != null) throw ProtocolException("expected 1 message but got multiple")
+      return result
+    }
+  }
 }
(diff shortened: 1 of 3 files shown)


wire-grpc-tests/src/test/java/com/squareup/wire/GrpcClientTest.kt
```diff
@@ -621,7 +620,7 @@ class GrpcClientTest {
     }
   }
 
-  @Test @Ignore("flaky")
+  @Test
```
jrodbx: 🎉

wire-grpc-client/src/main/java/com/squareup/wire/internal/GrpcMessageSource.kt
```diff
@@ -60,4 +60,13 @@ internal class GrpcMessageSource<T : Any>(
...
   }
+
+  fun readExactlyOneAndClose(): T {
+    use { reader ->
+      val result = reader.read() ?: throw ProtocolException("expected 1 message but got none")
+      val end = reader.read()
+      if (end != null) throw ProtocolException("expected 1 message but got multiple")
+      return result
+    }
+  }
```
jrodbx: nice

wire-grpc-client/src/main/java/com/squareup/wire/internal/RealGrpcCall.kt
```diff
@@ -62,19 +61,13 @@ internal class RealGrpcCall<S : Any, R : Any>(
           var message: R? = null
           var exception: IOException? = null
           try {
-            response.use {
-              response.messageSource(grpcMethod.responseAdapter).use { reader ->
-                message = reader.read()
-                    ?: throw ProtocolException("expected one message in response")
-                exception = response.grpcStatusToException()
```
oldergod: What happens to the trailers? we still have to check, don't we?

swankjesse: We do! I deleted too much! Can fix.

PR #1222, opened 2019-10-03
Small fixes for gRPC doc

oldergod:


docs/wire_grpc.md
```diff
@@ -137,12 +138,14 @@ interface RouteGuideServer : Service {
...
-### Sample
+Sample
+------
 
 Check out our [sample project][sampleProject] demonstrating bidirectional communication between an
 Android application and a server.
 
-### Get it
+Get it
+------
```
jrodbx: This will change from an h3 tag to an h2 tag, making them into their own sections instead of being listed under "gRPC for Servers".  is that what you want?

docs/wire_grpc.md
```diff
@@ -11,7 +11,8 @@ You can configure how your services are generated via the [Gradle plugin][gradle
 
 ```groovy
 wire {
-  kotlin { // Kotlin target will generate code for services.
+  // Kotlin target will generate code for services.
+  kotlin {
```
jrodbx: nice

docs/wire_grpc.md
oldergod: yes

PR #1227, opened 2019-10-07
Use gzip encoding for Grpc's communication

oldergod:
Just replaced all "identity" with "gzip", added the reading header, and tests pass.


swankjesse:
This changes behavior to always compress. I think this is fine for now but in follow up (3.1!) we can compress conditionally based on the content of the message.

PR #1234, opened 2019-10-15
Pre-allocate buffers for toString generation

eeSeeGee:


wire-java-generator/src/test/java/com/squareup/wire/java/JavaGeneratorTest.java
```diff
@@ -444,4 +445,24 @@ public void enclosingTypeIsNotMessage() {
...
+
+  @Test
+  public void toStringGeneration() throws IOException {
+    assertThat(new RepoBuilder()
+        .add("message.proto", ""
+            + "message Message {\n"
+            + "  required float long = 1;\n"
+            + "  required string str = 2;\n"
+            + "}\n")
+        .generateCode("Message")).contains("@Override\n"
```
jrodbx: line break before `.contains`

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -108,6 +109,12 @@
   static final ClassName NULLABLE = ClassName.get("androidx.annotation", "Nullable");
   static final ClassName CREATOR = ClassName.get("android.os", "Parcelable", "Creator");
 
+  static final String REDACTED_TOSTRING_ASSIGN = "\", $N=██\"";
+  static final int REDACTED_TOSTRING_ASSIGN_LENGTH = REDACTED_TOSTRING_ASSIGN.length() - 2;
+
+  static final String TOSTRING_ASSIGN = "\", $N=\"";
+  static final int TOSTRING_ASSIGN_LENGTH = TOSTRING_ASSIGN.length() - 2;
```
jrodbx: inline the escaped quotes back at the use sites, avoiding the need to offset by 2 here?

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -1426,16 +1433,32 @@ private MethodSpec messageHashCode(NameAllocator nameAllocator, MessageType type
     return result.build();
   }
 
-  private MethodSpec messageToString(NameAllocator nameAllocator, MessageType type) {
+  //@VisibleForTesting
```
jrodbx: delete?

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -15,6 +15,7 @@
  */
 package com.squareup.wire.java;
 
+import com.google.common.annotations.VisibleForTesting;
```
jrodbx: delete?

jrodbx:
Need to run `./gradlew generateTests`, commit the results and repush

wire-java-generator/src/test/java/com/squareup/wire/java/JavaGeneratorTest.java
eeSeeGee: Done

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
eeSeeGee: It's not the quotes I'm subtracting, it's the `$N`, which I should perhaps add a comment for.

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
eeSeeGee: Whoops, yeah

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
eeSeeGee: 👍

JakeWharton:
This a fair amount of code. I'm tempted to pre-compute the buffer size as the sum of all names (regardless of whether they're optional) multiplied by 2. That seems like it would prevent approximately 95% of the expansion allocations which seems like a good enough amount given that it only results in a fixed constant flowing into the `StringBuilder` instead of trying to perfectly capture the exact amount.

PR #1237, opened 2019-10-17
Enables preservation of service documentation in generated Kotlin output.

### LINKED ISSUE #1236: Generated code for services is missing proto's comments.

rharter:
Fixes #1236

I checked that Java generator but it looks like it that doesn't support services, so this should be good to go.

I updated all of the tests just to make sure every permutation was accounted for.


oldergod:
Merged into master via 11ff915001e7744cf7f739ac925ba0a9cf8c8742

PR #1238, opened 2019-10-17
Fix the adapter name for services with java packages

### LINKED ISSUE #1235: WireRpc annotation populated with incorrect adapter locations
In the proto file, you may specify the java_package in addition to the package. For example:
```
package Company.Proto.Auth
option java_package com.example.proto.auth
```
When Wire auto-generates the classes, it will put them under the java_package. For example `com.example.proto.auth.LoginRequest#ADAPTER`.

The auto-generated service will instead use the package, and say something like:
```
  @WireRpc(
    path = "/Company.Proto.Auth.Authentication/Login",
    requestAdapter = "Company.Proto.Auth.LoginRequest#ADAPTER",
    responseAdapter = "Company.Proto.Auth.LoginResponse#ADAPTER"
  )
  suspend fun Login(request: LoginRequest): LoginResponse
```

GrpcMethod.toGrpc then does
```
val requestAdapter = ProtoAdapter.get(wireRpc.requestAdapter) as ProtoAdapter<S>
```
and ProtoAdapter does
```
val hash = adapterString.indexOf('#')
val className = adapterString.substring(0, hash)
val fieldName = adapterString.substring(hash + 1)
return Class.forName(className).getField(fieldName).get(null) as ProtoAdapter<Any>
```

This results in `Class.forName("Company.Proto.Auth.LoginRequest#ADAPTER")` when the actual auto-generated class exists as `com.example.proto.auth.LoginRequest#ADAPTER`; causing ClassNotFoundException.

swankjesse:
Closes #1235

### DIFF
--- a/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
+++ b/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
@@ -162,8 +162,8 @@ class KotlinGenerator private constructor(
     val wireRpcAnnotationSpec = AnnotationSpec.builder(WireRpc::class.asClassName())
         .addMember("path = %S", "/$packageName$serviceName/${rpc.name()}")
         // TODO(oldergod|jwilson) Lets' use Profile for this.
-        .addMember("requestAdapter = %S", "$packageName${rpc.requestType().simpleName()}#ADAPTER")
-        .addMember("responseAdapter = %S", "$packageName${rpc.responseType().simpleName()}#ADAPTER")
+        .addMember("requestAdapter = %S", rpc.requestType().adapterString())
+        .addMember("responseAdapter = %S", rpc.responseType().adapterString())
         .build()
     val funSpecBuilder = FunSpec.builder(rpc.name())
         .addModifiers(KModifier.ABSTRACT)
@@ -687,7 +687,7 @@ class KotlinGenerator private constructor(
   }
 
   private fun ProtoType.adapterString() = when {
-    isScalar -> ProtoAdapter::class.qualifiedName + '#' + toString().toUpperCase(Locale.US)
+    isScalar -> ProtoAdapter::class.java.name + '#' + toString().toUpperCase(Locale.US)
     else -> typeName.reflectionName() + "#ADAPTER"
   }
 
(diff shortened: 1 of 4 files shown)


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -162,8 +162,8 @@ class KotlinGenerator private constructor(
     val wireRpcAnnotationSpec = AnnotationSpec.builder(WireRpc::class.asClassName())
         .addMember("path = %S", "/$packageName$serviceName/${rpc.name()}")
         // TODO(oldergod|jwilson) Lets' use Profile for this.
```
oldergod: Can remove the TODO I guess?

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -687,7 +687,7 @@ class KotlinGenerator private constructor(
   }
 
   private fun ProtoType.adapterString() = when {
```
oldergod: Well, that's handy...

wire-kotlin-generator/src/test/java/com/squareup/wire/kotlin/KotlinGeneratorTest.kt
```diff
@@ -276,8 +276,8 @@ class KotlinGeneratorTest {
           |interface RouteGuideBlockingServer : Service {
```
oldergod: Remove the comment line 269

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
swankjesse: Nope, we still wanna hook that up.

wire-kotlin-generator/src/test/java/com/squareup/wire/kotlin/KotlinGeneratorTest.kt
swankjesse: `@WireRpc` _path_ doesn’t use the Java package, but the adapters do!

oldergod: hmmmm

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
oldergod: hmmmm

Issue #1243, opened 2019-10-29
ProtoPruner


Egorand:
- [x] in which case should we keep the `google/protobuf/descriptor` and other around?
- [x] default are pruned out.
- [x] Need to support non-scalar default values
- [x] `google.protobuf.*` pruning doesn't seem to work.
- [x] even though options' type might be pruned, the options itself are not when excluding `google.protobuf.*`. That's because we prune `Options#map` but we print `Options#elements`

oldergod:
- [ ] in which case should we keep the `google/protobuf/descriptor` and other around?

oldergod:
done.

PR #1253, opened 2019-11-11
Update ProtoPruner to includes used extends containing proto files.

oldergod:
- Updated ProtoPruner to also includes proto files which contains used extensions.
- Updated `letter.proto` w/ `options.proto` for more relevant tests.
- Regenerated the tests.

part of #1243


oldergod:
build is green on Travis.

PR #1254, opened 2019-11-11
Retained used imports only

oldergod:
part of #1243 

We don't prune imports and this can create broken proto filed code base. The technique I'm using is that once we retained all types/services/extends (first round), we do another round on each proto files, and check if
　a) the referenced imported proto has any retained types, services, or extends, or
　b) we're extending some google protobuf thing. (_Not super happy with the string comparison but couldn't find better._)

### DIFF
--- a/wire-schema/src/main/java/com/squareup/wire/schema/ProtoFile.java
+++ b/wire-schema/src/main/java/com/squareup/wire/schema/ProtoFile.java
@@ -185,7 +185,7 @@ public final class ProtoFile {
           && importedProtoFile.extendList().isEmpty()) {
 
         // If we extend a google protobuf type, we should keep the import.
-        if (path.startsWith("google/protobuf")) {
+        if (path.startsWith("google/protobuf/")) {
           for (Extend extend : extendList) {
             if (extend.getName().startsWith("google.protobuf")) {
               retainedImportsBuilder.add(path);
--- a/wire-schema/src/main/java/com/squareup/wire/schema/ProtoFile.java
+++ b/wire-schema/src/main/java/com/squareup/wire/schema/ProtoFile.java
@@ -185,7 +185,7 @@ public final class ProtoFile {
           && importedProtoFile.extendList().isEmpty()) {
 
         // If we extend a google protobuf type, we should keep the import.
-        if (path.startsWith("google/protobuf/")) {
+        if (path.equals("google/protobuf/descriptor.proto")) {
           for (Extend extend : extendList) {
             if (extend.getName().startsWith("google.protobuf.")) {
               retainedImportsBuilder.add(path);
--- a/wire-schema/src/main/java/com/squareup/wire/schema/ProtoFile.java
+++ b/wire-schema/src/main/java/com/squareup/wire/schema/ProtoFile.java
@@ -187,7 +187,7 @@ public final class ProtoFile {
         // If we extend a google protobuf type, we should keep the import.
         if (path.startsWith("google/protobuf/")) {
           for (Extend extend : extendList) {
-            if (extend.getName().startsWith("google.protobuf")) {
+            if (extend.getName().startsWith("google.protobuf.")) {
               retainedImportsBuilder.add(path);
               break;
             }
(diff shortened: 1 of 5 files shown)


wire-schema/src/main/java/com/squareup/wire/schema/ProtoFile.java
```diff
@@ -174,6 +174,50 @@ ProtoFile retainAll(Schema schema, MarkSet markSet) {
...
+    for (String path : imports) {
+      ProtoFile importedProtoFile = findProtoFile(retained, path);
+      if (importedProtoFile == null) continue;
+
+      if (importedProtoFile.types().isEmpty()
+          && importedProtoFile.services().isEmpty()
+          && importedProtoFile.extendList().isEmpty()) {
+
+        // If we extend a google protobuf type, we should keep the import.
+        if (path.startsWith("google/protobuf")) {
```
JakeWharton:
```suggestion
        if (path.startsWith("google/protobuf/")) {
```

wire-schema/src/main/java/com/squareup/wire/schema/ProtoFile.java
```diff
@@ -174,6 +174,50 @@ ProtoFile retainAll(Schema schema, MarkSet markSet) {
...
+          for (Extend extend : extendList) {
+            if (extend.getName().startsWith("google.protobuf")) {
```
JakeWharton:
```suggestion
            if (extend.getName().startsWith("google.protobuf.")) {
```

wire-schema/src/main/java/com/squareup/wire/schema/ProtoFile.java
```diff
@@ -174,6 +174,50 @@ ProtoFile retainAll(Schema schema, MarkSet markSet) {
...
+    for (String path : imports) {
+      ProtoFile importedProtoFile = findProtoFile(retained, path);
+      if (importedProtoFile == null) continue;
+
+      if (importedProtoFile.types().isEmpty()
+          && importedProtoFile.services().isEmpty()
+          && importedProtoFile.extendList().isEmpty()) {
+
+        // If we extend a google protobuf type, we should keep the import.
+        if (path.startsWith("google/protobuf/")) {
```
swankjesse: I think you can actually use an exact match: `path.equals("google/protobuf/descriptor.proto")`

wire-schema/src/main/java/com/squareup/wire/schema/Pruner.java
```diff
@@ -46,9 +46,16 @@ public Schema prune() {
...
-    ImmutableList.Builder<ProtoFile> retained = ImmutableList.builder();
+    // We do a first round to detect later unused imports.
+    ImmutableList.Builder<ProtoFile> retainedFirstRoundBuilder = ImmutableList.builder();
     for (ProtoFile protoFile : schema.protoFiles()) {
-      retained.add(protoFile.retainAll(schema, marks));
+      retainedFirstRoundBuilder.add(protoFile.retainAll(schema, marks));
+    }
+    ImmutableList<ProtoFile> firstRound = retainedFirstRoundBuilder.build();
+
+    ImmutableList.Builder<ProtoFile> retained = ImmutableList.builder();
```
swankjesse: Dig

wire-schema/src/main/java/com/squareup/wire/schema/Pruner.java
```diff
@@ -46,9 +46,16 @@ public Schema prune() {
     markRoots();
     markReachable();
 
-    ImmutableList.Builder<ProtoFile> retained = ImmutableList.builder();
+    // We do a first round to detect later unused imports.
+    ImmutableList.Builder<ProtoFile> retainedFirstRoundBuilder = ImmutableList.builder();
```
swankjesse: should each round be its own function?

wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
```diff
@@ -878,4 +878,40 @@ class PrunerTest {
     val enumType = pruned.getType("Enum") as EnumType
     assertThat(enumType.allowAlias()).isTrue()
   }
+
+  @Test
+  fun excludeUnusedImports() {
```
swankjesse: Awesome

PR #1255, opened 2019-11-13
Improve options marking

oldergod:
In the case we have two ProtoMember of same type with different values, say
`ProtoMember(type = "google.protobuf.EnumValueOptions", value = A)` and
`ProtoMember(type = "google.protobuf.EnumValueOptions", value = B)`

Our current logic would add `A`, then say: we already have a type of `EnumValueOptions` so no need to enqueue its value which means we'd be missing `B`.

part of #1243

### DIFF
--- a/wire-schema/src/main/java/com/squareup/wire/schema/MarkSet.java
+++ b/wire-schema/src/main/java/com/squareup/wire/schema/MarkSet.java
@@ -37,7 +37,11 @@ import static com.google.common.base.Preconditions.checkArgument;
  */
 final class MarkSet {
   final IdentifierSet identifierSet;
+
+  /** The types to retain. We may retain a type but not all of its members. */
   final Set<ProtoType> types = new LinkedHashSet<>();
+
+  /** The members to retain. Any member not in here should be pruned! */
   final Multimap<ProtoType, ProtoMember> members = LinkedHashMultimap.create();
 
   MarkSet(IdentifierSet identifierSet) {
@@ -45,30 +49,26 @@ final class MarkSet {
   }
 
   /**
-   * Marks {@code protoMember}, throwing if it is explicitly excluded, or if its enclosing type is
-   * also specifically included. This implicitly excludes other members of the same type.
+   * Marks {@code protoMember}, throwing if it is explicitly excluded. This implicitly excludes
+   * other members of the same type.
    */
   void root(ProtoMember protoMember) {
     if (protoMember == null) throw new NullPointerException("protoMember == null");
     checkArgument(!identifierSet.excludes(protoMember));
-    checkArgument(!types.contains(protoMember.type()));
+    types.add(protoMember.type());
     members.put(protoMember.type(), protoMember);
   }
 
-  /**
-   * Marks {@code type}, throwing if it is explicitly excluded, or if any of its members are also
-   * specifically included.
-   */
+  /** Marks {@code type}, throwing if it is explicitly excluded. */
   void root(ProtoType type) {
     if (type == null) throw new NullPointerException("type == null");
     checkArgument(!identifierSet.excludes(type));
-    checkArgument(!members.containsKey(type));
     types.add(type);
   }
 
   /**
    * Marks a type as transitively reachable by the includes set. Returns true if the mark is new,
-   * the type will be retained, and its own dependencies should be traversed.
+   * the type will be retained, and reachable objects should be traversed.
    */
   boolean mark(ProtoType type) {
     if (type == null) throw new NullPointerException("type == null");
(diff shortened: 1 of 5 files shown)


wire-schema/src/main/java/com/squareup/wire/schema/MarkSet.java
```diff
@@ -83,9 +83,17 @@ boolean mark(ProtoType type) {
...
     if (protoMember == null) throw new NullPointerException("type == null");
     if (identifierSet.excludes(protoMember)) return false;
-    return members.containsKey(protoMember.type())
+    boolean result = members.containsKey(protoMember.type())
         ? members.put(protoMember.type(), protoMember)
         : types.add(protoMember.type());
+
+    // An option can hide another one.
+    String enclosingTypeOrPackage = protoMember.type().enclosingTypeOrPackage();
+    if (enclosingTypeOrPackage != null && enclosingTypeOrPackage.equals("google.protobuf")) {
```
JakeWharton:
```suggestion
    if ("google.protobuf".equals(enclosingTypeOrPackage)) {
```
Not sure how you feel about "Yoda" conditionals...

Pretty sure @swankjesse doesn't like them...

swankjesse: We should just convert to Kotlin then who cares

wire-schema/src/main/java/com/squareup/wire/schema/MarkSet.java
```diff
@@ -83,9 +83,17 @@ boolean mark(ProtoType type) {
   boolean mark(ProtoMember protoMember) {
     if (protoMember == null) throw new NullPointerException("type == null");
     if (identifierSet.excludes(protoMember)) return false;
-    return members.containsKey(protoMember.type())
+    boolean result = members.containsKey(protoMember.type())
```
swankjesse: this should probably be something other than `result` because it isn’t. Maybe `newMark` ?

wire-schema/src/main/java/com/squareup/wire/schema/MarkSet.java
```diff
@@ -83,9 +83,17 @@ boolean mark(ProtoType type) {
...
         ? members.put(protoMember.type(), protoMember)
         : types.add(protoMember.type());
+
+    // An option can hide another one.
```
swankjesse: This is no longer consistent with the declaration doc

swankjesse: ... I think this might trigger an infinite loop?

wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
```diff
@@ -914,4 +914,39 @@ class PrunerTest {
...
+             |}
+             |
+             """.trimMargin()
+        )
+        .schema()
+    val pruned = schema.prune(IdentifierSet.Builder()
+        .include("CurrencyCode")
+        .build())
+
+    assertThat(pruned.getType("RoundingMode")).isNotNull
```
swankjesse: isNotNull() please

wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
```diff
@@ -878,4 +880,75 @@ class PrunerTest {
...
+             |
+             |extend google.protobuf.EnumValueOptions {
+             |  optional int32 cash_rounding = 54000;
+             |  optional RoundingMode rounding_mode = 54002;
+             |}
+             |
+             |enum CurrencyCode {
+             |  AFN = 971;
+             |  ANG = 532;
+             |  NZD = 554 [(cash_rounding) = 10, (rounding_mode) = DOWN_ON_HALF];
```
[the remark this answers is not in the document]
swankjesse: Regardless these should be independent thingies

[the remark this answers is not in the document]
oldergod: So what happens is
`NZD` is marked, and we mark all options `EnumValueOptions#cash_rounding` and `EnumValueOptions#rounding_mode`. That's all. `RoundingMode` is marked later on when we traverse the option's member.

For the options, the proto members are both
`ProtoMember(type = "EnumValueOptions", member = "cash_rounding")` and
`ProtoMember(type = "EnumValueOptions", member = "rounding_mode")`.

Now when the code does
```java
    if ("google.protobuf".equals(protoMember.type().enclosingTypeOrPackage())) {
      return members.add(protoMember.member());
    } else {
      return isNewMark;
    }
```
we wanna traverse `member` only if new. If there was two options of the same type like
`ProtoMember(type = "EnumValueOptions", member = "rounding_mode")`; we wouldn't want to traverse `rounding_mode` twice. That's why we are tracking members.

wire-schema/src/main/java/com/squareup/wire/schema/MarkSet.java
```diff
@@ -38,7 +38,8 @@
 final class MarkSet {
   final IdentifierSet identifierSet;
   final Set<ProtoType> types = new LinkedHashSet<>();
-  final Multimap<ProtoType, ProtoMember> members = LinkedHashMultimap.create();
+  final Multimap<ProtoType, ProtoMember> protoMembers = LinkedHashMultimap.create();
+  private final Set<String> members = new LinkedHashSet<>();
```
swankjesse: I don’t think this is a sufficiently narrow namespace? What are the entries in this set?

oldergod: I agree, I thought about that too this morning

wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
```diff
@@ -13,6 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:Suppress("UsePropertyAccessSyntax")
```
swankjesse: thank you

PR #1257, opened 2019-11-14
ProtoPruner to use NewSchemaLoader

oldergod:


proto-pruner/src/main/java/com/squareup/wire/prune/ProtoPruner.kt
```diff
@@ -102,7 +106,7 @@ class ProtoPruner(
         throw IllegalArgumentException("No output path specified")
       }
 
-      ProtoPruner(inPaths, outPath, identifierSetBuilder.build()).run()
+      ProtoPruner(FileSystems.getDefault(), inPaths, outPath, identifierSetBuilder.build()).run()
```
swankjesse: fs

oldergod: What does that mean?

PR #1258, opened 2019-11-15
Print default values for scalar types for proto target within the options 

oldergod:
Because `default`s aren't options., we strip them on [parsing](https://github.com/square/wire/blob/5d4ea6f20042ce829bc258cebefc506dbe0a1bf9/wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ProtoParser.java#L362-L376).
Now, we wanna print them back when emitting protos, that is what this PR is doing.

One trouble is that the current implementation doesn't respect the order, when printing protos, the default value will always be added last. I don't think that's a blocker anyway.

part of #1243


oldergod:
Maybe I could change the parsing and actually store the default value, not as a string, but as an option?

PR #1252, opened 2019-11-09
.proto enum variable named the same as enumerator causes compilation …

mellster2012:
…error #299

Issue #299, opened 2015-08-06
.proto enum variable named the same as enumerator causes compilation error


maregrabik:
We got proto file:

```
enum Gender {
   Gender_Male = 0;
   Gender_Female = 1;
}
message SanctionPerson {
   optional string FirstName = 1;
   optional Gender Gender = 2;
}

After compiling to java:
<b>Gender.java</b>

public enum <b>Gender</b>
    implements ProtoEnum {
  Gender_Male(0),
  Gender_Female(1);

  private final int value;

  Gender(int value) {
    this.value = value;
  }

  public int getValue() {
    return value;
  }
```

<b>SanctionPerson.java:</b>

```
public final class SanctionPerson extends Message {
  private static final long serialVersionUID = 0L;

  public static final String DEFAULT_FIRSTNAME = "";
 <b> public static final Gender DEFAULT_GENDER = Gender.Gender_Male;// Compile error Cannot make a static reference to the non-static field Gender</b>

  @ProtoField(tag = 1, type = STRING)
  public final String FirstName;

  @ProtoField(tag = 2, type = ENUM)
  public final <b>Gender Gender; //Field got the same name</b>

  public SanctionPerson(String FirstName, Gender Gender) {
    this.FirstName = FirstName;
    this.Gender = Gender;
  }
```

The problem is that 
 public static final Gender DEFAULT_GENDER = <b>Gender.Gender_Male</b> 
is referring to local variable instead Gender enum in Gender.java.

Or is there any possibility to force compiler to use for example camelCase for variables in .proto like 
<b>optional Gender Gender = 2;</b>?
Error from console: "non-static variable Gender cannot be referenced from a static context" in compile time.

JakeWharton:
This naming convention violates the proto style guide: https://developers.google.com/protocol-buffers/docs/style. You are able to rename the field without affecting any clients, as an easy way to quickly eliminate the problem.

maregrabik:
Thanks for quick response!
Unfortunately i cant change anything in base .proto , i have to work on given files. I am aware of that it`s braking conventions, but still prottobuff with it`s own compiler transform it to non error java class. Hoped to switch for wire`s one, but that situation block me out.

mellster2012:
Facing the same issue. Wonder how hard a PR would be, doesn't seem like an overly complex problem to fix but it would help greatly for those that don't have control over the proto files given to them.

mellster2012:
[...]

PR #1260, opened 2019-11-20
.proto enum variable named the same as enumerator causes compilation error #299 (Kotlin version for Java pull request #1252)

mellster2012:


oldergod:
You probably created this PR by mistake so I'm closing it.

mellster2012:
@oldergod The Kotlin generator has the same issue when producing Kotlin sources (they don't compile if type name is the same as variable name), I don't know if you already incorporated a fix for Kotlin, that's why I provided it.

oldergod:
Oops !

oldergod:
Then please, fix the conflict.

mellster2012:
@oldergod see https://github.com/square/wire/pull/1276

PR #1269, opened 2019-11-24
Prepare to lazy-load and lazy-evaluate the proto path

swankjesse:
We currently aggressively load and validate everything on the proto path.
Making it load dependencies as necessary should make builds faster and
simplify configuration for sprawling declarations.

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
+++ b/wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
@@ -147,11 +147,20 @@ data class WireRun(
     val schemaLoader = NewSchemaLoader(fs, sourcePath, protoPath)
     val protoFiles = schemaLoader.load()
 
-    // 2. Load descriptor proto
-    val protoFilesPlusDescriptor = protoFiles.plus(schemaLoader.loadDescriptorProto())
+    // 2. Split source files (completely linked) from path files (linked as necessary).
+    // TODO(jwilson): replace with an interface that loads path files on-demand.
+    val sourceFiles = mutableListOf<ProtoFile>()
+    val pathFiles = mutableListOf<ProtoFile>()
+    for (protoFile in protoFiles) {
+      if (schemaLoader.sourceLocationPaths.contains(protoFile.location().path)) {
+        sourceFiles += protoFile
+      } else {
+        pathFiles += protoFile
+      }
+    }
 
     // 3. Validate the schema and resolve references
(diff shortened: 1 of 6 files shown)


wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
```diff
@@ -147,11 +147,20 @@ data class WireRun(
     val schemaLoader = NewSchemaLoader(fs, sourcePath, protoPath)
     val protoFiles = schemaLoader.load()
 
-    // 2. Load descriptor proto
-    val protoFilesPlusDescriptor = protoFiles.plus(schemaLoader.loadDescriptorProto())
```
swankjesse: This was unnecessary since descriptor is now loaded in step 1

wire-schema/src/main/java/com/squareup/wire/schema/FileLinker.java
```diff
@@ -0,0 +1,113 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire.schema;
+
+import java.util.Collection;
+
+final class FileLinker {
```
swankjesse: For follow-up, FileLinker will have phases like `IMPORTS/TYPES/MEMBERS/OPTIONS/VALIDATED` and we’ll only link to the level necessary.

sourcePath will be VALIDATED
direct dependencies will be TYPES
transitive dependencies will be nothing, not even loaded

There’s of course an annoying special case. When we have options we will need to link direct dependencies to the MEMBERS phase, and possibly transitive dependencies also.

wire-schema/src/main/java/com/squareup/wire/schema/Linker.java
```diff
@@ -28,117 +28,79 @@
...
+          new FileLinker(protoFile, withContext(protoFile)));
     }
 
-    // Link extensions. This depends on type registration.
-    for (ProtoFile protoFile : protoFiles) {
-      Linker linker = withContext(protoFile);
-      for (Extend extend : protoFile.extendList()) {
-        extend.link(linker);
-      }
+    for (FileLinker fileLinker : fileLinkers.values()) {
```
swankjesse: Unlikely-to-be-observed behavior change here: I’m doing imports early now. Previously we did ’em late. It’s necessary to do ’em early if we’re going to defer loading everything.

PR #1272, opened 2019-11-25
Lazily load from the proto path.

swankjesse:
Previously the proto path and source path were symmetric: all types and members and
options were always loaded, parsed, and linked.

With this change we lazily load, parse, and link from the proto path.

The biggest fallout of this change is that some things are newly null when
pruning including the options map and field types. This should be cleaned up
in a follow-up PR, but this PR is big enough as-is.

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/schema/NewSchemaLoader.kt
+++ b/wire-compiler/src/main/java/com/squareup/wire/schema/NewSchemaLoader.kt
@@ -16,118 +16,90 @@
 package com.squareup.wire.schema
 
 import com.google.common.io.Closer
-import com.squareup.wire.schema.internal.parser.ProtoParser
-import okio.buffer
-import okio.source
+import com.squareup.wire.schema.internal.parser.ProtoFileElement
 import java.io.Closeable
 import java.io.IOException
 import java.nio.file.FileSystem
-import java.util.ArrayDeque
+import java.util.TreeSet
 
 /**
  * Load proto files and their transitive dependencies and parse them. Keep track of which files were
  * loaded from where so that we can use that information later when deciding what to generate.
  */
 class NewSchemaLoader(
-  private val fs: FileSystem,
-
-  /** See [com.squareup.wire.WireRun.sourcePath]. */
-  private val sourcePath: List<Location>,
-
-  /** See [com.squareup.wire.WireRun.protoPath]. */
-  private val protoPath: List<Location> = listOf()
-) : Closeable {
+  private val fs: FileSystem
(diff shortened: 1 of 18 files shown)


wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt
```diff
@@ -400,7 +400,6 @@ class WireRunTest {
    * [SchemaException]. But we no longer do this to make Wire both faster and to eliminate the need
    * to place all transitive dependencies in the proto path.
    */
-  @Ignore("WireRun currently evaluates path files entirely")
```
swankjesse: This is the payoff!

wire-compiler/src/main/java/com/squareup/wire/schema/NewSchemaLoader.kt
```diff
@@ -16,118 +16,89 @@
...
+  ) {
+    sourcePathRoots = allRoots(closer, sourcePath)
+    protoPathRoots = allRoots(closer, protoPath)
+  }
 
-    loaded[DESCRIPTOR_PROTO] = loadDescriptorProto()
+  /** Returns the files in the source path. */
+  @Throws(IOException::class)
+  fun loadSourcePathFiles(): List<ProtoFile> {
+    check(::sourcePathRoots.isInitialized && ::protoPathRoots.isInitialized) {
```
Egorand: I recommend going with nullables, it's easy to forget placing an `isInitialized` check.

wire-compiler/src/main/java/com/squareup/wire/schema/NewSchemaLoader.kt
```diff
@@ -158,12 +128,27 @@ class NewSchemaLoader(
...
+
+  private fun checkForErrors() {
+    require(errors.isEmpty()) { errors.joinToString(separator = "\n") }
+  }
+
   override fun close() {
     return closer.close()
   }
 
   companion object {
```
Egorand: nit: Delete empty companion?

wire-compiler/src/main/java/com/squareup/wire/schema/NewSchemaLoader.kt
swankjesse: SGTM! Fixed.

wire-compiler/src/main/java/com/squareup/wire/schema/NewSchemaLoader.kt
swankjesse: Fixed!

PR #1273, opened 2019-11-25
Fix logictypo around Pruner's reachable objects

oldergod:


wire-schema/src/main/java/com/squareup/wire/schema/Pruner.java
```diff
@@ -231,7 +231,7 @@ private void markReachable() {
 
     for (ProtoMember member : options.fields().values()) {
       // If it's an extension, don't consider the entire enclosing type to be reachable.
-      if (isExtensionField(member)) {
+      if (!isExtensionField(member)) {
```
swankjesse: Tests didn’t catch this?

oldergod: The content's logic of the method has been inverted as well. This isn't really a behavior change, just testing what the method said it was.

wire-schema/src/main/java/com/squareup/wire/schema/Pruner.java
```diff
@@ -241,6 +241,7 @@ private void markReachable() {
 
   private boolean isExtensionField(ProtoMember protoMember) {
     Type type = schema.getType(protoMember.type());
-    return type instanceof MessageType && ((MessageType) type).field(protoMember.member()) != null;
+    return type instanceof MessageType
+        && ((MessageType) type).extensionField(protoMember.member()) != null;
```
oldergod: here, I changed to `extensionField()` instead of `field()`

PR #1274, opened 2019-11-25
Excludes properly prunes options

oldergod:
part of #1243

### DIFF
--- a/wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
+++ b/wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
@@ -1060,7 +1060,7 @@ class PrunerTest {
         .include("CurrencyCode")
         .build())
 
-    assertThat(pruned.getType("RoundingMode")).isNotNull() // TODO
+    assertThat(pruned.getType("RoundingMode")).isNotNull()
   }
 
   /**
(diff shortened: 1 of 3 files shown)


wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
```diff
@@ -1183,4 +1183,109 @@ class PrunerTest {
...
+        .schema()
+    val pruned = schema.prune(IdentifierSet.Builder()
+        .include("squareup.Author")
+        .exclude("google.protobuf.*")
+        .build())
+
+    assertThat(pruned.getType("squareup.Style")).isNull()
+    assertThat(pruned.getType("squareup.MessageOption")).isNull()
+    val authorType = pruned.getType("squareup.Author") as EnumType
+    assertThat(authorType.constant("ZEUS")!!.options.fields(null).size()).isZero()
```
swankjesse: the null parameter here isn’t a beautiful API ... maybe provide a default value?

wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
```diff
@@ -1183,4 +1183,109 @@ class PrunerTest {
...
+    val pruned = schema.prune(IdentifierSet.Builder()
+        .include("squareup.Letter")
+        .exclude("google.protobuf.*")
+        .build())
+
+    assertThat(pruned.getType("squareup.Style")).isNull()
+    assertThat(pruned.getType("squareup.MessageOption")).isNull()
+
+    val letterType = pruned.getType("squareup.Letter") as MessageType
+    assertThat(letterType.field("header").options().fields(null).size()).isZero()
```
swankjesse: nit: rather than this:
```
assertThat(someCollection.size()).isZero()
```
do this?
```
assertThat(someCollection).isEmpty()
```
That way if it fails the error message tells you which elements weren’t expected

PR #1275, opened 2019-11-26
Convert IdentifierSet to Kotlin

swankjesse:


swankjesse:
This PR is split into separate rename and conversion steps. Reviewing each commit independently should be much easier.

PR #1276, opened 2019-11-26
.proto enum variable named the same as enumerator causes compilation error #299 (Kotlin version for Java pull request #1252)

mellster2012:


mellster2012:
@oldergod Clean PR for the Kotlin generator.

PR #1277, opened 2019-11-28
Support custom targets in WireRun and the Gradle plugin

lkerford:


jrodbx:
I'm likely missing some context, but why do we want this?

swankjesse:
I wanna replace our internal RPC service generators with the Wire gradle plugin. But I don’t want a hard dependency from Wire on Square’s internal RPC stuff.

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireOutput.kt
```diff
@@ -41,11 +42,12 @@ abstract class WireOutput {
...
 
   /** Configure the project's graph to compile the files emitted. */
-  abstract fun applyToProject(
+  open fun applyToProject(
     project: Project,
     wireTask: TaskProvider<WireTask>?,
     kotlin: Boolean
-  )
+  ) {
+  }
```
oldergod: I wouldn't mind keep that as abstract, and let the implementer no-op it.

oldergod:
CI is sad because it cannot find `Path#writeString`. Are we using a different/wrong java version on CI?

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireOutput.kt
lkerford: fixed

lkerford:
> CI is sad because it cannot find `Path#writeString`. Are we using a different/wrong java version on CI?

has been addressed

PR #1278, opened 2019-11-28
Fix tag in test proto

oldergod:


wire-tests/src/jvmJavaTest/proto-java/com/squareup/wire/protos/custom_options/FooBar.java
```diff
@@ -46,8 +46,8 @@
       .build();
 
   public static final FieldOptions FIELD_OPTIONS_QUX = new FieldOptions.Builder()
-      .my_field_option_one(18)
       .my_field_option_two(34.5f)
+      .my_field_option_one(18)
```
jrodbx: nit: odd ordering

wire-tests/src/jvmJavaTest/proto-java/com/squareup/wire/protos/custom_options/FooBar.java
```diff
@@ -563,15 +563,15 @@ public More redact(More value) {
   }
 
   public enum FooBarBazEnum implements WireEnum {
-    FOO(1, 17, new More.Builder()
+    FOO(1, new More.Builder()
```
jrodbx: why does the argument ordering change?

oldergod: Let me check, I probably mixed up some unrelated code.

oldergod:
Thanks @jrodbx it should be good now

oldergod:
That's a really good question Egor ! And it seems we can when we are using `allow_alias` which we do here.

from https://developers.google.com/protocol-buffers/docs/proto#enum
```
You can define aliases by assigning the same value to different enum constants. To do this you need to set the allow_alias option to true, otherwise protocol compiler will generate an error message when aliases are found.

enum EnumAllowingAlias {
  option allow_alias = true;
  UNKNOWN = 0;
  STARTED = 1;
  RUNNING = 1;
}
enum EnumNotAllowingAlias {
  UNKNOWN = 0;
  STARTED = 1;
  // RUNNING = 1;  // Uncommenting this line will cause a compile error inside Google and a warning message outside.
}
```

I confirmed that by removing the `allow_alias`, this would generate `SchemaException: multiple enum constants share tag 3`. Cool.

Egorand:
Nice!

PR #1279, opened 2019-11-28
Keep ServiceOptions if used when pruning

oldergod:
part of #1243 

wire-schema/src/main/java/com/squareup/wire/schema/Field.java is where the logic is.

### DIFF
--- a/proto-pruner/src/test/expected/options.proto
+++ b/proto-pruner/src/test/expected/options.proto
@@ -13,3 +13,9 @@ extend google.protobuf.MessageOptions {
 extend google.protobuf.EnumValueOptions {
   optional int32 text_alignment = 54000;
 }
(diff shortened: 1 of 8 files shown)


proto-pruner/src/test/expected/options.proto
```diff
@@ -13,3 +13,9 @@ extend google.protobuf.MessageOptions {
 extend google.protobuf.EnumValueOptions {
   optional int32 text_alignment = 54000;
 }
+extend google.protobuf.ServiceOptions {
+  optional bool httpOneOnly = 56000;
+}
+extend google.protobuf.MethodOptions {
+  optional bool hide = 56001;
```
jrodbx: any special reason behind 56000?

oldergod: nope

PR #1280, opened 2019-11-29
Add a failing test case for linker pruning

swankjesse:


wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt
```diff
@@ -20,6 +20,9 @@ import com.google.common.jimfs.Jimfs
 import com.squareup.wire.StringWireLogger
 import com.squareup.wire.kotlin.RpcCallStyle
 import com.squareup.wire.kotlin.RpcRole
+import com.squareup.wire.testing.add
```
lkerford: I don't think these imports are required

wire-test-utils/src/main/java/com/squareup/wire/testing/files.kt
```diff
@@ -0,0 +1,78 @@
...
+import java.nio.file.FileVisitResult
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.SimpleFileVisitor
+import java.nio.file.attribute.BasicFileAttributes
+import java.util.zip.ZipEntry
+import java.util.zip.ZipOutputStream
+
+fun FileSystem.add(pathString: String, contents: String) {
+  val path = getPath(pathString)
```
lkerford: Should we be checking is 'pathString' is empty? Should we be handling invalid paths?

wire-test-utils/src/main/java/com/squareup/wire/testing/files.kt
```diff
@@ -0,0 +1,78 @@
...
+  if (path.parent != null) {
+    Files.createDirectories(path.parent)
+  }
+  Files.write(path, contents.toByteArray(Charsets.UTF_8))
+}
+
+fun FileSystem.symlink(linkPathString: String, targetPathString: String) {
+  val linkPath = getPath(linkPathString)
```
lkerford: Should we be checking is 'pathString' is empty? Should we be handling invalid paths?

wire-test-utils/src/main/java/com/squareup/wire/testing/files.kt
```diff
@@ -0,0 +1,78 @@
...
+        result.add(path.toString())
+      }
+      return FileVisitResult.CONTINUE
+    }
+  })
+  return result
+}
+
+fun FileSystem.addZip(pathString: String, vararg contents: Pair<String, String>) {
+  val path = getPath(pathString)
```
lkerford: Should we be checking is 'pathString' is empty? Should we be handling invalid paths?

wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt
swankjesse: I moved files out of this package, so they used to not need imports and now they do!

wire-test-utils/src/main/java/com/squareup/wire/testing/files.kt
swankjesse: nah, for test code like this if it fails it fails

wire-test-utils/src/main/java/com/squareup/wire/testing/files.kt
swankjesse: I think it’s mostly use-case dependent. This isn’t a public API so we don’t anticipate invalid inputs.

wire-test-utils/src/main/java/com/squareup/wire/testing/files.kt
swankjesse: same.

PR #1283, opened 2019-11-29
New mechanism to retain only linked things in the schema

lkerford:
When we load a schema with a separate source path and proto path, the returned
schema only contains the bits of the proto path that are used.


wire-compiler/src/test/java/com/squareup/wire/schema/OptionsLinkingTest.kt
```diff
@@ -0,0 +1,161 @@
...
+        mapOf(
+            formattingOptionsField to mapOf(
+                languageField to "English"
+            )
+        )
+    )
+  }
+
+  @Test
+  fun fieldsOfExtensions() {
```
lkerford: this was a crashing bug before this PR. We weren't retaining enough stuff and that caused a crash when resolving the name field of language.

PR #1285, opened 2019-11-30
Convert Field.java to Kotlin

swankjesse:


wire-codegen-sample/src/main/java/com/squareup/wire/sample/ServiceGenerator.java
```diff
@@ -45,19 +45,19 @@ public TypeSpec api(Service service) {
     }
 
     for (Rpc rpc : service.rpcs()) {
-      ProtoType requestType = rpc.requestType();
+      ProtoType requestType = rpc.getRequestType();
```
swankjesse: A bunch of these accessors should disappear in follow-up when I convert these types to Kotlin.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/FieldExtensions.kt
```diff
@@ -23,10 +23,10 @@ import com.squareup.wire.schema.Field
 import com.squareup.wire.schema.ProtoType
 
 internal val Field.isMap: Boolean
-  get() = type().isMap
+  get() = type!!.isMap
```
swankjesse: ... might change this to be lateinit later. It never escapes the Linker without being non-null.

wire-schema/src/main/java/com/squareup/wire/schema/ProtoType.kt
```diff
@@ -0,0 +1,155 @@
...
+ * fully-qualified name using the protocol buffer package.
+ */
+class ProtoType {
+  val isScalar: Boolean
+
+  private val string: String
+
+  val isMap: Boolean
+
+  /** The type of the map's keys. Only present when [.isMap] is true.  */
```
Egorand: nit: `[isMap]`

wire-schema/src/main/java/com/squareup/wire/schema/ProtoType.kt
```diff
@@ -0,0 +1,155 @@
...
+  val keyType: ProtoType?
+
+  /** The type of the map's values. Only present when [.isMap] is true.  */
```
Egorand: nit: `[isMap]`

wire-schema/src/main/java/com/squareup/wire/schema/ProtoType.kt
```diff
@@ -0,0 +1,155 @@
...
+  val valueType: ProtoType?
+
+  val simpleName: String
+    get() {
```
Egorand: Why not compute once and assign?

wire-schema/src/main/java/com/squareup/wire/schema/ProtoType.kt
```diff
@@ -0,0 +1,155 @@
...
+      val dot = string.lastIndexOf('.')
+      return string.substring(dot + 1)
+    }
+
+  /** Returns the enclosing type, or null if this type is not nested in another type.  */
+  val enclosingTypeOrPackage: String?
+    get() {
```
Egorand: Same

wire-schema/src/main/java/com/squareup/wire/schema/ProtoType.kt
```diff
@@ -0,0 +1,155 @@
...
+      val dot = string.lastIndexOf('.')
+      return if (dot == -1) null else string.substring(0, dot)
+    }
+
+  /** Creates a scalar or message type.  */
+  private constructor(isScalar: Boolean, string: String) {
+    checkNotNull(string, "string == null")
```
Egorand: Null check is probably extraneous.

wire-schema/src/main/java/com/squareup/wire/schema/Field.kt
```diff
@@ -0,0 +1,261 @@
...
+  val isRequired: Boolean
+    get() = label == Label.REQUIRED
+
+  /**
+   * Returns this field's name, prefixed with its package name. Uniquely identifies extension
+   * fields, such as in options.
+   */
+  val qualifiedName: String
+    get() {
+      return when {
```
Egorand: Compute once and assign? Same for fields below.

wire-schema/src/main/java/com/squareup/wire/schema/ProtoType.kt
swankjesse: Fixed!

wire-schema/src/main/java/com/squareup/wire/schema/ProtoType.kt
swankjesse: Fixed!

wire-schema/src/main/java/com/squareup/wire/schema/ProtoType.kt
swankjesse: To save allocations. This isn’t usually needed.

wire-schema/src/main/java/com/squareup/wire/schema/ProtoType.kt
swankjesse: I’d want to make that kind of change independent of the language conversion.

wire-schema/src/main/java/com/squareup/wire/schema/ProtoType.kt
swankjesse: Fixed. Good catch.

wire-schema/src/main/java/com/squareup/wire/schema/Field.kt
swankjesse: I think it reads better as computed-on-demand state. Lets discuss offline; I’m don’t hold strong feelings here.

PR #1291, opened 2019-12-03
Service and ProtoFile migration to Kotlin

oldergod:


wire-schema/src/main/java/com/squareup/wire/schema/ProtoFile.kt
```diff
@@ -0,0 +1,211 @@
...
+  private val packageName: String?,
+  private val types: List<Type>,
+  private val services: List<Service>,
+  private val extendList: List<Extend>,
+  private val options: Options,
+  private val syntax: Syntax?
+) {
+  private var javaPackage: Any? = null
+
+  // TODO(Benoit) Set all `val` as public and remove the redundant getters.
```
oldergod: This would have created way too much noise so decided to do it separately.

wire-schema/src/main/java/com/squareup/wire/schema/ProtoFile.java
```diff
@@ -13,64 +13,40 @@
...
+  private val packageName: String?,
+  private val types: List<Type>,
+  private val services: List<Service>,
+  private val extendList: List<Extend>,
+  private val options: Options,
+  private val syntax: Syntax?
+) {
+  private var javaPackage: Any? = null
+
+  // TODO(Benoit) Set all `val` as public and remove the redundant getters.
```
swankjesse: what worked for me was renaming the accessors to use a `get` prefix, _then_ doing the Convert to Kotlin shortcut

oldergod: Yes, I wanna do that. It's just that more than 15 files get updated. I thought that deserved its own PR

PR #1293, opened 2019-12-03
Konvert Util

jrodbx:


wire-schema/src/main/java/com/squareup/wire/schema/internal/Util.java
```diff
@@ -13,61 +13,64 @@
...
-  public static <T> ImmutableList<T> concatenate(List<T> a, T b) {
-    return ImmutableList.<T>builder()
-        .addAll(a)
-        .add(b)
-        .build();
-  }
+  /** True if the supplied value is in the valid tag range and not reserved.  */
+  fun isValidTag(value: Int) =
+    value in MIN_TAG_VALUE until RESERVED_TAG_VALUE_START ||
+        value in (RESERVED_TAG_VALUE_END + 1)..MAX_TAG_VALUE
```
swankjesse: off topic: we assign `MAX_TAG_VALUE` with a `- 1` above and then use it with a `+ 1` here

swankjesse: would rename to say, `UNRESERVED_TAG_VALUE_START` ?!

jrodbx: Updated to `until +1`.

The unreserved range is non-contiguous (given the intermediate reserved range), so indicating a START seems confusing to me.

jrodbx:
Hmm, tests failing due to a behavior change from Java String.split to Kotlin String.split.

PR #1295, opened 2019-12-03
Migrate some Guava collections to Kotlin collections

jrodbx:


wire-schema/src/main/java/com/squareup/wire/schema/Field.kt
```diff
@@ -214,67 +213,46 @@ class Field private constructor(
...
-      for (field in fields) {
-        // If the type is non-null, then the field has been linked.
-        if (field.type != null) {
-          result.add(field.withOptions(field.options.retainLinked()))
-        }
-      }
-      return result.build()
-    }
+    fun retainLinked(fields: List<Field>) = fields
+        .filter { it.type != null } // If the type is non-null, then the field has been linked.
```
swankjesse: awesome

PR #1297, opened 2019-12-04
No more Guava in Kotlin code

jrodbx:
Next up: Konversion


swankjesse:
I’d like to consider keeping Guava just for Multimap. You can kind of fake it with `getAndPut()` but it’s clumsy, and especially clumsy to do things like deep immutability on ’em.

jrodbx:
I see.  I thought the goal was to rid the dependency.  Should we just copy that code into this project?  On the other hand, if MultiMap usage is limited to what's changed here, maybe it's OK to migrate to getAndPut()?

swankjesse:
I don’t feel too strongly. Lets proceed with what you have here.

But if later we need a multimap, I think it’s probably worthwhile to bring it back! The dependency doesn’t hurt.

wire-schema/src/main/java/com/squareup/wire/schema/Options.kt
```diff
@@ -171,37 +160,40 @@ class Options(
...
+      is List<*> -> a + b as List<*>
 
       is Map<*, *> -> {
         @Suppress("UNCHECKED_CAST") // All maps have this type.
         union(linker, a as Map<ProtoMember, Any?>, b as Map<ProtoMember, Any>)
       }
 
-      else -> {
-        linker.addError("conflicting options: %s, %s", a, b)
-        a // Just return any placeholder.
```
oldergod: Not needed?

PR #1303, opened 2019-12-05
Prune options properly instead of retaining them all

oldergod:
Part of #1243 Let's go !

### DIFF
--- a/proto-pruner/src/test/java/com/squareup/wire/prune/ProtoPrunerTest.kt
+++ b/proto-pruner/src/test/java/com/squareup/wire/prune/ProtoPrunerTest.kt
@@ -19,7 +19,6 @@ import com.squareup.wire.schema.IdentifierSet
 import com.squareup.wire.schema.Location
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Before
-import org.junit.Ignore
 import org.junit.Rule
 import org.junit.Test
 import org.junit.rules.TemporaryFolder
@@ -87,7 +86,7 @@ class ProtoPrunerTest {
     assertOutputs(outputs)
   }
 
-  @Test @Ignore("Options are not properly pruned yet. See #1243.")
+  @Test
   fun testOptionsExcludingProtobuf() {
     val identifierSet = IdentifierSet.Builder()
         .include("squareup.options.poem.Poem")
@@ -97,7 +96,8 @@ class ProtoPrunerTest {
     invokeProtoPruner(identifierSet)
 
     val outputs = arrayOf(
-        "poem.proto"
+        "poem.proto",
+        "options.proto"
     )
     assertOutputs(outputs)
   }
--- a/wire-schema/src/main/java/com/squareup/wire/schema/Options.kt
+++ b/wire-schema/src/main/java/com/squareup/wire/schema/Options.kt
@@ -15,8 +15,11 @@
  */
 package com.squareup.wire.schema
 
+import com.google.common.collect.LinkedHashMultimap
+import com.google.common.collect.Multimap
 import com.squareup.wire.schema.ProtoMember.Companion.get
 import com.squareup.wire.schema.internal.parser.OptionElement
+import java.util.LinkedHashMap
 import java.util.regex.Pattern
 import kotlin.collections.component1
 import kotlin.collections.component2
@@ -30,15 +33,24 @@ import kotlin.collections.set
  */
 class Options(
   private val optionType: ProtoType,
-  val elements: List<OptionElement> = emptyList()
+  private val optionElements: List<OptionElement>
 ) {
   // Null until this options is linked.
-  var map: Map<ProtoMember, Any?>? = null
-    private set
+  private var entries: List<LinkedOptionEntry>? = null
 
-  fun retainLinked() = Options(optionType)
+  val elements: List<OptionElement>
+    get() {
+      return entries?.map { it.optionElement } ?: optionElements
+    }
 
-  operator fun get(protoMember: ProtoMember): Any? = map!![protoMember]
+  val map: Map<ProtoMember, Any?>
+    get() = entries!!.toMap()
+
+  fun retainLinked() = Options(optionType, emptyList())
+
+  operator fun get(protoMember: ProtoMember): Any? {
+    return entries?.find { it.protoMember == protoMember }?.value
+  }
 
   /**
    * Returns true if any of the options in `options` matches both of the regular expressions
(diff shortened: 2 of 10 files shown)


wire-schema/src/main/java/com/squareup/wire/schema/Options.kt
```diff
@@ -36,17 +37,24 @@ import kotlin.collections.set
...
-  val elements: ImmutableList<OptionElement> = ImmutableList.copyOf(elements)
-
   // Null until this options is linked.
-  var map: ImmutableMap<ProtoMember, Any?>? = null
-    private set
+  private var entries: List<LinkedOptionEntry>? = null
+
+  val elements: List<OptionElement>
+    get() {
+      return entries?.map { it.optionElement } ?: optionElements
```
swankjesse: neat!

wire-schema/src/main/java/com/squareup/wire/schema/Options.kt
```diff
@@ -55,26 +63,30 @@ class Options(
...
-    for (option in elements) {
-      val canonicalOption = canonicalizeOption(linker, optionType, option) ?: continue
-      map = union(linker, map, canonicalOption)
+    var entries: List<LinkedOptionEntry> = emptyList()
+
+    for (option in optionElements) {
+      val canonicalOption: List<LinkedOptionEntry> = canonicalizeOption(linker, optionType, option)
+              ?: continue
+
+      entries = union(linker, entries, canonicalOption)
```
swankjesse: For what it’s worth, this is `O(N*N)` which is bad if anyone has lots of options

wire-schema/src/main/java/com/squareup/wire/schema/Options.kt
```diff
@@ -198,6 +214,33 @@ class Options(
...
+    for (entry in bMap) {
+      val bValue = entry.value
+      val aValue = result[entry.key]
+      result[entry.key] = when {
+        aValue != null -> union(linker, aValue, bValue)
+        else -> bValue
+      }
+    }
+
+    return a.map { it.optionElement to it.protoMember }
```
swankjesse: this reads really awkwardly, especially as we have a.map and aMap

wire-schema/src/main/java/com/squareup/wire/schema/Options.kt
```diff
@@ -198,6 +214,33 @@ class Options(
     }
   }
 
+  private fun union(
+    linker: Linker, a: List<LinkedOptionEntry>, b: List<LinkedOptionEntry>
+  ): List<LinkedOptionEntry> {
+    val aMap: Map<ProtoMember, Any?> = a.toMap()
+    val bMap: Map<ProtoMember, Any> = b.toMap() as Map<ProtoMember, Any>
+
+    val result: MutableMap<ProtoMember, Any?> = LinkedHashMap(aMap)
```
swankjesse: should this be named result?

wire-schema/src/main/java/com/squareup/wire/schema/Options.kt
```diff
@@ -256,12 +297,20 @@ class Options(
   }
 
   fun retainAll(schema: Schema, markSet: MarkSet): Options {
-    if (map == null || map!!.isEmpty()) return this // Nothing to prune.
+    if (entries.isNullOrEmpty()) return this // Nothing to prune.
```
swankjesse: is this invoked in both null and empty cases?

wire-schema/src/main/java/com/squareup/wire/schema/Options.kt
```diff
@@ -353,4 +402,8 @@ class Options(
       return null
     }
   }
+
+  private fun List<LinkedOptionEntry>.toMap(): Map<ProtoMember, Any?> {
```
swankjesse: do you mean associateBy ?

wire-schema/src/main/java/com/squareup/wire/schema/Schema.java
```diff
@@ -125,12 +132,19 @@ static Schema fromFiles(Iterable<ProtoFile> sourceProtoFiles, Loader pathFilesLo
...
 
-  private static ImmutableMap<String, Type> buildTypesIndex(Iterable<ProtoFile> protoFiles) {
+  private static ImmutableMap<String, Type> buildTypesIndex(Iterable<ProtoFile> protoFiles,
+      Map<ProtoType, ProtoFile> protoFilesIndex) {
     Map<String, Type> result = new LinkedHashMap<>();
     for (ProtoFile protoFile : protoFiles) {
       for (Type type : protoFile.getTypes()) {
         index(result, type);
       }
+      for (Type type : result.values()) {
```
swankjesse: this loop should be out outside the enclosing for loop

swankjesse: I don’t think this is the best way to index these

swankjesse: ... instead, make the `index()` method take the `protoFilesIndex` ?

wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/LinkedOptionEntry.kt
```diff
@@ -0,0 +1,9 @@
+package com.squareup.wire.schema.internal.parser
+
+import com.squareup.wire.schema.ProtoMember
+
+data class LinkedOptionEntry(
```
swankjesse: Is this in the wrong package?

swankjesse:
I’m a little confused by the flipping between Lists and Maps.

wire-schema/src/main/java/com/squareup/wire/schema/Schema.java
oldergod: How could I get my hands on the right protoFile if I'm outside the enclosing loop?

wire-schema/src/main/java/com/squareup/wire/schema/Options.kt
oldergod: yes

wire-schema/src/main/java/com/squareup/wire/schema/Options.kt
oldergod: I agree

wire-schema/src/main/java/com/squareup/wire/schema/Options.kt
oldergod: We can look to optimise later. It's already N*N I believe.

wire-schema/src/main/java/com/squareup/wire/schema/Schema.java
oldergod: I can pass both the protoFile and the `protoFilesIndex` to `index()`

oldergod:
> I’m a little confused by the flipping between Lists and Maps.

Yes this sucks. I cheated by creating an early barrier between the change I needed and the refactoring that is much required.

proto-pruner/src/test/java/com/squareup/wire/prune/ProtoPrunerTest.kt
```diff
@@ -97,7 +96,8 @@ class ProtoPrunerTest {
     invokeProtoPruner(identifierSet)
 
     val outputs = arrayOf(
-        "poem.proto"
+        "poem.proto",
+        "options.proto"
```
oldergod: this is a bug I wanna fix later.

swankjesse:
Glad this works. Let's land and iterate!

Issue #1264, opened 2019-11-22
Trailing comments on reserved fields get carried over the next field.


kirillzh:
Steps:
1. In sample module replace one of the enum fields with reserved one and add a comment:
```proto
enum Period {
  /** 145.5 million years ago — 66.0 million years ago. */
  CRETACEOUS = 1;

  /** This is deprecated field. */
  reserved = 2;

  /** 252.17 million years ago — 201.3 million years ago. */
  TRIASSIC = 3;
}
```
2. Generate Java protos

Result (stripped down version):
```java
public enum Period implements WireEnum {
  /**
   * 145.5 million years ago — 66.0 million years ago.
   */
  CRETACEOUS(1),

  /**
   * This is deprecated field.
   * 252.17 million years ago — 201.3 million years ago.
   */
  TRIASSIC(3);

  private final int value;

  Period(int value) {
    this.value = value;
  }
}
```

Note that the `TRIASSIC` has comment carried over from previous reserved field: 
```
/**
* This is deprecated field.
* 252.17 million years ago — 201.3 million years ago.
*/
```

oldergod:
Wanna try fixing it @kirillzh ?

PR #1305, opened 2019-12-06
Adding failing tests for inlined reserved documentation affecting the next field.

oldergod:
part of #1264


oldergod:
Wanna verify the spec before merging.

PR #1307, opened 2019-12-06
Check marking of extension imports as well

oldergod:
part of #1243


swankjesse:
Ooooh I think it’s exactly where we want.

PR #1310, opened 2019-12-09
Create the profile with  the current schema

lkerford:


wire-compiler/src/main/java/com/squareup/wire/schema/NewSchemaLoader.kt
```diff
@@ -148,6 +152,76 @@ class NewSchemaLoader(
   override fun close() {
     return closer.close()
   }
+
+  override fun loadProfile(name: String, schema: Schema): Profile {
+    val allLocations = schema.protoFiles.map { it.location }
+    val locationsToCheck = locationsToCheck(name, allLocations)
+
+    val profileElements = mutableListOf<ProfileFileElement>()
+    for (location in locationsToCheck) {
```
pavlospt: I think it could be replaced with a `mapNotNull` and a `flatMap` :)

wire-compiler/src/main/java/com/squareup/wire/schema/Root.kt
```diff
@@ -59,8 +63,9 @@ internal fun Location.roots(
     }
     throw IllegalArgumentException("unable to resolve $this")
   } else {
-    val path = fs.getPath(path)
-    return path.roots(closer, this)
+    return baseToRoots.computeIfAbsent(path) {
+      fs.getPath(this.path).roots(closer, this)
```
jrodbx: nit: keep the temp `val path` for readability

wire-compiler/src/main/java/com/squareup/wire/schema/NewSchemaLoader.kt
```diff
@@ -16,10 +16,13 @@
 package com.squareup.wire.schema
 
 import com.google.common.io.Closer
+import com.squareup.wire.java.Profile
+import com.squareup.wire.java.internal.ProfileFileElement
```
jrodbx: nit: should these types be moved out of the `internal` subpackage?

wire-compiler/src/main/java/com/squareup/wire/schema/Root.kt
lkerford: done

wire-compiler/src/main/java/com/squareup/wire/schema/NewSchemaLoader.kt
lkerford: I am uncertain we can do that is this case, because we need to variable used in the `mapNotNull` in the `flatMap`, but I might be miss understanding

wire-compiler/src/main/java/com/squareup/wire/schema/NewSchemaLoader.kt
lkerford: @swankjesse  has plans for that we, we will follow up

PR #1314, opened 2019-12-12
Update CustomHandlerBeta interface to provide newProfileLoader

lkerford:
Custom implementations need to be able to provide the newProfileLoader


swankjesse:
Nice!

PR #1315, opened 2019-12-12
We don't prune google.protobuf.* options unless it's an extension

oldergod:
part of #1243 

will add more tests but wanna confirm it's not too bad as a solution now.


wire-schema/src/main/java/com/squareup/wire/schema/internal/Util.kt
```diff
@@ -78,7 +88,16 @@ object Util {
...
   fun isValidTag(value: Int) =
-    value in MIN_TAG_VALUE until RESERVED_TAG_VALUE_START ||
-        value in (RESERVED_TAG_VALUE_END + 1) until MAX_TAG_VALUE + 1
+      value in MIN_TAG_VALUE until RESERVED_TAG_VALUE_START ||
+          value in (RESERVED_TAG_VALUE_END + 1) until MAX_TAG_VALUE + 1
+}
 
-}
\ No newline at end of file
+internal fun ProtoMember.isExtensionField(schema: Schema): Boolean {
```
swankjesse: I don’t like this API. The ProtoMember is an identifier, not a value. It’d be like `"Canada".isCountry()`

swankjesse: Would be better to define an API on `Schema`.

wire-schema/src/main/java/com/squareup/wire/schema/internal/Util.kt
```diff
@@ -78,7 +88,16 @@ object Util {
...
+  val type = schema.getType(this.type)
+  return type is MessageType && type.extensionField(this.member) != null
+}
+
+internal fun ProtoMember.isGoogleProtobufType(): Boolean {
+  return arrayOf(FILE_OPTIONS, MESSAGE_OPTIONS, FIELD_OPTIONS, ENUM_OPTIONS, ENUM_VALUE_OPTIONS,
```
swankjesse: we should have a constant list for these

oldergod:
One side effect of this PR is that if `descriptor` is part of the schema, whatever the pruning configuration or their usage, the option types of `GOOGLE_PROTOBUF_OPTION_TYPES` will always be generated.

wire-schema/src/main/java/com/squareup/wire/schema/Field.kt
```diff
@@ -125,7 +126,9 @@ class Field private constructor(
 
     val memberName = if (isExtension) qualifiedName else name
     val protoMember = ProtoMember.get(enclosingType, memberName)
-    if (!markSet.contains(protoMember)) return null
+
+    if (!markSet.contains(protoMember) &&
```
oldergod: maybe I should add the `(GOOGLE_PROTOBUF_OPTION_TYPES.contains(enclosingType)` exception within the `markSet#contains` method...

wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
```diff
@@ -383,7 +383,12 @@ class PrunerTest {
     val pruned = schema.prune(IdentifierSet.Builder()
         .include("Message#g")
         .build())
-    assertThat(pruned.getType("google.protobuf.FieldOptions") as MessageType?).isNull()
```
oldergod: the option types are never pruned now, only their extension can be.

oldergod:
The same `descriptor.proto` is used for proto 3.

PR #1316, opened 2019-12-15
Basic since and until support for message fields

swankjesse:


JakeWharton:
Half-Life 3 confirmed!

wire-schema/src/main/java/com/squareup/wire/schema/IdentifierSet.kt
```diff
@@ -47,12 +47,31 @@ package com.squareup.wire.schema
...
   private val usedExcludes = mutableSetOf<String>()
 
   val isEmpty: Boolean
     get() = includes.isEmpty() && excludes.isEmpty()
 
+  /** Returns true unless [options] specifies a version that is outside of the configured range. */
+  fun isRetainedVersion(options: Options): Boolean {
+    if (newest != null) {
+      val sinceOption = options[SINCE]
+      val since = (sinceOption as? String)?.toLongOrNull()
```
JakeWharton:
```suggestion
      val since = (sinceOption as? String)?.toLong()
```
Shouldn't a specified but invalid format throw?

wire-schema/src/main/java/com/squareup/wire/schema/IdentifierSet.kt
```diff
@@ -47,12 +47,31 @@ package com.squareup.wire.schema
...
+      if (since != null && since > newest) return false
+    }
+
+    if (oldest != null) {
+      val untilOption = options[UNTIL]
+      val until = (untilOption as? String)?.toLongOrNull()
```
JakeWharton:
```suggestion
      val until = (untilOption as? String)?.toLong()
```
Same as above

wire-schema/src/main/java/com/squareup/wire/schema/IdentifierSet.kt
```diff
@@ -149,10 +170,28 @@ class IdentifierSet private constructor(builder: Builder) {
...
+
+    /**
+     * The inclusive upper bound of the version range. Fields with `since` values less than or equal
+     * to this are retained.
+     */
+    fun newest(newest: Long?) = apply {
+      this.newest = newest
+    }
+
     fun build() = IdentifierSet(this)
```
JakeWharton: Validate `oldest <= newest`?

wire-gradle-plugin/src/test/projects/since-until/src/main/proto/squareup/media/news.proto
```diff
@@ -0,0 +1,15 @@
...
+package squareup.media;
+
+option java_package = "com.squareup.media";
+
+message NewsFlash {
+  optional string title = 1;
+  optional string author = 2;
+  optional string radio = 3 [wire.until=1990];   // Omit this field on versions >= 1990.
+  optional string tv = 4 [wire.since=1950];      // Omit this field on versions < 1950.
+  optional string website = 5 [wire.since=2000]; // Omit this field on versions < 2000.
```
jrodbx: is the wire prefix necessary?  `since` and `until` read nicer.

wire-gradle-plugin/src/test/projects/since-until/build.gradle
```diff
@@ -0,0 +1,15 @@
...
+  id 'org.jetbrains.kotlin.jvm'
+  id 'com.squareup.wire'
+}
+
+wire {
+  // Support versions 1995 through 2000 inclusive. This will prune:
+  //  * declarations preceding the beginning of the range (wire.until=1995 and earlier)
+  //  * declarations following the end of the range (wire.since=2001 and later)
+  oldest 1995
+  newest 2000
```
jrodbx: interesting, my initial thoughts were to expose a single value representing the time from which to begin.  but this makes more sense.

wire-schema/src/main/java/com/squareup/wire/schema/Pruner.kt
```diff
@@ -90,16 +90,29 @@ internal class Pruner(
...
+      if (identifierSet.includes(reachable)) {
+        marks.root(reachable)
+        marks.mark(reachable.type) // Consider this type as visited.
+        queue.add(reachable)
       }
     }
   }
 
+  /** Returns true if this member survives `since` and `until` pruning. */
+  private fun isRetainedVersion(protoMember: ProtoMember): Boolean {
```
jrodbx: nit: extension function?

wire-schema/src/main/resources/wire/extensions.proto
```diff
@@ -0,0 +1,38 @@
...
+   */
+  optional int64 since = 22103;
+
+  /**
+   * Annotates a field that ceases to be available with a specified version.
+   *
+   * The value is the version that drops the annotated field. It is not available in that version or
+   * any subsequent version.
+   */
+  optional int64 until = 22104;
```
jrodbx: is there a use case to support other types?  for example, api version codes that are represented as strings?

wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
```diff
@@ -1334,4 +1334,144 @@ class PrunerTest {
...
+            |message Message {
+            |  optional string radio = 1 [wire.until=1950];
+            |  optional string video = 2 [wire.since=1950];
+            |}
+            """.trimMargin())
+        .schema()
+    val pruned = schema.prune(IdentifierSet.Builder()
+        .oldest(1949L)
+        .newest(1949L)
+        .build())
```
jrodbx: will the pruner always run as part of wire codegen?  otherwise, i'd imagine this would also be useful in non-pruning scenarios.

wire-compiler/src/main/java/com/squareup/wire/schema/NewSchemaLoader.kt
```diff
@@ -59,9 +63,23 @@ class NewSchemaLoader(
     protoPathRoots = allRoots(closer, protoPath)
   }
 
+  @Throws(IOException::class)
+  fun loadSchema(): Schema {
+    sourcePathFiles = loadSourcePathFiles()
+    val result = try {
+      Schema.fromFiles(sourcePathFiles, this)
```
Egorand: nit: name the `this` arg

wire-test-utils/src/main/java/com/squareup/wire/schema/RepoBuilder.kt
```diff
@@ -40,14 +38,10 @@ import java.nio.file.Files
...
-  private val schemaLoader = SchemaLoader().addSource(root)
+  private val schemaLoader = NewSchemaLoader(fs)
 
   fun add(name: String, protoFile: String): RepoBuilder {
-    if (name.endsWith(".proto")) {
-      schemaLoader.addProto(name)
-    } else if (!name.endsWith(".wire")) {
-      throw IllegalArgumentException("unexpected file extension: $name")
-    }
+    check(name.endsWith(".proto") || name.endsWith(".wire")) { "unexpected file extension: $name" }
```
Egorand:
```suggestion
    require(name.endsWith(".proto") || name.endsWith(".wire")) { "unexpected file extension: $name" }
```
To keep throwing `IllegalArgumentException`

wire-compiler/src/main/java/com/squareup/wire/schema/NewSchemaLoader.kt
Egorand: Oh, that's in Java, nevermind.

wire-gradle-plugin/src/test/projects/since-until/src/main/proto/squareup/media/news.proto
oldergod: Future proofing from conflicts.

Also, it looks like the spec required the parentheses for extensions https://developers.google.com/protocol-buffers/docs/proto

wire-schema/src/main/java/com/squareup/wire/schema/IdentifierSet.kt
swankjesse: Good call.

wire-schema/src/main/java/com/squareup/wire/schema/IdentifierSet.kt
swankjesse: Good call.

wire-schema/src/main/java/com/squareup/wire/schema/IdentifierSet.kt
swankjesse: Good idea.

wire-gradle-plugin/src/test/projects/since-until/src/main/proto/squareup/media/news.proto
swankjesse: Fixed!

wire-schema/src/main/java/com/squareup/wire/schema/Pruner.kt
swankjesse: I don’t like extension functions on things that rely on state that those things don’t have? In this case ProtoMember is an identifier, but this method needs to use the schema to make the decision.

wire-schema/src/main/resources/wire/extensions.proto
swankjesse: Yeah, I think we should address this before we start using it.

wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
swankjesse: It’ll always run as a part of code gen. I think "pruning" is the implementation, whether it’s tree shaking or version stripping.

wire-test-utils/src/main/java/com/squareup/wire/schema/RepoBuilder.kt
swankjesse: Fixed!

PR #1318, opened 2019-12-17
Prevent class/variable name clash during compilation of generated sources when a field has the same name as a type within the schema

mellster2012:


wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -164,6 +164,7 @@
         Set<String> collidingNames = collidingFieldNames(fieldsAndOneOfFields);
         for (Field field : fieldsAndOneOfFields) {
           String suggestion = collidingNames.contains(field.getName())
+            || schema.getType(field.getPackageName() + "." + field.getName()) != null
```
oldergod: qualifiedName

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -262,7 +262,7 @@ class KotlinGenerator private constructor(
               newName("CREATOR", "CREATOR")
             }
             message.fieldsAndOneOfFields.forEach { field ->
-              if (field.name == field.type!!.simpleName) {
+              if (field.name == field.type!!.simpleName || schema.getType("${field.packageName ?: ""}.${field.name}") != null) {
```
oldergod: qualifiedName

mellster2012:
Fixed - thanks!

PR #1322, opened 2019-12-18
Prune enum constants with wire.since and wire.until

swankjesse:


wire-schema/src/main/resources/wire/extensions.proto
```diff
@@ -34,3 +34,21 @@ extend google.protobuf.FieldOptions {
...
+   */
+  optional int64 since = 1076;
+
+  /**
+   * Annotates an enum constant that ceases to be available with a specified version.
+   *
+   * The value is the version that drops the annotated constant. It is not available in that version
+   * or any subsequent version.
+   */
+  optional int64 until = 1077;
```
oldergod: I guess we are fine with using the same ID as for the `FieldOptions`. But aren't we suppose to register those tag to some worldwideweb places?
Will we do that later before releasing kind of thing?

swankjesse: https://github.com/protocolbuffers/protobuf/pull/7020

oldergod: Dope

PR #1326, opened 2019-12-19
Remove wire-maven-plugin

jrodbx:
RIP


build.gradle
```diff
@@ -79,15 +80,6 @@ buildscript {
           'gradle': "ru.vyarus:gradle-animalsniffer-plugin:${versions.animalSnifferGradle}",
           'annotations': "org.codehaus.mojo:animal-sniffer-annotations:${versions.animalSniffer}",
       ],
-      'maven': [
-          'core': "org.apache.maven:maven-core:${versions.maven}",
-          'plugin': [
-              'api': "org.apache.maven:maven-plugin-api:${versions.maven}",
-              'annotations': "org.apache.maven.plugin-tools:maven-plugin-annotations:${versions.mavenAnnotations}",
```
Egorand: Cleanup `versions` map?

wire-codegen-sample/build.gradle
```diff
@@ -1,4 +1,3 @@
-apply plugin: 'com.liferay.maven.plugin.builder'
```
Egorand: We should rewrite `wire-codegen-sample` to use Gradle, now it won't generate code.

jrodbx: wanna tackle that in a follow-up?

Egorand: For sure!

Issue #1327, opened 2019-12-19
Including an extension of an option type should prune the other fields of the same type.


oldergod:
part of #1243 

In a big proto schema, one might wanna keep only a few options among many. Right now, whitelisting options would prune the unused ones but would keep non-whitelisted options as well.
We need a way to exclusively whitelist options.

oldergod:
fixed with #1362

PR #1328, opened 2019-12-19
Add failing test for incorrectly retained option's extension

oldergod:
For #1327

### DIFF
--- a/wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
+++ b/wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
@@ -420,6 +420,42 @@ class PrunerTest {
     assertThat((pruned.getType("SomeFieldOptions") as MessageType).field("c")).isNull()
   }
 
+  @Test @Ignore("Need to address #1327.")
+  fun includeExtensionOfOptionPrunesOtherExtensionFields() {
+    val schema = RepoBuilder()
+        .add("lecture.proto", """
+             |package squareup.options.lecture;
+             |
+             |import "google/protobuf/descriptor.proto";
+             |
+             |extend google.protobuf.FieldOptions {
+             |  optional bool relevant = 22301;
+             |  optional bool unrelevant = 22302;
+             |  optional bool unused = 22303;
+             |}
+             |
+             |message Lecture {
+             |  optional string title = 1 [(squareup.options.lecture.relevant) = true];
+             |  optional string content = 2 [(squareup.options.lecture.unrelevant) = true];
+             |}
+             """.trimMargin())
+        .schema()
+    val pruned = schema.prune(PruningRules.Builder()
+        .include("squareup.options.lecture.Lecture")
+        .include("google.protobuf.FieldOptions#squareup.options.lecture.relevant")
+        .build())
+    val fieldOptions = pruned.getType("google.protobuf.FieldOptions") as MessageType
+    assertThat(fieldOptions.extensionField("squareup.options.lecture.relevant")).isNotNull()
+    assertThat(fieldOptions.extensionField("squareup.options.lecture.unused")).isNull()
+    assertThat(fieldOptions.extensionField("squareup.options.lecture.unrelevant")).isNull()
+
+    val messageType = pruned.getType("squareup.options.lecture.Lecture") as MessageType
+    assertThat(messageType.field("title")).isNotNull()
+    assertThat(messageType.field("title")!!.options.elements).hasSize(1)
+    assertThat(messageType.field("content")).isNotNull()
+    assertThat(messageType.field("content")!!.options.elements).isEmpty()
+  }
+
   @Test
   fun optionRetainsType() {
     val schema = RepoBuilder()


wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
```diff
@@ -420,6 +420,42 @@ class PrunerTest {
...
+             |
+             |message Lecture {
+             |  optional string title = 1 [(squareup.options.misc.relevant) = true];
+             |  optional string content = 2 [(squareup.options.misc.unrelevant) = true];
+             |}
+             """.trimMargin())
+        .schema()
+    val pruned = schema.prune(PruningRules.Builder()
+        .include("squareup.options.lecture.Lecture")
+        .include("google.protobuf.FieldOptions#relevant")
```
oldergod: we do that for fields like here https://github.com/square/wire/blob/master/wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt#L471-L496

Is it fair to expect the same to happen? Otherwise, it would be really hard to whitelist the options we would wanna consume in a big code base.

wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
```diff
@@ -420,6 +420,42 @@ class PrunerTest {
...
+             |
+             |message Lecture {
+             |  optional string title = 1 [(squareup.options.lecture.relevant) = true];
+             |  optional string content = 2 [(squareup.options.lecture.unrelevant) = true];
+             |}
+             """.trimMargin())
+        .schema()
+    val pruned = schema.prune(PruningRules.Builder()
+        .include("squareup.options.lecture.Lecture")
+        .include("google.protobuf.FieldOptions#squareup.options.lecture.relevant")
```
oldergod: we do that for fields like here https://github.com/square/wire/blob/master/wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt#L471-L496

Is it fair to expect the same to happen? Otherwise, it would be really hard to whitelist the options we would wanna consume in a big code base.

Egorand: I think it shouldn't be needed. Including `Lecture` should automatically include all options it uses.

wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
```diff
@@ -420,6 +420,42 @@ class PrunerTest {
...
+        .build())
+    val fieldOptions = pruned.getType("google.protobuf.FieldOptions") as MessageType
+    assertThat(fieldOptions.extensionField("squareup.options.lecture.relevant")).isNotNull()
+    assertThat(fieldOptions.extensionField("squareup.options.lecture.unused")).isNull()
+    assertThat(fieldOptions.extensionField("squareup.options.lecture.unrelevant")).isNull()
```
Egorand: This should be non-null.

wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
```diff
@@ -420,6 +420,42 @@ class PrunerTest {
...
+
+    val messageType = pruned.getType("squareup.options.lecture.Lecture") as MessageType
+    assertThat(messageType.field("title")).isNotNull()
+    assertThat(messageType.field("title")!!.options.elements).hasSize(1)
+    assertThat(messageType.field("content")).isNotNull()
+    assertThat(messageType.field("content")!!.options.elements).isEmpty()
```
Egorand: Should not be empty. It should only be empty if `excludes` contains `google.protobuf.FieldOptions#squareup.options.lecture.unrelevant`.

wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
```diff
@@ -420,6 +420,42 @@ class PrunerTest {
...
+  fun includeExtensionOfOptionPrunesOtherExtensionFields() {
+    val schema = RepoBuilder()
+        .add("lecture.proto", """
+             |package squareup.options.lecture;
+             |
+             |import "google/protobuf/descriptor.proto";
+             |
+             |extend google.protobuf.FieldOptions {
+             |  optional bool relevant = 22301;
+             |  optional bool unrelevant = 22302;
```
Egorand: grammar nazi nit: `irrelevant`

oldergod: might be old english https://www.lexico.com/en/definition/unrelevant

oldergod:
We don't agree on the expected behavior so I'm surprised you approved it :p

wire-schema/src/test/java/com/squareup/wire/schema/PrunerTest.kt
swankjesse: Agreed with Egor that this include shouldn’t do anything? I’m trying to think of what the right syntax is to include/exclude a set of extensions

PR #1329, opened 2019-12-20
Change wire.since and wire.until to use semver versions

swankjesse:


wire-schema/src/main/java/com/squareup/wire/schema/PruningRules.kt
```diff
@@ -208,20 +210,20 @@ class PruningRules private constructor(builder: Builder) {
...
      */
-    fun newest(newest: Long?) = apply {
-      this.newest = newest
+    fun newest(newest: String?) = apply {
+      this.newest = newest?.toLowerCaseSemVer()
     }
 
     fun build(): PruningRules {
-      check((oldest ?: Long.MIN_VALUE) <= (newest ?: Long.MAX_VALUE)) {
+      check(oldest == null || newest == null || oldest!! <= newest!!) {
```
jrodbx: worth splitting into 3 separate checks for debugging purposes?

swankjesse: Unfortunately I only have the nulls there to guard the comparison.

PR #1330, opened 2019-12-20
Gradle 6.0.1

Egorand:


wire-gradle-plugin/build.gradle
```diff
@@ -48,14 +48,7 @@ tasks.named('compileKotlin').configure {
...
   into "${rootDir.path}/build/localMaven"
-
-  into("com/squareup/wire/dinosaur/all-protos/1.0") {
-    from "${projectDir.path}/src/test/libs/all-protos-1.0.jar"
-  }
-
-  into("com/squareup/wire/dinosaur/only-dinosaur/1.0") {
-    from "${projectDir.path}/src/test/libs/only-dinosaur-1.0.jar"
-  }
+  from "${projectDir.path}/src/test/libs"
```
Egorand: Gradle 6 was complaining that it can't find `maven-metadata.xml` in `localMaven`, so I've generated it with `mvn install` and copying it over into `localMaven` along with the jars and POMs.

jrodbx: should `mvn install` be run as part of the test setup script?

Egorand: I don't think there's any value in doing that, I don't expect the test proto jars to change that often.

PR #1343, opened 2019-12-27
Migrate internal parser types to common

JakeWharton:


JakeWharton:
The two `expect class`es should quickly die as the real types move to common.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Location.kt
```diff
@@ -47,7 +46,7 @@ data class Location(
   override fun toString() = buildString {
     if (!base.isEmpty()) {
       append(base)
-      append(File.separator)
+      append('/')
```
Egorand: Do we care about Windows? Does Kotlin have something similar to `File.separator`?

JakeWharton: This just matches the hardcoded `/` that we were already using down below in the factory method for the trim. If we want to make it work properly on Windows there's at least some existing behavior that needs changed.

There is no multiplatform way to get the separator, so we would have to deal with it ourselves.

PR #1351, opened 2020-01-02
Rename PruningRules#includes to pruningRoots

oldergod:


jrodbx:
WireRun has a concept of treeShakingRoots, making this rename a bit confusing.  What makes `includes` no longer an adequate descriptive name?

oldergod:
PruningRules is used both as pruning rules and as emitting rules. I wanna
split that into two different classes. For the pruning one, include means
root so I renamed accordingly.

WireRun will be using the new EmittingRules in follow-up

PR #1353, opened 2020-01-06
Add EmittingRules to be used on generation.

oldergod:
`PruningRules` was also used for emission decision. Since no pruning is involved, I've splitten the logic into its own class. Moreover, `member`s were not supported, and it's explicit now.
The logic is taken from `PruningRules` minus what concerns pruning only.

From the point of view of the Gradle Plugin, it's an explicit split between `root/prune` at the root, which is modelled by the pruning rules, and `include/exclude` inside the target modelled by the emitting rules.

### DIFF
--- a/wire-schema/src/jvmMain/java/com/squareup/wire/schema/EmittingRules.kt
+++ b/wire-schema/src/jvmMain/java/com/squareup/wire/schema/EmittingRules.kt
@@ -1,0 +1,154 @@
+/*
+ * Copyright (C) 2020 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire.schema
+
+/**
+ * A set of rules that describes which types to generate.
+ *
+ * Despite the builder, instances of this class are not safe for concurrent use.
+ *
+ * ### Identifier Matching
+ *
+ * Identifiers in this set may be in the following forms:
+ *
+ *  * Package names, followed by `.*`, like `squareup.protos.person.*`. This matches types and
+ *    services defined in the package and its descendant packages.
+ *
+ *  * Fully qualified type and service names, like `squareup.protos.person.Person`.
+ *
+ * Identifiers should not contain member names.
+ *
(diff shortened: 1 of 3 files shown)


wire-schema/src/jvmMain/java/com/squareup/wire/schema/EmittingRules.kt
```diff
@@ -0,0 +1,154 @@
...
+  private val usedExcludes = mutableSetOf<String>()
+
+  val isEmpty: Boolean
+    get() = includes.isEmpty() && excludes.isEmpty()
+
+  /** Returns true if `type` should be generated. */
+  fun includes(type: ProtoType) = includes(type.toString())
+
+  private fun includes(identifier: String): Boolean {
+    if (includes.isEmpty()) return !exclude(identifier)
```
oldergod: we could get rid of the exclude method if we were to set `includes` to `*` if empty on building.

wire-schema/src/jvmMain/java/com/squareup/wire/schema/EmittingRules.kt
```diff
@@ -0,0 +1,154 @@
...
+
+    var includeMatch: String? = null
+    var excludeMatch: String? = null
+    var rule: String? = identifier
+    while (rule != null) {
```
swankjesse: For follow-up: should this prefer the most narrow one?

wire-schema/src/jvmMain/java/com/squareup/wire/schema/EmittingRules.kt
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  private fun exclude(identifier: String): Boolean {
+    var excludeMatch: String? = null
+    var rule: String? = identifier
+    while (rule != null) {
+      if (excludes.contains(rule)) {
+        excludeMatch = rule
+      }
+      rule = PruningRules.enclosing(rule)
```
swankjesse: Should we move this method somewhere else?

wire-schema/src/jvmMain/java/com/squareup/wire/schema/EmittingRules.kt
```diff
@@ -0,0 +1,154 @@
...
+ * Identifiers in this set may be in the following forms:
+ *
+ *  * Package names, followed by `.*`, like `squareup.protos.person.*`. This matches types and
+ *    services defined in the package and its descendant packages.
+ *
+ *  * Fully qualified type and service names, like `squareup.protos.person.Person`.
+ *
+ * Identifiers should not contain member names.
+ *
+ * This set has *included identifiers* and *excluded identifiers*, with excludes taking precedence
```
Egorand: Is this in line with the idea of honouring the most granular rule, regardless of whether it's an include or exclude? In that case I expect us to fail if a type is both included and excluded explicitly at the same time.

oldergod: Yeah it's not doing the most precise rule yet

PR #1355, opened 2020-01-08
Pruning rules: Most precise rule takes precedence, and we crash for conflicting rules.

### LINKED ISSUE #1349: Update pruning behaviour to give precedence to the most precise rule
part of #1243

Right now excluding rules take precedence however precise they are compared to including ones.
We wanna update that to let the most precise rule get priority when pruning. If exclusion and inclusion are equal, exclusion takes precedence.

oldergod:
related to #1243 
fixes #1349

### DIFF
--- a/wire-schema/src/jvmMain/java/com/squareup/wire/schema/MarkSet.kt
+++ b/wire-schema/src/jvmMain/java/com/squareup/wire/schema/MarkSet.kt
@@ -34,6 +34,10 @@ class MarkSet(
 
   /** The members to retain. Any member not in here should be pruned! */
   val members = mutableMapOf<ProtoType, MutableSet<ProtoMember>>()
+  /** The root members which are never to be pruned, including their referenced type. */
+  private val rootMemberTypes = mutableMapOf<ProtoMember, ProtoType>()
+  /** The members this MarkSet have seen. The values should be non-null once the marking is done. */
+  private val memberTypes = mutableMapOf<ProtoMember, ProtoType>()
 
   /**
    * Marks `protoMember`, throwing if it is explicitly excluded. This implicitly excludes other
@@ -42,6 +46,7 @@ class MarkSet(
   fun root(protoMember: ProtoMember) {
     check(!pruningRules.excludes(protoMember))
     types.add(protoMember.type)
+    rootMemberTypes[protoMember] = UNKNOWN_TYPE
     val memberSet = members.getOrPut(protoMember.type, { mutableSetOf() })
     memberSet += protoMember
   }
@@ -50,6 +55,25 @@ class MarkSet(
   fun root(type: ProtoType) {
     check(!pruningRules.excludes(type))
     types.add(type)
+  }
+
+  /**
+   * Marks a type as transitively reachable by the includes set. Returns true if the mark is new,
+   * the type will be retained, and reachable objects should be traversed.
+   *
+   * If there is an exclude for [type], non-root members referencing it will be pruned. The type
+   * itself will also be pruned unless it is referenced by a root member.
+   */
+  fun mark(type: ProtoType, reference: ProtoMember): Boolean {
+    memberTypes[reference] = type
+
+    if (rootMemberTypes.containsKey(reference)) {
+      rootMemberTypes[reference] = type
+      // We keep.
+      return types.add(type)
+    }
+
+    return mark(type)
   }
 
   /**
@@ -79,8 +103,27 @@ class MarkSet(
 
   /** Returns true if `member` is marked and should be retained. */
   operator fun contains(protoMember: ProtoMember): Boolean {
+    val memberType: ProtoType? = memberTypes[protoMember]
+
+    // We do not contain non-root members whose referenced type is excluded.
+    if (!rootMemberTypes.containsKey(protoMember) &&
+        memberType != null && pruningRules.excludes(memberType)) {
+      return false
+    }
+
+    // A member cannot be included if its referencing type is excluded unless a root member of this
+    // referenced type exists.
+    if (pruningRules.excludes(protoMember.type) &&
+        rootMemberTypes.containsValue(protoMember.type)) {
+      return true
+    }
+
     if (pruningRules.excludes(protoMember)) return false
     val memberSet = members[protoMember.type]
     return memberSet != null && memberSet.contains(protoMember)
+  }
+
+  companion object {
+    private val UNKNOWN_TYPE = ProtoType.get("(unknown type)")
   }
 }
--- a/wire-schema/src/jvmMain/java/com/squareup/wire/schema/PruningRules.kt
+++ b/wire-schema/src/jvmMain/java/com/squareup/wire/schema/PruningRules.kt
@@ -46,9 +46,10 @@ import com.squareup.wire.schema.SemVer.Companion.toLowerCaseSemVer
  * as `Movie#name` and `Movie#release_date`). It contains the type `Actor` and one member
  * `Actor#name`, but not `Actor#birth_date` or `Actor#oscar_count`.
  *
- * This set has *included identifiers* and *excluded identifiers*, with excludes taking precedence
- * over includes. That is, if a type `Movie` is in both the includes and the excludes, it is not
- * contained in the set.
+ * This set has *included identifiers* and *excluded identifiers*, with the most precise identifier
+ * taking precedence over the other ones. For instance, if there is one included identifier
+ * `a.Movie` along an excluded identifier `a.*`, the type `a.Movie` is considered included in the
+ * set.
  *
  * If the includes set is empty, that implies that all elements should be included. Use this to
  * exclude unwanted types and members without also including everything else.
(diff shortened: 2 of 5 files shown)


wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/PruningRulesTest.kt
```diff
@@ -223,18 +237,15 @@ class PruningRulesTest {
...
         .build()
     set.includes(ProtoMember.get("a.IncludedType#member"))
-    assertThat(set.unusedExcludes()).containsExactly("a.IncludedType")
+    assertThat(set.unusedExcludes()).containsExactly("a.*")
   }
 
-  @Test
-  fun trackingUnusedPrecedence() {
-    val set = PruningRules.Builder()
+  @Test(expected = IllegalStateException::class)
```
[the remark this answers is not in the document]
oldergod: addressed

wire-schema/src/jvmMain/java/com/squareup/wire/schema/PruningRules.kt
```diff
@@ -226,6 +230,10 @@ class PruningRules private constructor(builder: Builder) {
       check(oldest == null || newest == null || oldest!! <= newest!!) {
         "expected oldest $oldest <= newest $newest"
       }
+      val conflictingRules = includes.intersect(excludes)
+      check(conflictingRules.isEmpty()) {
+        "same rule(s) defined in both includes and excludes: ${conflictingRules.joinToString()}"
```
Egorand: supernit: `Same`

wire-schema/src/jvmMain/java/com/squareup/wire/schema/PruningRules.kt
```diff
@@ -125,28 +126,27 @@ class PruningRules private constructor(builder: Builder) {
...
-      }
-      includeMatch != null -> {
-        usedIncludes.add(includeMatch)
-        true
-      }
-      else -> {
-        false
-      }
+    val isRoot = when {
+      excludeMatch != null && includeMatch != null -> excludeMatch.length < includeMatch.length
```
swankjesse: I think it matters if one is a type and the other is not. This is relevant when pruning the field type vs. the enclosing type. In other words, `a.b.C#d` is effectively longer than `a.b.e.Elephant` even though the string is shorter.

wire-schema/src/jvmMain/java/com/squareup/wire/schema/PruningRules.kt
```diff
@@ -125,28 +126,27 @@ class PruningRules private constructor(builder: Builder) {
...
+      excludeMatch != null -> false
+      includeMatch != null -> true
+      else -> false
     }
+    if (isRoot) {
+      usedIncludes.add(includeMatch!!)
+    } else {
```
swankjesse: should this be `else if` ?

wire-schema/src/jvmMain/java/com/squareup/wire/schema/PruningRules.kt
```diff
@@ -160,24 +160,28 @@ class PruningRules private constructor(builder: Builder) {
...
-    return when {
-      excludeMatch != null -> {
-        usedExcludes.add(excludeMatch)
-        true
-      }
-      else -> {
-        false
-      }
+    val excluded = when {
+      excludeMatch != null && includeMatch != null -> excludeMatch.length >= includeMatch.length
```
swankjesse: similar to above, strict string length is too simple

wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/PruningRulesTest.kt
```diff
@@ -223,18 +238,20 @@ class PruningRulesTest {
...
   @Test
-  fun trackingUnusedPrecedence() {
-    val set = PruningRules.Builder()
-        .include("a.*")
-        .exclude("a.*")
-        .build()
-    set.includes(ProtoType.get("a.Message"))
-    assertThat(set.unusedExcludes()).isEmpty()
-    assertThat(set.unusedIncludes()).containsExactly("a.*")
+  fun crashForConflictingRules() {
```
swankjesse: nice test

wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/PruningRulesTest.kt
```diff
@@ -223,18 +238,20 @@ class PruningRulesTest {
         .exclude("a.IncludedType")
         .build()
     set.includes(ProtoMember.get("a.IncludedType#member"))
-    assertThat(set.unusedExcludes()).containsExactly("a.IncludedType")
+    assertThat(set.unusedExcludes()).containsExactly("a.*")
   }
 
   @Test
-  fun trackingUnusedPrecedence() {
```
swankjesse: what happened to this?

wire-schema/src/jvmMain/java/com/squareup/wire/schema/PruningRules.kt
oldergod: When would that happen though? the longest identifier will always contain the short one if we have both an `excludeMatch` and an `includeMatch`, they cannot both match without one containing the other.

wire-schema/src/jvmMain/java/com/squareup/wire/schema/PruningRules.kt
oldergod: We don't capitalize the error message just above so I just did the _same_.

wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/PruningRulesTest.kt
oldergod: We cannot have conflicting rules anymore (this crashes). We already have tests when include/exclude are of different precision so I killed it.

wire-schema/src/jvmMain/java/com/squareup/wire/schema/PruningRules.kt
oldergod: When would that happen though? the longest identifier will always contain the short one if we have both an excludeMatch and an includeMatch, they cannot both match without one containing the other.

wire-schema/src/jvmMain/java/com/squareup/wire/schema/PruningRules.kt
oldergod: Fixed!

wire-schema/src/jvmMain/java/com/squareup/wire/schema/MarkSet.kt
```diff
@@ -52,6 +59,28 @@ class MarkSet(
     types.add(type)
   }
 
+  /**
+   * Marks a type as transitively reachable by the includes set. Returns true if the mark is new,
+   * the type will be retained, and reachable objects should be traversed.
+   * Even though `type` might be excluded, it will be retained if it is one the root member's
+   * actual type.
+   */
+  fun mark(type: ProtoType, reference: ProtoMember): Boolean {
```
oldergod: TODO(benoit) add test if people approves this approach.

wire-schema/src/jvmMain/java/com/squareup/wire/schema/MarkSet.kt
```diff
@@ -31,9 +31,15 @@ class MarkSet(
...
   val types = mutableSetOf<ProtoType>()
+  /** The types excluded by the pruning rules. */
+  private val excludedTypes = mutableSetOf<ProtoType>()
 
   /** The members to retain. Any member not in here should be pruned! */
   val members = mutableMapOf<ProtoType, MutableSet<ProtoMember>>()
+  /** The root members which are never to be pruned, including their actual type. */
+  private val rootMemberTypes = mutableMapOf<ProtoMember, ProtoType?>()
+  /** The members this MarkSet have seen. The values should be non-null once the marking is done. */
+  private val memberTypes = mutableMapOf<ProtoMember, ProtoType?>()
```
oldergod: Maybe create a class to handle that

wire-schema/src/jvmMain/java/com/squareup/wire/schema/MarkSet.kt
```diff
@@ -34,6 +34,10 @@ class MarkSet(
 
   /** The members to retain. Any member not in here should be pruned! */
   val members = mutableMapOf<ProtoType, MutableSet<ProtoMember>>()
+  /** The root members which are never to be pruned, including their actual type. */
```
swankjesse: “actual type“ means referenced type? I want a better word but I can’t think of one

wire-schema/src/jvmMain/java/com/squareup/wire/schema/MarkSet.kt
```diff
@@ -42,6 +46,7 @@ class MarkSet(
   fun root(protoMember: ProtoMember) {
     check(!pruningRules.excludes(protoMember))
     types.add(protoMember.type)
+    rootMemberTypes[protoMember] = null
```
swankjesse: Worthwhile using a value other than null as a placeholder?

val UNKNOWN_TYPE = ProtoType.get("(unknown type)")

wire-schema/src/jvmMain/java/com/squareup/wire/schema/MarkSet.kt
swankjesse: I prefer _referenced type_ here, which is what you use on line 66 anyway

wire-schema/src/jvmMain/java/com/squareup/wire/schema/MarkSet.kt
```diff
@@ -52,6 +57,24 @@ class MarkSet(
     types.add(type)
   }
 
+  /**
+   * Marks a type as transitively reachable by the includes set. Returns true if the mark is new,
+   * the type will be retained, and reachable objects should be traversed.
+   * Even though `type` might be excluded, it will be retained if it is one the root member's
```
swankjesse: Can I rephrase this sentence?

```
If there is an exclude for [type], non-root members referencing it will be pruned. The type itself will also be pruned unless is referenced by a root member.
```

oldergod: fixed

wire-schema/src/jvmMain/java/com/squareup/wire/schema/MarkSet.kt
oldergod: fixed

wire-schema/src/jvmMain/java/com/squareup/wire/schema/MarkSet.kt
oldergod: fixed

PR #1358, opened 2020-01-09
Move the majority of model types into common

JakeWharton:


wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Field.kt
```diff
@@ -1,3 +1,4 @@
+/*
```
JakeWharton: File deleted in follow-up. Kept to avoid git diff explosion.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
```diff
@@ -1,3 +1,4 @@
+/*
```
JakeWharton: File deleted in follow-up. Kept to avoid git diff explosion.

wire-schema/src/jvmMain/java/com/squareup/wire/schema/Schema.kt
```diff
@@ -25,8 +24,8 @@ import kotlin.collections.set
  *
  * Use [SchemaLoader] to load a schema from source files.
  */
-class Schema internal constructor(protoFiles: Iterable<ProtoFile>) {
-  val protoFiles = protoFiles.sortedBy { it.location.path }.toImmutableList()
+actual class Schema internal actual constructor(protoFiles: Iterable<ProtoFile>) {
+  actual val protoFiles: List<ProtoFile> = protoFiles.sortedBy { it.location.path }
```
JakeWharton: Breaking API change: `ImmutableList` --> `List`

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Linker.kt
```diff
@@ -415,45 +416,44 @@ class Linker {
     return Linker(this, context)
   }
 
-  fun addError(format: String, vararg args: Any) {
+  fun addError(message: String) {
```
JakeWharton: Breaking API change

Egorand: Alternative: call into an expect/actual `format(String, vararg Any)` that throws on JS and Native.

JakeWharton: That would mean you couldn't use the Linker and thus couldn't actually load protos on those platforms which was the whole purpose of moving this artifact in the first place.

I can leave a `@Deprecated(level=ERROR)` function that only works on the JVM for binary compatibility if we want to maintain compatibility.

Egorand: Yeah, I meant you can leave the method with the old signature and make it call into the new one. That would obviously fail if somebody calls the old method on unsupported platforms.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Linker.kt
```diff
@@ -382,11 +383,11 @@ class Linker {
           var index = 1
           append("multiple enums share constant $constant:")
           for (enumType in value) {
-            append(String.format("\n  %s. %s.%s (%s)",
-                index++, enumType.type, constant, enumType.constant(constant)!!.location))
+            append("\n  ${index++}. ${enumType.type}.$constant (%s)",
```
swankjesse: mutating ops inside a string format is unfortunate!

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Linker.kt
swankjesse: I think we take breaking changes here.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Schema.kt
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire.schema
+
+expect class Schema internal constructor(protoFiles: Iterable<ProtoFile>) {
```
swankjesse: Can we later put the whole schema into commonMain?

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/internal/Util.kt
```diff
@@ -82,3 +82,9 @@ internal fun Int.isValidTag() =
 internal expect fun Char.isDigit(): Boolean
 
 internal expect fun String.toEnglishLowerCase(): String
+
+expect interface MutableQueue<T> : MutableCollection<T> {
+  fun poll(): T
```
swankjesse: `T?`

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Schema.kt
JakeWharton: That's the plan!

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/internal/Util.kt
JakeWharton: Probably, yes. And the `T` should be `T : Any`. Although I'm not sure if the `typealias` will match then. I'll try!

JakeWharton:
👍

PR #1361, opened 2020-01-10
KotlinPoet 1.5.0

Egorand:


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -317,14 +317,14 @@ class KotlinGenerator private constructor(
       addAndroidCreator(type, companionBuilder)
     }
 
+    addMessageConstructor(type, classBuilder)
```
Egorand: https://github.com/square/kotlinpoet/pull/867

PR #1362, opened 2020-01-13
Do not prune nested type

lkerford:
Currently we are pruning protos which provided nested types if no other types in the file is used.

### DIFF
--- a/wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt
+++ b/wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt
@@ -526,6 +526,36 @@ class WireRunTest {
           """.trimMargin())
   }
 
+  @Test
+  fun includeSubTypes() {
+    writeOrangeProto()
+    writeTriangleProto()
+
+    val wireRun = WireRun(
+        sourcePath = listOf(Location.get("colors/src/main/proto")),
+        protoPath = listOf(Location.get("polygons/src/main/proto")),
+        targets = listOf(KotlinTarget(outDirectory = "generated/kt"))
+    )
+    wireRun.execute(fs, logger)
+
+    assertThat(fs.find("generated")).containsExactly(
+        "generated/kt/squareup/colors/Orange.kt")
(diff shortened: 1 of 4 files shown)


JakeWharton:
Hm what's the motivation here? The existing behavior seems correct. While nesting implies a relationship, that relationship should not be the sole reason a nested type is kept. It should have an actual path to a prune root through the type system.

jrodbx:
Agree with Jake here.  What's the use case here @lkerford ?

lkerford:
@JakeWharton @jrodbx  we have found cases were we get NPE when compiling protos.

 If protos A depends on a nested type in proto B but none of the none nested type. We remove proto B from the schema. We shouldn't be removing proto B from the schema in this case as it is use. This currently result in NPE mentioned above.

wire-schema/src/jvmMain/java/com/squareup/wire/schema/ProtoFile.kt
```diff
@@ -66,6 +66,19 @@ actual class ProtoFile private constructor(
...
   }
 
+  /**
+   * Returns all types and subtypes which are found in the proto file.
+   */
+  fun typesAndNestedTypes(): List<Type> {
+    val typesAndNestedTypes = mutableListOf<Type>()
+    typesAndNestedTypes.addAll(types)
+    for (type in types) {
+      typesAndNestedTypes.addAll(type.nestedTypes)
```
swankjesse: change this to be recursive?

wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt
```diff
@@ -558,6 +588,11 @@ class WireRunTest {
           |package squareup.polygons;
           |message Triangle {
           |  repeated double angles = 1;
+          |   enum Type {
```
oldergod: nittty as hell but there are 1 too many space on the indention

lkerford: good catch, will fix

PR #1363, opened 2020-01-14
Move protofile handling logic into the SchemaHandler.

oldergod:
This is required to create a proto target which needs protofiles information to generate the schema properly.

The consumed set is wack and weak though.

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
+++ b/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
@@ -64,6 +64,33 @@ sealed class Target {
   interface SchemaHandler {
     fun handle(type: Type)
     fun handle(service: Service)
+    /**
+     * This will handle all [Type]s and [Service]s of the `protoFile` in respect to the emitting
+     * rules. If exclusive, the handled [Type]s and [Service]s should be added to the consumed set.
+     * Consumed types and services themselves are to be omitted by this handler.
+     */
+    fun handle(
+      protoFile: ProtoFile,
+      emittingRules: EmittingRules,
+      consumedTypesAndServices: MutableSet<Any>,
+      isExclusive: Boolean
+    ) {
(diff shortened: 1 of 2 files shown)


wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -64,6 +64,33 @@ sealed class Target {
   interface SchemaHandler {
     fun handle(type: Type)
     fun handle(service: Service)
```
swankjesse: remove the above methods?

wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -64,6 +64,33 @@ sealed class Target {
...
+    /**
+     * This will handle all [Type]s and [Service]s of the `protoFile` in respect to the emitting
+     * rules. If exclusive, the handled [Type]s and [Service]s should be added to the consumed set.
+     * Consumed types and services themselves are to be omitted by this handler.
+     */
+    fun handle(
+      protoFile: ProtoFile,
+      emittingRules: PruningRules,
+      consumedTypesAndServices: MutableSet<Any>,
```
swankjesse: I’d prefer a purpose-built type for this

wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
oldergod: Actually I kind of need them for Java/Kotlin/Custom target otherwise the default implementation of `handle(protofile)` is doomed

wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
oldergod: will do in follow up

PR #1364, opened 2020-01-15
Don't fully-qualify names of fields in generated Java

swankjesse:
This backwards-incompatible change was introduced since Wire 3.0.
This reverts that change for Java which doesn't break the corresponding
test case.

See https://github.com/square/wire/pull/1318/files


swankjesse:
FYI @mellster2012

PR #1365, opened 2020-01-15
Add rudimentary proto target

oldergod:
will follow up with emitting rules application + exclusivity logic.

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
+++ b/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
@@ -76,7 +76,7 @@ sealed class Target {
       isExclusive: Boolean
     ) {
       protoFile.types
-          .filter { !consumedTypesAndServices.contains(it) && emittingRules.includes(it.type!!) }
+          .filter { !consumedTypesAndServices.contains(it) && emittingRules.includes(it.type) }
           .forEach { type ->
             handle(type)
             // We don't let other targets handle this one.
@@ -336,8 +336,7 @@ interface CustomHandlerBeta {
     outDirectory: String,
     logger: WireLogger,
     newProfileLoader: NewProfileLoader
-  ):
-      Target.SchemaHandler
+  ): Target.SchemaHandler
 }
 
 // TODO: merge this interface with Loader.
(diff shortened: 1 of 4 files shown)


wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -257,6 +260,55 @@ data class KotlinTarget(
...
+                writer.write(protoFile.toSchema())
+              }
+        } catch (e: IOException) {
+          throw IOException("Error emitting $outputFilePath to $outDirectory", e)
+        }
+      }
+    }
+  }
+
+  private fun ProtoFile.isEmpty() = types.isEmpty() && services.isEmpty() && extendList.isEmpty()
```
swankjesse: neat

wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -257,6 +260,55 @@ data class KotlinTarget(
...
+        val outputDirectory = fs.getPath(outDirectory, relativePath)
+
+        require(Files.notExists(outputDirectory) || Files.isDirectory(outputDirectory)) {
+          "path $outputDirectory exists but is not a directory."
+        }
+        Files.createDirectories(outputDirectory)
+
+        val outputFilePath = outputDirectory.resolve("${protoFile.name()}.proto")
+        try {
+          OutputStreamWriter(Files.newOutputStream(outputFilePath), StandardCharsets.UTF_8)
```
swankjesse: I wonder if you can use Okio here

wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt
```diff
@@ -151,6 +152,264 @@ class WireRunTest {
...
+    assertThat(fs.find("generated")).containsExactly(
+        "generated/proto/squareup/colors/blue.proto",
+        "generated/proto/squareup/colors/red.proto")
+    assertThat(fs.get("generated/proto/squareup/colors/blue.proto"))
+        .contains("message Blue {")
+    assertThat(fs.get("generated/proto/squareup/colors/red.proto"))
+        .contains("message Red {")
+  }
+
+  @Test @Ignore("To be implemented")
```
swankjesse: Nuke these tests and fix Target to specify the emit types only when they make sense

wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
oldergod: oh yeah

wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt
oldergod: nuked

oldergod: will refactor the target in followup

PR #1366, opened 2020-01-16
Type#type is non-nullable

oldergod:
All the implementing classes set as non-null.


wire-schema/src/jvmMain/java/com/squareup/wire/schema/Type.kt
```diff
@@ -23,7 +23,7 @@ import com.squareup.wire.schema.internal.parser.TypeElement
 
 abstract class Type {
   abstract val location: Location
-  abstract val type: ProtoType?
```
Egorand: I wonder what the use-case for null `ProtoType` was. Or was it just a Java to Kotlin conversion artifact?

oldergod: My guess is that it was a confusion around the lazy resolution Jesse has been doing recently.

PR #1369, opened 2020-01-21
Built ClaimedDefinitions to manage claimed types and services on generation

oldergod:


wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -261,11 +261,12 @@ data class KotlinTarget(
 }
 
 data class ProtoTarget(
-  override val includes: List<String> = listOf(),
-  override val excludes: List<String> = listOf(),
-  override val exclusive: Boolean = true,
   val outDirectory: String
 ) : Target() {
+  override val includes: List<String> = listOf()
```
Egorand: This change is unrelated to `ClaimedDefinitions`, wanna move it to a separate PR?

oldergod: yeah removed it

PR #1372, opened 2020-01-21
KotlinGenerator: put option fields generation behind an plugin option…

oldergod:
… defaulted to false.

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
+++ b/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
@@ -182,7 +182,10 @@ data class KotlinTarget(
   val rpcRole: RpcRole = RpcRole.CLIENT,
 
   /** True for emitted services to implement one interface per RPC. */
-  val singleMethodServices: Boolean = false
+  val singleMethodServices: Boolean = false,
+
+  /** True for option fields of emitted types' members to be generated. */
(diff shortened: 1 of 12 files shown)


wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -182,7 +182,10 @@ data class KotlinTarget(
   val rpcRole: RpcRole = RpcRole.CLIENT,
 
   /** True for emitted services to implement one interface per RPC. */
-  val singleMethodServices: Boolean = false
+  val singleMethodServices: Boolean = false,
+
+  /** True for option fields of emitted types' members to be generated. */
+  val generateOptionFields: Boolean = false
```
Egorand: I don't like this in the API. As a consumer of this library, it's not clear to me when and why I should use it. I just want stuffs to work.

oldergod: Would you rather just hide this feature for now? never generate those for Kotlin until we meet someone with a use case ?

Egorand: If the feature is broken we should fix it.

oldergod:
We're looking into getting rid of those static fields so obsolete.

PR #1376, opened 2020-01-22
Remove proto-pruner module

oldergod:
the proto target is taking over.


JakeWharton:
Can the proto target be used without Gradle?

oldergod:
I don't think the protoTarget can be used without Gradle; we have the same constraints for Kotlin, don't we ?

JakeWharton:
I don't know, but the point of proto-pruner was to be something that you could use with other tools, such as iOS toolchains.

Issue #1373, opened 2020-01-21
Generated options field can generate not compiling code.


oldergod:
Say we depend on Java written/generated/whatever `FieldOptions`. Now, if we generate via the `KotlinTarget` a message and its options, the generated code might look like
```
val FIELD_OPTIONS_FOO: FieldOptions = 	
    FieldOptions(	
      my_field_option_one = 17	
    )
```
but this code can only compile if `FieldOptions` is in Kotlin, otherwise named parameter are forbidden and this will forbid any compilation of the generated code.

Egorand:
Can you suggest steps to repro? Does this manifest when you're generating different protos containing custom options in different targets? So `FieldOptions` gets generated multiple times?

oldergod:
I'll try to write a not-compiling test in the gradle plugin module.

oldergod:
We're thinking of removing the custom options fields generation from all target.

PR #1378, opened 2020-01-22
Showcase of generated code which would fail at compilation.

oldergod:
related to #1372 and #1373 

The generated code cannot compile here. The field `FIELD_OPTIONS_NAME` is calling the constructor `FieldOptions` with a named parameter but `FieldOptions` cannot use named parameter, moreover doesn't have a parameter of the type boolean in its constructor.


oldergod:
We're looking into getting rid of those static fields so obsolete.

PR #1379, opened 2020-01-22
Adds logging for ProtoTarget generation

oldergod:


wire-compiler/src/test/java/com/squareup/wire/StringWireLogger.kt
```diff
@@ -31,6 +31,10 @@ internal class StringWireLogger : WireLogger {
     this.quiet = quiet
   }
 
+  @Synchronized  override fun artifact(outputPath: Path, filePath: String) {
```
Egorand: supernit: extra space after `@Synchronized`

oldergod: fixed!

PR #1390, opened 2020-01-27
Add task to delete all generated files on wire-tests:clean

oldergod:
I put the two `all_types.proto` and `collections.proto` because we were using generated file manually added into the code base.
Only `wire-grpc-sample` is out of the scope because the sample doesn't build with the parent project yet. All the rest is gone on cleaning.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_script:
   - if [ ! -z "$(git status --porcelain)" ]; then git diff; echo -e "\nTest files changed. Did you run ./gradlew generateTests?"; exit 1; fi
 
 script:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -13,7 +13,7 @@ before_script:
   - if [ ! -z "$(git status --porcelain)" ]; then git diff; echo -e "\nTest files changed. Did you run ./gradlew generateTests?"; exit 1; fi
 
 script:
-  - ./gradlew clean build
+  - ./gradlew build
```
oldergod: we clean in `before_script`

PR #1394, opened 2020-01-30
Add validation to check proto3 Enum has always a 0 value at the first position

### LINKED ISSUE #575: Make enums proto3 compliant for default value
This is called out in the proto3 working document (https://github.com/square/wire/issues/279).

From the [proto3 spec](https://developers.google.com/protocol-buffers/docs/proto3#default):

> Note that for scalar message fields, once a message is parsed there's no way of telling whether a field was explicitly set to the default value (for example whether a boolean was set to false) or just not set at all: you should bear this in mind when defining your message types. For example, don't have a boolean that switches on some behaviour when set to false if you don't want that behaviour to also happen by default. Also note that if a scalar message field is set to its default, the value will not be serialized on the wire.

It looks like a few small updates to the compiled files would fix this issue:
- `Builder` methods should initialize any enum fields to their default value
- `ProtoAdapter#encode` methods should not encode anything if an enum field is set to the default value.
- `ProtoAdapter#decode` methods should set enum values on the builder to `null` when catching `EnumConstantNotFoundException` (since these values no longer default to `null`).

This could break backwards compatibility with proto2 users. I'm looking to see if there's any way to specify to the compiler that files should be built for proto3 vs. proto2, but can't find any.

monkey-mas:
fixes #575 
As the official Google document says, proto3 Enum must have a zero value. To support this, we'll add a validation for this in `EnumType#validate(...)`. Since we need to know the proto version of `ProtoFile` we validate, `Syntax?` should be passed to the function.

Ref: https://developers.google.com/protocol-buffers/docs/proto3#enum
```
There must be a zero value, so that we can use 0 as a numeric default value.
```

### DIFF
--- a/wire-compiler/src/test/java/com/squareup/wire/schema/LinkerTest.kt
+++ b/wire-compiler/src/test/java/com/squareup/wire/schema/LinkerTest.kt
@@ -22,6 +22,7 @@ import com.google.common.jimfs.Jimfs
 import com.squareup.wire.testing.add
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Test
+import kotlin.test.fail
 
 class LinkerTest {
   private val fs = Jimfs.newFileSystem(Configuration.unix())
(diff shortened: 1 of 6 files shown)


wire-schema/src/jvmMain/java/com/squareup/wire/schema/EnumType.kt
```diff
@@ -49,12 +51,21 @@ class EnumType private constructor(
...
       validateTagUniqueness(linker)
     }
+    if (syntax == PROTO_3) {
+      validateTagZero(linker)
+    }
+  }
+
+  private fun validateTagZero(linker: Linker) {
+    if (constants.none { it.tag == 0 }) {
+      linker.addError("missing a tag 0 constant")
```
monkey-mas: Is `tag` a wire-specific word? Should we use `value zero` like the official doc says?

wire-schema/src/jvmMain/java/com/squareup/wire/schema/EnumType.kt
```diff
@@ -49,12 +51,21 @@ class EnumType private constructor(
...
 
     if ("true" != allowAlias) {
       validateTagUniqueness(linker)
     }
+    if (syntax == PROTO_3) {
+      validateTagZero(linker)
+    }
+  }
+
+  private fun validateTagZero(linker: Linker) {
```
monkey-mas: `validateTagZeroExistence` is better? 🙄

monkey-mas:
To be clear, this PR doesn't support another proto3 enum requirement: `The zero value needs to be the first element, for compatibility with the proto2 semantics where the first enum value is always the default.`

wire-schema/src/jvmMain/java/com/squareup/wire/schema/EnumType.kt
```diff
@@ -49,12 +51,21 @@ class EnumType private constructor(
...
   }
 
-  override fun validate(linker: Linker) {
+  override fun validate(linker: Linker, syntax: Syntax?) {
     val linker = linker.withContext(this)
 
     if ("true" != allowAlias) {
       validateTagUniqueness(linker)
     }
+    if (syntax == PROTO_3) {
```
oldergod: `===`

wire-schema/src/jvmMain/java/com/squareup/wire/schema/EnumType.kt
oldergod: Let's check that the first constant is a zero as well.

wire-schema/src/jvmMain/java/com/squareup/wire/schema/EnumType.kt
oldergod: For the error, message we could do something like `missing a zero value at the first element`

wire-compiler/src/test/java/com/squareup/wire/schema/LinkerTest.kt
```diff
@@ -22,6 +22,7 @@ import com.google.common.jimfs.Jimfs
 import com.squareup.wire.testing.add
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Test
+import kotlin.test.fail
```
oldergod: Let's use `org.junit.Assert.fail()`. We might have to add the dependency in the module.

wire-schema/src/jvmMain/java/com/squareup/wire/schema/Type.kt
```diff
@@ -29,7 +30,7 @@ abstract class Type {
   abstract val nestedTypes: List<Type>
   abstract fun linkMembers(linker: Linker)
   abstract fun linkOptions(linker: Linker)
-  abstract fun validate(linker: Linker)
+  abstract fun validate(linker: Linker, syntax: Syntax?)
```
oldergod: neat

wire-compiler/src/test/java/com/squareup/wire/schema/LinkerTest.kt
```diff
@@ -171,6 +172,72 @@ class LinkerTest {
     assertThat(schema.protoFile("b.proto")!!.javaPackage()).isEqualTo("com.squareup.b")
   }
 
+  @Test
+  fun linkerThrowsSchemaExceptionIfProto3EnumIsMissingTagZero() {
```
oldergod: Let's add a happy proto3 path and another one which fails if 0 isn't the first element.

wire-schema/src/jvmMain/java/com/squareup/wire/schema/EnumType.kt
oldergod: `validateFirstElement()` might be enough?

wire-schema/src/jvmMain/java/com/squareup/wire/schema/EnclosingType.kt
```diff
@@ -29,7 +30,9 @@ class EnclosingType internal constructor(
 
   override fun linkMembers(linker: Linker) {}
   override fun linkOptions(linker: Linker) = nestedTypes.forEach { it.linkOptions(linker) }
-  override fun validate(linker: Linker) = nestedTypes.forEach { it.validate(linker) }
+  override fun validate(linker: Linker, syntax: Syntax?) {
```
swankjesse: Can you replace Syntax the enum with an interface, SyntaxRules?

oldergod:
Let's build a SyntaxRule interface in follow up instead of passing the syntax around.
Something like

```kotlin
interface SyntaxRule {
  fun requiresEnumZeroTagValueAtFirstPosition(): Boolean
}
```

kind of thing.

PR #1396, opened 2020-01-31
Add validation to check proto3 extentions are only used for custom options

monkey-mas:
fixes: #1384 

https://developers.google.com/protocol-buffers/docs/proto3#custom_options
`Note that creating custom options uses extensions, which are permitted only for custom options in proto3.`

### DIFF
--- a/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Extend.kt
+++ b/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Extend.kt
@@ -16,6 +16,7 @@
 package com.squareup.wire.schema
 
 import com.squareup.wire.schema.Field.Companion.retainAll
+import com.squareup.wire.schema.Options.Companion.GOOGLE_PROTOBUF_OPTION_TYPES
 import com.squareup.wire.schema.internal.parser.ExtendElement
 import kotlin.jvm.JvmStatic
 
@@ -38,9 +39,13 @@ class Extend private constructor(
     }
   }
 
-  fun validate(linker: Linker) {
+  fun validate(linker: Linker, syntaxRules: SyntaxRules) {
     val linker = linker.withContext(this)
     linker.validateImport(location, type!!)
+
+    if (!syntaxRules.canExtend(ProtoType.get(name))) {
+      linker.addError("extensions are not allowed [proto3]")
+    }
   }
 
   fun retainAll(schema: Schema, markSet: MarkSet): Extend? {
--- a/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
+++ b/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
@@ -21,6 +21,7 @@ import com.squareup.wire.schema.ProtoFile.Syntax.PROTO_3
 
 /** A set of rules which defines schema requirements for a specific [Syntax]. */
 interface SyntaxRules {
+  fun canExtend(protoType: ProtoType): Boolean
   fun enumRequiresZeroValueAtFirstPosition(): Boolean
 
   companion object {
@@ -33,10 +34,14 @@ interface SyntaxRules {
     }
 
     internal val PROTO_2_SYNTAX_RULES = object : SyntaxRules {
+      override fun canExtend(protoType: ProtoType): Boolean = true
       override fun enumRequiresZeroValueAtFirstPosition(): Boolean = false
     }
 
     internal val PROTO_3_SYNTAX_RULES = object : SyntaxRules {
+      override fun canExtend(protoType: ProtoType): Boolean {
+        return protoType in Options.GOOGLE_PROTOBUF_OPTION_TYPES
+      }
       override fun enumRequiresZeroValueAtFirstPosition(): Boolean = true
     }
   }
--- a/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/SchemaTest.kt
+++ b/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/SchemaTest.kt
@@ -1674,4 +1674,49 @@ class SchemaTest {
     assertThat(enum.constant("JURASSIC")).isNotNull
     assertThat(enum.constant("TRIASSIC")).isNotNull
   }
+
+  @Test
+  fun proto3CanExtendCustomOption() {
+    val schema = RepoBuilder()
+        .add("test.proto", """
+            |syntax = "proto3";
+            |import "google/protobuf/descriptor.proto";
+            |
+            |extend google.protobuf.FieldOptions {
+            |  string a = 22101;
+            |}
+            |message Message {
+            |  string title = 1 [(a) = "hello"];
+            |}
+            """.trimMargin())
+        .schema()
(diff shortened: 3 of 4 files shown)


wire-schema/src/jvmMain/java/com/squareup/wire/schema/Extend.kt
```diff
@@ -37,9 +40,16 @@ class Extend private constructor(
...
 
-  fun validate(linker: Linker) {
+  fun validate(
+    linker: Linker,
+    syntax: Syntax?
+  ) {
     val linker = linker.withContext(this)
     linker.validateImport(location, type!!)
+
+    if (syntax === PROTO_3 && GOOGLE_PROTOBUF_OPTION_TYPES.none { it.toString() == name }) {
```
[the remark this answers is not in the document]
monkey-mas: I just thought `none {...}` is performant since we don't create a `ProtoType` instance every time we call `validate(...)`. With that said, I'll follow Kotlin-way :) > `ProtoType.get(name) !in GOOGLE_PROTOBUF_OPTION_TYPES`

monkey-mas:
TODO: 
- [ ] merge #1399

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Extend.kt
```diff
@@ -38,9 +39,17 @@ class Extend private constructor(
     }
   }
 
-  fun validate(linker: Linker) {
+  fun validate(
+    linker: Linker,
+    syntaxRules: SyntaxRules
+  ) {
```
oldergod: nit: could we put them back on one line?

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
```diff
@@ -22,6 +22,7 @@ import com.squareup.wire.schema.ProtoFile.Syntax.PROTO_3
 /** A set of rules which defines schema requirements for a specific [Syntax]. */
 interface SyntaxRules {
   fun enumRequiresZeroValueAtFirstPosition(): Boolean
+  fun extensionIsPermittedOnlyForCustomOptions(): Boolean
```
oldergod: How about turning it upside down and name it `allowsMessageExtensions()`? Plus, we're testing elsewhere, not here, if the type is an option so I'd rather take it out of the method name. I also wanna make it easy to read and understand.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Extend.kt
```diff
@@ -38,9 +39,17 @@ class Extend private constructor(
...
     val linker = linker.withContext(this)
     linker.validateImport(location, type!!)
+
+    if (syntaxRules.extensionIsPermittedOnlyForCustomOptions() &&
+        ProtoType.get(name) !in GOOGLE_PROTOBUF_OPTION_TYPES) {
+      linker.addError("extensions are permitted only for custom options [proto3]")
```
oldergod: Maybe simply `"extensions are not allowed [proto3]"`? I feel, same as the method name, that trying to convey we can actually extend on options makes thing awkward.

wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/SchemaTest.kt
```diff
@@ -1546,4 +1546,49 @@ class SchemaTest {
       )
     }
   }
+
+  @Test
```
oldergod: neat

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Extend.kt
monkey-mas: Just copy-and-pasted from the Google proto document lol
I'm not familiar with the wire terminology, so I'd be happy to fix it!

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Extend.kt
```diff
@@ -38,9 +39,14 @@ class Extend private constructor(
     }
   }
 
-  fun validate(linker: Linker) {
+  fun validate(linker: Linker, syntaxRules: SyntaxRules) {
     val linker = linker.withContext(this)
     linker.validateImport(location, type!!)
+
+    if (!syntaxRules.allowsMessageExtensions() &&
+        ProtoType.get(name) !in GOOGLE_PROTOBUF_OPTION_TYPES) {
```
oldergod: I was thinking, how about we try passing the proto type into the method? would that be such a bad thing...?

oldergod: Yeah what do you think about ? I think that's good

```kotlin
// proto2
fun allowsExtensions(type: ProtoType) = true
// proto3
fun allowsExtensions(type: ProtoType) {
  return type in GOOGLE_PROTOBUF_OPTION_TYPES
}
```

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Extend.kt
monkey-mas: [memo]
SyntaxRule is not just a config, e.g. a list of booleans for a particular syntax rule. It can have some logic to tell the API user whether or not a specific syntax is met.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Extend.kt
monkey-mas: [memo]
SyntaxRule is not just a config, e.g. a list of booleans for a particular syntax rule. It can have some logic to tell the API user whether or not a specific syntax is met.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Extend.kt
monkey-mas: [memo]
`Linker` doesn't have not know if the log is proto3 specific. It might be a good idea that `SyntaxRule` could have an API that generates some useful message to explain why the syntax error happened to the API user. This is more user-friendly and the user can figure out how to fix the error. It should be ok for SyntaxRule to know how to fix the error (since it concluded to fail the syntax for some reason).

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
```diff
@@ -33,10 +34,14 @@ interface SyntaxRules {
     }
 
     internal val PROTO_2_SYNTAX_RULES = object : SyntaxRules {
+      override fun canExtend(protoType: ProtoType): Boolean = true
       override fun enumRequiresZeroValueAtFirstPosition(): Boolean = false
     }
 
     internal val PROTO_3_SYNTAX_RULES = object : SyntaxRules {
+      override fun canExtend(protoType: ProtoType): Boolean {
+        return protoType !in Options.GOOGLE_PROTOBUF_OPTION_TYPES
```
oldergod: This is backward

Issue #1385, opened 2020-01-25
Proto3: Forbid user-defined default values


oldergod:
- [x] crash when default is set on schema. #1397 
- [ ] don't print default for proto3 schemas (when doing `Element.toSchema()`)

PR #1397, opened 2020-01-31
Add validation to check default values in proto3

monkey-mas:
fixes: #1385 

In proto3, users are not allowed to set a default value unlike proto2.

https://developers.google.com/protocol-buffers/docs/proto3#default
`When a message is parsed, if the encoded message does not contain a particular singular element, the corresponding field in the parsed object is set to the default value for that field.`

### DIFF
--- a/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
+++ b/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
@@ -21,6 +21,7 @@ import com.squareup.wire.schema.ProtoFile.Syntax.PROTO_3
 
 /** A set of rules which defines schema requirements for a specific [Syntax]. */
 interface SyntaxRules {
+  fun allowUserDefinedDefaultValue(): Boolean
   fun canExtend(protoType: ProtoType): Boolean
   fun enumRequiresZeroValueAtFirstPosition(): Boolean
 
@@ -34,11 +35,13 @@ interface SyntaxRules {
     }
 
     internal val PROTO_2_SYNTAX_RULES = object : SyntaxRules {
+      override fun allowUserDefinedDefaultValue(): Boolean = true
       override fun canExtend(protoType: ProtoType): Boolean = true
       override fun enumRequiresZeroValueAtFirstPosition(): Boolean = false
     }
 
     internal val PROTO_3_SYNTAX_RULES = object : SyntaxRules {
+      override fun allowUserDefinedDefaultValue(): Boolean = false
       override fun canExtend(protoType: ProtoType): Boolean {
         return protoType in Options.GOOGLE_PROTOBUF_OPTION_TYPES
       }
--- a/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/SchemaTest.kt
+++ b/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/SchemaTest.kt
@@ -1715,4 +1715,28 @@ class SchemaTest {
       )
     }
   }
+
+  @Test
+  fun proto3DoesNotAllowUserDefinedDefaultValue() {
+    try {
+      RepoBuilder()
+          .add("dinosaur.proto", """
+              |syntax = "proto3";
+              |
+              |message Dinosaur {
+              |  string name = 1 [default = "T-Rex"];
+              |}
+              |""".trimMargin()
+          )
+          .schema()
+      fail()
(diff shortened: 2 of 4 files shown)


monkey-mas:
TODO: 
- [ ] merge #1399

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/NewField.kt
```diff
@@ -102,14 +102,20 @@ class Field private constructor(
     isRedacted = options.optionMatches(".*\\.redacted", "true")
   }
 
-  fun validate(linker: Linker) {
+  fun validate(
+    linker: Linker,
+    syntaxRules: SyntaxRules
+  ) {
```
oldergod: nit: let's keep it on one line?

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
```diff
@@ -21,6 +21,7 @@ import com.squareup.wire.schema.ProtoFile.Syntax.PROTO_3
 
 /** A set of rules which defines schema requirements for a specific [Syntax]. */
 interface SyntaxRules {
+  fun allowUserDefinedDefaultValue(): Boolean
```
monkey-mas: I don't know for sure how much logic `SyntaxRules` can have. i.e. Is it good to pass `userDefinedDefaultValue: String?` to the func to compute the result?

In this case, do we want to check that
a) can I have a user-defined default value? e.g. `allowUserDefinedDefaultValue()` + is this default value set by the user?
or
b) can I use this default value? e.g. `allowUserDefinedDefaultValue(defaultValue: String?)` or `canUseUserDefinedDefaultValue(defaultValue: String?)`

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/NewField.kt
```diff
@@ -102,14 +102,17 @@ class Field private constructor(
...
+  fun validate(linker: Linker, syntaxRules: SyntaxRules) {
     val linker = linker.withContext(this)
     if (isPacked && !isPackable(linker, type!!)) {
       linker.addError("packed=true not permitted on $type")
     }
     if (isExtension && isRequired) {
       linker.addError("extension fields cannot be required")
     }
+    if (!syntaxRules.allowUserDefinedDefaultValue() && default != null) {
+      linker.addError("user-defined default value is not permitted [proto3]")
```
oldergod:
```suggestion
      linker.addError("user-defined default values are not permitted [proto3]")
```

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/NewField.kt
```diff
@@ -102,14 +102,17 @@ class Field private constructor(
...
-  fun validate(linker: Linker) {
+  fun validate(linker: Linker, syntaxRules: SyntaxRules) {
     val linker = linker.withContext(this)
     if (isPacked && !isPackable(linker, type!!)) {
       linker.addError("packed=true not permitted on $type")
     }
     if (isExtension && isRequired) {
       linker.addError("extension fields cannot be required")
     }
+    if (!syntaxRules.allowUserDefinedDefaultValue() && default != null) {
```
oldergod:
```suggestion
    if (default != null && !syntaxRules.allowUserDefinedDefaultValue()) {
```

No point in check the rules if the default isn't set.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
oldergod: I like a better in this case.

wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/SchemaTest.kt
```diff
@@ -1715,4 +1715,28 @@ class SchemaTest {
...
+              |message Dinosaur {
+              |  string name = 1 [default = "T-Rex"];
+              |}
+              |""".trimMargin()
+          )
+          .schema()
+      fail()
+    } catch (expected: SchemaException) {
+      assertThat(expected).hasMessage("""
+            |user-defined default value is not permitted [proto3]
```
oldergod:
```suggestion
            |user-defined default values are not permitted [proto3]
```

PR #1398, opened 2020-01-31
Delete old Field/ProtoFile and rename new ones.

oldergod:
follow-up of #1358


JakeWharton:
Can we split this to two commits? The idea was to not destroy the history of these files.

oldergod:
Split in two commits.

PR #1399, opened 2020-01-31
Add SyntaxRules to avoid passing Syntax

monkey-mas:
Follow-up for https://github.com/square/wire/pull/1394

SyntaxRules is used to have logic for a particular proto syntax.

### DIFF
--- a/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/FileLinker.kt
+++ b/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/FileLinker.kt
@@ -137,11 +137,11 @@ internal class FileLinker(
     protoFile.linkOptions(linker)
   }
 
-  fun validate() {
+  fun validate(syntaxRules: SyntaxRules) {
     protoFile.validate(linker)
 
     for (type in protoFile.types) {
-      type.validate(linker, protoFile.syntax)
+      type.validate(linker, syntaxRules)
     }
     for (service in protoFile.services) {
       service.validate(linker)
--- a/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
+++ b/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
@@ -1,0 +1,43 @@
+/*
+ * Copyright (C) 2020 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire.schema
+
+import com.squareup.wire.schema.ProtoFile.Syntax
+import com.squareup.wire.schema.ProtoFile.Syntax.PROTO_2
+import com.squareup.wire.schema.ProtoFile.Syntax.PROTO_3
+
+/** A set of rules which defines schema requirements for a specific [Syntax]. */
+interface SyntaxRules {
+  fun enumRequiresZeroValueAtFirstPosition(): Boolean
+
+  companion object {
+    fun get(syntax: Syntax?): SyntaxRules {
+      return when (syntax) {
+        PROTO_3 -> PROTO_3_SYNTAX_RULES
+        PROTO_2,
+        null -> PROTO_2_SYNTAX_RULES
+      }
+    }
+
+    internal val PROTO_2_SYNTAX_RULES = object : SyntaxRules {
+      override fun enumRequiresZeroValueAtFirstPosition(): Boolean = false
+    }
+
+    internal val PROTO_3_SYNTAX_RULES = object : SyntaxRules {
+      override fun enumRequiresZeroValueAtFirstPosition(): Boolean = true
+    }
+  }
+}
--- a/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/EnclosingType.kt
+++ b/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/EnclosingType.kt
@@ -15,7 +15,6 @@
  */
 package com.squareup.wire.schema
 
-import com.squareup.wire.schema.ProtoFile.Syntax
 import com.squareup.wire.schema.internal.parser.MessageElement
 
 /** An empty type which only holds nested types.  */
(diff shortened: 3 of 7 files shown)


wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
```diff
@@ -0,0 +1,26 @@
+package com.squareup.wire.schema
+
+import com.squareup.wire.schema.ProtoFile.Syntax
+import com.squareup.wire.schema.ProtoFile.Syntax.PROTO_2
+import com.squareup.wire.schema.ProtoFile.Syntax.PROTO_3
+
+/** A set of rules which defines schema requirements for a specific [Syntax]. */
+interface SyntaxRules {
+  fun enumRequiresZeroValueAtFirstPosition(): Boolean = false
```
monkey-mas: Would it be better to define this as `val` instead of `fun`? Thoughts?

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
```diff
@@ -0,0 +1,26 @@
...
+
+  companion object {
+    fun get(syntax: Syntax?): SyntaxRules {
+      return when (syntax) {
+        PROTO_2 -> PROTO_2_SYNTAX_RULES
+        PROTO_3 -> PROTO_3_SYNTAX_RULES
+        null -> PROTO_2_SYNTAX_RULES
```
oldergod:
```kotlin
PROTO_3 -> PROTO_3_SYNTAX_RULES
PROTO_2,
null -> PROTO_2_SYNTAX_RULES
```

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
```diff
@@ -0,0 +1,26 @@
+package com.squareup.wire.schema
+
+import com.squareup.wire.schema.ProtoFile.Syntax
+import com.squareup.wire.schema.ProtoFile.Syntax.PROTO_2
+import com.squareup.wire.schema.ProtoFile.Syntax.PROTO_3
+
+/** A set of rules which defines schema requirements for a specific [Syntax]. */
+interface SyntaxRules {
```
oldergod: internal

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
```diff
@@ -0,0 +1,26 @@
...
+  companion object {
+    fun get(syntax: Syntax?): SyntaxRules {
+      return when (syntax) {
+        PROTO_2 -> PROTO_2_SYNTAX_RULES
+        PROTO_3 -> PROTO_3_SYNTAX_RULES
+        null -> PROTO_2_SYNTAX_RULES
+      }
+    }
+
+    internal val PROTO_2_SYNTAX_RULES = object : SyntaxRules {}
```
oldergod: I wonder if we should avoid default values and always set it explicitly.

swankjesse: When designing systems like this I prefer obviousness over terseness. Default values don’t save much code but do introduce indirection so I’d avoid ’em

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/FileLinker.kt
```diff
@@ -141,7 +141,7 @@ internal class FileLinker(
     protoFile.validate(linker)
 
     for (type in protoFile.types) {
-      type.validate(linker, protoFile.syntax)
+      type.validate(linker, SyntaxRules.get(protoFile.syntax))
```
swankjesse: For follow up, probably better to create the syntax rules once rather than once on each iteration of the loop. I’d potentially pass-it-into the validate() call, that way you can later make syntax rules strict or lenient using external configuration.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
```diff
@@ -0,0 +1,26 @@
+package com.squareup.wire.schema
```
swankjesse: nit: SQ copyright on all open source .kt and .java files

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
swankjesse: I don’t feel strongly either way. And you can always change it.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
monkey-mas: Seems like we can't use `internal` interface as an argument in a public function. Let's fix this in a future follow-up.
```
e: wire/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/EnclosingType.kt: (32, 41): 'public' function exposes its 'internal' parameter type SyntaxRules
e: wire/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/EnumType.kt: (53, 41): 'public' function exposes its 'internal' parameter type SyntaxRules
e: wire/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/MessageType.kt: (134, 41): 'public' function exposes its 'internal' parameter type SyntaxRules
e: wire/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Type.kt: (33, 41): 'public' function exposes its 'internal' parameter type SyntaxRules
```

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
monkey-mas: Ok! Fixed 'em.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/FileLinker.kt
monkey-mas: Sure, let's talk about it next week!

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
monkey-mas: Oooops! Sorry about that!

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/FileLinker.kt
oldergod: In `Linker.kt#108`, we can create the syntax and pass it down for now.
```
    for (fileLinker in sourceFiles) {
      val syntaxRules = SyntaxRules.get(fileLinker.protoFile.syntax)
      fileLinker.validate(syntaxRules)
    }
```

PR #1400, opened 2020-02-03
Full support for extensions syntax.

### LINKED ISSUE #927: Wire doesn't accept multiple extension ranges declared in one `extensions` statement
The Protobuf spec allows multiple extension ranges to specified in a single `extensions` statement: [The spec specifies the following grammar:](https://developers.google.com/protocol-buffers/docs/reference/proto2-spec#extensions)

```
extensions = "extensions" ranges ";"
ranges = range { "," range }
range =  intLit [ "to" ( intLit | "max" ) ]
```

Here's an example protobuf message using this grammar:
```
message Msg {
    extensions 1, 3 to 5;
}
```

Google's protoc compiler accepts the above.
However, Wire only allows a single range; attempting to compile the above yields a syntax error. This discrepancy means a risk when using both Wire and Google's protoc in the same code base, or when trying to move from one to the other.

I'm not familiar with the Wire codebase, but it looks like the parsing code in [ProtoParser.readExtensions](https://github.com/square/wire/blob/master/wire-schema/src/main/java/com/squareup/wire/schema/internal/parser/ProtoParser.java#L471) is what doesn't allow multiple ranges. I'm not sure what other pieces of code may need to change to accommodate multiple ranges.

oldergod:
fixes #927

### DIFF
--- a/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Extensions.kt
+++ b/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Extensions.kt
@@ -22,23 +22,38 @@ import kotlin.jvm.JvmStatic
 class Extensions private constructor(
   val location: Location,
   val documentation: String,
-  val start: Int,
-  val end: Int
+  val values: List<Any>
 ) {
   fun validate(linker: Linker) {
-    if (!start.isValidTag() || !end.isValidTag()) {
-      linker.withContext(this)
-          .addError("tags are out of range: $start to $end")
+    val linker = linker.withContext(this)
+    val outOfRangeTags = mutableListOf<String>()
+    values.forEach { value ->
+      when (value) {
+        is Int -> {
+          if (!value.isValidTag()) {
+            outOfRangeTags.add("$value")
+          }
+        }
+        is IntRange -> {
+          if (!value.first.isValidTag() || !value.last.isValidTag()) {
+            outOfRangeTags.add("${value.first} to ${value.last}")
+          }
+        }
+        else -> throw AssertionError()
+      }
+    }
+    if (outOfRangeTags.isNotEmpty()) {
+      linker.addError("tags are out of range: ${outOfRangeTags.joinToString(separator = ", ")}")
     }
   }
 
   companion object {
     @JvmStatic
     fun fromElements(elements: List<ExtensionsElement>) =
-      elements.map { Extensions(it.location, it.documentation, it.start, it.end) }
+        elements.map { Extensions(it.location, it.documentation, it.values) }
 
     @JvmStatic
     fun toElements(extensions: List<Extensions>) =
-      extensions.map { ExtensionsElement(it.location, it.documentation, it.start, it.end) }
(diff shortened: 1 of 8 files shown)


wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Extensions.kt
```diff
@@ -22,23 +22,38 @@ import kotlin.jvm.JvmStatic
 class Extensions private constructor(
   val location: Location,
   val documentation: String,
-  val start: Int,
-  val end: Int
+  val values: List<Any>
```
swankjesse: document what the potential values are?

swankjesse: I guess that’s implied on line 31

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Extensions.kt
```diff
@@ -22,23 +22,38 @@ import kotlin.jvm.JvmStatic
...
 ) {
   fun validate(linker: Linker) {
-    if (!start.isValidTag() || !end.isValidTag()) {
-      linker.withContext(this)
-          .addError("tags are out of range: $start to $end")
+    val linker = linker.withContext(this)
+    val outOfRangeTags = mutableListOf<String>()
+    values.forEach { value ->
```
swankjesse: why not use a for loop?

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Extensions.kt
```diff
@@ -22,23 +22,38 @@ import kotlin.jvm.JvmStatic
...
+      when (value) {
+        is Int -> {
+          if (!value.isValidTag()) {
+            outOfRangeTags.add("$value")
+          }
+        }
+        is IntRange -> {
+          if (!value.first.isValidTag() || !value.last.isValidTag()) {
```
swankjesse: check they’re non-decreasing?

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Extensions.kt
swankjesse: ahh, found it in the element

oldergod: Yeah, it wrote it the same way `ReservedElement` was documented.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Extensions.kt
oldergod: no reason really

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Extensions.kt
oldergod: Why not

PR #1402, opened 2020-02-05
Use Kotlin's (implictly imported) LinkedHashMap

JakeWharton:


JakeWharton:
We need to get this repo onto actions. Travis CI is pain.

PR #1403, opened 2020-02-05
Move Schema and Pruner to common

JakeWharton:


wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Schema.old.kt
```diff
@@ -13,8 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+/*
 package com.squareup.wire.schema
 
 expect class Schema internal constructor(protoFiles: Iterable<ProtoFile>) {
   val protoFiles: List<ProtoFile>
 }
+*/
```
JakeWharton: Done this way to preserve the history of the `Schema.kt` file. Will push a quick follow-up to fix the names.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/CoreLoader.kt
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (C) 2019 Square, Inc.
```
oldergod: super nit: 2020?

JakeWharton: I put 2019 because this file is extracted from the existing CoreLoader (which has 2019). I don't really know which is right...

PR #1404, opened 2020-02-05
Replace JVM Pattern with Kotlin Regex

JakeWharton:


JakeWharton:
Last blocker to commonizing.

PR #1409, opened 2020-02-05
SchemaLoader API and documentation improvements

JakeWharton:


JakeWharton:
Nope, it only works on the JVM. I need something that works in multiplatform.

PR #1412, opened 2020-02-06
Stop generating options static fields for both Java and Kotlin

### LINKED ISSUE #1373: Generated options field can generate not compiling code.

oldergod:
Previously talked about between Jesse, Egor, and I.

would fix #1373


oldergod:
Will do!

PR #1416, opened 2020-02-14
Add BYTES into NUMERIC_SCALAR_TYPES

oldergod:


oldergod:
@swankjesse Is `bool` a numeric scalar as well?

monkey-mas:
I might be wrong but I think `bytes` is not numeric. My understanding is that `bytes` is a sequence but not a primitive-ish numeric value. In other words, I think `Byte` is a numeric value but I don't think `List<Byte>` is for example.

[The doc: Scalar Value Types](https://developers.google.com/protocol-buffers/docs/proto3#scalar) says that `bytes May contain any arbitrary sequence of bytes no longer than 2^32.`

swankjesse:
`bytes` is not-numeric. For whatever reason I thought proto had a standalone `byte` type (int8); it does not.

PR #1417, opened 2020-02-15
Introduce wire-protoc-tests and a simple test to start

### LINKED ISSUE #1415: Round trip between protoc and Wire
I thought it'd be cool to have tests like protoc encoding -> wire decoding, wire encoding -> protoc decoding, wire compiled of a schema vs protoc compiled of a schema (as in does Wire act the same, are we more permissive/restrictive), etc.

oldergod:
fixes #1415


build.gradle
```diff
@@ -13,7 +13,8 @@ buildscript {
       'okio': '2.4.3',
       'okhttp': '4.2.2',
       'moshi': '1.6.0',
-      'protobuf': '0.8.10',
+      'protobuf': '3.11.4',
+      'protobufGradlePlugin': '0.8.10',
       'protoc': '3.0.0',
```
monkey-mas: a) Let's keep the `protobuf` and `protoc` versions consistent.
e.g.
```
  String protobufVersion = '3.11.0'
  ext.versions = [
      'protobuf': protobufVersion,
      'protobufGradlePlugin': '0.8.10',
      'protoc': protobufVersion,
  ]
```
If we explicitly put the version, such as `3.11.0`, for each key, it'd be nice to have a comment
e.g. `'protoc': '3.11.0', // 'protoc' version should be the same as 'protobuf' version.`

b) Given that grpc-java still uses `3.11.0`, we use the same version as of now? (maybe they have their _own_ reason why they fix the version to 3.11.0 though.)

ref [grpc-java has the same version setting](https://github.com/grpc/grpc-java/blob/v1.27.1/build.gradle#L51-L52)

build.gradle
```diff
@@ -78,7 +79,8 @@ buildscript {
           'annotations': "org.codehaus.mojo:animal-sniffer-annotations:${versions.animalSniffer}",
       ],
       'protoc': "com.google.protobuf:protoc:${versions.protoc}",
-      'protobufGradlePlugin': "com.google.protobuf:protobuf-gradle-plugin:${versions.protobuf}",
+      'protocJava': "com.google.protobuf:protobuf-java:${versions.protoc}",
+      'protobufGradlePlugin': "com.google.protobuf:protobuf-gradle-plugin:${versions.protobufGradlePlugin}",
```
monkey-mas: nit: alphabetical order `protobufGradlePlugin` should be put before `protoc`

settings.gradle
```diff
@@ -2,14 +2,15 @@ include ':sample'
...
 include ':wire-gradle-plugin'
-include ':wire-gson-support'
 include ':wire-grpc-client'
 include ':wire-grpc-tests'
+include ':wire-gson-support'
 include ':wire-java-generator'
 include ':wire-kotlin-generator'
 include ':wire-moshi-adapter'
-include ':wire-runtime'
+include ':wire-protoc-tests'
```
monkey-mas: just an idea, `wire-protoc-compatibility-tests` are more informative as a module name? WDYT?

wire-protoc-tests/src/main/proto/google/protobuf/descriptor.proto
```diff
@@ -0,0 +1,885 @@
...
+// without any other information (e.g. without reading its imports).
+
+
+syntax = "proto2";
+
+package google.protobuf;
+
+option go_package = "github.com/golang/protobuf/protoc-gen-go/descriptor;descriptor";
+option java_package = "com.google.protobuf";
+option java_outer_classname = "DescriptorProtos";
```
monkey-mas: I'm not sure but do you use this proto file in this PR? If not, we can add this in a future PR when needed.

wire-protoc-tests/src/test/java/com/squareup/wire/ProtocWireTests.kt
```diff
@@ -0,0 +1,54 @@
...
+            SimpleMessageOuterClass.SimpleMessage.NestedMessage.newBuilder().setBb(806).build()
+        )
+        .setNoDefaultNestedEnum(SimpleMessageOuterClass.SimpleMessage.NestedEnum.BAR)
+        .addAllRepeatedDouble(listOf(1.0, 33.0))
+        .setRequiredInt32(46)
+        .setOther("hello")
+        .build()
+
+    val encodedWireMessage: ByteArray = wireMessage.encode()
+    val encodedGoogleMessage:ByteArray = googleMessage.toByteArray()
```
monkey-mas:
```suggestion
    val encodedGoogleMessage: ByteArray = googleMessage.toByteArray()
```

wire-protoc-tests/src/test/java/com/squareup/wire/ProtocWireTests.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    assertThat(encodedWireMessage).isEqualTo(encodedGoogleMessage)
+
+    val newWireMessage = SimpleMessage.ADAPTER.decode(encodedGoogleMessage)
+    val newGoogleMessage = SimpleMessageOuterClass.SimpleMessage.parseFrom(encodedWireMessage)
```
monkey-mas: nit: IMO, `newXxx` is a bit ambiguous for others to figure it out. i.e. what does it mean `new`?

```suggestion
    val wireMessageDecodedFromGoogleMessage = SimpleMessage.ADAPTER.decode(encodedGoogleMessage)
    val googleMessageParsedFromWireMessage =
        SimpleMessageOuterClass.SimpleMessage.parseFrom(encodedWireMessage)
```

wire-protoc-tests/src/test/java/com/squareup/wire/ProtocWireTests.kt
```diff
@@ -0,0 +1,54 @@
+/*
```
monkey-mas: I think the class pass under `test` should be `kotlin` instead of `java`. WDYT?

wire-protoc-tests/gradle.properties
```diff
@@ -0,0 +1,3 @@
+POM_ARTIFACT_ID=wire-protoc-tests
```
Egorand: Probably doesn't make sense to publish this module so this file is not needed

oldergod: We publish all the other tests module `wire-grpc-tests` and `wire-tests`. I could remove them all in follow up

wire-protoc-tests/src/test/java/com/squareup/wire/ProtocWireTests.kt
oldergod: I don't think it matters. We have kotlin code under `java` all over the place for now.

build.gradle
oldergod: the gradle plugin itself seems to be out of sync with the protobuf versioning.
I will create a hierarchy for the dependencies though

settings.gradle
oldergod: will do

wire-protoc-tests/src/main/proto/google/protobuf/descriptor.proto
oldergod: removed

wire-protoc-tests/src/test/java/com/squareup/wire/ProtocWireTests.kt
oldergod: fixed

wire-protoc-tests/src/test/java/com/squareup/wire/ProtocWireTests.kt
oldergod: yes, thank you

wire-protoc-tests/gradle.properties
oldergod: removed it from this one anyway

PR #1421, opened 2020-02-21
Proto3: convert Wire Schema Packed options to proto files

### LINKED ISSUE #1383: Proto3: Packed is default for repeated scalars
- [x] set packed to true when linking if proto3 for number scalars.
- [ ] don't print `packed="true"` for those on proto3 schemas (when doing `Element.toSchema()`)

monkey-mas:
Fixes #1383: don't print packed="true" for those on proto3 schemas (when doing Element.toSchema())

As the doc says, packed option is `false` by default in proto2[1]. On the other hand, the option is `true` for scalar numeric scalar types by default in proto3[2].
```
In proto3, repeated fields of scalar numeric types use packed encoding by default.
```

With this change, packed option values are properly generated in both proto2 and proto3 files.

Note that we *DO NOT* add `[packed = true]` option in numeric repeated fields in proto3 files since Wire Schema doesn't _know_ whether or not the "true" value was set by the user(proto3 file) or by Wire. These generated proto3 files follow the proto3 specs correctly as scalar numeric types are packed by default. (We can possibly fix this behavior later.)

[1] https://developers.google.com/protocol-buffers/docs/proto#options
[2] https://developers.google.com/protocol-buffers/docs/proto3#specifying-field-rules

### DIFF
--- a/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
+++ b/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
@@ -25,7 +25,7 @@ interface SyntaxRules {
   fun allowUserDefinedDefaultValue(): Boolean
   fun canExtend(protoType: ProtoType): Boolean
   fun enumRequiresZeroValueAtFirstPosition(): Boolean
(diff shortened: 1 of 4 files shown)


monkey-mas:
Closes since #1422 was merged instead.

PR #1422, opened 2020-02-21
Change default of Field#isPacked when linking proto3

### LINKED ISSUE #1383: Proto3: Packed is default for repeated scalars

oldergod:
Fixes #1383 
We stop adding packed option on parsing for we cannot know later if the packed was actually defined or not.

I think it's better than adding packed option on parsing because `FieldElement` stays a pure representation of what the proto file is, which allows us to print it back in its original form. I also like the symmetry we have with the existing `Type.validate(syntaxRules)`.

This kind of refactored #1414 and make #1421 obsolete.


wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Options.new.kt
```diff
@@ -42,7 +42,7 @@ class Options(
 
   fun retainLinked() = Options(optionType, emptyList())
 
-  operator fun get(protoMember: ProtoMember): Any? {
```
monkey-mas: Is this change related to this PR? If it's not, I'd prefer to do it in a separated PR.

wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/SchemaTest.kt
```diff
@@ -13,10 +13,18 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:Suppress("UsePropertyAccessSyntax")
```
monkey-mas: It feels to me that this is not necessary. Correct me if I'm wrong.

wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/SchemaTest.kt
```diff
@@ -13,10 +13,18 @@
...
 package com.squareup.wire.schema
 
 import com.squareup.wire.schema.Options.Companion.FIELD_OPTIONS
 import com.squareup.wire.schema.internal.isValidTag
+import com.squareup.wire.schema.internal.parser.FieldElement
+import com.squareup.wire.schema.internal.parser.MessageElement
+import com.squareup.wire.schema.internal.parser.OptionElement
+import com.squareup.wire.schema.internal.parser.OptionElement.Companion.PACKED_OPTION_ELEMENT
+import com.squareup.wire.schema.internal.parser.ProtoFileElement
+import com.squareup.wire.schema.internal.parser.ProtoParser
```
monkey-mas: nit: unnecessary imports.

wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/SchemaTest.kt
oldergod: We want it otherwise we'll have warning for methods like `isNull()` or `isEmpty()`, etc.

PR #1429, opened 2020-02-24
Organise usage of Wire plugin on module basis

oldergod:
How about that so we use the latest wire gradle plugin (always one build away but good enough?) save the sample which stays on regular build?

The goal would be for `wire-tests` to use that as well, and generate what it needs via the Gradle plugin (not `WireCompiler` )

### DIFF
--- a/sample/build.gradle
+++ b/sample/build.gradle
@@ -1,3 +1,12 @@
+buildscript {
+  dependencies {
+    classpath deps.plugins.releasedWire
+  }
+  repositories {
+    mavenCentral()
+  }
+}
+
 apply plugin: 'application'
 apply plugin: 'com.squareup.wire'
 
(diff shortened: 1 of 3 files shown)


sample/build.gradle
```diff
@@ -13,6 +22,11 @@ dependencies {
   implementation deps.releasedWire.runtime
 }
 
+sourceSets {
+  // TODO(benoit): we do this to make IntelliJ happy but the Wire Gradle plugin should do that.
```
swankjesse: HELL YES IT SHOULD

wire-protoc-compatibility-tests/build.gradle
```diff
@@ -1,11 +1,15 @@
 buildscript {
   dependencies {
     classpath deps.protobuf.gradlePlugin
+    classpath deps.plugins.snapshotWire
   }
 
   repositories {
     mavenCentral()
     gradlePluginPortal()
+    maven {
```
swankjesse: snapshots?

JakeWharton:
I'm opposed to this because it means you can't build old SHAs. If we want to use the latest we should use a nested Gradle build.

oldergod:
Yeah, that sucks.

PR #1430, opened 2020-02-25
Use KDoc square brackets to link arguments in function instead of ``

monkey-mas:
- Use [] instead of ``
- Fix wrong argument reference (options -> entries).

### DIFF
--- a/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Options.kt
+++ b/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Options.kt
@@ -47,7 +47,7 @@ class Options(
   }
 
   /**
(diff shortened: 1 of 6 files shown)


wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Options.kt
```diff
@@ -47,7 +47,7 @@ class Options(
   }
 
   /**
-   * Returns true if any of the options in `options` matches both of the regular expressions
+   * Returns true if any of the options in [entries] matches both of the regular expressions
```
monkey-mas: I think this refers to `private var entries: List<LinkedOptionEntry>? = null`

PR #1431, opened 2020-02-25
Migrate Wire to use composite builds

oldergod:
You can read about composite builds [here](https://docs.gradle.org/current/userguide/composite_builds.html)

The project would now be split in two
1. The root (still called wire) which contains what we don't publish, usually all our tests.
2. The `wire-library` which is all artifacts we publish (except for test-utils)

`wire-library` is self-contained, and thanks to composite builds, the `root` project can depend on `wire-library` as if it was externally published. This means we can use the Wire Gradle Plugin® *built from the source*. :applause:

There should be *no* behavior change with this PR.

Left tasks:
- [ ] make sure the publish script isn't broken.
- [ ] make sure the mkdoc thing isn't broken.


local.properties
```diff
@@ -0,0 +1,8 @@
+## This file must *NOT* be checked into Version Control Systems,
```
Egorand: Add to gitignore?

wire-tests/build.gradle
```diff
@@ -8,7 +8,7 @@ kotlin {
         kotlin.srcDirs += 'src/jvmKotlinInteropTest/proto-kotlin'
         dependencies {
           implementation(compilations.main.compileDependencyFiles)
-          implementation project(':wire-grpc-client')
+          implementation "com.squareup.wire:wire-grpc-client:3.2.0-SNAPSHOT"
```
Egorand: Should this reference coordinates inside `deps`?

Egorand: Should this reference coordinates inside `deps`?

oldergod: yes!

oldergod: yes!

local.properties
oldergod: fixed

oldergod:
Actually failing because of https://youtrack.jetbrains.com/issue/KT-36784

PR #1440, opened 2020-03-02
Prepare for codegen gRPC services

JakeWharton:
This makes GrpcMethod a dumb carrier type for an RPC endpoint. Two new methods on GrpcClient, newCall and newStreamingCall allow passing this model object to create a new call.


wire-library/wire-grpc-client/src/commonMain/kotlin/com/squareup/wire/GrpcMethod.kt
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright 2019 Square Inc.
```
JakeWharton: Derived from gutting the old type. Git just can't track it.

wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/GrpcClient.kt
```diff
@@ -33,27 +34,48 @@ class GrpcClient private constructor(
...
           }
         }
     ) as T
   }
 
+  fun <S : Any, R : Any> newCall(method: GrpcMethod<S, R>): GrpcCall<S, R> {
+    return RealGrpcCall(this, method)
+  }
+
+  fun <S : Any, R : Any> newStreamingCall(method: GrpcMethod<S, R>): GrpcStreamingCall<S, R> {
```
JakeWharton: These two methods are meant for the common version of this type. Just didn't move them there because there's no impl on the other platforms yet.

oldergod: In this kind of scenario, is it an option to move it to common and `actual stuff { TODO() }` on the to-be implemented side?

JakeWharton: Oh yeah absolutely. I just figured we'd be porting JVM to codegen first. We can do this to unblock other MPP work if we need to though.

oldergod: Was just asking for now, don't think it blocks anyone yet. 👍

PR #1442, opened 2020-03-02
Add tests for "required" custom option in proto3

oldergod:
Scalars cannot be optional anymore so I just wanna live a path that our code will have to go through for that use case without going mad.


wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/Proto3WireProtocCompatibilityTests.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    val googleMessage = RequiredExtension.RequiredExtensionMessage.newBuilder()
+        .setStringField("Yo")
+        .build()
+
+    assertThat(wireMessage.encode()).isEqualTo(googleMessage.toByteArray())
+
+    // Although the custom options has no label, it shouldn't be "required" to instantiate
+    // `FieldOptions`.
+    DescriptorProtos.FieldOptions.newBuilder().build()
```
Egorand: Missing asserts here? If expected behavior is "no exception thrown" then try/catch and `fail()` inside the catch block.

JakeWharton: I also use `assertNotNull` for cases like this with a comment.

oldergod: went with that. the try/catch adds nesting, and many lines.
Thanks to you both

PR #1445, opened 2020-03-03
Sanitize KDoc for generated code

### LINKED ISSUE #1443: Generated Kotlin classes contain errors caused by comments
## Description
If the comment of the message or field contains sequence `/*` or `*/` it will cause that the generated Kotlin class will be invalid, since they will be missing parts due to the un-escaped KDoc comments.

## Example:
```proto
//This is escaped in *Java*//
message Broken {

    // Some leading text //*Code-like comment here*
    optional string message = 1;
}
```

This will result in Kotlin class where both class comment and field comment are not ended:

```kotlin
/**
 * This is escaped in *Java*//
 */
class Broken(
  /**
   * Some leading text //*Code-like comment here*
   */
  @field:WireField(
    tag = 1,
    adapter = "com.squareup.wire.ProtoAdapter#STRING"
  )
  val message: String? = null,
  unknownFields: ByteString = ByteString.EMPTY
) // shortened...
```
while the generated Java class is handling this well:
```java
/**
 * This is escaped in *Java&#42;//
 */
public final class Broken extends Message<Broken, Broken.Builder> {
  /**
   * Some leading text //*Code-like comment here*
   */
  @WireField(
      tag = 1,
      adapter = "com.squareup.wire.ProtoAdapter#STRING"
  )
  public final String message;
 // shortened
}
```

Escaping the `*` to `&#42;` would also solve this problem for Kotlin.

I have encountered this issue while using wire gradle plugin in version `3.1.0`

oldergod:
fixes #1443

Also opened an issue in Kotlinpoet because I think it could do it for us ( https://github.com/square/kotlinpoet/issues/887 )

### DIFF
--- a/wire-library/wire-tests/src/jvmKotlinInteropTest/proto-kotlin/com/google/protobuf/SourceCodeInfo.kt
+++ b/wire-library/wire-tests/src/jvmKotlinInteropTest/proto-kotlin/com/google/protobuf/SourceCodeInfo.kt
@@ -298,11 +298,11 @@ class SourceCodeInfo(
      *   // Detached comment for corge paragraph 2.
      *
      *   optional string corge = 5;
-     *   /* Block comment attached
+     *   /&#42; Block comment attached
      *    * to corge.  Leading asterisks
-     *    * will be removed. */
-     *   /* Block comment attached to
-     *    * grault. */
+     *    * will be removed. &#42;/
+     *   /&#42; Block comment attached to
+     *    * grault. &#42;/
      *   optional int32 grault = 6;
      *
      *   // ignored detached comments.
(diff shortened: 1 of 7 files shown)


oldergod:
The thing for nested comments is that if we have a closing a nested `*/`, the tag are actually valid. That seems like a lot of work to assure the closing tag exists...

wire-library/wire-tests/src/jvmKotlinInteropTest/proto-kotlin/com/google/protobuf/SourceCodeInfo.kt
```diff
@@ -298,11 +298,11 @@ class SourceCodeInfo(
...
      *   optional string corge = 5;
-     *   /* Block comment attached
+     *   /&#42; Block comment attached
      *    * to corge.  Leading asterisks
-     *    * will be removed. */
-     *   /* Block comment attached to
-     *    * grault. */
+     *    * will be removed. &#42;/
+     *   /&#42; Block comment attached to
+     *    * grault. &#42;/
```
oldergod: Like these were valids.

PR #1452, opened 2020-03-06
Migrate GrpcDecoder and GrpcEncoder to common

JakeWharton:


wire-library/wire-grpc-client/src/commonMain/kotlin/com/squareup/wire/internal/GrpcEncoder.kt
```diff
@@ -29,7 +27,7 @@ internal sealed class GrpcEncoder(val name: String) {
   }
 
   internal object GzipGrpcEncoder : GrpcEncoder("gzip") {
-    override fun encode(sink: BufferedSink) = sink.gzip().buffer()
+    override fun encode(sink: BufferedSink) = sink.asGzip().buffer()
```
JakeWharton: @oldergod any idea why this one is buffered but the decoder is not?

oldergod: I don't recall. Maybe @swankjesse does.

swankjesse: Typically it’s ’cause that’s what the callers want.

PR #1455, opened 2020-03-06
Move GrpcMessageSink/Source to common

JakeWharton:
This is also the start of moving OkHttp's HTTP model in common code.


wire-library/wire-grpc-client/src/commonMain/kotlin/com/squareup/wire/internal/util.kt
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Executes the given [block] function on this resource and then closes it down correctly using
+ * [close] whether an exception is thrown or not.
+ *
+ * Differs from the JVM `use` in that it can be called on any type provided that you supply a
+ * [close] function.
+ *
+ * @return the result of [block] function invoked on this resource.
+ */
```
jrodbx: nice

Issue #1387, opened 2020-01-25
Proto3: Json Support


oldergod:
Both moshi-adapter and gson-support are concerned, maybe we can decide to only support moshi...

The task is about to change the way we encode/decode JSON to match the proto3 spec. We also wanna guarantee backward compatibility. Be aware of casing!

- [x] moshi
- [x] gson
- [x] backward compat
- [x] Fix moshi to look up the proper type on primitive (It's looking for `Long` when the generated code is of `long`)
- [x] name mapping: `order_ids` become `orderIds` in Json.
- [x] we shouldn't print empty lists (how about other default values?)
- [x] _If a value is missing in the JSON-encoded data or if **its value is null**, it will be interpreted as the appropriate default value_
- [ ] uint64 map keys

swankjesse:
Hold the phone. We need to use strings for these three: int64, fixed64, uint64:
https://developers.google.com/protocol-buffers/docs/proto3#json

oldergod:
Let's follow in #279

PR #1458, opened 2020-03-09
Json support for AnyMessage

oldergod:
part of #1387

Only covers Moshi.


oldergod:
will clean up Moshi tests first

PR #1457, opened 2020-03-07
Fix 404s

ligi:

PR #1459, opened 2020-03-09
Update README.md

iamrare:
update links: moved to wire-library/


iamrare:
just realized that it's already there:
https://github.com/square/wire/pull/1457

PR #1465, opened 2020-03-11
Revert changes to server side gRPC codegen

Egorand:


wire-library/wire-kotlin-generator/src/test/java/com/squareup/wire/kotlin/KotlinGeneratorTest.kt
```diff
@@ -147,25 +147,29 @@ class KotlinGeneratorTest {
   }
 
   @Test fun blockingSingleRequestSingleResponse() {
+    //language=kotlin
```
oldergod: What are those comments? I don't remember seeing them.

wire-library/wire-kotlin-generator/src/test/java/com/squareup/wire/kotlin/KotlinGeneratorTest.kt
```diff
@@ -147,25 +147,29 @@ class KotlinGeneratorTest {
...
+           */
+          @WireRpc(
+            path = "/routeguide.RouteGuide/GetFeature",
+            requestAdapter = "routeguide.Point#ADAPTER",
+            responseAdapter = "routeguide.Feature#ADAPTER"
+          )
+          fun GetFeature(request: Point): Feature
+        }
+        
+        """.trimIndent()
```
oldergod: Was the code written without margin before too? Reads nicely. Wonder if that can confuse anyone that this is actually a string and not real code when reading in the middle of it.

wire-library/wire-kotlin-generator/src/test/java/com/squareup/wire/kotlin/KotlinGeneratorTest.kt
oldergod: should we get rid of them?

Egorand: This hints the IDE to add Kotlin highlighting to the multiline string, which makes it nicer to read/edit those fixtures in the IDE.

wire-library/wire-kotlin-generator/src/test/java/com/squareup/wire/kotlin/KotlinGeneratorTest.kt
Egorand: This has changed from `trimMargin()` with pipe prefixes to `trimIndent()`. I don't think it's confusing.

PR #1466, opened 2020-03-11
Add --rpc_role and --rpc_code_style args to WireCompiler

Egorand:


Egorand:
Server codegen is completely broken at the moment and won't compile, will have to rebase once other fixes are merged.

oldergod:
How about starting a new module in root which uses the plugin, instead of the WireCompiler ?
I wanna move all the `gen_tests` to the gradle plugin (but JS is blocking me right now).

We could still start a fresh module for new tests though? and migrate `wire-tests` bit by bit.

Egorand:
It'd be great to migrate fixtures to be generated with the Gradle plugin, but it's outside of the scope of this PR. If you feel like `KotlinGeneratorTest` fixtures are enough and better than codegened ones, provided by this PR, I'll close it.

PR #1467, opened 2020-03-12
Fix README links

oldergod:


monkey-mas:
LGTM! We might fix `sed` command when deploying website, but that's another stuff. 
https://github.com/square/wire/pull/1457#issuecomment-596728302

oldergod:
I think the commands to deploy the website are done

Le mer. 11 mars 2020 à 23:10, monkey-mas <notifications@github.com> a
écrit :

PR #1472, opened 2020-03-27
Fix kotlin MessageType allocated name list

### LINKED ISSUE #1446: Generated Kotlin code with member `adapter` is hiding supertype member
## About:
If the proto definition contains field named `adapter` it will cause that the generated Kotlin code will contain conflicting member.

This is caused by the fact that `nameAllocatorStore` in `KotlinGenerator` is [set only with the constant variant (uppercase)](https://github.com/square/wire/blob/7aab9195bf1c898d355b117a9b5e3c954219789f/wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt#L257) of the member name but in the supertype `Message` it is defined lowercase.

## Example proto:
```proto
message Broken {

    // This will hide supertype member
    optional string adapter = 1;

    // This is changed to `unknownFields_`
    repeated string unknownFields = 2;
}
```

This issue is present in current master (3.2.0-SNAPSHOT).

pavelreiter:
Add `adapter` to allocated name list for MessageType in KotlinGenerator, since it is already used in Message interface.

This fixes #1446

### DIFF
--- a/wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
+++ b/wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
@@ -295,6 +295,7 @@ class KotlinGenerator private constructor(
           is MessageType -> {
             newName("unknownFields", "unknownFields")
             newName("ADAPTER", "ADAPTER")
+            newName("adapter", "adapter")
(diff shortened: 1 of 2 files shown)


Egorand:
Not sure what's up with Travis, but the build is green - somehow it failed to sync the status back to Github. Merging.

PR #1474, opened 2020-03-28
Generate interface and implementation for gRPC clients

### LINKED ISSUE #1473: Generate interfaces for gRPC client implementations
Generated client implementations aren't very ergonomic when it comes to tests, they can't be faked and there's no fake `GrpcClient` implementation either. For now we want to generate an interface for every implementation:

```kotlin
interface ClientSyncClient : Service {
  fun AddSyncEntity(): GrpcCall<AddSyncEntityRequest, AddSyncEntityResponse>
}

class RealClientSyncClient(
  private val client: GrpcClient
) : ClientSyncClient {
  override fun AddSyncEntity(): GrpcCall<AddSyncEntityRequest, AddSyncEntityResponse> =
      client.newCall(GrpcMethod(
      path = "/clientsync.service.ClientSync/AddSyncEntity",
      requestAdapter = AddSyncEntityRequest.ADAPTER,
      responseAdapter = AddSyncEntityResponse.ADAPTER
  ))
}
```

Interface and implementation should live in separate files but within the same package.

Egorand:
Closes #1473


wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -143,43 +144,79 @@ class KotlinGenerator private constructor(
...
     }
 
     return builder.build()
   }
 
+  /**
+   * If [onlyRpc] isn't null, this will generate code only for this onlyRpc; otherwise, all RPCs of
+   * the [service] will be code generated.
+   */
+  @Deprecated(
```
swankjesse: Nice.

wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -143,43 +144,79 @@ class KotlinGenerator private constructor(
...
     val serviceName = generatedServiceName(service, onlyRpc)
-    val builder = if (rpcRole == RpcRole.SERVER) {
+    val builder = if (rpcRole == RpcRole.SERVER || !generateClientImplementation) {
       TypeSpec.interfaceBuilder(serviceName)
+          .addSuperinterface(com.squareup.wire.Service::class.java)
     } else {
-      TypeSpec.classBuilder(serviceName)
+      val serviceImplementationName = ClassName(
+          serviceName.packageName,
+          "Real${serviceName.simpleName}"
```
JakeWharton: I thought we were going with "Grpc" as a prefix.

Egorand: Ah, true, I missed that comment. Submitting a PR shortly.

PR #1477, opened 2020-03-29
Use Grpc prefix for gRPC client implementations

Egorand:


Egorand:
Hm, I just discovered a compilation error that this PR introduced, meaning our CI is not working. I'm gonna look into moving us over to Actions this week.

PR #1479, opened 2020-03-30
Fix tests in KotlinGeneratorTest

Egorand:


Egorand:
This was a lot of work which goes to show that generated fixtures are better than handwritten ones. Let's get rid of `KotlinGeneratorTest` and rely on `WireCompilerTest` to verify codegen correctness.

PR #1481, opened 2020-03-30
Renamed oldest/newest to sinceVersion/untilVersion

oldergod:


oldergod:
Added a new commit to update since/until pruning rules to be inclusive/exclusive …

PR #1483, opened 2020-03-31
Encode uint64 as string on the wire via moshi

oldergod:
part of #1387

Just doing what protoc is.

We have a change in the way we print json but the parsing isn't affected so maybe that's fine?

- [ ] support int64, fixed64, uint64
- [ ] tests that confirm we handle unquoted strings (and confirming it's consistent with protoc)
- [ ] check if protos support weird stuff like `{ orders: "123.0" } or { orders: "123e1" }`


wire-library/wire-moshi-adapter/src/main/java/com/squareup/wire/WireJsonAdapterFactory.kt
```diff
@@ -135,10 +135,10 @@ class WireJsonAdapterFactory private constructor(
       @Throws(IOException::class)
       override fun toJson(writer: JsonWriter, value: Long?) {
         if (value!! < 0) {
-          val unsigned = power64.add(BigInteger.valueOf(value))
+          val unsigned = power64.add(BigInteger.valueOf(value)).toString()
```
swankjesse, line 138: I don’t think this is correct

swankjesse: Though perhaps in this incorrectness it is consistent with proto3? What do the docs advise?

swankjesse:
Hold the phone. We need to use strings for these three: int64, fixed64, uint64:
https://developers.google.com/protocol-buffers/docs/proto3#json

oldergod:
will come back. I think I need to fix wireField first.

PR #1488, opened 2020-04-02
Rename folder that contains only Kotlin sources

JakeWharton:


JakeWharton:
https://travis-ci.org/github/square/wire/builds/670289243

PR #1490, opened 2020-04-03
Fix generated Kotlin code is malformed for long identifiers

### LINKED ISSUE #1451: Generated Kotlin code is malformed for long identifiers
## About
If the proto definition contains member with name longer then 82 chars, it will cause that generated
Kotlin will contain incorrect wrapping.
As described [here](https://github.com/square/kotlinpoet#spaces-wrap-by-default), the KotlinPoet wraps the code by default and unfortunately the KotlinGenerator.kt contains large number of templates like `%L =`  where the space has to be replaced with `·` in order to prevent wrapping at incorrect location, eg.: `%L·=`
This issue is also present in case of message names, but the char limit is higher.

Issue is present in current master

## Sample proto:
```proto
syntax = "proto2";

package squareup.sample;

message Broken {

    required string membernamewhichis82charactersloooooooooooooooooooooooooooooooooooooooooooooooooong = 1;

}
```
### Generated Kotlin class
```kotlin
      // ..shortened
      override fun decode(reader: ProtoReader): Broken {
          // ..shortened
          when (tag) {
            1 -> membernamewhichis82charactersloooooooooooooooooooooooooooooooooooooooooooooooooong
                = ProtoAdapter.STRING.decode(reader)

```

pavelreiter:
About:
---------
* Replace space in KotlinGenerator in critical places with KotlinPoet non-breaking space symbol `·` to prevent generation of broken code for long message names or member names

This PR fixes #1451

### DIFF
--- a/wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
+++ b/wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
@@ -486,12 +486,12 @@ class KotlinGenerator private constructor(
 
     val body = buildCodeBlock {
       addStatement("if (%N === this) return true", otherName)
-      addStatement("if (%N !is %T) return false", otherName, kotlinType)
+      addStatement("if (%N !is %T) return·false", otherName, kotlinType)
       add("«return unknownFields == %N.unknownFields", otherName)
       val fields = type.fieldsAndOneOfFields
       for (field in fields) {
         val fieldName = localNameAllocator[field]
-        add("\n&& %1L == %2N.%1L", fieldName, otherName)
+        add("\n&& %1L·== %2N.%1L", fieldName, otherName)
       }
       add("\n»")
     }
@@ -904,7 +904,7 @@ class KotlinGenerator private constructor(
       val nameAllocator = nameAllocator(schema.getType(protoType)!!)
       if (!first) add(",")
       first = false
-      add("\n⇥%L = %L⇤", nameAllocator[field], valueInitializer)
+      add("\n⇥%L·= %L⇤", nameAllocator[field], valueInitializer)
     }
     add("\n)")
   }
@@ -1067,7 +1067,7 @@ class KotlinGenerator private constructor(
     }
 
     val returnBody = buildCodeBlock {
-      addStatement("return %T(⇥", className)
+      addStatement("return·%T(⇥", className)
 
       val missingRequiredFields = MemberName("com.squareup.wire.internal", "missingRequiredFields")
       message.fieldsAndOneOfFields.forEach { field ->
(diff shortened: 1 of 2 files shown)


Egorand:
Awesome! Would be great to add some test for these cases.

oldergod:
Indeed, a test within `KotlinGeneratorTest` would be nice

pavelreiter:
I have one prepared but I am unable to run it locally, so basically I would be blindly submit them and hope CI passes, which is not ideal.

If I try to run existing test I get following error for both cases if I try to run them from CLI or when try it from IDE, the error I get is:
```
> Task :wire-runtime:compileKotlinJvm FAILED
e: /home/user/IdeaProjects/wire/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/AnyMessage.kt: (18, 8): Unresolved reference: okio
e: /home/user/IdeaProjects/wire/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/AnyMessage.kt: (27, 14): Unresolved reference: ByteString
e: /home/user/IdeaProjects/wire/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/AnyMessage.kt: (27, 27): Unresolved reference: ByteString
e: /home/user/IdeaProjects/wire/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/AnyMessage.kt: (28, 43): Unresolved reference: ByteString
e: /home/user/IdeaProjects/wire/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/AnyMessage.kt: (67, 12): Unresolved reference: ByteString
e: /home/user/IdeaProjects/wire/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/AnyMessage.kt: (94, 21): Unresolved reference: ByteString
e: /home/user/IdeaProjects/wire/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/AnyMessage.kt: (97, 35): Overload resolution ambiguity: 
@Throws public final actual fun decode(source: [ERROR : BufferedSource]): String defined in com.squareup.wire.ProtoAdapter
@Throws public final actual fun decode(bytes: [ERROR : ByteString]): String defined in com.squareup.wire.ProtoAdapter
@Throws public abstract actual fun decode(reader: ProtoReader): String defined in com.squareup.wire.ProtoAdapter
e: /home/user/IdeaProjects/wire/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/AnyMessage.kt: (98, 32): Overload resolution ambiguity: 
@Throws public final actual fun decode(source: [ERROR : BufferedSource]): [ERROR : ByteString] defined in com.squareup.wire.ProtoAdapter
```
It is longer but it repeats across multiple classes.

Can you give me a hint how to run tests locally so I can run them before I submit it to PR?

Also to be honest, maybe best test for this is to attempt to compile the generated Kotlin class string.

pavelreiter:
Ok, so I found it in CI configuration.

For the record to run tests you have to:

### Prerequisites

-  have chrome/chromium installed and export CHROME_BIN env variable with path to binary 

### Steps:
1. run `./gradlew clean generateTests`
2. run `./gradlew -p wire-library build`
3. run tests eg. `./gradlew build`

oldergod:
When it doesn't work from the IDE, I can still run them via the gradle tab. This allow you to also run in debug mode.

PR #1462, opened 2020-03-10
Proto3: scalar/enum are required but message are optional

oldergod:
part of #1386
All implicitly. And we don't wanna mess up with oneofs, they don't have labels but their required logic is bound among their peers.

The same way we store wether a field is redacted, we store if a field _should_ be optional or _should_ be required.
This was a bit tough but couldn't come up with a better way for now.

Issue #1386, opened 2020-01-25
Proto3: Scalars become non-nullable, but message fields remain nullable.


oldergod:
- [x] crash if label (required|optional) is set
- [x] deal with oneof
- [x] don't print required or optional for proto3 schemas (`Element.toSchema()`)
- [x] skip default values on encoding/autoset them on decoding.
  - [x] single values
  - [x] needs to work on list
- [x] Test case: empty string, default enum, 0 int in a repeated field should all be encoded.
- [x] Test case: map entries that have default values.
- [x] throw when default is set like `> protoc: stdout: . stderr: squareup/proto3/pizza/pizza.proto: Explicit default values are not allowed in proto3.`

Adapter and generated code (set required/optional with default as fit).

PR #1494, opened 2020-04-11
Deprecate isOptional/isRequired in Field and introduce encodeMode

oldergod:
Remake of #1462, part of #1386

The goal is to rely on encode mode from within the generators to fully support proto3 as well, for optionality/requiredness is a proto2 only concept.

The fact that the golden files didn't change is my test for not breaking old code. We'll test proto3 in follow-ups when we update the generators.

### DIFF
--- a/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Field.kt
+++ b/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Field.kt
@@ -26,7 +26,8 @@ class Field private constructor(
 
   val location: Location,
 
-  /** May be null for proto3 fields. */
+  // TODO(oldergod) Make private, we wanna expose encodeMode instead.
+  /** May be null for proto3 fields, or one-of's. */
   val label: Label?,
 
   val name: String,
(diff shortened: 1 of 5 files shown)


wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Field.kt
```diff
@@ -50,22 +51,25 @@ class Field private constructor(
...
-  private var packed: Any? = null
-
   // Null until this field is linked.
   var isRedacted: Boolean = false
     private set
 
   val isRepeated: Boolean
     get() = label == Label.REPEATED
 
+  @Deprecated("Proto2 concept inexistant in proto3. Use EncodeMode instead.")
```
swankjesse, line 61: Did you mean: **_inexistent_**

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Field.kt
```diff
@@ -50,22 +51,25 @@ class Field private constructor(
...
   val isOptional: Boolean
     get() = label == Label.OPTIONAL
 
+  @Deprecated("Proto2 concept inexistant in proto3. Use EncodeMode instead.")
```
swankjesse, line 65: Did you mean: **_inexistent_**

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Field.kt
```diff
@@ -209,10 +215,28 @@ class Field private constructor(
...
   }
 
+  enum class EncodeMode {
+    /** Optional from proto2. */
+    NULL_IF_ABSENT,
+
+    /** Required from proto2. */
+    THROW_IF_ABSENT,
+
+    /** Non-repeated fields in proto3. */
```
swankjesse, line 230: Can you say `0`, `false`, or `null` here?

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Field.kt
oldergod: I like to speak obsolete words
image

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Field.kt
oldergod: fixed, and added `""` as well.

PR #1501, opened 2020-04-16
Adds a MAP EncodeMode

oldergod:
Goal is to improve readability
Hopefully this should also make things easier to read in Java.


wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Field.kt
```diff
@@ -224,7 +224,7 @@ class Field private constructor(
   }
 
   enum class EncodeMode {
-    /** Optional, Map, or OneOf from proto2. */
+    /** Optional, or OneOf from proto2. */
```
JakeWharton, line 227:
```suggestion
    /** Optional or OneOf from proto2. */
```

PR #1502, opened 2020-04-16
Add OneOf EncodeMode

oldergod:
Goal is to improve readability.


oldergod:
Hmm, I need to first check how encoding is supposed to happen for oneof on proto2. If proto3 and proto2 differs here, that sucks.

PR #1505, opened 2020-04-16
Adds some test for kotlin keywords in Json passes.

oldergod:


wire-library/wire-moshi-adapter/src/test/java/com/squareup/wire/MoshiTest.java
```diff
@@ -293,4 +295,14 @@
     assertThat(parsed.pack_int32).isEmpty();
     assertThat(parsed.map_int32_int32).isEmpty();
   }
+
+  @Test public void usedKeywordsInKotlin() throws IOException {
+    JsonAdapter<Keyword> adapter = moshi.adapter(Keyword.class);
+
+    Keyword keyword = new Keyword.Builder().object_("object").when_(1).build();
+    String json = adapter.toJson(keyword);
+    JsonUtils.assertJsonEquals("{\"object_\":\"object\",\"when_\":1}", json);
```
swankjesse, line 304: These expectations are wrong, aren’t they? Shouldn’t this be a failing test with `@Ignore` ?

oldergod: You're 200% right. That's what reviews are for.

PR #1508, opened 2020-04-19
Add originalName field to WireField

oldergod:


wire-library/wire-tests/src/jvmJavaTest/proto-java/com/google/protobuf/FieldOptions.java
```diff
@@ -212,7 +212,8 @@
    */
   @WireField(
       tag = 22101,
-      adapter = "com.squareup.wire.ProtoAdapter#STRING"
+      adapter = "com.squareup.wire.ProtoAdapter#STRING",
+      originalName = "a"
   )
   public final String squareup_protos_extension_collision_1_a;
```
oldergod, line 218: Yo the Json must have been crazy

wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/WireField.kt
```diff
@@ -44,7 +45,13 @@ annotation class WireField(
   /**
    * Redacted fields are omitted from toString() to protect sensitive data. Defaults to false.
    */
-  val redacted: Boolean = false
+  val redacted: Boolean = false,
+  /**
+   * Original name of this field as defined in the proto schema. This value is set to a non-empty
+   * string only when the original name differs from the generated one; for instance, a proto field
+   * named `object` generated in Java will be renamed `object_`.
```
Egorand, line 52: Is this actually the case with JavaGenerator (since `object` is not a Java keyword)?

oldergod: Java shouldn't renamed `object` no

oldergod: oh I see...

Egorand: I can submit a PR.

PR #1509, opened 2020-04-19
Use WireField.originalName for json passing.

### LINKED ISSUE #1503: Can wire generate `object` for fields called object in Kotlin instead of object_?
Some of the square APIs include an `object` field ie https://developer.squareup.com/docs/disputes-api/process-disputes#webhooksubscription

We use wire and compile some protos to Kotlin, however wire generates code for for that field with an identifier of `object_`. We use Gson / json serialization and kotlin supports keywords-as-identifiers via quasiquotes so it would be nice if the compiler could generate a field called

```
`object`
```

It'd be fine as a compiler param for backwards compatibility with any of your java code that needs to use object_.

Here's the same info from my colleague

> I got an interesting issue. I added new proto Kotlin SquareWebHookEvent it has optional Object object = 3; and Kotlin generates field object_ because object is a keywork in Kotlin. But object_ is not papulated because we use Gson for de-serialization. Does anyone know how to de-serialize json with com.squareup.wire & ProtoAdapter?

oldergod:
fixes #1503


JakeWharton:
declaredName? protoName?

PR #1513, opened 2020-04-22
Use name allocation for missingRequiredFields params

### LINKED ISSUE #1504: Finishing support around keywords in Kotlin
- [x] required field won't get compiling code. `when_ = when_ ?: throw missingRequiredFields(when, "when"),`
- [x] properly rename fields when going to/from json

See #1505 #1506

oldergod:
Also added some repeated fields for good measures.

fixes #1504

### DIFF
--- a/wire-library/wire-moshi-adapter/src/test/java/squareup/keywords/KeywordKotlin.kt
+++ b/wire-library/wire-moshi-adapter/src/test/java/squareup/keywords/KeywordKotlin.kt
@@ -8,11 +8,14 @@ import com.squareup.wire.ProtoAdapter
 import com.squareup.wire.ProtoReader
 import com.squareup.wire.ProtoWriter
 import com.squareup.wire.WireField
+import com.squareup.wire.internal.checkElementsNotNull
+import com.squareup.wire.internal.missingRequiredFields
 import com.squareup.wire.internal.sanitize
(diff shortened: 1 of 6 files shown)


wire-library/wire-moshi-adapter/src/test/java/squareup/keywords/KeywordKotlin.kt
```diff
@@ -98,12 +114,15 @@ class KeywordKotlin(
     @JvmField
     var fun_: Map<String, String> = emptyMap()
 
+    @JvmField
+    var private: List<Boolean> = emptyList()
```
JakeWharton, line 118: So this is either:
 1. Broken because it's not escaped, or
 2. Not a keyword so shouldn't be used to test this case in Kotlin

oldergod: it compiles so.... will change.

oldergod: image

oldergod: used `return` instead for both.

PR #1514, opened 2020-04-22
Initial support for identity if absent in JSON

swankjesse:


wire-library/wire-moshi-adapter/src/main/java/com/squareup/wire/IdentityIfAbsent.kt
```diff
@@ -0,0 +1,11 @@
+package com.squareup.wire
```
oldergod, line 1: copyright

oldergod: hmm I noticed our `internal annotation class Uint64` doesn't have it either. Not needed for them?

swankjesse: I like it for everything.

swankjesse: Fixed x 2.

PR #1516, opened 2020-04-22
Implement proto3 durations for the JVM

swankjesse:
Rather than using the generated proto type we use javax.time.Duration.

### DIFF
--- a/wire-library/wire-moshi-adapter/src/main/java/com/squareup/wire/DurationJsonAdapter.kt
+++ b/wire-library/wire-moshi-adapter/src/main/java/com/squareup/wire/DurationJsonAdapter.kt
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2020 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire
+
+import com.squareup.moshi.JsonAdapter
+import com.squareup.moshi.JsonDataException
+import com.squareup.moshi.JsonReader
+import com.squareup.moshi.JsonWriter
+import java.time.Duration
+
+/**
+ * Encode a duration as a JSON string like "1.200s". From the spec:
+ *
+ * > Generated output always contains 0, 3, 6, or 9 fractional digits, depending on required
+ * > precision, followed by the suffix "s". Accepted are any fractional digits (also none) as long
+ * > as they fit into nano-seconds precision and the suffix "s" is required.
+ *
+ * Note that [Duration] always returns a positive nanosPart, so "-1.200s" is represented as -2
+ * seconds and 800_000_000 nanoseconds.
+ */
+internal object DurationJsonAdapter : JsonAdapter<Duration>() {
+  override fun toJson(out: JsonWriter, duration: Duration?) {
+    out.value(durationToString(duration!!))
+  }
+
+  override fun fromJson(input: JsonReader): Duration? {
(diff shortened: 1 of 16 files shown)


wire-library/wire-moshi-adapter/src/main/java/com/squareup/wire/WireJsonAdapterFactory.kt
```diff
@@ -94,6 +95,9 @@ class WireJsonAdapterFactory private constructor(
     if (rawType == AnyMessage::class.java) {
       return AnyMessageJsonAdapter(moshi, typeUrlToAdapter)
     }
+    if (rawType == Duration::class.java) {
```
JakeWharton, line 98: This will require users be using AGP 4.0 and have `coreLibraryDesugaring true`. You might want to wrap this in try/catch for NoClassDefFoundError and ignore so it works on AGP 3.x and/or AGP 4 without core library desugaring.

wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -215,5 +216,6 @@ actual abstract class ProtoAdapter<E> actual constructor(
     @JvmField actual val DOUBLE: ProtoAdapter<Double> = commonDouble()
     @JvmField actual val BYTES: ProtoAdapter<ByteString> = commonBytes()
     @JvmField actual val STRING: ProtoAdapter<String> = commonString()
+    @JvmField val DURATION: ProtoAdapter<Duration> = DurationProtoAdapter
```
JakeWharton, line 219: This again will require AGP 4.0 and core library desugaring enabled because it's initialized on class load and `DurationProtoAdapter` eager-initializes as a singleton whose constructor then makes reference to `Duration::class` to invoke the superclass constructor.

I don't think you even need this field. Kotlin users can refer to the object and Java users can call `DurationProtoAdapter.INSTANCE`.

wire-library/wire-moshi-adapter/src/main/java/com/squareup/wire/DurationJsonAdapter.kt
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Encode a duration as a JSON string like "1.200s". From the spec:
+ *
+ * > Generated output always contains 0, 3, 6, or 9 fractional digits, depending on required
+ * > precision, followed by the suffix "s". Accepted are any fractional digits (also none) as long
+ * > as they fit into nano-seconds precision and the suffix "s" is required.
+ *
+ * Note that [Duration] always returns a positive nanosPart, so "-1.200s" is represented as -2
+ * seconds and 800_000_000 nanoseconds.
```
oldergod, line 32: :sweatingtowelguy:

wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/DurationProtoAdapter.kt
```diff
@@ -0,0 +1,80 @@
...
+   * Returns a value like 1 for 1.200s and -1 for -1.200s. This is different from the Duration
+   * seconds field which is always the integer floor when seconds is negative.
+   */
+  private val Duration.sameSignSeconds: Long
+    get() {
+      return when {
+        seconds < 0L && nano != 0 -> seconds + 1L
+        else -> seconds
+      }
+    }
```
oldergod, line 67: nice

wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/DurationTest.kt
```diff
@@ -0,0 +1,78 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire
+
+import com.google.protobuf.Duration
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+
+class DurationTest {
```
oldergod, line 22: Super nice!

JakeWharton:
Good point. The MPP date/time library is still progressing on the JetBrains side but it won't be stable for a while. I would think ideally we would want to use that when it's available over java.time. Perhaps we should gait this behind some JVM-specific experimental flag?

oldergod:
Superseeded by #1579

PR #1524, opened 2020-04-24
Split CI build into a JVM matrix job on Ubuntu and a single job on macOS

Egorand:


.github/workflows/build.yml
```diff
@@ -49,11 +49,40 @@ jobs:
...
+          if [ ! -z "$(git status --porcelain)" ]; then git diff; echo -e "\nTest files changed. Did you run ./gradlew generateTests?"; exit 1; fi
+
+      - name: Install Chrome
+        run: |
+          brew update
+          brew cask install google-chrome
+
+      - name: Test
+        run: |
+          ./gradlew -p wire-library build
```
Egorand, line 79: Can further improve this to only run JVM tests on Ubuntu to reduce the time spent on downloading dependencies for JS and Native, and run Native and JS tests on macOS only.

JakeWharton: Would be nice if this was built-in somehow, but I suspect we want to do this ourselves with properties.

Something like `./gradlew ... -Pkjs=false -Pknative=false` and then we do like
```groovy
if (getProperty("kjs") != "false")) {
  js {
    browser()
  }
}
```

That way everything is on-by-default but you can disable a platform if need-be.

Egorand: Hmm, I thought about approaching this by just invoking platform-specific tasks instead of `build`, I wonder though if there'll be something that runs during configuration that'll break the builds. I'm gonna try this first cause it won't require modifying any build files, if it fails I'll try introducing properties.

Egorand:
It seems to work well on macOS, so we can avoid running JS tests on Ubuntu instead, especially since we don't really need to run them for every JVM version.

.github/workflows/build.yml
JakeWharton: My concern with executing platform-specific tasks is that it requires constant maintenance to opt-in rather than opt-out. With the system property approach, every build is run by default and if we want to optimize CI we disable targets. This ensures that we never forget a target and that as new modules come in that they're covered.

.github/workflows/build.yml
```diff
@@ -39,6 +39,45 @@ jobs:
...
+            wire-kotlin-generator:build \
+            wire-moshi-adapter:build \
+            wire-runtime:jvmTest \
+            wire-schema:build \
+            wire-tests:jvmTest
+          ./gradlew \
+            sample:build \
+            wire-codegen-sample:build \
+            wire-grpc-tests:build \
+            wire-protoc-compatibility-tests:build
```
JakeWharton, line 59: Yeah this whole thing seems very error-prone and not future-proof.

.github/workflows/build.yml
Egorand: That's a great point, definitely see the value of opt-out vs opt-in in this case. I'll introduce properties for JS and Native; not sure how to got about JVM, but ultimately it'd be nice to avoid running tests for JVM-only modules on macOS.

wire-library/wire-tests/src/commonTest/kotlin/com/squareup/wire/IgnoreJs.kt
```diff
@@ -1,6 +1,4 @@
 package com.squareup.wire
 
-@OptionalExpectation
```
Egorand, line 3: `OptionalExpectation` doesn't work well with dynamically included targets, so better get rid of it and use no-op actuals.

Egorand:
@JakeWharton PTAL bca856f

PR #1527, opened 2020-04-28
Bring back GrpcClient.create() for JVM users

swankjesse:


wire-library/wire-grpc-client/build.gradle
```diff
@@ -35,6 +35,7 @@ kotlin {
     jvmMain {
       dependencies {
         api deps.okhttp
+        api project(':wire-runtime')
```
Egorand, line 38: Shouldn't be needed since it's already declared in `commonMain`.

wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/GrpcClient.kt
```diff
@@ -23,11 +23,39 @@ import okhttp3.HttpUrl.Companion.toHttpUrl
...
+  /** Returns a [service] that makes gRPC calls using this client. */
+  fun <T : Service> create(service: KClass<T>): T {
+    // Use reflection to find the implementing class like "routeguide.GrpcRouteGuideClient" for a
+    // generated interface like "routeguide.RouteGuideClient".
+    try {
+      val implementationClass = implementationClass(service)
+      val onlyConstructor = implementationClass.declaredConstructors.single()
+      val instance = onlyConstructor.newInstance(this)
+      return service.java.cast(instance)
+    } catch (e: Exception) {
```
Egorand, line 45: nit:
```suggestion
    } catch (_: Exception) {
```

wire-library/wire-grpc-client/build.gradle
swankjesse: Fixed. IntelliJ is still grumpy but Gradle is happy.

wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/GrpcClient.kt
swankjesse: Fixed.

wire-library/wire-grpc-client/build.gradle
Egorand: Hmmm, is that because reflection? Although I don't think we package `kotlin-reflect` with `wire-runtime`. Anyway, Kotlin reflection is tricky, some APIs are present in stdlib but aren't backed by an actual implementation, so they'll blow up at runtime unless `kotlin-reflect` is on the classpath. May not be the case here, but worth checking.

PR #1528, opened 2020-04-30
Proto3 allows optional

### LINKED ISSUE #1497: Field presence support for proto3
Google will update the proto3 syntax to support "optional" for field presence:

*  Summary
Add field presence to proto3 in a backward-compatible way by introducing the optional keyword with the same semantics as proto2.

* .proto syntax changes
Users can put `optional` on any proto3 field, with syntax and semantics identical to proto2:

```
syntax = "proto3";
message Foo {
  // Optional fields track presence.
  optional int32 a = 1;
  optional int64 b = 2;
  optional string c = 3;
  optional double d = 4;
```

Public announcement will come soon. Can we support this for proto3? 

Removing line 299 - 301 would be OK.
https://github.com/square/wire/blob/master/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/internal/parser/ProtoParser.kt#L299

oldergod:
https://github.com/protocolbuffers/protobuf/blob/master/docs/implementing_proto3_presence.md

fixes #1497

### DIFF
--- a/wire-library/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/internal/parser/ProtoParserTest.kt
+++ b/wire-library/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/internal/parser/ProtoParserTest.kt
@@ -715,21 +715,34 @@ class ProtoParserTest {
   }
 
   @Test
-  fun proto3MessageFieldsForbidOptional() {
+  fun proto3MessageFieldsAllowOptional() {
     val proto = """
         |syntax = "proto3";
         |message Message {
         |  optional string a = 1;
         |}
         """.trimMargin()
-    try {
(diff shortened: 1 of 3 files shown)


wire-library/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/internal/parser/ProtoParserTest.kt
```diff
@@ -715,21 +715,34 @@ class ProtoParserTest {
   }
 
   @Test
-  fun proto3MessageFieldsForbidOptional() {
+  fun proto3MessageFieldsAllowOptional() {
```
swankjesse, line 718: nice

PR #1529, opened 2020-04-30
Dependencies in one file only

oldergod:


wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireOutput.kt
```diff
@@ -138,7 +138,7 @@ open class KotlinOutput @Inject constructor() : WireOutput() {
       "To generate Kotlin protos, please apply a Kotlin plugin."
     }
     compileKotlinTasks.configureEach {
-      it.source(out)
+      it.source(out!!)
```
JakeWharton, line 141: ?

oldergod: 1.3.61
image

1.3.71
image

PR #1530, opened 2020-05-01
Change Location#toString to help IDE

oldergod:
`path/file:line:column` when clicked but the cursor at the right position.

### DIFF
--- a/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Location.kt
+++ b/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Location.kt
@@ -53,7 +53,8 @@ data class Location(
       append(':')
       append(line)
       if (column != -1) {
-        append(":$column")
+        append(':')
+        append(column)
       }
     }
   }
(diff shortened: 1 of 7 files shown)


wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Location.kt
```diff
@@ -50,7 +50,7 @@ data class Location(
     }
     append(path)
     if (line != -1) {
-      append(" at $line")
+      append(":$line")
```
JakeWharton, line 53: This should really be
```suggestion
      append(':')
      append(line)
```
and the same below. Otherwise you're creating a nested StringBuilder instance to create a String only to append it to the enclosing StringBuilder.

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Location.kt
```diff
@@ -50,7 +50,7 @@ data class Location(
...
       if (column != -1) {
         append(":$column")
```
JakeWharton, line 55:
```suggestion
        append(':')
        append(column)
```

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Location.kt
JakeWharton: There is an R8 feature request to optimize this case, but it's not enabled yet: https://issuetracker.google.com/issues/113859361

oldergod: I keep forgetting about this. Thank you.

PR #1531, opened 2020-05-01
Add onlyVersion to target a unique pruning version

oldergod:


wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
```diff
@@ -148,6 +148,12 @@ data class WireRun(
    */
   val until: String? = null,
 
+  /**
+   * The only version of the version range. Fields with `until` values greater than this, as well as
+   * fields with `since` values less than or equal to this, are retained.
+   */
+  val onlyVersion: String? = null,
```
JakeWharton, line 155: I don't like that this has the suffix of "Version" in the DSL whereas "since" and "until" do not. They should match.

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/PruningRules.kt
```diff
@@ -226,7 +233,18 @@ class PruningRules private constructor(builder: Builder) {
...
+    /**
+     * The only version of the version range. Fields with `until` values greater than this, as well
+     * as fields with `since` values less than or equal to this, are retained.
+     */
+    fun only(only: String?) = apply {
+      this.only = only?.toLowerCaseSemVer()
+    }
+
     fun build(): PruningRules {
+      check(only == null || since == null && until == null) {
```
JakeWharton, line 245: Would prefer clarifying parenthesis to clearly denote precedence

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
```diff
@@ -148,6 +148,12 @@ data class WireRun(
    */
   val until: String? = null,
 
+  /**
+   * The only version of the version range. Fields with `until` values greater than this, as well as
+   * fields with `since` values less than or equal to this, are retained.
```
JakeWharton, line 153: Document that it's mutually exclusive with `since`/`until` as well, please.

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
```diff
@@ -242,6 +248,7 @@ data class WireRun(
         .prune(treeShakingRubbish)
         .since(since)
         .until(until)
+        .only(onlyVersion)
```
JakeWharton, line 251: Similarly, it feels like all three of these need the version suffix.

oldergod: I think `pruningRules` can keep the 3 of them without suffix, but WireRun would have them all 3 with the suffix. Sounds good to you?

swankjesse: Works for me

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
oldergod: fixed

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/PruningRules.kt
oldergod: fixed

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
oldergod: fixed

PR #1532, opened 2020-05-04
Updating gRPC sample

oldergod:
Goal is to followin-up and make it so it builds with the whole wire project


wire-grpc-sample/build.gradle
```diff
@@ -2,31 +2,30 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
...
+          'coroutinesAndroid': "org.jetbrains.kotlinx:kotlinx-coroutines-android:1.3.5",
       ],
       'androidx': [
           'appcompat': "androidx.appcompat:appcompat:1.0.2",
           'constraintLayout': "androidx.constraintlayout:constraintlayout:1.1.3",
           'ktx': 'androidx.core:core-ktx:1.0.2',
       ],
       'wire': [
-          'runtime': "com.squareup.wire:wire-runtime-multiplatform:${versions.wire}",
+          'runtime': "com.squareup.wire:wire-runtime:${versions.wire}",
```
Egorand, line 23: FWIW `wire-runtime-multiplatform` and `wire-runtime` both work here, although `wire-runtime` needs less resolution steps.

oldergod: So the project would NOT compile with `wire-runtime-multiplatform` by the way.

oldergod: None of the Wire types would get resolved
image

Egorand: This is strange, I would expect Gradle to resolve the dependencies properly.

PR #1535, opened 2020-05-04
Treat the wire-grpc-sample as a Wire module

oldergod:


wire-grpc-sample/client/build.gradle
```diff
@@ -30,10 +30,10 @@ wire {
 }
 
 dependencies {
-  implementation project(':protos')
-  implementation deps.kotlin.stdlib
-  implementation deps.kotlin.coroutines
-  implementation deps.kotlin.coroutinesAndroid
+  implementation project(':wire-grpc-sample:protos')
+  implementation deps.kotlin.coroutines.core.common
```
Egorand, line 34: Can we use `core` transitively through `android`? Also, probably just `core`, not `core.common`.

oldergod: fixed

PR #1538, opened 2020-05-07
camelCaseJsonKeysForMoshi

oldergod:
~This would also change the printing for proto2.~ We wanna do that only for proto3.

part of #1387

### DIFF
--- a/wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
+++ b/wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
@@ -1169,7 +1169,7 @@ public final class JavaGenerator {
     }
 
     if (!field.getJsonName().equals(field.getName())) {
-      result.addMember("jsonName = %s", field.getJsonName());
+      result.addMember("jsonName", "$S", field.getJsonName());
     }
 
     return result.build();
--- a/wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
+++ b/wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
@@ -1168,6 +1168,10 @@ public final class JavaGenerator {
       result.addMember("declaredName", "$S", field.getName());
     }
 
+    if (!field.getJsonName().equals(field.getName())) {
+      result.addMember("jsonName = %s", field.getJsonName());
+    }
+
     return result.build();
   }
 
--- a/wire-library/wire-runtime/src/jsMain/kotlin/com/squareup/wire/internal/-Platform.kt
+++ b/wire-library/wire-runtime/src/jsMain/kotlin/com/squareup/wire/internal/-Platform.kt
@@ -38,7 +38,7 @@ actual fun camelCase(string: String): String {
     while (index < string.length) {
       var char = string[index]
       if (index == 0) {
-        if (char in 'A'..'Z') char = char.toLowerCase()
+        if (char in 'A'..'Z') char -= 'A' - 'a'
       }
 
       index++
@@ -48,7 +48,7 @@ actual fun camelCase(string: String): String {
         continue
       }
       if (uppercase) {
-        if (char in 'a'..'z') char = char.toUpperCase()
+        if (char in 'a'..'z') char += 'A' - 'a'
       }
       append(char)
       uppercase = false
--- a/wire-library/wire-runtime/src/jsMain/kotlin/com/squareup/wire/internal/-Platform.kt
+++ b/wire-library/wire-runtime/src/jsMain/kotlin/com/squareup/wire/internal/-Platform.kt
@@ -29,3 +29,29 @@ actual inline fun <T> MutableList<T>.toUnmodifiableList(): List<T> = this
 
 @Suppress("NOTHING_TO_INLINE") // Syntactic sugar.
 actual inline fun <K, V> MutableMap<K, V>.toUnmodifiableMap(): Map<K, V> = this
+
+// TODO: Use code points to process each char.
+actual fun camelCase(string: String): String {
+  return buildString(string.length) {
+    var index = 0
+    var uppercase = false
+    while (index < string.length) {
+      var char = string[index]
+      if (index == 0) {
+        if (char in 'A'..'Z') char = char.toLowerCase()
+      }
+
+      index++
+
+      if (char == '_') {
+        uppercase = true
+        continue
+      }
+      if (uppercase) {
+        if (char in 'a'..'z') char = char.toUpperCase()
+      }
+      append(char)
+      uppercase = false
+    }
+  }
+}
--- a/wire-library/wire-runtime/src/nativeMain/kotlin/com/squareup/wire/internal/-Platform.kt
+++ b/wire-library/wire-runtime/src/nativeMain/kotlin/com/squareup/wire/internal/-Platform.kt
@@ -38,7 +38,7 @@ actual fun camelCase(string: String): String {
     while (index < string.length) {
       var char = string[index]
       if (index == 0) {
-        if (char in 'A'..'Z') char = char.toLowerCase()
+        if (char in 'A'..'Z') char -= 'A' - 'a'
       }
 
       index++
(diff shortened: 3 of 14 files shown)


wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/CamelCase.kt
```diff
@@ -0,0 +1,42 @@
...
+// This mirrors the logic of `fieldNameToJsonName` in `com/google/protobuf/Descriptors.java`, except
+// it lower case the first character as well (protoc seems to do this step earlier).
+// See https://github.com/protocolbuffers/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/Descriptors.java
+fun camelCase(string: String): String {
+  return buildString {
+    var index = 0
+    var uppercase = false
+    while (index < string.length) {
+      var codePoint = string.codePointAt(index)
+      if (index == 0){
```
JakeWharton, line 27: nit
```suggestion
      if (index == 0) {
```

wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/CamelCase.kt
```diff
@@ -0,0 +1,42 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire
+
+// This mirrors the logic of `fieldNameToJsonName` in `com/google/protobuf/Descriptors.java`, except
+// it lower case the first character as well (protoc seems to do this step earlier).
+// See https://github.com/protocolbuffers/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/Descriptors.java
+fun camelCase(string: String): String {
+  return buildString {
```
JakeWharton, line 22: It'll never be larger so might as well
```suggestion
  return buildString(string.length) {
```

wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/CamelCase.kt
```diff
@@ -0,0 +1,42 @@
...
+    var uppercase = false
+    while (index < string.length) {
+      var codePoint = string.codePointAt(index)
+      if (index == 0){
+        if (codePoint in 'A'.toInt()..'Z'.toInt()) codePoint += 'a' - 'A'
+      }
+      index += Character.charCount(codePoint)
+      if (codePoint == '_'.toInt()) {
+        uppercase = true
+        continue
```
JakeWharton, line 33: The control flow here is super messed up. Can we make it:
```kotlin
while (..) {
  var codePoint = ..
  index += ..
  if (codePoint == "_") {
    ..
    continue
  }
  if (index == 0) {
    ..
  } else if (uppercase) {
    ..
  }
  appendCodePoint(codePoint)
  uppercase = false
}
```

oldergod: hmm I'll check but moving `index +=` at the top breaks the logic

JakeWharton: I don't see how it could. The `if` statement that precedes it doesn't affect the loop or index.

JakeWharton: Ah, nevermind. Idiot... I get it now. It's fine as-is.

wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/CamelCase.kt
oldergod: fixed

wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/CamelCase.kt
oldergod: fixed

oldergod:
I'll follow up because we don't wanna break our proto2 JSON. This will have to happen only for proto3.

wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -1168,6 +1168,10 @@ private AnnotationSpec wireFieldAnnotation(NameAllocator nameAllocator, Field fi
       result.addMember("declaredName", "$S", field.getName());
     }
 
+    if (!field.getJsonName().equals(field.getName())) {
+      result.addMember("jsonName = %s", field.getJsonName());
```
swankjesse, line 1172: I would have expected
```
       result.addMember("jsonName", "$S", field.getJsonName()); 
```

wire-library/wire-runtime/src/jsMain/kotlin/com/squareup/wire/internal/-Platform.kt
```diff
@@ -29,3 +29,29 @@ actual inline fun <T> MutableList<T>.toUnmodifiableList(): List<T> = this
...
+
+// TODO: Use code points to process each char.
+actual fun camelCase(string: String): String {
+  return buildString(string.length) {
+    var index = 0
+    var uppercase = false
+    while (index < string.length) {
+      var char = string[index]
+      if (index == 0) {
+        if (char in 'A'..'Z') char = char.toLowerCase()
```
swankjesse, line 41: I think you meant `char -= 'A' - 'a'`

wire-library/wire-runtime/src/jsMain/kotlin/com/squareup/wire/internal/-Platform.kt
```diff
@@ -29,3 +29,29 @@ actual inline fun <T> MutableList<T>.toUnmodifiableList(): List<T> = this
...
+      }
+
+      index++
+
+      if (char == '_') {
+        uppercase = true
+        continue
+      }
+      if (uppercase) {
+        if (char in 'a'..'z') char = char.toUpperCase()
```
swankjesse, line 51: I think you meant char += 'A' - 'a'

wire-library/wire-runtime/src/jsMain/kotlin/com/squareup/wire/internal/-Platform.kt
swankjesse: toLowerCase() is a broken function and should not be used.

wire-library/wire-runtime/src/nativeMain/kotlin/com/squareup/wire/internal/-Platform.kt
```diff
@@ -29,3 +29,29 @@ actual inline fun <T> MutableList<T>.toUnmodifiableList(): List<T> = this
...
+
+// TODO: Use code points to process each char.
+actual fun camelCase(string: String): String {
+  return buildString(string.length) {
+    var index = 0
+    var uppercase = false
+    while (index < string.length) {
+      var char = string[index]
+      if (index == 0) {
+        if (char in 'A'..'Z') char = char.toLowerCase()
```
swankjesse, line 41: don’t use `char.toLowerCase()`

wire-library/wire-runtime/src/nativeMain/kotlin/com/squareup/wire/internal/-Platform.kt
```diff
@@ -29,3 +29,29 @@ actual inline fun <T> MutableList<T>.toUnmodifiableList(): List<T> = this
...
+      }
+
+      index++
+
+      if (char == '_') {
+        uppercase = true
+        continue
+      }
+      if (uppercase) {
+        if (char in 'a'..'z') char = char.toUpperCase()
```
swankjesse, line 51: don’t use `char.toUpperCase()`

wire-library/wire-runtime/src/nativeMain/kotlin/com/squareup/wire/internal/-Platform.kt
oldergod: You might have written a post about this, and it also talked about Locale... maybe

wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
oldergod: Yes, it's fixed in follow up!

oldergod: but fixed now

PR #1539, opened 2020-05-07
Skips proto3 files in SchemaLoader (WireCompiler)

oldergod:


wire-library/wire-schema/src/jvmMain/kotlin/com/squareup/wire/schema/SchemaLoader.kt
```diff
@@ -150,14 +151,27 @@ class SchemaLoader {
...
+      loaded[proto] = protoFile
 
       // Queue dependencies to be loaded.
       for (importPath in element.imports) {
         protos.addLast(importPath)
       }
     }
 
+    if (skippedForSyntax.isNotEmpty()) {
+      println("""Skipped .proto files with unsupported syntax. Add this line to fix:
```
JakeWharton, line 169: I don't think this class should directly print.

oldergod: It's a best effort to avoid crashing. I can relook tomorrow if I can come up with a good cheap idea.

JakeWharton: We have no problem _parsing_ proto3. I'd rather see the callers fail if possible by looking into the schema. There will be duplication, but the problem is codegen and not parsing.

oldergod:
#1546 is taking over.

PR #1542, opened 2020-05-08
Correct some of our test protos

oldergod:
I tested them all via `protoc` and that's the result of what had to be fixed.

- Enum used in Maps needs a 0 tag value.
- custom options need parentheses.

### DIFF
--- a/wire-library/wire-tests/src/commonTest/proto/java/samebasename/single_level.proto
+++ b/wire-library/wire-tests/src/commonTest/proto/java/samebasename/single_level.proto
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package single_level;
+package samebasename.single_level;
 
 option java_package = "com.squareup.wire.protos.single_level";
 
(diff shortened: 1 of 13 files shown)


wire-library/wire-tests/src/commonTest/proto/java/samebasename/single_level.proto
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package single_level;
+package samebasename.single_level;
```
oldergod, line 16: btw two files with same name + same java_package will crash on protoc because protoc will want to generate outer classes for each and they'll have the same name (even though each file might have different messages.)

oldergod: there is another `single_level.proto` one directory below.

oldergod: Related to #1544

PR #1543, opened 2020-05-08
NewSchemaLoader from WireCompiler

oldergod:
Second round....


wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/MessageType.kt
```diff
@@ -98,7 +98,15 @@ class MessageType private constructor(
...
 
   fun addExtensionFields(fields: List<Field>) {
-    extensionFields.addAll(fields)
+    fields.forEach { field ->
+      if (field !in extensionFields) {
+        // We keep ordering.
+        extensionFields.remove(field)
+        extensionFields.add(field)
+      } else {
+        println("We're duplicating extension fields: $field to $this")
```
JakeWharton, line 107: Remove or throw. Printing is not reasonable from this module.

oldergod:
#1546 is taking over.

PR #1546, opened 2020-05-13
Migrate WireCompiler to use WireRun

swankjesse:
This deletes the old SchemaLoader class and so we finally only have
one again.


oldergod:
You can delete `ProfileLoaderTest.kt` and `ProfileLoader.kt`

PR #1553, opened 2020-05-18
Gson support

morkovkin:
Add gson support. (For Serializing/Deserializing use jsonName) 
add Serializing/Deserializing tests

part of #1387


oldergod:
Thank you! There are a few missing pieces in the PR which makes it incomplete (See #1538).
We intend to support proto3 for Gson as well, part of #1387 but this is going to happen once we finished proto3 support for Moshi.

morkovkin:
Hello Thank you for this project! Sorry i do't understand what im missing? this pr about naming of proto generated class (like some_field) to json ( camelCase ). Yah proto3 for Gson is working but when you try to run test GsonAdapterNameMappingTest from pr add fields will be null. Im see #1387 and don't find functional like in this pr. Or a question in the design pr? i try to find rules about it but can't find.. sorry about it.

oldergod:
fixed in #1616

PR #1556, opened 2020-05-20
Proto2 roundtrip tests with protoc

oldergod:
Safeguard against possible side-effect we might unintentionally cause when adding support to proto3.

~Wasn't able to check extensions because of https://github.com/protocolbuffers/protobuf/issues/7534~ used Registry.

### DIFF
--- a/wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/Proto2WireProtocCompatibilityTests.kt
+++ b/wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/Proto2WireProtocCompatibilityTests.kt
@@ -15,6 +15,7 @@
  */
 package com.squareup.wire
 
+import com.google.protobuf.ExtensionRegistry
 import com.squareup.wire.proto2.simple.SimpleMessage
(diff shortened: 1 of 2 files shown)


oldergod:
I guess we should want that for Java as well.

wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/Proto2WireProtocCompatibilityTests.kt
```diff
@@ -53,4 +60,306 @@ class Proto2WireProtocCompatibilityTests {
...
+
+    assertThat(AllTypes.ADAPTER.decode(byteArrayProtoc)).isEqualTo(identityAllTypesWire)
+
+    assertThat(AllTypesOuterClass.AllTypes.parseFrom(byteArrayWire, allTypesRegistry))
+        .isEqualTo(identityAllTypesProtoc)
+  }
+
+  companion object {
+    private val allTypesRegistry = ExtensionRegistry.newInstance().apply {
+    add(extOptBool)
```
swankjesse, line 85: nit: ktlint

PR #1557, opened 2020-05-20
Expose encode mode via `WireField#encodeMode`

oldergod:
The diff is overwhelming but split in 3 commits.
1- Refactoring oneOfs, they are `NULL_IF_ABSENT`.
2- Adding the field to `WireField`
3- generating tests.

### DIFF
--- a/wire-grpc-tests/src/test/proto-grpc/routeguide/FeatureDatabase.kt
+++ b/wire-grpc-tests/src/test/proto-grpc/routeguide/FeatureDatabase.kt
@@ -28,7 +28,8 @@ class FeatureDatabase(
   @field:WireField(
     tag = 1,
     adapter = "routeguide.Feature#ADAPTER",
-    label = WireField.Label.REPEATED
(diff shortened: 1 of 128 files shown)


JakeWharton:
Why can't this be inferred from the type and label? Could it be if we knew the syntax version? Maybe we should annotate types with that.

oldergod:
`type + label + syntax` is enough for most, not for oneOfs though. OneOf fields don't have any label to them, as regular field in proto3, so we need something more to know to treat it as "null if absent" (oneOf), or "identity if absent" (regular proto3 member).
We could cheat and set `WireField.Label.ONE_OF` to those fields even though the schema doesn't have that label. In that case, adding the syntax would be enough yes.

with EncodeMode, we'd have only `type + encodeMode` and this should cover all cases. We would deprecate `WireField#label` and maybe kill it later.

swankjesse:
One vote for setting the label to ONE_OF

Issue #1515, opened 2020-04-22
Wire runtime issue with instant execution at execution time


stephanenicolas:
After solving: #1496, there was another bug preventing me from testing the full build:

This bug is solved on the nightly version of gradle:https://issuetracker.google.com/issues/148400854

Using both the snapshot with the previous fix and the gradle nightly:`6.4-20200421230107+0000` and `AGP 4.0 beta 04`, I got a new stacktrace at execution time for instant execution:

```
(note to self: `failing a proto generation task`...)
Caused by: java.lang.NullPointerException
        at com.squareup.wire.gradle.WireInput.toLocations(WireInput.kt:127)
        at com.squareup.wire.gradle.WireTask.generateWireFiles(WireTask.kt:101)
        at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:104)
```

I will try to address the issue soon.

Egorand:
Nice, thanks!

stephanenicolas:
I can repro using 
* AGP:4.1.0-alpha09
* Kotlin:1.3.72
* Gradle:6.5-20200519143217+0000
```
6 instant execution problems were found, 3 of which seem unique.
- field 'configuration' from type 'com.squareup.wire.gradle.WireInput': cannot deserialize object of type 'org.gradle.api.artifacts.Configuration' as these are not supported with instant execution.
  See https://docs.gradle.org/6.5-20200519143217+0000/userguide/configuration_cache.html#disallowed_types
- field 'configuration' from type 'com.squareup.wire.gradle.WireInput': value 'file collection' is not assignable to 'org.gradle.api.artifacts.Configuration'
- field 'project' from type 'com.squareup.wire.gradle.WireInput': cannot deserialize object of type 'org.gradle.api.Project' as these are not supported with instant execution.
  See https://docs.gradle.org/6.5-20200519143217+0000/userguide/configuration_cache.html#disallowed_types
```

oldergod:
Is it fixed then @stephanenicolas ?

stephanenicolas:
Yes. Sorry, I forgot to close it.

PR #1560, opened 2020-05-22
Change WireTask inputs to a serializable type

stephanenicolas:
Solves issue #1515.

### DIFF
--- a/wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
+++ b/wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
@@ -78,7 +78,7 @@ class WirePlugin : Plugin<Project> {
     project: Project,
     extension: WireExtension
   ) {
-    val sourceInput = WireInput(project.configurations.getByName("protoSource"))
+    val sourceInput = WireInput(project.configurations.named("protoSource"))
     if (extension.sourcePaths.isNotEmpty() ||
         extension.sourceTrees.isNotEmpty() ||
         extension.sourceJars.isNotEmpty()) {
@@ -89,7 +89,7 @@ class WirePlugin : Plugin<Project> {
       sourceInput.addPaths(project, setOf("src/main/proto"))
     }
 
-    val protoInput = WireInput(project.configurations.getByName("protoPath"))
+    val protoInput = WireInput(project.configurations.named("protoPath"))
     if (extension.protoPaths.isNotEmpty() ||
         extension.protoTrees.isNotEmpty() ||
         extension.protoJars.isNotEmpty()) {
@@ -122,8 +122,8 @@ class WirePlugin : Plugin<Project> {
       }
       it.outputDirectories = outputs.map { output -> File(output.out!!) }
       it.source(sourceInput.configuration)
-      it.sourceInput = sourceInput.toLocations()
-      it.protoInput = protoInput.toLocations()
+      it.sourceInput.set(sourceInput.toLocations())
+      it.protoInput.set(protoInput.toLocations())
       it.roots = extension.roots.toList()
       it.prunes = extension.prunes.toList()
       it.sinceVersion = extension.sinceVersion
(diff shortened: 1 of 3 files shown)


wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireTask.kt
```diff
@@ -34,10 +35,10 @@ open class WireTask : SourceTask() {
   var pluginVersion: String = VERSION
 
   @get:Internal
-  internal lateinit var sourceInput: WireInput
+  internal lateinit var sourceInput: List<Location>
```
stephanenicolas, line 38: This is the main change: WireInput cannot be serialized during instant execution (a.k.a config caching) because it has a configuration which is not serializable.
The workaround is to use the ` Locations` data class which seem to be handled well by Gradle. I faced an another bug that prevented me to make sure it works 100%.

wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -114,13 +114,16 @@ class WirePlugin : Plugin<Project> {
...
     wireTask.configure {
+      if (it.logger.isDebugEnabled) {
+        sourceInput.debug(it.logger)
+        protoInput.debug(it.logger)
+      }
       it.outputDirectories = outputs.map { output -> File(output.out!!) }
       it.source(sourceInput.configuration)
-      it.sourceInput = sourceInput
-      it.protoInput = protoInput
+      it.sourceInput = sourceInput.toLocations()
```
stephanenicolas, line 125: This means the `toLocations` method is now called during configuration. Before, it was run during an `afterEvaluate` block which is pretty much the same. There might be a better way to organize things so that the dependency resolution would be performed at execution time, as it should but I don't know the plugin enough to refactor it this way.

eskatos: You could try make `WireTask.sourceInput` a `ListProperty<Location>` and wire a `Provider<Location>` to it here. Maybe changing `WireInput.toLocation()` to directly return the provider. This would defer dependency resolution to the task execution on a vintage build, and to the task graph calculation on a configuration-cache enabled build.

stephanenicolas: I agree that it would be better, and I thought about using properties but all fields indeed should use lazy properties when it applies and it does for most of them. 
That would be a nice refresh for the wire plugin. Can it be done in a different PR though @eskatos ? My goal is not really to modernize the plugin but to have instant execution working.

oldergod:
I think we don't wanna make `toLocations` on configuration. I wonder if we could just pass a list of `string`, then within the Plugin, do something like `File(string).toLocation`

stephanenicolas:
@swankjesse @eskatos , I adopted the solution of @eskatos after talking with Rodrigo from Gradle. I didn't know that providers are evaluated after the configuration of the task, when configurations are resolved and the tasks are serialized. 

So,in this case, @oldergod , we don't resolve configurations (via the call to `toLocations()`) during configuration, but right after. When the task is serialized, that's where gradle wants to resolve configuration and we use this information in the plugin exactly at this moment to provide the locations.

@swankjesse I am also wondering if we shouldn't change the `@Internal` annotation for `@Input`, and maybe apply some kind of normalizer to help the up-to-date check / caching.

stephanenicolas:
Note that I didn't have time yet to make sure it works with the current 6.5 rc 1 of gradle and instant execution. I will try tonight.

stephanenicolas:
I could test and make sure it works:
* gradle nightly `6.6-20200603220033+0000`
* AGP 4.1.0-alpha09
* KGP 1.3.72
* plus this PR for `3.3.0-SNAPSHOT` 

And everything works perfectly :D

oldergod:
Looks good to me. @jrodbx could you give a look?

jrodbx:
LGTM

oldergod:
Thank you Stéphane

PR #1561, opened 2020-05-26
Add IDENTITY_IF_ABSENT label

oldergod:
I don't think we need the ONE_OF value for oneOfs. Right now, they have the `OPTIONAL` label (which is a lie) but they are treated as optional fields in both proto2 and 3 so I am leaving them as is. Let me know if you wanna set their label to `ONE_OF`.

### DIFF
--- a/wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
+++ b/wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
@@ -1151,12 +1151,28 @@ public final class JavaGenerator {
       result.addMember("adapter", "$S", adapterString(field.getType()));
     }
 
-    if (!field.isOptional()) {
-      if (field.isPacked()) {
-        result.addMember("label", "$T.PACKED", WireField.Label.class);
-      } else if (field.getLabel() != null) {
-        result.addMember("label", "$T.$L", WireField.Label.class, field.getLabel());
-      }
+    WireField.Label wireFieldLabel;
+    //noinspection ConstantConditions
+    switch (field.getEncodeMode()) {
+      case THROW_IF_ABSENT:
+        wireFieldLabel = WireField.Label.REQUIRED;
+        break;
+      case IDENTITY_IF_ABSENT:
+        wireFieldLabel = WireField.Label.IDENTITY_IF_ABSENT;
+        break;
+      case REPEATED:
+        wireFieldLabel = WireField.Label.REPEATED;
+        break;
+      case PACKED:
+        wireFieldLabel = WireField.Label.PACKED;
+        break;
+      case MAP:
+      case NULL_IF_ABSENT:
+      default:
+        wireFieldLabel = null;
+    }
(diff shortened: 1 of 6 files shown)


wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/WireField.kt
```diff
@@ -66,7 +66,14 @@ annotation class WireField(
...
     ONE_OF,
     /** Implies [REPEATED]. */
-    PACKED;
+    PACKED,
+    /**
+     * Special label to defines proto3 fields which should not be emitted if their value is equal
+     * to their type's respective identity value. E.g.: a field of type `int32` will not get emitted
+     * if its value is `0`.
+     */
+    IDENTITY_IF_ABSENT,
```
JakeWharton, line 75: The doc describes "absent if identity" which is a property of encoding and not "identity if absent" which would be a property of decoding. Should this be named `ABSENT_IF_IDENTITY`?

oldergod: This actually go both ways, right. I think all those labels can be read for both encoding and decoding. Maybe just `IDENTITY`?

JakeWharton: Ah, I see.

`IGNORE_IDENTITY`? `OMIT_IDENTITY`?

wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -1151,12 +1151,28 @@ private AnnotationSpec wireFieldAnnotation(NameAllocator nameAllocator, Field fi
...
-    if (!field.isOptional()) {
-      if (field.isPacked()) {
-        result.addMember("label", "$T.PACKED", WireField.Label.class);
-      } else if (field.getLabel() != null) {
-        result.addMember("label", "$T.$L", WireField.Label.class, field.getLabel());
-      }
+    WireField.Label wireFieldLabel;
+    //noinspection ConstantConditions
+    switch (field.getEncodeMode()) {
+      case THROW_IF_ABSENT:
```
swankjesse, line 1157: should we rename `THROW_IF_ABSENT` to `REQUIRED` ? Fewer names is better

wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/WireField.kt
swankjesse: I like `OMIT_IDENTITY`

wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
oldergod: fixed

oldergod:
I went with `OMIT_IDENTITY`

PR #1568, opened 2020-05-28
Cleaning up Field's API

oldergod:


wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Field.kt
```diff
@@ -60,11 +59,6 @@ class Field private constructor(
   val isRepeated: Boolean
     get() = label == Label.REPEATED
 
-  @Deprecated("Proto2 concept inexistent in proto3. Use EncodeMode instead.")
-  val isOptional: Boolean
-    get() = label == Label.OPTIONAL
-
-  @Deprecated("Proto2 concept inexistent in proto3. Use EncodeMode instead.")
   val isRequired: Boolean
     get() = encodeMode == EncodeMode.REQUIRED
```
oldergod, line 63: or I could inline this one? whatever.

PR #1572, opened 2020-05-29
add OmitValueJsonAdapter to omit identities

oldergod:
- What do we want to do about list? Leave it to the serializer, or explicitly omit empty list/maps?
- `uint64` needs an identity version as well.

### DIFF
--- a/wire-library/wire-moshi-adapter/src/main/java/com/squareup/wire/OmitValueJsonAdapter.kt
+++ b/wire-library/wire-moshi-adapter/src/main/java/com/squareup/wire/OmitValueJsonAdapter.kt
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2020 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


wire-library/wire-moshi-adapter/src/main/java/com/squareup/wire/OmitValueJsonAdapter.kt
```diff
@@ -0,0 +1,55 @@
...
+  }
+}
+
+/**
+ * Returns a JSON adapter equal to this JSON adapter, except this will not emit if the value is
+ * equal to [omittedValue].
+ */
+internal fun <T : Any> JsonAdapter<T>.omitValue(omittedValue: T): JsonAdapter<T> {
+  return if (this is OmitValueJsonAdapter<*>) {
+    this
```
JakeWharton, line 51: Throw or just allow re-wrap. This behavior as-is is wrong for `omit(0).omit(1)` when you write a 1, though.

oldergod: Went with
```kotlin
internal fun <T : Any> JsonAdapter<T>.omitValue(omittedValue: T): JsonAdapter<T> {
  return if (this is OmitValueJsonAdapter<*>) {
    throw IllegalArgumentException("JsonAdapter $this is already of type OmitValueJsonAdapter")
  } else {
    OmitValueJsonAdapter(this, omittedValue)
  }
}
```

oldergod:
Needs to merge #1570 first.

PR #1573, opened 2020-05-29
Add failing test for all64s

oldergod:
We need a way to say that's proto3, that's proto2; otherwise we'll end up breaking our JSON behavior for proto2. The problem doesn't show for OMIT_IDENTITY but for list/map/oneof where we cannot tell which syntax it is.

### DIFF
--- a/wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/Proto3WireProtocCompatibilityTests.kt
+++ b/wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/Proto3WireProtocCompatibilityTests.kt
@@ -32,6 +32,8 @@ import org.assertj.core.api.Assertions.assertThat
 import org.junit.Assert.fail
 import org.junit.Ignore
 import org.junit.Test
+import squareup.proto3.alltypes.All64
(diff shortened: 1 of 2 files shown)


wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/Proto3WireProtocCompatibilityTests.kt
```diff
@@ -566,10 +629,35 @@ class Proto3WireProtocCompatibilityTests {
...
+        |  }
+        |}""".trimMargin()
+
     private const val IDENTITY_ALL_TYPES_JSON = "{}"
 
     private fun <T : kotlin.Any> list(t: T): List<T> {
       return listOf(t, t)
     }
+
+    private const val BIG_POSITIVE_NUMBER = -9223372036854775807L - 1L
```
JakeWharton, line 661: Isn't this `Long.MIN_VALUE`? Maybe we could just inline that rather than have this (somewhat misleading) named constant.

oldergod: Thank you. Fixed.

PR #1578, opened 2020-06-02
Setup Duration actual and expected + its proto adapter

oldergod:
@swankjesse the only thing I took from your PR #1516 is the proto adapter I defined in `commonMain` with updated getters.

### DIFF
--- a/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/Duration.kt
+++ b/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/Duration.kt
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2020 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire
+
+expect class Duration {
+  fun getSeconds(): Long
+  fun getNano(): Int
+}
+
+expect fun durationOfSeconds(seconds: Long, nanoAdjustment: Long): Duration
--- a/wire-library/wire-runtime/src/jsMain/kotlin/com/squareup/wire/Duration.kt
+++ b/wire-library/wire-runtime/src/jsMain/kotlin/com/squareup/wire/Duration.kt
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2020 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/Duration.kt
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+package com.squareup.wire
+
+expect class Duration {
+  fun getSeconds(): Long
+  fun getNano(): Int
+}
+
+expect fun durationOfSeconds(seconds: Long, nanoAdjustment: Long): Duration
```
JakeWharton, line 23: nanos should be `Int` not `Long`.

Also, this seems like it wants to be
```kotlin
fun Duration(seconds: Long, nanoAdjustment: Int)
```
so we get constructor-like behavior. Because it's not really "of seconds" since we can specify nanos as well. Plus maybe we specify nanos to be 0 by default?

JakeWharton: Ah crap the Java method is named "of seconds"...

wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/Duration.kt
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire
+
+actual typealias Duration = java.time.Duration
+
+actual fun durationOfSeconds(seconds: Long, nanoAdjustment: Long): Duration {
```
JakeWharton, line 20: nit:
```suggestion
actual inline fun durationOfSeconds(seconds: Long, nanoAdjustment: Long): Duration {
```

oldergod:
Builds will fail because it's missing actual adapters from the other PR.

wire-library/wire-runtime/src/jsMain/kotlin/com/squareup/wire/Duration.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.squareup.wire.internal.NANOS_PER_SECOND
+import com.squareup.wire.internal.addExactLong
+import com.squareup.wire.internal.floorDivLong
+import com.squareup.wire.internal.floorModLong
+
+actual class Duration internal constructor(
+  private val seconds: Long,
+  private val nanos: Int
+) {
+  actual fun getSeconds(): Long = seconds
```
JakeWharton, line 27: Ugh. Making the API worse just to satisfy the JVM.

wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/Duration.kt
oldergod: and `ofSeconds` takes both as longs :O

JakeWharton: Weird. I suspect that's because they want to be friendly to people using `nanoTime()` and doing math? What happens if you exceed a second's worth of nanos, I wonder.

oldergod:
Used in #1579

PR #1580, opened 2020-06-04
Make map adapters lazy

### LINKED ISSUE #1545: map with a recursive value cannot be initialized.
Example proto:

```
message ModelEvaluation {
  optional string name = 1;
  optional double score = 2;
  map<string, double> signals = 3;
  map<string, ModelEvaluation> models = 4;
}
```

When trying to initialize a `ModelEvaluation` with an empty map, an error is thrown, so a recursive map can not be used.
```
 ModelEvaluation.Builder()
  .name(name)  
   .score(score)
   .signals(mapOf()) // works
  .models(mapOf()) // crashes
  .build()
```

oldergod:
fixes #1545


wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -284,7 +284,7 @@ public TypeName typeName(ProtoType protoType) {
 
   private CodeBlock singleAdapterFor(Field field, NameAllocator nameAllocator) {
     return field.getType().isMap()
-        ? CodeBlock.of("$N", nameAllocator.get(field))
+        ? CodeBlock.of("$NAdapter()", nameAllocator.get(field))
```
JakeWharton, line 287: Any reason for the suffix? You can name methods the same as fields.

wire-library/wire-tests/src/jvmKotlinInteropTest/proto-kotlin/com/squareup/wire/protos/kotlin/map/Mappy.kt
```diff
@@ -83,8 +85,8 @@ class Mappy(
       Mappy::class, 
       "type.googleapis.com/com.squareup.wire.protos.kotlin.map.Mappy"
     ) {
-      private val thingsAdapter: ProtoAdapter<Map<String, Thing>> =
-          ProtoAdapter.newMapAdapter(ProtoAdapter.STRING, Thing.ADAPTER)
+      private val thingsAdapter: ProtoAdapter<Map<String, Thing>> by lazy(PUBLICATION) {
```
JakeWharton, line 88: I would prefer the same approach be used as Java. This requires a lock, an allocation, and causes creation of `KProperty` objects and a `KProperty[]` array.

wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
oldergod: Doesn't it read better in the generated code? The adapter field named after its type's name is I think more confusing.

wire-library/wire-tests/src/jvmKotlinInteropTest/proto-kotlin/com/squareup/wire/protos/kotlin/map/Mappy.kt
oldergod: I've addressed.

PR #1583, opened 2020-06-05
Crash when generated files conflict.

### LINKED ISSUE #1544: Wire should fail if multiple messages would overwrite one another
Consider two proto files in the same directory:
au.proto
```
package one
option java_package = "same.package"
message Owner {}
```
and jp.proto
```
package two
option java_package = "same.package"
message Owner {}
```
Running wire on these would result in same.package.Owner being written then overwritten. It should fail so as to not let folks think it generated successfully.

oldergod:
fixes #1544 

If this looks fine, I could add it for services.


wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -77,12 +78,25 @@ sealed class Target : Serializable {
...
     ) {
       protoFile.types
           .filter { it !in claimedDefinitions && emittingRules.includes(it.type) }
           .forEach { type ->
-            handle(type)
+            val generatedFilePath = handle(type)
+
+            if (generatedFilePath != null) {
+              if (generatedFilePath in claimedPaths.keys) {
+                val other = claimedPaths[generatedFilePath]!!
```
JakeWharton, line 91: You shouldn't call `contains` and then `get` on a map. Just call `get` and then check if the result is non-null. Otherwise you pay the hashing and traversal cost twice. Here you're actually paying it twice in both branches because the other codepath then inserts an entry. This can all be consolidated into a single interaction with the map:
```kotlin
claimedPaths.putIfAbsent(generatedFilePath, type)?.let { other ->
  throw ISE(...)
}
```

wire-library/wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt
```diff
@@ -627,6 +627,32 @@ class WireRunTest {
         .contains("class Orange")
   }
 
+  @Test fun crashWhenGenerationConflicts() {
```
swankjesse, line 630: 👍🏻

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -66,7 +67,7 @@ sealed class Target : Serializable {
   ): SchemaHandler
 
   interface SchemaHandler {
-    fun handle(type: Type)
+    fun handle(type: Type): Path?
```
swankjesse, line 70: Document what the return value is

oldergod:
Added code to assess conflicting services as well. I didn't check if a type would conflict with a service because it wouldn't make sense.

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
oldergod: fixed

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
swankjesse: Nice, thank you!

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -77,20 +81,42 @@ sealed class Target : Serializable {
       protoFile: ProtoFile,
       emittingRules: EmittingRules,
       claimedDefinitions: ClaimedDefinitions,
+      claimedTypePaths: MutableMap<Path, Type>,
```
swankjesse, line 84: single map so they can’t mutually conflict?

oldergod: I am not sure it makes sense but I'll try anyway. `MutableMap<Path, Any>` would be gross though.

oldergod: fixed! not so gross afterall

pforhan:
Nice work, thanks!

PR #1584, opened 2020-06-05
Remove Maven references from docs

jrodbx:
Still need to update wire-compiler command line options and maybe fix/re-add wire-compiler-with-dependencies, but for now, let's stop mentioning the since-removed Maven plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -501,44 +501,15 @@ Since version 3.0.0, Wire supports [gRPC][grpc_docs].
 Generating Code With Wire
 -------------------------
 
-Wire's compiler is available via a Maven plugin. Put `.proto` sources in your project's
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -565,16 +536,6 @@ Get Wire
 The `wire-runtime` package contains runtime support libraries that must be included in applications
 that use Wire-generated code.
 
-With Maven:
```
Egorand, line 568: This should still be valid.

jrodbx: True.  Should we even mention it though?

Egorand: Don't have a strong opinion.

oldergod: Let's keep it IMO

PR #1585, opened 2020-06-08
Add syntax to ProtoAdapter

oldergod:
We need it to know in JSON how to serialize some fields. We can use it as well to know when to not emit empty list/maps.

- I don't feel good adding it to `commonBool`, `commonInt`, etc.; this should be 1) `null` (and say that only Message/Enum can have it?) or 2) the value of their enclosing type, but seems difficult to do so.
- I'm using a string but we could obviously use a `class enum Syntax`. We have one in `ProtoFile.Syntax` which we could duplicate or move. I think I'm gonna move it once we feel happy with the overall implementation path.


oldergod:
Will reopen a fresh one.

PR #1586, opened 2020-06-10
Encoding and decoding of struct protos

swankjesse:
We use Map, List, String, Double, Boolean and Null instead of our own
competing runtime types.


swankjesse:
For follow-ups:

 * Builders should canonicalize values. Numbers should be converted to Doubles, lists and maps should be immutable, other types rejected.
 * JSON encoding and decoding

PR #1588, opened 2020-06-12
JsonAdapter for Struct

oldergod:


wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -1494,6 +1495,10 @@ class KotlinGenerator private constructor(
 
   private val Field.typeName: TypeName
     get() {
+      if (type == ProtoType.STRUCT_MAP) return type!!.typeName
+      if (type == ProtoType.STRUCT_LIST) return type!!.typeName
+      if (type == ProtoType.STRUCT_VALUE) return type!!.typeName.copy(nullable = true)
+      if (type == ProtoType.STRUCT_NULL) return type!!.typeName.copy(nullable = true)
```
oldergod, line 1501: Maybe we could change their encodeMode?
Something like `STRUCT_MAP` -> `MAP`, `STRUCT_LIST` -> `REPEATED` (not packed? can it be packed?), then `STRUCT_VALUE` and `STRUCT_NULL` would be `OMIT_IDENTITY`

wire-library/wire-moshi-adapter/src/main/java/com/squareup/wire/StructJsonAdapter.kt
```diff
@@ -0,0 +1,75 @@
...
+    when (any) {
+      null -> out.nullValue()
+      is Number -> out.value(any)
+      is String -> out.value(any)
+      is Boolean -> out.value(any)
+      is Map<*, *> -> {
+        if (any.isEmpty()) {
+          skip(out)
+          return
+        }
```
oldergod, line 37: Just mimicking what protoc is doing here, same for list. It helps with the tests for protoc/wire round trip but I don't really care if we print it or not.

wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
JakeWharton: Does that also eliminate the special casing for identity value and accepts null below? Because if so, i'm supportive.

wire-library/wire-moshi-adapter/src/main/java/com/squareup/wire/StructJsonAdapter.kt
JakeWharton: This doesn't appear to be tested though

oldergod: Without it, some tests in `Proto3WireProtocCompatibilityTests` will fail now that we have a Struct adapter.

oldergod: I'll add one explicit test case in StructTests

oldergod: done

wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
oldergod: will follow up with that

wire-library/wire-moshi-adapter/src/main/java/com/squareup/wire/StructJsonAdapter.kt
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Json adapter for structs, list values, values, or null values as defined in
+ * `google/protobuf/struct.proto`.
+ */
+internal object StructJsonAdapter : JsonAdapter<Any>() {
+  override fun toJson(out: JsonWriter, any: Any?) {
+    when (any) {
+      null -> out.nullValue()
+      is Number -> out.value(any)
```
swankjesse, line 30: Double

wire-library/wire-moshi-adapter/src/main/java/com/squareup/wire/StructJsonAdapter.kt
```diff
@@ -0,0 +1,75 @@
...
+      is String -> out.value(any)
+      is Boolean -> out.value(any)
+      is Map<*, *> -> {
+        if (any.isEmpty()) {
+          skip(out)
```
swankjesse, line 35: I don’t like this name. Maybe `out.skippedNullValue()` ?

wire-library/wire-moshi-adapter/src/main/java/com/squareup/wire/StructJsonAdapter.kt
```diff
@@ -0,0 +1,75 @@
...
+
+  override fun fromJson(input: JsonReader): Any? {
+    return input.readJsonValue()
+  }
+
+  private fun skip(out: JsonWriter) {
+    val serializeNulls: Boolean = out.serializeNulls
+    out.serializeNulls = false
+    try {
+      out.nullValue()
```
swankjesse, line 70: This is wrong. If the List is a member of another List, you want `[]`, not `null`.

wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/StructTest.kt
```diff
@@ -311,4 +316,68 @@ class StructTest {
...
+        value_c = emptyList<Any>(),
+        value_d = emptyMap<String, Any>()
+    )
+
+    val moshi = Moshi.Builder().add(WireJsonAdapterFactory()).build()
+    val allStructAdapter = moshi.adapter(AllStructs::class.java)
+    assertJsonEquals(allStructAdapter.toJson(wireAllStruct), json)
+    assertThat(allStructAdapter.fromJson(json))
+        // Parsed non-printed map and list are null now.
+        .isEqualTo(wireAllStruct.copy(value_c = null, value_d = null))
```
swankjesse, line 381: ew

swankjesse: What does protoc do?

oldergod: This was a bad interpretation from my part. Ignore please.

wire-library/wire-moshi-adapter/src/main/java/com/squareup/wire/StructJsonAdapter.kt
oldergod: You're correct, thank you. I fixed it in 8bec13f

PR #1594, opened 2020-06-13
Check the grpc-status for an early failure

swankjesse:


wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/grpc.kt
```diff
@@ -134,22 +134,25 @@ internal fun <S : Any> ReceiveChannel<S>.writeToRequestBody(
...
+  if (grpcStatus != null && grpcStatus != "0") {
+    throw IOException("grpc failed: status=${code}, grpc-status=$grpcStatus")
+  }
 
-  return false
+  val contentType = body!!.contentType()
+  if (code != 200 ||
+      contentType == null ||
+      contentType.type != "application" ||
+      contentType.subtype != "grpc" && contentType.subtype != "grpc+proto") {
```
oldergod, line 153: remembering using the two isn't fun. Would a extension function help?

swankjesse: I don't expect it to be necessary beyond this file.

PR #1596, opened 2020-06-14
Encoding and decoding of empty.proto

oldergod:
Not sure yet if we want an `object Empty` or if we should write it all


wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/Empty.kt
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire
+
+/**
+ * A generic empty message that you can re-use to avoid defining duplicated empty messages in your
+ * APIs.
+ */
+object Empty
```
JakeWharton, line 22: If this doesn't have to be a subtype of `Message` can we use `kotlin.Unit`?

oldergod: We can!

oldergod: fixed

PR #1600, opened 2020-06-18
Make StructMap and StructList nullable

oldergod:
`NullValue` is still a special case because it's an ENUM but the others are regular messages and treating them as so is cool.


oldergod:
This isn't what I understood from our discussion.
They represent a Json object right, and those two objects are not the same:
```json
{
  "struct": null
}
```
vs
```json
{
  "struct": {}
}
```

swankjesse:
Yes, but the way I think we represent a nullable thing in a proto schema declaration is as a google.protobuf.Value (capable of representing any JSON thing including object and null), not as a google.protobuf.Struct (capable of representing objects only).

swankjesse:
I did some thinking on this. Because not supporting null is a divergence from the proto spec, we should have a very good reason to not support null. If this feature was widely useful, then that would be our very good reason. But structs are probably going to be a bit of a niche use case, and adding some `!!` or `?: emptyMap()` calls there is of minimal cost to the developer.

PR #1607, opened 2020-06-27
Omit enum 0 values when WireEnum is OmitIdentity

oldergod:


wire-library/wire-moshi-adapter/src/main/java/com/squareup/wire/OmitWireEnumValueJsonAdapter.kt
```diff
@@ -0,0 +1,55 @@
...
+ * limitations under the License.
+ */
+package com.squareup.wire
+
+import com.squareup.moshi.JsonAdapter
+import com.squareup.moshi.JsonReader
+import com.squareup.moshi.JsonWriter
+import java.io.IOException
+
+private class OmitWireEnumValueJsonAdapter<T : WireEnum>(
```
swankjesse, line 23: Rather than composing Moshi’s built-in enum adapter would you consider just having a WireEnumJsonAdapter? That way we get the behavior we want, regardless of what Moshi’s default does. In particular, the default one is configurable, and I don’t like thinking about what happens if it’s configured to do something weird, like lowercase enum constants.

PR #1610, opened 2020-06-30
read next Long for SInt64 adapter instead of delegating to uint64Adapter

oldergod:


wire-library/wire-moshi-adapter/src/main/java/com/squareup/wire/WireJsonAdapterFactory.kt
```diff
@@ -190,10 +190,11 @@ class WireJsonAdapterFactory private constructor(
       @Throws(IOException::class)
       override fun fromJson(reader: JsonReader): Long? {
         val bigInteger = BigInteger(reader.nextString())
-        return if (bigInteger > maxLong)
+        return if (bigInteger > maxLong) {
```
swankjesse, line 193: ty

PR #1611, opened 2020-07-01
Load JSON from files in wire-tests to allow sharing among multiple modules.

oldergod:


oldergod:
You're right. I'll try to think of something.

swankjesse:
Might be just a thin interface that does to and from JSON on either library.

oldergod:
I've copied current JSON used in Moshi and Gson to create shared `.json` used by a parameterized test suite testing both Gson and Moshi.

PR #1614, opened 2020-07-02
Wire configurations cannot be consumed by other projects

swankjesse:


swankjesse:
https://docs.gradle.org/current/userguide/declaring_dependencies.html#sec:resolvable-consumable-configs

PR #1615, opened 2020-07-03
Add break lines between messages and members

oldergod:


JakeWharton:
Motivation? I'm supportive of the type spacing but unsure about member spacing.

oldergod:
It feels a bit crowded if you have a big type with members having comments. I don't really mind either way.

```protobuf
message MyRequest {
  // Some comment in and there blah blah
  // Some comment in and there blah blah
  // Some comment in and there blah blah
  optional string investment_entity_token = 1;
  // Some comment in and there blah blah
  // Some comment in and there blah blah
  // Some comment in and there blah blah
  optional string investment_entity_token = 2;
  // Some comment in and there blah blah
  // Some comment in and there blah blah
  // Some comment in and there blah blah
  optional string investment_entity_token = 3;
```
VS
```protobuf
message MyRequest {
  // Some comment in and there blah blah
  // Some comment in and there blah blah
  // Some comment in and there blah blah
  optional string investment_entity_token = 1;

  // Some comment in and there blah blah
  // Some comment in and there blah blah
  // Some comment in and there blah blah
  optional string investment_entity_token = 2;

  // Some comment in and there blah blah
  // Some comment in and there blah blah
  // Some comment in and there blah blah
  optional string investment_entity_token = 3;
```

rayokota:
Hi @oldergod @JakeWharton , we're storing the `toSchema` representation in the Confluent Schema Registry (https://github.com/confluentinc/schema-registry) and this is a breaking change.  Would it be possible to add back a way to generate the old format, possibly with flags to indicate which format, like `boolean addLinesBetweenMessagesAndMembers`?

JakeWharton:
No, the output is not stable nor guaranteed to be in any particular form. At present we try to guarantee the semantics of the output are exactly that as the input, but we've had slight variances in the past. Relying on the text output to be byte-for-byte compatible will break. It should always be representationally compatible. It makes me extremely nervous to think that there's somehow tooling dependent on the text form of a proto schema rather than its semantic form.

PR #1621, opened 2020-07-05
Duration support for gson

oldergod:


wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/Proto3WireProtocCompatibilityTests.kt
```diff
@@ -158,6 +158,36 @@ class Proto3WireProtocCompatibilityTests {
     assertThat(jsonAdapter.fromJson(json)).isEqualTo(pizzaDelivery)
   }
 
+  @Test fun gsonJson() {
```
swankjesse, line 161: It’d be great if these tests were JSON-library agnostic

PR #1625, opened 2020-07-08
Reject the schema if there's a package cycle problem

swankjesse:


wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/CycleChecker.kt
```diff
@@ -0,0 +1,163 @@
...
+
+  /**
+   * In good programming languages like Java and Kotlin it's completely fine to have dependency
+   * cycles between packages. But Go forbids cycles between packages and that effectively means
+   * that proto must as well when it targets Go.
+   *
+   * This checks for cycles between packages that would offend the Go compiler if these .proto
+   * sources were ever generated with it.
+   */
+  fun checkForPackageCycles() {
```
oldergod, line 90: That was an unintuitive one....

PR #1629, opened 2020-07-09
Emit qualified proto type in file header

JakeWharton:


wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -166,9 +166,8 @@ data class JavaTarget(
         val javaTypeName = javaGenerator.generatedTypeName(type)
         val javaFile = JavaFile.builder(javaTypeName.packageName(), typeSpec)
             .addFileComment("\$L", WireCompiler.CODE_GENERATED_BY_WIRE)
-            .apply {
```
oldergod, line 169: What was this apply doing here....

JakeWharton: I assume it had some conditional logic in it at one point. Dunno.

PR #1631, opened 2020-07-10
Use TypeName instead of ClassName for prototype to java type conversion

oldergod:
We'll need it for struct because struct will be `TypeName` ( `ParameterizedTypeName ` precisely) and not `ClassName`.

### DIFF
--- a/wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
+++ b/wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
@@ -283,9 +283,15 @@ public final class JavaGenerator {
 
     TypeName typeName = nameToJavaName.get(protoType);
     Type type = schema.getType(protoType);
-    if (!(typeName instanceof ClassName)) return null;
 
-    ClassName javaName = (ClassName) typeName;
+    ClassName javaName;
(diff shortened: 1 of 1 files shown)


wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -281,8 +281,11 @@ public TypeName typeName(ProtoType protoType) {
     TypeName profileJavaName = profile.getTarget(protoType);
     if (profileJavaName == null) return null;
 
-    ClassName javaName = nameToJavaName.get(protoType);
+    TypeName typeName = nameToJavaName.get(protoType);
     Type type = schema.getType(protoType);
+    if (!(typeName instanceof ClassName)) return null;
```
JakeWharton, line 286: Throw instead? Or maybe do
```java
ClasssName javaName;
if (typeName instanceof ClassName) {
  javaName = (ClassName) typeName;
} else if (typeName instanceof ParameterizedTypeName) {
  javaName = ((ParameterizedTypeName) typeName).rawType;
} else {
  throw new IllegalArgumentException(..);
}
```

oldergod: Yes, done!

PR #1632, opened 2020-07-11
Add Struct support in Java + Fix Adapters for built in proto3 types

oldergod:
Next step is proper tests for all that shit


wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -1775,4 +1812,8 @@ static long valueToLong(@Nullable Object value) {
       return new BigInteger(string).longValue(); // Decimal.
     }
   }
+
+  private static boolean isStructNull(ProtoType type) {
```
JakeWharton, line 1816: Will this be used in more places in the future? It's only used once right now so it could just be inlined.

oldergod: Inlined.

oldergod:
Hmm, why can't it resolve Duration or Instant...

PR #1634, opened 2020-07-13
Execute WireRun in the scope of the Closer

JakeWharton:
When directoryPaths encouters a zip it opens it as an FS, registers the FS as a closable, and returns a path pointer inside the zip. This is aggregated into a list of paths which is handed to WireRun. Historically, the closer is then closed before executing, but that would render the paths to zips as unusable. Instead, execute the WireRun inside the context of the closer to ensure those paths remain valid.


JakeWharton:
Obviously no one is using this functionality, and neither am I, but it simplifies some other work I'm doing inside this method.

JakeWharton:
Failure was a flake downloading a dep

PR #1639, opened 2020-07-15
Use map for scalar identity values

oldergod:


oldergod:
I think I like it as is for now.

PR #1646, opened 2020-07-16
Testing negative double zero just in case

oldergod:


wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/Proto3WireProtocCompatibilityTests.kt
```diff
@@ -724,6 +724,25 @@ class Proto3WireProtocCompatibilityTests {
...
+    val wireKotlin = AllTypesK(squareup_proto3_kotlin_alltypes_double = -0.0)
+
+    val protocByteArray = protoc.toByteArray()
+    assertThat(AllTypesK.ADAPTER.encode(wireKotlin)).isEqualTo(protocByteArray)
+    assertThat(AllTypesK.ADAPTER.decode(protocByteArray)).isEqualTo(wireKotlin)
+
+    val protocJson = JsonFormat.printer().print(protoc)
+    val wireKotlinJson = moshi.adapter(AllTypesK::class.java).toJson(wireKotlin)
+    assertJsonEquals("{\"double\": -0.0}", protocJson)
+    assertThat(wireKotlinJson).contains("\"double\":-0.0")
```
JakeWharton, line 740: Why is this a different mechanism than above?

oldergod:
```suggestion
    // Wire prints empty lists and empty maps so we check that it contains the concerned value only.
    assertThat(wireKotlinJson).contains("\"double\":-0.0")
```

PR #1649, opened 2020-07-16
Add BOM-aware decoding of file contents

### LINKED ISSUE #1647: Support UTF-8 with BOM in proto files please
Hey guys,

I have a bunch of 3rd-party proto files that are UTF-8 with BOM and without BOM:

Example #1 (with BOM):
```
d@OMEN:.../WireTest/app/src/main/proto$ hexdump -C MachineId.proto
00000000  ef bb bf 73 79 6e 74 61  78 20 3d 20 22 70 72 6f  |...syntax = "pro|
```

Example #2 (without BOM):
```
d@OMEN:.../WireTest/app/src/main/proto$ hexdump -C StatDoc.proto
00000000  73 79 6e 74 61 78 20 3d  20 22 70 72 6f 74 6f 33  |syntax = "proto2|
```

Wire compiler v3.2.2 apparently doesn't like BOM:
```
Syntax error in .../MachineId.proto:1:1: expected a word
```

Would it be feasible for Wire to support this?

JakeWharton:
Closes #1647

### DIFF
--- a/wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Root.kt
+++ b/wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Root.kt
@@ -20,9 +20,13 @@ import com.squareup.wire.java.internal.ProfileFileElement
 import com.squareup.wire.java.internal.ProfileParser
 import com.squareup.wire.schema.CoreLoader.isWireRuntimeProto
 import com.squareup.wire.schema.internal.parser.ProtoParser
+import okio.BufferedSource
+import okio.ByteString.Companion.decodeHex
+import okio.Options
 import okio.buffer
 import okio.source
 import java.io.IOException
+import java.nio.charset.Charset
 import java.nio.file.FileSystem
 import java.nio.file.FileSystems
 import java.nio.file.FileVisitOption
@@ -127,7 +131,8 @@ internal class ProtoFilePath(
   fun parse(): ProtoFile {
     try {
       path.source().buffer().use { source ->
-        val data = source.readUtf8()
+        val charset = source.readBomAsCharset()
+        val data = source.readString(charset)
         val element = ProtoParser.parse(location, data)
         return ProtoFile.get(element)
       }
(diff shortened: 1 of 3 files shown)


wire-library/wire-schema/src/jvmMain/kotlin/com/squareup/wire/schema/internal/UtilJVM.kt
```diff
@@ -28,3 +38,24 @@ internal actual inline fun String.toEnglishLowerCase() = toLowerCase(Locale.US)
...
+private val UNICODE_BOMS = Options.of(
+    "efbbbf".decodeHex(), // UTF-8
+    "feff".decodeHex(), // UTF-16BE
+    "fffe".decodeHex(), // UTF-16LE
+    "0000ffff".decodeHex(), // UTF-32BE
+    "ffff0000".decodeHex() // UTF-32LE
+)
+
+@Throws(IOException::class)
+fun BufferedSource.readBomAsCharset(default: Charset = UTF_8): Charset {
```
JakeWharton, line 51: I'm going to move this so it's not public. It was being used in two places but now it's only one so it can just live close to its site of use.

@swankjesse should this live in Okio? Maybe as `{Path|File}.bomAwareSource()` or something?

swankjesse: Not yet. The problem is Okio sources don’t know how to read anything other than UTF-8. If we ever were to add such support? That’d be a lot of complexity.

I was hoping to just wait out the universe until UTF-8 was the only thing that existed.

JakeWharton: Nevermind that can't be the signature 🤦 

Maybe `BufferedSource.readStringBomAware()`, although that's a mouthful.

JakeWharton: Even if UTF-8 is everywhere you still need to conditionally handle the presence of its BOM

JakeWharton:
Failure is a flake

PR #1652, opened 2020-07-17
Throw on unresolved options

oldergod:
This is what Protoc does.
I think we should just turn it on without going through an option.
It might break some schema but as we can see in the files I had to fix, something like `redacted` could be NOT applied without nobody noticing it. It actually happened a few days ago internally.

### DIFF
--- a/wire-library/wire-compiler/src/test/java/com/squareup/wire/WireCompilerErrorTest.kt
+++ b/wire-library/wire-compiler/src/test/java/com/squareup/wire/WireCompilerErrorTest.kt
@@ -140,6 +140,7 @@ class WireCompilerErrorTest {
           |multiple enums share constant QUIX:
           |  1. com.squareup.protos.test.Bar.QUIX (/source/test.proto:4:3)
           |  2. com.squareup.protos.test.Bar2.QUIX (/source/test.proto:10:3)
+          |  for file /source/test.proto
           """.trimMargin())
   }
 
--- a/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Linker.kt
+++ b/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Linker.kt
@@ -473,6 +473,10 @@ class Linker {
           error.append("$prefix message ${context.type} (${context.location})")
         }
 
+        is EnumConstant -> {
+          error.append("$prefix constant ${context.name} (${context.location})")
+        }
+
         is EnumType -> {
           error.append("$prefix enum ${context.type} (${context.location})")
         }
--- a/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Linker.kt
+++ b/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Linker.kt
@@ -94,14 +94,14 @@ class Linker {
     for (fileLinker in sourceFiles) {
       // TODO(benoit) Create a map to cache those.
       val syntaxRules = SyntaxRules.get(fileLinker.protoFile.syntax)
-      fileLinker.linkOptions(syntaxRules)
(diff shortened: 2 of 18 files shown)


wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Options.kt
```diff
@@ -96,6 +96,7 @@ class Options(
         path = resolveFieldPath(packageName + "." + option.name, extensionsForType.keys)
       }
       if (path == null) {
+        linker.addError("unable to resolve ${option.name}")
```
JakeWharton, line 99:
```suggestion
        linker.addError("unable to resolve option '${option.name}'")
```

oldergod: Fixed

oldergod:
Added rule for file options

wire-library/wire-compiler/src/test/java/com/squareup/wire/WireCompilerErrorTest.kt
```diff
@@ -140,6 +140,7 @@ class WireCompilerErrorTest {
           |multiple enums share constant QUIX:
           |  1. com.squareup.protos.test.Bar.QUIX (/source/test.proto:4:3)
           |  2. com.squareup.protos.test.Bar2.QUIX (/source/test.proto:10:3)
+          |  for file /source/test.proto
```
oldergod, line 143: That's unfortunate.
I cannot come up with a fix for this specific case, and I also think it's not a big of a deal with we add this line for this error?

JakeWharton: Seems fine

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Linker.kt
```diff
@@ -449,6 +449,10 @@ class Linker {
     val error = StringBuilder()
     error.append(message)
 
+    val contextStack = this.contextStack.toMutableList()
+    if (contextStack.size > 1) {
+      contextStack.removeAll { it is ProtoFile }
```
swankjesse, line 454: perfect

swankjesse: Screen Shot 2020-07-17 at 21 52 56

I spoke too soon! Needs to check if any is not a ProtoFile.
```
    if (contextStack.any { it !is ProtoType }) {
      contextStack.removeAll { it is ProtoFile }
    }
```

oldergod: Fixed, thanks for checking!

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Linker.kt
```diff
@@ -94,14 +94,14 @@ class Linker {
...
-      fileLinker.linkOptions(syntaxRules)
+      fileLinker.linkOptions(syntaxRules, validate = true)
     }
 
     // For compactness we'd prefer to link the options of source files only. But we link file
     // options on referenced files to make sure that java_package is populated.
     while (fileOptionsQueue.isNotEmpty()) {
       val fileLinker = fileOptionsQueue.poll()!!
-      fileLinker.requireFileOptionsLinked()
+      fileLinker.requireFileOptionsLinked(validate = false)
```
swankjesse, line 104: NICE

swankjesse:
Confirmed this works with our internal protos.

oldergod:
Thank you!

PR #1653, opened 2020-07-17
map fields cannot be extensions

### LINKED ISSUE #1651: protoc rejects extensions whose type is a map
We should do the same. This occurs in `all_types`.

oldergod:
fixes #1651

### DIFF
--- a/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Field.kt
+++ b/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Field.kt
@@ -117,8 +117,13 @@ class Field private constructor(
     if (isPacked && !isPackable(linker, type!!)) {
       linker.addError("packed=true not permitted on $type")
     }
-    if (isExtension && isRequired) {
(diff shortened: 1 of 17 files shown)


wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Field.kt
```diff
@@ -117,8 +117,13 @@ class Field private constructor(
...
       linker.addError("packed=true not permitted on $type")
     }
-    if (isExtension && isRequired) {
-      linker.addError("extension fields cannot be required")
+    if (isExtension) {
+      if (isRequired) {
+        linker.addError("extension fields cannot be required")
+      }
+      if (type!!.isMap) {
+        linker.addError("map fields cannot be extensions")
```
JakeWharton, line 125:
```suggestion
        linker.addError("extension fields cannot be a map")
```
For consistency with above message

oldergod: fixed

PR #1654, opened 2020-07-17
enum value in map must define 0 as the first value

### LINKED ISSUE #1650: protoc rejects maps whose value type is an enum without a zero case
We should do the same validation.

oldergod:
fixes #1650


swankjesse:
This constraint only applies to proto3?

oldergod:
No, proto2 as well.

swankjesse:
I believe this is a departure from the proto2 language spec.

oldergod:
Is it? I know protoc fails in that case in proto2 schema.

image

swankjesse:
Ahhh... It's the “in a map” part I missed. My bad!

PR #1656, opened 2020-07-19
Promote some JSON behavior to FieldBinding

swankjesse:
This way it's easier to share between JSON frameworks.

### DIFF
--- a/wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/FieldBinding.kt
+++ b/wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/FieldBinding.kt
@@ -17,9 +17,16 @@ package com.squareup.wire.internal
 
 import com.squareup.wire.Message
 import com.squareup.wire.ProtoAdapter
+import com.squareup.wire.Syntax
+import com.squareup.wire.Syntax.PROTO_2
+import com.squareup.wire.WireEnum
 import com.squareup.wire.WireField
+import okio.ByteString
+import okio.ByteString.Companion.decodeBase64
 import java.lang.reflect.Field
 import java.lang.reflect.Method
+import java.math.BigDecimal
+import java.math.BigInteger
 
 /**
  * Read, write, and describe a tag within a message. This class knows how to assign fields to a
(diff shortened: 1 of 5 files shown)


wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/StructTest.kt
```diff
@@ -416,7 +414,6 @@ class StructTest {
       |}""".trimMargin()
     // Protoc doesn't print those unless explicitly set.
     val moshiJson = """{
-      |  "nullValue": null,
```
swankjesse, line 419: I didn’t set out to make these deltas; they came out accidentally. Any concerns? I think probably the thing we want to be most deliberate about is when to omit null. (Always?)

wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/FieldBinding.kt
```diff
@@ -142,4 +181,83 @@ class FieldBinding<M : Message<M, B>, B : Message.Builder<M, B>> internal constr
...
 
   internal fun getFromBuilder(builder: B): Any? = builderField.get(builder)
+
+  fun jsonStringAdapter(syntax: Syntax): JsonFormatter<*>? {
+    return when (adapterString) {
+      "com.squareup.wire.ProtoAdapter#BYTES" -> ByteStringJsonFormatter
+      "com.squareup.wire.ProtoAdapter#INT64",
+      "com.squareup.wire.ProtoAdapter#SFIXED64",
+      "com.squareup.wire.ProtoAdapter#SINT64" -> LongAsStringJsonFormatter
+      "com.squareup.wire.ProtoAdapter#FIXED64" -> UnsignedLongAsStringJsonFormatter
```
oldergod, line 191: Do we wanna use them for proto2 as well? I was thinking we'd add that for proto3 only.

wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/StructTest.kt
oldergod: All good

wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/FieldBinding.kt
swankjesse: Good call.

oldergod: That's was probably why the tests were failing. It's green now!

PR #1658, opened 2020-07-20
Expose set properties on PruningRules

JakeWharton:


wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/PruningRules.kt
```diff
@@ -71,16 +71,23 @@ import com.squareup.wire.schema.SemVer.Companion.toLowerCaseSemVer
  * clients. Such code will have both the `age` and `birth_date` fields.
  */
 class PruningRules private constructor(builder: Builder) {
-  private val roots = builder.roots.toSet()
-  private val prunes = builder.prunes.toSet()
-  private val since = builder.since
-  private val until = builder.until
-  private val only = builder.only
+  val roots = builder.roots.toSet()
+  val prunes = builder.prunes.toSet()
```
JakeWharton, line 75: I only need these two, but figured I'd do 'em all?

swankjesse: works for me.

PR #1659, opened 2020-07-21
Use JsonFormatter for Instant and Duration

swankjesse:
Also only handle structs when they're fields of other message types.


wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/InstantJsonFormatter.kt
```diff
@@ -0,0 +1,38 @@
...
+import java.time.format.DateTimeFormatter.ISO_INSTANT
+import java.time.format.DateTimeFormatter.ISO_OFFSET_DATE_TIME
+
+/**
+ * Encode an instant as a JSON string like "1950-01-01T00:00:00Z". From the spec:
+ *
+ * > Uses RFC 3339, where generated output will always be Z-normalized and uses 0, 3, 6 or 9
+ * > fractional digits. Offsets other than "Z" are also accepted.
+ */
+object InstantJsonFormatter : JsonFormatter<Instant> {
```
swankjesse, line 29: Not new code, just renamed & refactored InstantJsonAdapter.

PR #1661, opened 2020-07-22
Resolve options relative paths

oldergod:


swankjesse:
Nice fix.

oldergod:
CI is blocking but it all passes.

PR #1663, opened 2020-07-23
Resolve options used inside extensions

oldergod:
Not too bad?


oldergod:
It's probably too weak. I'll test when we can "resolve" the option but the type is different (string vs int32). And how about when there's one option in the same file as the caller site, and another one in another file? We're gonna say it's ambiguous but protoc probably treat it nicely.

OR we could never validate options used in extensions... which sucks.

oldergod:
I think it's good enough for now.

swankjesse:
Definitely an improvement.

PR #1664, opened 2020-07-28
Add ProtoAdapter.identity

swankjesse:


swankjesse:
The interesting bits are in the first commit; the rest is just regenerating things.
https://github.com/square/wire/pull/1664/commits/b729e6d0f6ccfa155586df052e04fd96355321c9

swankjesse:
Needs tests. In particular:
 - test that an enum compiled with an old version gets the right identity when it exists
 - test that an enum compiled with an old version gets the right identity when it doesn’t exist
 - identity of repeated and packed fields should be listOf()
 - identity of maps should be mapOf()
 - loop over the scalar types and confirm they all have identity values
 - generated message adapters should have identity null
 - runtime message adapters should have identity null

Note that this API includes the identity even for proto2 types. I think this is a reasonable simplification, though callsites will need to know to check the syntax before using the identity.

wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -242,4 +256,12 @@ actual abstract class ProtoAdapter<E> actual constructor(
...
     @JvmField actual val STRUCT_VALUE: ProtoAdapter<Any?> = commonStructValue()
   }
+
+  internal val isStruct: Boolean
+    get() {
+      return this == STRUCT_MAP ||
+          this == STRUCT_LIST ||
+          this == STRUCT_VALUE ||
+          this == STRUCT_NULL
+    }
```
oldergod, line 266: unused?

swankjesse: Yeah, mistake. Fixed.

wire-library/wire-tests/src/jvmJavaTest/kotlin/com/squareup/wire/ProtoAdapterIdentityTest.kt
```diff
@@ -0,0 +1,142 @@
...
+import com.squareup.wire.protos.person.Person
+import com.squareup.wire.protos.person.Person.PhoneType
+import com.squareup.wire.protos.simple.SimpleMessage.NestedEnum
+import okio.ByteString
+import org.assertj.core.api.Assertions.assertThat
+import java.lang.reflect.Modifier
+import kotlin.reflect.KClass
+import kotlin.test.Test
+
+class ProtoAdapterIdentityTest {
```
swankjesse, line 28: NEW!

wire-library/wire-tests/src/jvmJavaTest/kotlin/com/squareup/wire/ProtoAdapterTypeUrlTest.kt
```diff
@@ -0,0 +1,70 @@
...
+
+import com.squareup.wire.protos.person.Person
+import com.squareup.wire.protos.person.Person.PhoneType
+import okio.ByteString
+import org.assertj.core.api.Assertions.assertThat
+import java.lang.reflect.Modifier
+import kotlin.reflect.KClass
+import kotlin.test.Test
+
+class ProtoAdapterTypeUrlTest {
```
swankjesse, line 26: NEW!

PR #1665, opened 2020-07-30
AnyMessageTypeAdapter

oldergod:
Tests in followup (trying to split PRs)


wire-library/wire-gson-support/src/main/java/com/squareup/wire/AnyMessageTypeAdapter.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    val jsonObject = delegate.toJsonTree(value.unpack(protoAdapter) as Message<*, *>).asJsonObject
+    for ((name: String, element: JsonElement) in jsonObject.entrySet()) {
+      writer.name(name)
+      elementAdapter.write(writer, element)
+    }
+    writer.endObject()
+  }
+
+  @Throws(java.io.IOException::class)
```
swankjesse, line 56: import?

PR #1666, opened 2020-07-30
GsonJsonIntegration

oldergod:


wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/FieldBinding.kt
```diff
@@ -65,10 +65,14 @@ class FieldBinding<M : Message<M, B>, B : Message.Builder<M, B>> internal constr
...
   /** The identity value is safe to omit during encoding. */
   val identity: Any? = run {
-    if (wireField.label != WireField.Label.OMIT_IDENTITY) return@run null
+    // TODO(benoit) Use ProtoAdapter.identity to avoid using that hack.
+    // Only set for maps.
+    if (wireField.keyAdapter.isNotEmpty()) return@run emptyMap<Any, Any?>()
 
     if (wireField.label.isRepeated) return@run listOf<Any?>()
 
+    if (wireField.label != WireField.Label.OMIT_IDENTITY) return@run null
```
oldergod, line 74: That was the fix for emptyList. The first `if` statement was taking us out for list and maps.

swankjesse: When we use the ProtoAdapter.identity field, that will thankfully go away.

PR #1667, opened 2020-07-30
Fix tests, delete StructJsonAdapter, and misc

oldergod:


wire-grpc-sample/protos/src/main/java/com/squareup/wire/whiteboard/WhiteboardUpdate.kt
```diff
@@ -292,7 +294,8 @@ class WhiteboardUpdate(
         FieldEncoding.LENGTH_DELIMITED, 
         UpdatePoints::class, 
         "type.googleapis.com/com.squareup.wire.whiteboard.WhiteboardUpdate.UpdatePoints", 
-        PROTO_2
+        PROTO_2,
```
swankjesse, line 297: thanks for fixing these

PR #1671, opened 2020-08-01
Make immutable copies of structs.

swankjesse:


swankjesse:
For follow-up, we’re still incorrect for a few cases:

 * [ ] Regular lists and maps are not immutable in generated Kotlin
 * [ ] No immutable copies of map and repeated struct values

PR #1672, opened 2020-08-01
Make lists and maps immutable from Kotlin

swankjesse:
This has a big problem - it causes fields to be reordered, which impacts
on-the-wire encoding of both .proto messages and JSON.

I'm tempted to force all of our properties to have separate parameter and
property declarations to prevent this shuffling.

### DIFF
--- a/wire-library/wire-tests/src/commonTest/kotlin/com/squareup/wire/KotlinMapTest.kt
+++ b/wire-library/wire-tests/src/commonTest/kotlin/com/squareup/wire/KotlinMapTest.kt
@@ -22,6 +22,7 @@ import okio.ByteString.Companion.decodeHex
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertNotNull
+import kotlin.test.fail
 
 class KotlinMapTest {
   private val adapter = Mappy.ADAPTER
@@ -37,6 +38,22 @@ class KotlinMapTest {
 
     val empty = adapter.decode(ByteArray(0))
     assertNotNull(empty.things)
+  }
+
+  @Test fun mapsAreImmutable() {
+    val map = mutableMapOf("one" to Thing("One"))
+
+    val mappy = Mappy(things = map)
+    try {
+      (mappy.things as MutableMap<*, *>).clear()
+      fail()
+    } catch (_: UnsupportedOperationException) {
+      // Mutation failed as expected.
+    }
+
+    // Mutate the values used to create the map. Wire should have defensive copies.
+    map.clear()
+    assertEquals(mapOf("one" to Thing("One")), mappy.things)
   }
 
   companion object {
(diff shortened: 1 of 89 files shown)


swankjesse:
The easier-to-review diff is the 2nd commit only:
https://github.com/square/wire/pull/1672/commits/662c7585d3173116a1a5caf8098744e3ca87ce36

wire-library/wire-tests/src/jvmJsonTest/kotlin/com/squareup/wire/WireJsonTest.kt
```diff
@@ -138,15 +138,16 @@ class WireJsonTest {
...
         .nested__message(NestedCamelCase.Builder().one_int32(1).build())
         ._Rep_int32(listOf(1, 2))
         .IDitIt_my_wAy("frank")
         .map_int32_Int32(mapOf(1 to 2))
         .build()
-    assertThat(jsonLibrary.toJson(camel, CamelCase::class.java)).isEqualTo(
-        """{"nestedMessage":{"oneInt32":1},"RepInt32":[1,2],"iDitItMyWAy":"frank","mapInt32Int32":{"1":2}}""")
+    assertThat(jsonLibrary.toJson(camel, CamelCase::class.java)).isIn(
+        """{"nestedMessage":{"oneInt32":1},"RepInt32":[1,2],"iDitItMyWAy":"frank","mapInt32Int32":{"1":2}}""",
+        """{"RepInt32":[1,2],"mapInt32Int32":{"1":2},"nestedMessage":{"oneInt32":1},"iDitItMyWAy":"frank"}""")
```
oldergod, line 150: This is why I made `assertJsonEquals`, you can use it to confirm JSON equality without caring about field orders.

`assertJsonEquals(a: String, b: String)`

swankjesse: Yeah, that’s much nicer. Fixed.

wire-library/wire-tests/src/commonTest/kotlin/com/squareup/wire/KotlinMapTest.kt
```diff
@@ -40,6 +40,7 @@ class KotlinMapTest {
     assertNotNull(empty.things)
   }
 
+  @IgnoreJs
```
oldergod, line 43: Is there a bug we could file on youtrack?

swankjesse: Maps casted to MutableMaps can be cleared. The test is exercising something JS isn’t particularly good at.

swankjesse: Hang on, I have it wrong here. It’s because our unmodifiable functions don’t work on Native or JS.

```
@Suppress("NOTHING_TO_INLINE") // Syntactic sugar.
actual inline fun <K, V> MutableMap<K, V>.toUnmodifiableMap(): Map<K, V> = this
```

swankjesse: (not going to fix this now, regardless)

PR #1676, opened 2020-08-02
Move all64 json tests in shared json tests.

oldergod:


wire-library/wire-tests/src/jvmJsonTest/kotlin/com/squareup/wire/WireJsonTest.kt
```diff
@@ -176,6 +177,9 @@ class WireJsonTest {
         .value_e(mapOf("a" to 1.0))
         .value_f(listOf("a", 3.0))
         .build()
+
+    assertJsonEquals(ALL_STRUCT_JSON, jsonLibrary.toJson(value, AllStructs::class.java))
```
oldergod, line 181: We were missing a few.

PR #1679, opened 2020-08-03
Don't generate boxed types for omit-identity proto3 fields

swankjesse:


wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -1190,11 +1192,21 @@ private TypeName fieldType(Field field) {
...
+
     TypeName messageType = typeName(type);
-    return field.isRepeated() ? listOf(messageType) : messageType;
+    switch (field.getEncodeMode()) {
+      case REPEATED:
+      case PACKED:
+        return listOf(messageType.box());
+      case NULL_IF_ABSENT:
+      case REQUIRED:
+        return messageType.box();
```
oldergod, line 1206: I realized too late that REQUIRED too could be unboxed.

oldergod: will PR

oldergod: Or maybe we don't wanna change proto2's API?

PR #1684, opened 2020-08-04
Throw if one sets a null to a omit-identity enum in Java

oldergod:
But we let `STRUCT_NULL` be free as the wind.


swankjesse:
image

PR #1685, opened 2020-08-04
Introduce modules for partitioning a schema 

JakeWharton:
This will allow you to split up code generation into distinct modules which are related to each other.


wire-library/wire-compiler/src/main/java/com/squareup/wire/Manifest.kt
```diff
@@ -0,0 +1,310 @@
...
+fun Schema.partition(manifest: Manifest): PartitionedSchema {
+  val modules = manifest.modules
+
+  if (debug) {
+    println("Modules: ${modules.keys}")
+    println("Roots: ${manifest.roots}")
+    println("Order: ${manifest.order}")
+  }
+
+  val disjointGraphs = manifest.roots
```
swankjesse, line 121: consider doing like DagChecker and splitting the graph theory work from the application of that work

swankjesse: (a bug in disjoinGraphs computation is pretty brutal to debug the way this function is structured)

wire-library/wire-compiler/src/main/java/com/squareup/wire/Manifest.kt
```diff
@@ -0,0 +1,310 @@
...
+
+    val pruningRules = PruningRules.Builder()
+        .apply {
+          for (dependencyName in transitiveDependencySet) {
+            val dependency = modules.getValue(dependencyName)
+            addRoot(dependency.roots)
+            prune(dependency.prunes)
+          }
+          addRoot(module.roots)
+          // TODO(jw): what do we do about prunes that apply to types in an upstream module? fail?
```
swankjesse, line 170: you should be able to detect this by seeing if any type is a subset of itself in a different module

wire-library/wire-compiler/src/main/java/com/squareup/wire/Manifest.kt
```diff
@@ -0,0 +1,310 @@
...
+        val currentTypes = partitions.getValue(currentName).types
+        val otherTypes = partitions.getValue(otherName).types
+        val duplicates = currentTypes.intersect(otherTypes)
+        if (duplicates.isNotEmpty()) {
+          val currentModule = modules.getValue(currentName)
+          val otherModule = modules.getValue(otherName)
+          for (duplicate in duplicates) {
+            val duplicateName = duplicate.toString()
+            if (duplicateName !in currentModule.roots || duplicateName !in otherModule.roots) {
+              warnings += """
```
swankjesse, line 258: neat

wire-library/wire-compiler/src/main/java/com/squareup/wire/Manifest.kt
```diff
@@ -0,0 +1,310 @@
...
+      }
+      .toMap()
+
+  return object : SchemaHandler {
+    override fun handle(type: Type): Path? {
+      val protoType = type.type
+      return handlers[protoType]?.let { (partition, delegate) ->
+        // The schema used to invoke this function is the global one which has not been pruned so we
+        // swap its type for the corresponding pruned one.
+        // TODO(jw): Split WireRun into two phases so we can insert our pruned schema before step 4.
```
swankjesse, line 294: Rad

swankjesse:
On YAML stuff . . . . 

WireRun is a bit weird. It’s purposefully separate from the Gradle plugin, and I’m pretty happy with how that worked out.

For this feature I propose something that might make this all fit better together:
 - make manifests a feature of WireRun, and remove the tight YAML integration.
 - Gradle users can do manifests without any YAML. Just add Gradle integration to get at the manifest features.
 - For non-Gradle users, we could offer a YAML structure for the entire build.

```
sourcePaths:
 - srcDir 'src/main/protos'
    includes:
      - 'com/example/pizza/pizza_delivery.proto'
      - 'com/example/pizza/pizza.proto'

modules:
  ModuleName:
    dependencies:
     - OtherModule
    roots:
     - example.feature.Message
    prunes:
     - example.logging.Settings
  OtherModule:
    roots:
     - example.other.Type

kotlin:
  out: build/kotlin/generated
```

swankjesse:
(Definitely the right call not trying to squeeze something this expressive into the CLI)

wire-library/wire-compiler/src/main/java/com/squareup/wire/Manifest.kt
JakeWharton: The intent is to get rid of this inheritance, actually. May be worth omitting it from this PR and inverting the behavior where prunes are local but we generate unnecessary types.

This manifests as the `@Ignore` in the partition test moving from the current method to the other one.

wire-library/wire-compiler/src/main/java/com/squareup/wire/Manifest.kt
```diff
@@ -0,0 +1,310 @@
...
+          append(" - ")
+          append(cycle)
+          append('\n')
+        }
+      }
+    }
+  }
+
+  /** Modules which have no dependencies. */
+  val roots: Set<String> = modules.filterValues { it.dependencies.isEmpty() }.keys
```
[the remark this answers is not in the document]
JakeWharton: Leaves are nodes with no dependents I think. This is the inverse. I don't have a good alternative name.

wire-library/wire-compiler/src/main/java/com/squareup/wire/Manifest.kt
```diff
@@ -0,0 +1,310 @@
...
+/** Module name to partition info. The iteration order of this map is the generation order. */
+class PartitionedSchema(
+  modules: Map<String, Partition>,
+  val warnings: List<String>,
+  val errors: List<String>
+) : Map<String, Partition> by modules {
+  class Partition(
+    val schema: Schema,
+    val types: Set<ProtoType>,
+    val upstreamTypes: Map<ProtoType, String>
```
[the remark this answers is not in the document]
JakeWharton: Yep. My kingdom for inline classes. At this rate it'll be stable in the JVM before Kotlin.

wire-library/wire-compiler/src/main/java/com/squareup/wire/Manifest.kt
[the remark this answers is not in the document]
swankjesse: I think both terms are a bit wrong because this is a directed graph. Wikipedia suggests 'sink', a word I don’t like here. Maybe seeds?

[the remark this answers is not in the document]
JakeWharton: Seeds

wire-library/wire-compiler/src/main/java/com/squareup/wire/Manifest.kt
```diff
@@ -0,0 +1,310 @@
...
+  val disjointGraphs = manifest.roots
+      .map { root ->
+        val reachableNames = mutableSetOf<String>()
+        val visitQueue = ArrayDeque<String>().apply { add(root) }
+        while (visitQueue.isNotEmpty()) {
+          val visitName = visitQueue.removeFirst()
+          reachableNames += visitName
+
+          val upstreamModules = modules.getValue(visitName).dependencies
+          val downstreamModules = modules.filterValues { visitName in it.dependencies }.keys
```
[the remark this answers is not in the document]
JakeWharton: Done

wire-library/wire-compiler/src/main/java/com/squareup/wire/Manifest.kt
JakeWharton: Split into its own type!

wire-library/wire-compiler/src/main/java/com/squareup/wire/Manifest.kt
JakeWharton: Remove this for now. It would have gone away anyway.

JakeWharton:
WireRun was likely always the target for this. Right now it's broken without shoving it between two WireRun steps. Can do in a follow-up.

I don't think path arguments should be moved to the manifest though. Input and output paths might be dynamically generated such as in Bazel for its sandboxing. Aside from that, I don't think we should necessarily conflate the inputs and outputs with the transform operation configuration.

swankjesse:
Was trying to suggest something else: that the YAML file represents a WireRun, not the module graph. For Bazel, you would express the module structure in the Starlark file and skip YAML altogether.

JakeWharton:
You will still have to propagate that information into the command-line compiler though so you're basically advocating we abandon all command line flags? It would be weird to support both regular arguments and manifest keys through the command-line compiler.

Internally I think we'll have the models unified, but I'm not sure it makes sense to pipe that up through the CLI.

JakeWharton:
GitHub Actions are not happy...

JakeWharton:
Oh infinite loop and no default test timeout in JUnit 😬 Whoooops

JakeWharton:
PTAL

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
```diff
@@ -170,9 +171,40 @@ data class WireRun(
...
   /** True to build proto3 artifacts. This is unsupported and does not work. */
   val proto3Preview: Boolean = false
 ) {
+  data class Module(
+    val dependencies: Set<String> = emptySet(),
+    val pruningRules: PruningRules? = null
+  )
+
+  init {
+    val dagChecker = DagChecker(modules.keys) { moduleName ->
```
swankjesse, line 193: For follow up, we can move `check()` into the new `DirectedAcyclicGraph` class!

wire-library/wire-compiler/src/test/java/com/squareup/wire/schema/PartitionedSchemaTest.kt
```diff
@@ -0,0 +1,122 @@
...
+
+    val partitionedSchema = schema.partition(modules)
+
+    val commonPartition = partitionedSchema.getValue("common")
+    // B has no field of type C because of its inclusion in common's prune list.
+    assertThat(commonPartition.schema.getMessage("B").field("c")).isNull()
+    // C is not generated in common because of its inclusion in the prune list.
+    assertThat(commonPartition.types).doesNotContain(ProtoType.get("C"))
+
+    // C is not in feature because its only dependant is B which is from common.
```
swankjesse, line 65: This is neat; potentially a bit surprising!

wire-library/wire-compiler/src/test/java/com/squareup/wire/schema/PartitionedSchemaTest.kt
```diff
@@ -0,0 +1,122 @@
...
+    val featurePartition = partitionedSchema.getValue("feature")
+    assertThat(featurePartition.types).doesNotContain(ProtoType.get("C"))
+  }
+
+  @Test fun upstreamPruneIsNotPrunedDownstream() {
```
swankjesse, line 70: Really useful tests in terms of explaining the goals

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/PartitionedSchema.kt
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  return PartitionedSchema(partitions, warnings, errors)
+}
+
+internal fun PartitionedSchema.buildSchemaHandler(
+  factory: (name: String, partition: Partition) -> SchemaHandler
+): SchemaHandler {
+  // TODO this means you cannot generate the same type in two unrelated modules! fix!
+  val handlers = this
```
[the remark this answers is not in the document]
JakeWharton: Oh! This whole nonsense can be removed now that it's baked into `WireRun`!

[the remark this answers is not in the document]
JakeWharton: Removed

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
JakeWharton: Yeah it lives in wire-schema and this in wire-compiler. I can move this up into wire-schema and combine afterward.

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/PartitionedSchema.kt
```diff
@@ -0,0 +1,151 @@
...
+ * limitations under the License.
+ */
+package com.squareup.wire.schema
+
+import com.squareup.wire.schema.PartitionedSchema.Partition
+import com.squareup.wire.schema.Target.SchemaHandler
+import com.squareup.wire.schema.WireRun.Module
+import java.nio.file.Path
+
+/** Module name to partition info. The iteration order of this map is the generation order. */
```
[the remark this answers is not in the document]
JakeWharton: The type implements Map, but i'll switch it to just expose a property.

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -60,6 +60,7 @@ sealed class Target : Serializable {
 
   internal abstract fun newHandler(
     schema: Schema,
+    moduleDir: String,
```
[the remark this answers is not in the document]
JakeWharton: Switched it to name and to pass `null` for the single/default case.

JakeWharton:
Not perfect but need forward progress.

PR #1686, opened 2020-08-05
Adds guards against duplication in message/service/rpcs

oldergod:


swankjesse:
You’re going to become known for writing nice tests.

PR #1688, opened 2020-08-06
Make schema models data classes

JakeWharton:
This allows users to create and copy them on their own. This shouldn't be a problem since they will be linked in Schema.fromFiles() to ensure their validity before being used anywhere. But if you have public API that takes one of these classes you could now receive an unlinked/invalidly-linked instance.

### DIFF
--- a/wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/PartitionedSchema.kt
+++ b/wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/PartitionedSchema.kt
@@ -60,10 +60,30 @@ internal fun Schema.partition(modules: Map<String, Module>): PartitionedSchema {
       }
     } as Map<ProtoType, String> // TODO use buildMap
 
-    val prunedSchema = if (module.pruningRules != null) {
-      prune(module.pruningRules)
-    } else {
+    val stubbedSchema = if (upstreamTypes.isEmpty()) {
       this
+    } else {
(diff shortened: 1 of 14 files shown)


sgarfinkel:
One problem I was having is that the CoreLoader would crash when attempting to resolve ProtoFiles, during the `Linker.link` process, that didn't exist on the file system. I'm not entirely sure why but there should definitely be a test for this.

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/PartitionedSchema.kt
```diff
@@ -111,6 +131,36 @@ internal fun Schema.partition(modules: Map<String, Module>): PartitionedSchema {
   return PartitionedSchema(partitions, warnings, errors)
 }
 
+/** Return a copy of this type with all possible type references removed. */
+private fun Type.asStub(): Type = when {
```
swankjesse, line 135: do we do top-level expressions elsewhere? Should we?

JakeWharton: I don't know many arguments against them, especially not private ones. These two are used by another top-level function that landed in the previous PR so for now I'll keep.

JakeWharton:
@sgarfinkel I would prefer to pursue that as a separate problem/fix. Perhaps re-open your bug and let's draw up a test there?

PR #1691, opened 2020-08-10
Avoid partitioning when no modules specified

JakeWharton:
Bypasses needing a special default module name and keeps the `null` module name logic contained inside `WireRun` rather than leaking it into the partitioning code.

### DIFF
--- a/wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
+++ b/wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
@@ -18,6 +18,7 @@ package com.squareup.wire.schema
 import com.squareup.wire.ConsoleWireLogger
 import com.squareup.wire.Syntax
 import com.squareup.wire.WireLogger
+import com.squareup.wire.schema.PartitionedSchema.Partition
 import com.squareup.wire.schema.internal.DagChecker
 import com.squareup.wire.schema.internal.TypeMover
 import java.nio.file.FileSystem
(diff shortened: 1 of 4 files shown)


wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
```diff
@@ -231,13 +232,20 @@ data class WireRun(
     }
     val targetsExclusiveLast = targets.sortedBy { it.exclusive }
 
-    val partitionedSchema = schema.partition(modules)
+    val partitions = if (modules.isNotEmpty()) {
+      val partitionedSchema = schema.partition(modules)
+      // TODO handle errors and warnings
```
JakeWharton, line 237: Got dropped in the refactoring of the module PR

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
```diff
@@ -175,11 +176,11 @@ data class WireRun(
    * A map from module dir to module info which dictates how the loaded types are partitioned and
    * generated.
    *
-   * By default a single module is used which includes everything in the root output directory.
+   * When empty everything is generated in the root output directory.
```
swankjesse, line 179:
```
When empty everything
```
 ->
```
When empty, everything
```

oldergod: This is the kind of comment hard to make for non-english-native speakers.

PR #1692, opened 2020-08-10
Migrate Service to expose properties

JakeWharton:
These remain binary compatible but do not include source-compatible trampolines.

Has been annoying me for weeks. Finally got around to it.


wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Service.kt
```diff
@@ -21,46 +21,28 @@ import com.squareup.wire.schema.ProtoMember.Companion.get
...
 import kotlin.jvm.JvmStatic
 
 data class Service(
-  private val protoType: ProtoType,
-  private val location: Location,
-  private val documentation: String,
-  private val name: String,
-  private val rpcs: List<Rpc>,
-  private val options: Options
+  @get:JvmName("type") // For binary compatibility.
```
swankjesse, line 28: honest question: do we care?

swankjesse: (do we care about binary compatibility for wire-schema?)

JakeWharton: I don't know. I maintained it in the past for a similar migration. It's low effort here so might as well leave it, but I don't really know moving forward. Doesn't Jackson depend on it?

oldergod:
Thank you!

PR #1694, opened 2020-08-10
Add syntax to extensions.proto

oldergod:


oldergod:
I wonder if that means nobody ever used our options...

PR #1698, opened 2020-08-11
Prefix EnumValue's wire.since/until with constant_

oldergod:
Protoc doesn't allow same named extensions on the same package even if they are extensions of different types.


wire-protoc-compatibility-tests/src/main/proto/squareup/proto3/kotlin/extensions/wire_message.proto
```diff
@@ -0,0 +1,32 @@
...
+ */
+syntax = "proto3";
+
+package squareup.proto3.kotlin.extensions;
+
+import "wire/extensions.proto";
+
+option (wire.wire_package) = "squareup.proto3.wire.extensions";
+
+message WireMessage {
```
swankjesse, line 24: nice

PR #1699, opened 2020-08-11
Add CLI flag for creating modules from a manifest

JakeWharton:
The manifest is a YAML file whose format is as follows:

```yaml
ModuleName:
  dependencies:
    - OtherModule
  roots:
    - example.feature.Message
  prunes:
    - example.logging.Settings

OtherModule:
  roots:
    - example.other.Type
```

All of the keys under the module name are optional.

The flag is prefixed with 'experimental' since we have yet to agree on its scope or format but want to allow its use for now.

### DIFF
--- a/wire-library/build.gradle
+++ b/wire-library/build.gradle
@@ -9,6 +9,7 @@ buildscript {
 
   dependencies {
     classpath deps.plugins.kotlin
+    classpath deps.plugins.kotlinSerialization
     classpath deps.plugins.shadow
     classpath deps.plugins.japicmp
     classpath deps.plugins.mavenPublish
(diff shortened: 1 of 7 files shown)


swankjesse:
I’d still like to either grow the YAML to model the whole build, or shrink the module to use the build tools’ API.

JakeWharton:
That means deprecating all other command-line arguments, right?

swankjesse:
I think the entire CLI is a wreck. I wouldn't deprecate anything, but I would make the YAML completely capable of representing what the Gradle Plugin can do.

JakeWharton:
For that I would want to do JSON which is better for machines. The yaml exists at the right level of abstraction for humans, but for compiler invocations involving all the file paths and nonsense I think JSON is a better universal transport. We can make the current usage of yaml an implementation detail of that build system and it can coalesce paths and such to produce the canonical JSON wire would accept.

PR #1700, opened 2020-08-12
Do not link after stubbing types

JakeWharton:
Our input is a fully-linked Schema so replacing a type with a stub does not require re-linking. More importantly, because we cannot stub `google.protobuf.*`, this prevents extensions from duplicating on the `google.protobuf.*` types.

### DIFF
--- a/wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/PartitionedSchema.kt
+++ b/wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/PartitionedSchema.kt
@@ -17,6 +17,7 @@ package com.squareup.wire.schema
 
 import com.squareup.wire.schema.PartitionedSchema.Partition
 import com.squareup.wire.schema.WireRun.Module
+import com.squareup.wire.schema.internal.withStubs
 
 internal class PartitionedSchema(
   /** Module name to partition info. The iteration order of this map is the generation order. */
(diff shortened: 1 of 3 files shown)


wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/internal/Util.kt
```diff
@@ -90,3 +96,55 @@ expect interface MutableQueue<T : Any> : MutableCollection<T> {
...
+/**
+ * Replace types in this schema which are present in [typesToStub] with empty shells that have no
+ * outward references. This has to be done in this module so that we can access the internal
+ * constructor to avoid re-linking.
+ */
+fun Schema.withStubs(typesToStub: Set<ProtoType>): Schema {
+  if (typesToStub.isEmpty()) {
+    return this
+  }
+  return Schema(protoFiles.map { protoFile ->
```
JakeWharton, line 110: This is the only change in this PR aside from code relocation. We call `Schema`'s internal constructor rather than `Schema.fromFiles` (which links).

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/PartitionedSchema.kt
```diff
@@ -60,25 +61,11 @@ internal fun Schema.partition(modules: Map<String, Module>): PartitionedSchema {
...
-            }
-        )
-      }
-      Schema.fromFiles(stubbedFiles)
-    }
+    // Replace types which have already been generated with stub types that have no external
+    // references. This ensures our types can still link. More critically, it ensures that
+    // transitive types which were pruned upstream will only be generated in this module if they
+    // are reachable from this module's types.
+    val stubbedSchema = withStubs(upstreamTypes.keys)
```
JakeWharton, line 68: For context, `this` is a `Schema`

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/internal/Util.kt
```diff
@@ -90,3 +96,55 @@ expect interface MutableQueue<T : Any> : MutableCollection<T> {
 }
 
 internal expect fun <T : Any> mutableQueueOf(): MutableQueue<T>
+
+// TODO internal and friend for wire-compiler: https://youtrack.jetbrains.com/issue/KT-20760
```
oldergod, line 100: The ticket is fixed in 1.3.60?

JakeWharton: I believe the compiler is fixed but the IDE is not. I just copied the issue link from above. The IDE issue is https://youtrack.jetbrains.com/issue/KT-34102. I can update all the links separately.

PR #1707, opened 2020-08-15
Generate Kotlin annotations for extension fields

swankjesse:


swankjesse:
This is the [interesting change only](https://github.com/square/wire/pull/1707/commits/c6b8515816c7664e8cadd2dc6180999545ad6230).

oldergod:
We generate them but don't assign them to their corresponding members?

swankjesse:
Follow up!

PR #1709, opened 2020-08-17
Fallback to stub protoadapters when java.time is missing

oldergod:
This allows clients, specially Android ones, not to be forced to
to address the issue if they don't have to. If a client is
actually using google.protobuf.Duration or g.p.Instant, the
generated code will not compile and clients will be able to
detect the missing parts at compile time, instead of having
the stub proto adapter failing at runtime.

### DIFF
--- a/wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/ProtoAdapter.kt
+++ b/wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/ProtoAdapter.kt
@@ -160,7 +160,7 @@ actual abstract class ProtoAdapter<E> actual constructor(
     @JvmField actual val value: Int,
     type: KClass<*>?
   ) : IllegalArgumentException("Unknown enum tag $value for ${type?.java?.name}") {
-    constructor(value: Int, type: Class<*>): this(value, type.kotlin)
+    constructor(value: Int, type: Class<*>) : this(value, type.kotlin)
   }
 
   actual companion object {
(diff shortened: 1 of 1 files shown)


wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -251,12 +251,37 @@ actual abstract class ProtoAdapter<E> actual constructor(
...
+
+    /**
+     * Stub [ProtoAdapter] for Wire types which are typeliased to `java.time` types on the JVM
+     * such as [Duration] and [Instant]. This proto adapter is used when the corresponding
+     * `java.time` type is missing from the JVM classpath.
+     */
+    class UnsupportedTypeProtoAdapter : ProtoAdapter<Nothing>(FieldEncoding.LENGTH_DELIMITED,
+        Nothing::class) {
+      override fun redact(value: Nothing) =
+          throw IllegalStateException("Operation not supported.")
```
JakeWharton, line 278: Feels like we should mention core library desugaring or minSdk 26 here.

oldergod:
I'll try setting up some tests.

JakeWharton:
For Retrofit I just copied a smoke JVM test into an Android module and added API 21, 24, 26, and 29 to the GitHub Action. You should be able to mostly do that here as well.

JakeWharton:
I also don't think it's a blocker. Test on Cash and if it works merge so it makes the beta release. We can land tests over time.

oldergod:
Yes, on Cash!

PR #1710, opened 2020-08-18
Apply annotations on Java messages and fields

swankjesse:


wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -2007,8 +2015,47 @@ public ClassName generatedTypeName(Field field) {
...
+      return false;
+    }
+
+    if (field.getPackageName().equals("google.protobuf")) {
+      return false; // Don't emit annotations for packed, etc.
+    }
+
+    if (field.getName().equals("redacted")) {
+      return false; // Redacted is built-in.
+    }
```
oldergod, line 2030: Probably need those 2 conditions in kotlinGenerator as well?

PR #1717, opened 2020-08-19
Emit compiler flags around Codable, Equatable, Hashable, and Redactable

JakeWharton:
This allows you to turn off features for a module which are not needed to save binary size.


wire-library/wire-tests-swift/src/main/swift/DeprecatedEnum.swift
```diff
@@ -10,9 +10,7 @@ public enum DeprecatedEnum : UInt32, CaseIterable, Codable {
     case OFF = 4
 
     public static var allCases: [DeprecatedEnum] {
-        get {
-            return [.DISABLED, .ENABLED, .ON, .OFF]
-        }
+        return [.DISABLED, .ENABLED, .ON, .OFF]
```
JakeWharton, line 13: Compact properties joined this change!

PR #1723, opened 2020-08-22
Configuration to turn on and off option annotations

swankjesse:
Currently declared options are emitted by default, and applied options
are not. In a follow-up release I'll turn everything on by default.
I've chosen this mode because it's easiest to upgrade. Otherwise you
can't upgrade downstream things without upgrading upstream things
first.

This required some changes to linking because we now need to recover
the package name of the file the option is declared in. Previously this
was not necessary.

### DIFF
--- a/wire-library/wire-gradle-plugin/src/test/projects/emit-java-options/src/main/proto/squareup/options/documentation.proto
+++ b/wire-library/wire-gradle-plugin/src/test/projects/emit-java-options/src/main/proto/squareup/options/documentation.proto
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2020 Square Inc.
+ *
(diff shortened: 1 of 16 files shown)


wire-library/wire-gradle-plugin/src/test/projects/emit-java-options/src/main/proto/squareup/options/documentation.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto2";
```
oldergod, line 1: copyright to new files?

swankjesse: Fixed.

PR #1726, opened 2020-08-26
Don't break line between lazy and {

oldergod:
This doesn't compile
image


swankjesse:
Yeah, sucks that significant whitespace is relevant in so many places.

PR #1728, opened 2020-08-26
Allow JSON read of field.name if name is a JVM keyword

oldergod:


wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/RuntimeMessageAdapter.kt
```diff
@@ -49,12 +49,19 @@ class RuntimeMessageAdapter<M : Message<M, B>, B : Builder<M, B>>(
...
    *
-   * In proto3 fields declared in snake_case like `customer_id` are written in camelCase like
-   * `customerId`, but can be read in either form. The alternate name will be absent if the snake
-   * and camel cases are the same, such as in single-word identifiers.
+   * For reserved keywords in Kotlin or Java, a field like `public` are written as so although the
+   * generated field name is `public_`. We wanna read in either form. As well, in proto3 fields
+   * declared in snake_case like `customer_id` are written in camelCase like `customerId`, but can
+   * be read in either form. We can use exclusive logic between the two cases because there's no
+   * keyword defined in scake_case. The alternate name will be absent if the field name isn't a
+   * keyword or if the snake and camel cases are the same, such as in single-word identifiers.
```
oldergod, line 57: I don't feel super happy about exclusively treat the two logic but providing an array of `jsonAlternameNames` which doesn't match the size of `jsonNames` seemed painful as well.

JakeWharton:
Who wins if I do
```proto
message Test {
  string final = 1;
  string final_ = 2;
}
```
?

oldergod:
I am not sure but the problem already exists right? they both would write `"final": "value"`.
I think that's the generated code by protoc for a schema like that will be broken as well since they camelCase every thing.
I don't think so it's worth worrying about.

oldergod:
Even Wire, I wonder what it would generate here.

wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/RuntimeMessageAdapter.kt
swankjesse: Yeah I like this observation!

swankjesse:
For follow-up at some point we should do a schema check on names that collide in surprising ways, including Jake’s example above.

wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/RuntimeMessageAdapter.kt
```diff
@@ -49,12 +49,19 @@ class RuntimeMessageAdapter<M : Message<M, B>, B : Builder<M, B>>(
...
    * name.
    *
-   * In proto3 fields declared in snake_case like `customer_id` are written in camelCase like
-   * `customerId`, but can be read in either form. The alternate name will be absent if the snake
-   * and camel cases are the same, such as in single-word identifiers.
+   * For reserved keywords in Kotlin or Java, a field like `public` are written as so although the
+   * generated field name is `public_`. We wanna read in either form. As well, in proto3 fields
+   * declared in snake_case like `customer_id` are written in camelCase like `customerId`, but can
+   * be read in either form. We can use exclusive logic between the two cases because there's no
+   * keyword defined in scake_case. The alternate name will be absent if the field name isn't a
```
Blisse, line 56: oops, `s/scake_case/snake_case`

oldergod: thank you, fixed

PR #1731, opened 2020-08-26
Throw if a proto2 enum is referenced in a proto3 message

oldergod:


wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
```diff
@@ -107,6 +109,12 @@ interface SyntaxRules {
       }
 
       override fun jsonName(name: String): String = camelCase(name, upperCamel = false)
+      override fun allowTypeReference(type: Type?): Boolean {
```
swankjesse, line 112: what does it mean if `type` is null? I’d prefer to not call the method at all in that case

oldergod: fixed

PR #1736, opened 2020-08-29
Option annotations on enum types and enum values

swankjesse:


wire-library/wire-tests/src/commonTest/proto-kotlin/com/squareup/wire/protos/custom_options/FooBar.kt
```diff
@@ -429,21 +430,26 @@ class FooBar(
     }
   }
 
+  @EnumOption(true)
   enum class FooBarBazEnum(
     override val value: Int,
     val enum_value_option: Int?,
     val complex_enum_value_option: More?,
     val foreign_enum_value_option: Boolean?
```
oldergod, line 438: Maybe it's a good time to stop generated those weird constructors for Enums

swankjesse: You don't have to ask me twice.

PR #1743, opened 2020-08-31
Fixing map tests

oldergod:


wire-library/wire-tests/src/commonTest/shared/json/all_64_max_proto3.json
```diff
@@ -44,5 +44,20 @@
     "9223372036854775807",
     "9223372036854775807"
   ],
-  "oneofInt64": "9223372036854775807"
+  "oneofInt64": "9223372036854775807",
+  "mapInt64Int64": {
+    "9223372036854775807": "9223372036854775807"
+  },
+  "mapInt64Uint64": {
+    "9223372036854775807": "9223372036854775807"
```
swankjesse, line 52: this is the max signed long . . . use the max unsigned long to exercise the difference?

wire-library/wire-tests/src/commonTest/shared/json/all_64_min_proto3.json
```diff
@@ -44,5 +44,20 @@
...
+    "-9223372036854775808": "-9223372036854775808"
+  },
+  "mapInt64Uint64": {
+    "-9223372036854775808": "9223372036854775808"
+  },
+  "mapInt64Sint64": {
+    "-9223372036854775808": "-9223372036854775808"
+  },
+  "mapInt64Fixed64": {
+    "-9223372036854775808": "9223372036854775808"
```
swankjesse, line 58: I would have expected this to be 0? unless we’re testing that it is backwards-compatible

oldergod: This is the JSON generated by protoc

oldergod: Right, will update what we test in followup

PR #1744, opened 2020-08-31
Include the importing file when an import fails

swankjesse:
This presents significantly larger errors because we include both
the failed imports and the failed resolution of the imported types.
Overall I think it's an improvement.

### DIFF
--- a/wire-library/wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt
+++ b/wire-library/wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt
@@ -890,6 +890,33 @@ class WireRunTest {
         .doesNotContain("@DocumentationUrl")
   }
 
+  @Test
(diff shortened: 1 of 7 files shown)


wire-library/wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt
```diff
@@ -890,6 +890,33 @@ class WireRunTest {
...
+
+    try {
+      wireRun.execute(fs, logger)
+      fail()
+    } catch (expected: SchemaException) {
+      assertThat(expected).hasMessage("""
+          |unable to find squareup/polygons/triangle.proto
+          |  searching 1 proto paths:
+          |    polygons/src/main/proto
+          |  for file colors/src/main/proto/squareup/colors/blue.proto
```
oldergod, line 912: :niiice:

PR #1745, opened 2020-08-31
Support for int32s in proto3

oldergod:
The work really happens in https://github.com/square/wire/commit/6c56a5982ff49cbe7d46cd7b926cf8963f02ba83


wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/JsonIntegration.kt
```diff
@@ -171,6 +173,25 @@ abstract class JsonIntegration<F, A> {
...
 
+  /** Encodes a unsigned value without quotes, like `123`. */
+  private object UnsignedIntAsNumberJsonFormatter : JsonFormatter<Int> {
+    // 2^32, used to convert sint32 values >= 2^31 to unsigned decimal form
+    private const val power32 = 1L shl 32
+
+    override fun fromString(value: String): Int {
+      val doubleValue = value.toDouble()
+      return if (doubleValue < 0.0) doubleValue.toInt()
+      else doubleValue.toUInt().toInt()
```
swankjesse, line 184: Do we use `UInt` elsewhere?

wire-library/wire-tests/src/jvmJsonTest/kotlin/com/squareup/wire/WireJsonTest.kt
```diff
@@ -423,6 +425,38 @@ class WireJsonTest {
     assertThat(unsignedJson).contains(""""repUint64":["456"]""")
   }
 
+  @Test fun `int32s are encoded as unsigned decoded with either`() {
```
swankjesse, line 428: I think the name is wrong? Should be uint32s ?

swankjesse: oh, no this is all of ’em. Works for me

wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/JsonIntegration.kt
JakeWharton: It's not ABI stable so we shouldn't use it regardless

oldergod: fixed

PR #1753, opened 2020-09-01
Stop generating google.protobuf types

oldergod:
We don't have to since we are not generating any fields for options anymore.


swankjesse:
Wow! Nice.

PR #1761, opened 2020-09-05
Use a compiler option to prevent crashes executing in Gradle

swankjesse:
It was crashing like this:

```
java.lang.NoSuchMethodError: 'void kotlin.jvm.internal.FunctionReferenceImpl.<init>(int, java.lang.Object, java.lang.Class, java.lang.String, java.lang.String, int)'
        at com.squareup.wire.kotlin.KotlinGenerator.<init>(KotlinGenerator.kt)
        at com.squareup.wire.kotlin.KotlinGenerator.redactFun(KotlinGenerator.kt:1244)
        at com.squareup.wire.kotlin.KotlinGenerator.addAdapter(KotlinGenerator.kt:1052)
```

See https://youtrack.jetbrains.com/issue/KT-37435

### DIFF
--- a/wire-library/build.gradle
+++ b/wire-library/build.gradle
@@ -48,6 +48,8 @@ subprojects { project ->
   tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach {
     kotlinOptions {
       jvmTarget = "1.8"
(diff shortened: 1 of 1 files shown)


JakeWharton (approved):

JakeWharton:
I was there Gandalf. I was there three thousand years ago when Gradle took the Kotlin. I was there the day the strength of Men failed.

PR #1764, opened 2020-09-08
Add a ProtoAdapter extension to decode NSData

Egorand:
Upstreaming from [wire-multiplatform-demo](https://github.com/Egorand/wire-multiplatform-sample/blob/master/protos/src/iosMain/kotlin/com/squareup/util/Okio.kt).

Example usage:

https://github.com/Egorand/wire-multiplatform-sample/blob/d4485faf3156c75d8bce2541a03470ad59ba01d2/ios/Dinosaurs/Dinosaurs/Resource.swift#L47-L49

Implementation borrowed from https://github.com/JetBrains/kotlin-native/issues/3172.

### DIFF
--- a/wire-library/wire-runtime/src/darwinMain/kotlin/com/squareup/wire/ProtoAdapter.kt
+++ b/wire-library/wire-runtime/src/darwinMain/kotlin/com/squareup/wire/ProtoAdapter.kt
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2020 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire
+
+import kotlinx.cinterop.addressOf
+import kotlinx.cinterop.usePinned
+import platform.Foundation.NSData
+import platform.posix.memcpy
+
+/**
+ * Read an encoded message from `data`.
+ *
+ * Note: this method is marked with [ExperimentalUnsignedTypes] annotation and requires an opt-in
+ * (e.g. through `@OptIn(ExperimentalUnsignedTypes::class)` to be used.
+ *
+ * @throws IllegalArgumentException if `data.length` is larger than [Int.MAX_VALUE].
+ */
+@ExperimentalUnsignedTypes
+fun <E> ProtoAdapter<E>.decode(data: NSData): E {
+  require(data.length <= Int.MAX_VALUE.toULong()) {
+    "Can't decode data with length ${data.length}, maximum length is ${Int.MAX_VALUE}."
+  }
+  val bytes = ByteArray(data.length.toInt()).apply {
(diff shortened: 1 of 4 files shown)


wire-library/wire-runtime/src/darwinMain/kotlin/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -0,0 +1,30 @@
...
+ */
+package com.squareup.wire
+
+import kotlinx.cinterop.addressOf
+import kotlinx.cinterop.usePinned
+import platform.Foundation.NSData
+import platform.posix.memcpy
+
+fun <E> ProtoAdapter<E>.decode(data: NSData): E {
+  val bytes = ByteArray(data.length.toInt()).apply {
```
JakeWharton, line 24: can we `require` the length be in bounds first rather than silently truncating?

wire-library/wire-runtime/src/darwinMain/kotlin/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -0,0 +1,30 @@
...
+    usePinned { pinned ->
+      memcpy(pinned.addressOf(0), data.bytes, data.length)
```
JakeWharton, line 26: Seems like we want an `NSData.asSink()` in Okio eventually.

wire-library/wire-runtime/src/darwinMain/kotlin/com/squareup/wire/ProtoAdapter.kt
JakeWharton: not sure if you can test that... does a 2.0000000001GiB empty NSData actually allocate?

JakeWharton (approved):

wire-library/wire-runtime/src/darwinMain/kotlin/com/squareup/wire/ProtoAdapter.kt
JakeWharton: asSource, obviously, dummy

wire-library/wire-runtime/src/darwinMain/kotlin/com/squareup/wire/ProtoAdapter.kt
Egorand: By truncating you mean going from unsigned to signed? There's [`dataWithCapacity`](https://developer.apple.com/documentation/foundation/nsmutabledata/1547236-datawithcapacity?language=objc) which doesn't allocate, but it looks like it sets `length` to 0.

JakeWharton: Yes. The `toInt()` to me indicates `NSData` can be larger than `ByteArray` so I want to make sure we throw in that case.

Egorand: Added the require, no tests cause there doesn't seem to be a way to create a large enough `NSData` without allocating a ton of memory. Also added the `@ExperimentalUnsignedTypes` annotation to the method since `NSData`'s `length` is an `NSUInteger`, which Kotlin converts to `ULong`. Does this work, or do we want to introduce our own opt in marker?

JakeWharton: This seems fine. Perhaps note it in the method docs (which should probably be added regardless).

Egorand: Yep, added the docs, will merge on green.

PR #1765, opened 2020-09-08
Rename AnyMessage.pack to Message.toAny

oldergod:
Couldn't find another name for `unpack`.


JakeWharton (approved):

JakeWharton:
> Couldn't find another name for `unpack`.

`decode`?

wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/AnyMessage.kt
```diff
@@ -15,21 +15,23 @@
...
 import kotlin.jvm.JvmField
+import kotlin.jvm.JvmName
+import kotlin.jvm.JvmStatic
 
 /**
  * Wire implementation of the `google.protobuf.Any` type. The `Any` type wraps an arbitrary
  * protobuf message. Field of `Any` generated by Wire will be typed [AnyMessage]. [typeUrl] is
  * used to identify the type of the message, defined as 'type.googleapis.com/full.type.name'.
- * [value] is the serialized representation of the wrapped message. The [pack] and [unpack] methods
+ * [value] is the serialized representation of the wrapped message. The [toAny] and [unpack] methods
```
swankjesse, line 28: We should find a replacement for `unpack` if we’re going to replace `pack`. . . . these methods are twins and should have complementary names!

oldergod: `fromAny`....

PR #1769, opened 2020-09-10
Suffix generated option annotations with 'Options'

swankjesse:
This avoids collisions when the options match the types they apply.


wire-library/wire-tests/src/commonTest/proto-kotlin/com/squareup/wire/protos/custom_options/EnumOptionOption.kt
```diff
@@ -13,6 +13,6 @@ import kotlin.annotation.Target
  */
 @Retention(AnnotationRetention.RUNTIME)
 @Target(AnnotationTarget.CLASS)
-annotation class EnumOption(
+annotation class EnumOptionOption(
```
JakeWharton, line 16: Yo dawg

JakeWharton (approved):

PR #1771, opened 2020-09-12
Changelog for 3.3.0

oldergod:
- Do we wanna add modules/manifest? And if yes, what do we wanna write?
- Do we wanna talk about the type mover?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,36 @@
 Change Log
 ==========
 
+Version 3.3.0
+-------------
+
+_2020-09-14_
+
+ * New: Proto3 support! This includes the new behaviors, the new types, and the JSON.
+ * New: Swift support for proto2 schemas. The details are in our [blog post][swiftblogpost].
+ * New: Wire will now throw an error when:
+   * two generated files end up overriding each other,
+   * imports form a cycle,
+   * packages form a cycle. This can be turned out with the flag `permitPackageCycles`,
+   * an option within the source set cannot be resolved,
+   * there are name duplications of members in a message, or of rpcs in a service,
+   * a map is used as an extension.
+ * New: Support for the `json_name` pseudo option.
+ * New: The `wire_package` file option allows one to set the JVM package where classes generated
+   from the concerned file will be placed. `wire_package` takes precedence over `java_package`.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,39 @@
 Change Log
 ==========
 
+Version 3.3.0
+-------------
+
+_2020-09-12_
+
+ * New: Proto3 support! This includes the new behaviors, the new types, and the JSON. Only for JVM
```
Egorand, line 9: Curious why Proto3 is JVM only, is that because of the duration types?

oldergod: I don't think proto3 has been addressed for Swift. I don't believe there are tests, JSON support, adequate equivalent generated type for the new protobuf things, etc. As far as I know anyway.

JakeWharton: Swift has lots of proto3 support. That aside, what about JS and native? Those should also have support already.

oldergod:
How about proto3 support for all. Json support for jvm?

Le dim. 13 sept. 2020 17 h 48, Jake Wharton <notifications@github.com> a
écrit :

CHANGELOG.md
```diff
@@ -1,6 +1,37 @@
...
+   * two generated files end up overriding each other,
+   * imports form a cycle,
+   * packages form a cycle. This can be turned out with the flag `permitPackageCycles`,
+   * an option within the source set cannot be resolved,
+   * there are name duplications of members in a message, or of rpcs in a service,
+   * a map is used as an extension.
+ * New: Support for the `json_name` pseudo option.
+ * New: The `wire_package` file option allows one to set the JVM package where classes generated
+   from the concerned file will be placed. `wire_package` takes precedence over `java_package`.
+ * New: Lists and maps in kotlin generated code are now immutable.
```
swankjesse, line 22: supernit: kotlin –> Kotlin

swankjesse (approved):

CHANGELOG.md
Egorand: FWIW I converted my demo to proto3 and it works flawlessly: https://github.com/Egorand/wire-multiplatform-sample/commit/872d9eee599f889005c0df51738a6a77322abb7e

PR #1775, opened 2020-09-14
Change readme's proto examples syntax to proto3

oldergod:
None intrusive way to say we promote it now for new comers.


Egorand (approved):

swankjesse (commented):
LGTM

Nice!

PR #1786, opened 2020-09-21
Pass errorCollector to SyntaxRules

oldergod:


wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/internal/parser/FieldElement.kt
```diff
@@ -55,9 +54,9 @@ data class FieldElement(
    * Both `default` and `json_name` are defined in the schema like options but they are actually
    * not options themselves as they're missing from `google.protobuf.FieldOptions`.
    */
-  private fun optionsWithSpecialValues(syntaxRules: SyntaxRules): List<OptionElement> {
+  private fun optionsWithSpecialValues(): List<OptionElement> {
     var options =
-        if (defaultValue == null || !syntaxRules.allowUserDefinedDefaultValue()) {
```
oldergod, line 60: Doesn't make sense to not parse it. The parsing shouldn't know about that and the linker, who does, will add an error for us. I so removed all syntax rules in the `toSchema` because not needed anymore.

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt
```diff
@@ -22,9 +22,11 @@ import com.squareup.wire.internal.camelCase
 
 /** A set of rules which defines schema requirements for a specific [Syntax]. */
 interface SyntaxRules {
-  fun allowUserDefinedDefaultValue(): Boolean
-  fun canExtend(protoType: ProtoType): Boolean
-  fun enumRequiresZeroValueAtFirstPosition(): Boolean
+  fun validateDefaultValue(hasDefaultValue: Boolean, errors: ErrorCollector)
```
swankjesse, line 25: I really like this structure

swankjesse (approved):

PR #1790, opened 2020-09-22
Make WireTask cacheable

### LINKED ISSUE #1784: [feature request] Make generateProtos Gradle task cacheable
The `generateProtos` task is not currently [cacheable](https://docs.gradle.org/current/userguide/build_cache.html#sec:task_output_caching_details). Adding cache support has the potential to save build time for users of Gradle's local and remote caches.

oldergod:
https://docs.gradle.org/current/userguide/build_cache.html#enable_caching_of_non_cacheable_tasks

fixes #1784 

Tested on our test module.
image


Egorand (approved):
Nice!

JakeWharton:
I like how everything good in Gradle is opt-in rather than opt-out and there's also zero validation that you do the right thing even when you opt-in.

Did I say like? No I hate it.

oldergod:
Maybe someday we could look at https://docs.gradle.org/current/userguide/test_kit.html

JakeWharton:
Yeah we use test kit in a bunch of other repos. It's okay. Testing things like incremental, caching, and cache relocatability, and caching across platforms is still really hard to do.

PR #1793, opened 2020-09-23
Fix enum values in geology period sample code

arturdryomov:
Relates to #1394.

Using the current sample produces the following error:

```
* What went wrong:
Execution failed for task ':module:generateProtos'.
> missing a zero value at the first element [proto3]
    for enum squareup.geology.Period (module/src/main/proto/squareup/geology/period.proto:7:1)
```

The `wire/sample` module uses PB 2 producing no error on builds. You might want to change it to PB 3 or have two samples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,13 +58,13 @@ option java_package = "com.squareup.geology";
 
 enum Period {
   // 145.5 million years ago — 66.0 million years ago.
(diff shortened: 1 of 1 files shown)


swankjesse:
Thanks!

PR #1812, opened 2020-09-30
Throw on ambiguous enums

### LINKED ISSUE #1805: We should fail on ambiguous enum constants as protoc do.

oldergod:
fix #1805

### DIFF
--- a/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/EnumType.kt
+++ b/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/EnumType.kt
@@ -59,7 +59,35 @@ data class EnumType(
     if ("true" != allowAlias) {
       validateTagUniqueness(linker)
     }
+    validateTagNameAmbiguity("true" == allowAlias, linker)
     syntaxRules.validateEnumConstants(constants, linker.errors)
+  }
+
+  private fun validateTagNameAmbiguity(allowAlias: Boolean, linker: Linker) {
+    val nameToConstants: Map<String, List<EnumConstant>> =
+        constants.groupBy {
+          buildString(it.name.length) {
+            for (char in it.name) {
+              if (char in 'A'..'Z') append(char - ('A' - 'a'))
+              else append(char)
+            }
+          }
+        }
+
+    for ((_, constants) in nameToConstants) {
+      if (constants.size > 1) {
+        if (allowAlias && constants.groupBy { it.tag }.size == 1) continue
+
+        val error = buildString {
+          append("Ambiguous constant names (if you are using allow_alias, use the same value for " +
+              "these constants):")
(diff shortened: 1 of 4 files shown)


wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/EnumType.kt
```diff
@@ -59,9 +60,30 @@ data class EnumType(
     if ("true" != allowAlias) {
       validateTagUniqueness(linker)
     }
+    validateTagNameAmbiguity("true" == allowAlias, linker)
     syntaxRules.validateEnumConstants(constants, linker.errors)
   }
 
+  private fun validateTagNameAmbiguity(allowAlias: Boolean, linker: Linker) {
+    val nameToConstants: Map<String, List<EnumConstant>> =
+        constants.groupBy { it.name.toEnglishLowerCase() }
```
swankjesse, line 69: What the heck is English? I think somebody means ASCII? English is definitely not stable or formally specified.

wire-library/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/SchemaTest.kt
```diff
@@ -2433,4 +2433,65 @@ class SchemaTest {
...
+            |  ZERO = 0;
+            |  zero = 1;
+            |}
+          """.trimMargin())
+          .schema()
+      fail()
+    } catch (e: SchemaException) {
+      assertThat(e).hasMessage(
+          """|Ambiguous constant names (if you are using allow_alias, use the same value for these constants):
+             |  1. ZERO (/source/message.proto:2:3)
```
swankjesse, line 2451: The 1. and 2. bullets in these is confusing and unnecessary? Maybe use the tag values instead?

swankjesse (approved):
Great tests.

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/EnumType.kt
oldergod: it uses `toLowerCase(Locale.US)` it's something we're already using in a few places.
I can do manual conversion here instead

wire-library/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/SchemaTest.kt
oldergod:
```
Ambiguous constant names (if you are using allow_alias, use the same value for these constants):
  ZERO:0 (/source/message.proto:3:3)
  zero:1 (/source/message.proto:4:3)
  for enum Foo (/source/message.proto:1:1)
```

swankjesse (approved):

PR #1813, opened 2020-09-30
Update camelcase to Protoc

### LINKED ISSUE #1801: protobuf changed definition of camelcase between 3.0.0 and 3.13.0
Protoc changed the JSON encoding of fields that start with an uppercase letter somewhere between 3.0.0 and 3.13.0.

```
message InteropCamelCase {
  optional string TitleCase = 5;
}
```

With 3.0.0:
```
 "{"titleCase":"hello"}"
```

With 3.13.0:
```
"{"TitleCase":"hello"}"
```

oldergod:
fix #1801 

This breaks behavior though. For `string EEee_ff_gGg = 4;`
Before, we would
- print `eEeeFfGGg` and
- read `eEeeFfGGg`.

now we
- cannot read `eEeeFfGGg`.

We could potentially just try to read all jsonNames we have by switching the case of the first letter...?


swankjesse (approved):

swankjesse:
I think we ship this only.

PR #1818, opened 2020-10-02
Add compact flag to toSchema

rayokota:
A recent PR https://github.com/square/wire/pull/1615 changed the format of `toSchema`.  This is a breaking change for Confluent Schema Registry (https://github.com/confluentinc/schema-registry).   This PR allows a client to request the old format in a way that is backward compatible with the recent change.

### DIFF
--- a/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/internal/parser/EnumElement.kt
+++ b/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/internal/parser/EnumElement.kt
@@ -30,7 +30,7 @@ data class EnumElement(
   // Enums do not allow nested type declarations.
   override val nestedTypes: List<TypeElement> = emptyList()
 
-  override fun toSchema() = buildString {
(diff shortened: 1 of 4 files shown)


rayokota:
@oldergod , @JakeWharton I kindly request your review.  Thank you very much in advance!

swankjesse (approved):
Wanna add a short test case so we don't inadvertently regress on this?

JakeWharton:
I am opposed to this change. I can explain later, I'm moving.

Also, ironically, this change is binary incompatible.

JakeWharton:
See https://github.com/square/wire/pull/1615#issuecomment-702993474

I meant to say "very opposed". I don't want to touch output byte-for-byte compatibility with a 20-foot pole.

swankjesse:
@rayokota is there anything in our API that prevents you from encoding these things externally? That way we don't have the awkward byte-for-byte constraint that is difficult to maintain, and you can still get out of the corner in which you're currently painted? Perhaps to promote visibility of some members?

In any case I'm closing this. Jake's right that committing ourselves to supporting this old encoded form is bad news.

Issue #1325, opened 2019-12-19
Unable to resolve import windows/android studio


johnnyzen:
Hi all

I have created another ticket, as the sample project kindly supplied by @Egorand does not work for me. I still get the same issue

`unable to resolve 1 imports`

Version 3.0.2
Windows 10
Android Studio

Example project here: 

https://github.com/Egorand/wire-tastes

Many thanks

Egorand:
Thanks! That's the link to my project though, please fix.

Would you be interested in implementing a fix? Most of us use Macs and since this is Windows-specific we won't be able to reproduce the issue.

johnnyzen:
I included your example project as thought it would be useful for someone to try on windows... but no-one has access to a windows machine? Okies, no problem, I am happy to try and fix myself, but I am not used to working with gradle plugins.. could you help me with how to turn the project debug mode on, as I can see from the source, the debug lists the paths, which would be helpful for me to try and work out whats going.

Egorand:
Whenever you build with `-i` or `-d` flags you should be seeing INFO or DEBUG output:
```
./gradlew generateProtos -d
```

johnnyzen:
Excellent!! thank you so much! I will see what I can do for everyone :)

johnnyzen:
Ok, need some help if anyone can assist:

with no sourcepath defined in build.gradle, the debugger outputs:

`wireSourceDependencies.files for dep: [C:\AndroidProjects\wire-tastes\src\main\proto]`

which seems like the default proto directory.

Which building I get:

`Cause: unable to resolve 1 imports:
  tastes/tastes.proto
searching 0 proto paths:`

BUT, when defining:

`sourcePath {
        srcDir 'src/main/proto'
    }`

The debug log shows: 

`wireSourceDependencies.files for dep: [C:\AndroidProjects\wire-tastes\src\main\proto\zen\people\person.proto, C:\AndroidProjects\wire-tastes\src\main\proto\zen\tastes\tastes.proto]`

So now its showing the the full path and filename of the proto.

But when building I get:

`Collection contains no element matching the predicate.` 

which I have no clue what this means??

So, can anyone assist with what looks right, showing and including the proto files and path in the `wireSourceDependencies.files` list or shouldnt it matter?

johnnyzen:
Also, there was a fix a while back https://github.com/square/wire/pull/1173

[...]

PR #1820, opened 2020-10-04
Fix location filtering for windows

SoMuchForSubtlety:
This should fix the `Collection contains no element matching the predicate.` error on windows described in https://github.com/square/wire/issues/1325#issuecomment-570569691

I believe imports across directories on windows will still not work because they also work based on file paths. Fixing that would require bigger changes in `wire-schema`.


oldergod:
Thanks

Issue #1817, opened 2020-10-02
Kotlin protos trip up on "open" keyword


yschimke:
Testing on this PR - https://github.com/yschimke/emulator-tools/pull/1

Source proto

```
message SocketControl {
    enum Sort {
        identity = 0;  // Indicates the identify of the channel on which this
                       // message is sent.
        open = 1;      // Request the client to open up a connection with the
                       // requested id.
        close = 2;     // Indicate that the channel with the given fd is to be
                    // closed. More specifically No new data will be placed on
                    // the wire after the sending of this message.
    }
```

yschimke:
Also getting a secondary error

e: C:\Users\yuri\workspace\emulator-tools\src\main\kotlin\com\android\emulator\control\LogMessage.kt: (187, 21): Type mismatch: inferred type is Unit but Serializable was expected

oldergod:
Fixed a while ago.

PR #1822, opened 2020-10-04
Testing with open keyword

yschimke:
Trying to reproduce a blocking issue https://github.com/square/wire/issues/1817


oldergod:
https://github.com/square/kotlinpoet/issues/991

PR #1824, opened 2020-10-05
Increase flexibility around proto keywords when parsing

### LINKED ISSUE #1797: Allow `service` for enum name
Service could not be declared inside enum block, but proto parser fails when encountering `service` as enum entry decalration.

Example:
```
enum ServiceType {
    service = 0; // parser fails, as it expects service declaration
    external = 1;
    mysql = 2;
    mdb_mysql = 3;
    conductor = 4;
    jenkins = 5;
}
```

oldergod:
We had a few cases where we would "teach" the consumer how to write protos, I don't think they're really valuable? so I removed them.

fix #1797

### DIFF
--- a/wire-library/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/internal/parser/ProtoParserTest.kt
+++ b/wire-library/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/internal/parser/ProtoParserTest.kt
@@ -2398,7 +2398,7 @@ class ProtoParserTest {
   }
 
   @Test fun protoKeywordAsEnumConstants() {
-    // Note: cases were protoc fails are commented out.
+    // Note: this is consistent with protoc.
     val proto = """
       |enum Foo {
       |  syntax = 0;
@@ -2440,12 +2440,44 @@ class ProtoParserTest {
     assertThat(ProtoParser.parse(location, proto)).isEqualTo(expected)
   }
 
-  @Test fun protoKeywordAsMessageNameAndField() {
-    // Note: cases were protoc fails are commented out.
+  @Test fun protoKeywordAsMessageNameAndFieldProto2() {
+    // Note: this is consistent with protoc.
     val proto = """
+      |message syntax {
+      |  optional syntax syntax = 1;
+      |}
+      |message import {
+      |  optional import import = 1;
+      |}
+      |message package {
+      |  optional package package = 1;
+      |}
+      |message option {
+      |  optional option option = 1;
+      |}
+      |message reserved {
+      |  optional reserved reserved = 1;
+      |}
       |message message {
       |  optional message message = 1;
-      |  // message message = 2;
+      |}
+      |message enum {
+      |  optional enum enum = 1;
+      |}
+      |message service {
+      |  optional service service = 1;
+      |}
+      |message extend {
+      |  optional extend extend = 1;
+      |}
+      |message rpc {
+      |  optional rpc rpc = 1;
+      |}
+      |message oneof {
+      |  optional oneof oneof = 1;
+      |}
+      |message extensions {
+      |  optional extensions extensions = 1;
       |}
       |""".trimMargin()
     val expected = ProtoFileElement(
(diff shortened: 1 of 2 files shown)


wire-library/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/internal/parser/ProtoParserTest.kt
```diff
@@ -618,21 +617,6 @@ class ProtoParserTest {
     }
   }
 
-  @Test
-  fun syntaxInWrongContextThrows() {
```
swankjesse, line 622: 👍

wire-library/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/internal/parser/ProtoParserTest.kt
```diff
@@ -2413,13 +2397,22 @@ class ProtoParserTest {
     assertThat(ProtoParser.parse(location, proto)).isEqualTo(expected)
   }
 
-  @Test @Ignore("Broken. See https://github.com/square/wire/issues/1797")
-  fun protoKeywordAsEnumConstants() {
+  @Test fun protoKeywordAsEnumConstants() {
+    // Note: cases were protoc fails are commented out.
```
swankjesse, line 2401: lets make Wire consistent with protoc here? Totally sucks to define an enum, use it with Wire, and then to find out your schema isn’t portable to protoc

wire-library/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/internal/parser/ProtoParserTest.kt
```diff
@@ -2429,16 +2422,62 @@ class ProtoParserTest {
...
+                    EnumConstantElement(location.at(13, 3), "extensions", 11),
+                )
+            )
+        )
+    )
+    assertThat(ProtoParser.parse(location, proto)).isEqualTo(expected)
+  }
+
+  @Test fun protoKeywordAsMessageNameAndField() {
+    // Note: cases were protoc fails are commented out.
```
swankjesse, line 2444: Is wire consistent with protoc?

wire-library/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/internal/parser/ProtoParserTest.kt
```diff
@@ -2429,16 +2422,62 @@ class ProtoParserTest {
...
+    val proto = """
+      |message message {
+      |  optional message message = 1;
```
swankjesse, line 2447: what about proto3?

swankjesse (approved):

wire-library/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/internal/parser/ProtoParserTest.kt
oldergod: I see. Need to make sure then fail in both or pass in both.

wire-library/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/internal/parser/ProtoParserTest.kt
oldergod: this is the one below commented out, it fails on both protoc/wire

wire-library/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/internal/parser/ProtoParserTest.kt
oldergod: And yes they are both behaving the same.

oldergod:
Added lot of tests. All is consistent with what Protoc does.

PR #1834, opened 2020-10-12
Replacing / with File.separator in wire-compiler

oldergod:
related to #1325


oldergod:
It's not that simple.

PR #1837, opened 2020-10-15
Use union-find for disjoint set algorithm

### LINKED ISSUE #1690: Use union-find for disjoint set algorithm

aperfilyev:
closes #1690


oldergod:
Sorry we didn't review that back then. I'll close but we can re-open if one is willing to complete it.

PR #1838, opened 2020-10-19
Fix deprecated funs

oldergod:
It's time.

### DIFF
--- a/wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
+++ b/wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
@@ -513,9 +513,7 @@ public final class JavaGenerator {
     throw new IllegalStateException("Unknown type: " + type);
   }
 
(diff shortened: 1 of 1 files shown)


Egorand:
Needs a major release?

oldergod:
3.5 is next

Egorand:
Major would be 4.0. Not sure though how strict we want to be with breaking the API in `JavaGenerator`, but strictly speaking this kind of change would require 4.0.

oldergod:
I don't think Wire is strict about these. We've broken APIs before without
doing the big change.
Those deprecated are also very old.

2020年10月19日(月) 11:10 Egor Andreevich <notifications@github.com>:

oldergod:
Nevertheless, we could also do a 4.0.

2020年10月19日(月) 11:28 Benoît Quenaudon <oldering@gmail.com>:

swankjesse:
I don’t expect anyone is running JavaGenerator outside of Wire itself. Anyone who is must upgrade to WireRun, as that’s the only thing we really want to maintain.

PR #1851, opened 2020-11-12
Very limited support for Kotlin Serialization

swankjesse:
This adds a new Gradle plugin option for Kotlin targets:

    kotlin {
      emitKotlinSerialization = "UNSUPPORTED"
    }

When used, generated code is annotated `@Serializable`. Using
this interface imposes many constraints on the target class.
To satisfy these constraints, generated messages do not
subclass Message with this option.

This is our best option to use serialization off JVM.


wire-library/wire-kotlin-serialization/src/commonMain/kotlin/com/squareup/wire/ByteStringSerializer.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.descriptors.PrimitiveKind
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import okio.ByteString
+import okio.ByteString.Companion.decodeBase64
```
oldergod, line 26: nice

oldergod (approved):

PR #1856, opened 2020-11-13
WIP: Sample of new handling for OneOf's

oldergod:


kimble:
Nice! This is also super useful for logging, tracing and metrics

PR #1860, opened 2020-11-26
Use WireOutput for task property

### LINKED ISSUE #1859: WireTask uses non-normalized paths and breaks Gradle build caching
After doing some simple local build caching tests with the same project in multiple locations, build scans revealed that `WireTask` is not up to date when it should be due to using non-normalized paths in its `transforms` property

Namely - `outDirectory`

Output from a simple snippet to print it

```kotlin
tasks.withType<WireTask>().configureEach {
  doFirst { println(transforms) }
}
```

```
KotlinTarget(includes=[*], excludes=[], exclusive=true, outDirectory=/Users/zsweers/dev/slack/slack-android-ng/telemetry-definitions/build/generated/source/wire, android=false, javaInterop=false, emitDeclaredOptions=true, emitAppliedOptions=false, rpcCallStyle=SUSPENDING, rpcRole=CLIENT, singleMethodServices=false)
```

ZacSweers:
Resolves #1859 by using WireOutput for the `WireTask` property and normalizing its path. I need to double check if `String` is valid for path normalizing though

### DIFF
--- a/wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireOutput.kt
+++ b/wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireOutput.kt
@@ -22,6 +22,9 @@ import com.squareup.wire.schema.JavaTarget
 import com.squareup.wire.schema.KotlinTarget
 import com.squareup.wire.schema.ProtoTarget
 import com.squareup.wire.schema.Target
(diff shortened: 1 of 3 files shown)


staktrace:
@ZacSweers do you have plans to finish this off? It would be nice to get this fixed for build caching purposes.

ZacSweers:
Sorry not at the moment. It's far enough out of date that it's probably best to start from scratch too

PR #1868, opened 2020-12-12
Windows support

### LINKED ISSUE #1325: Unable to resolve import windows/android studio

charleskorn:
This fixes #1325 (or at least does for the sample app). 

It also enables running the CI build on Windows to catch issues like this in the future. Sample builds are available at https://github.com/charleskorn/wire/actions?query=branch%3Awindows-support.

The issue appears to be that `Linker.fileLinkers` is used with paths in both OS-native path format (eg. with `\` separators on Windows) and source file format (eg. with `/` separators). If these match, there is no issue, but if they don't match (eg. loading a file with `/` separators in imports on Windows), imports aren't resolved correctly.

This may not be a complete fix - this was the bare minimum required to get things to work on Windows for the sample app, and the introduced `Linker.normaliseImportPath()` might not handle escape sequences correctly (and probably has other flaws). A more elegant solution may be to use a separator-agnostic type for the key to `Linker.fileLinkers`.

~~:warning: The build is currently failing because the 'check test files' step fails on Windows - each of the generated files has the source file name (eg. `differentpackage\bar.proto`) in the header, which is different on Windows due to the different path separator and so the check fails. I'm not sure what the best way to handle this is. There also seems to be some kind of encoding-related problem - [this](https://github.com/square/wire/pull/1868/checks?check_run_id=1543579818#step:5:1583) is just one example.~~ This is fixed, the multiplatform build is failing for the same reason as for `master`.

Very open to feedback or suggestions on how to improve this.


oldergod:
Thank you!

Maybe we could always print paths with `/` instead of `\` ? That'd be bad to have different outputs depending on which platform the code has been generated.

charleskorn:
I have this down to one failing Gradle plugin test. [`WirePluginTest.moveMessage`](https://github.com/charleskorn/wire/blob/windows-support/wire-library/wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt#L829) fails with:

```
Expecting:
 <"// squareup\dinosaurs\dinosaur.proto
syntax = "proto2";
package squareup.dinosaurs;

option java_package = "com.squareup.dinosaurs";

message Dinosaur {
  // Common name of this dinosaur, like "Stegosaurus".
  optional string name = 1;

  // URLs with images of this dinosaur.
  repeated string picture_urls = 2;

  optional double length_meters = 3;

  optional double mass_kilograms = 4;

  optional squareup.geology.Period period = 5;
}
">
to contain:
 <"import "squareup/geology/geology.proto";"> 

	at com.squareup.wire.gradle.WirePluginTest.moveMessage(WirePluginTest.kt:829)
        ...
```

Any ideas why the `import ...` line isn't getting emitted?

oldergod:
I'm not able to try locally for I don't have windows. If you feel like to, the debugging would happen around `WireRun#moves` and see how the path for the moves are handled.

charleskorn:
All of the tests now pass on Windows 🎉 The multiplatform build is failing for the same reason as for `master`.

I'm not 100% happy with this, I'm thinking of adding a `normalisedPath` property to `Location` that returns a `NormalisedPath` type to make the logic less clunky in a lot of places, but that would be quite a significant change - let me know what you think.

oldergod:
Charles, thanks again.
[Okio](https://github.com/square/okio) is in the middle of implementation some multiplatform filesystem. I think we'll try to use that in Wire first then rebase your PR on that to see what's left. We're hoping using Okio would fix a few problems for us for free.

charleskorn:
No worries, I think that makes a lot of sense, I'm not super happy with where this ended up, it's quite messy. 

If Okio had a type that represents a path that can be compared ignoring the directory separator used in the original string, that type could be used instead of `String` everywhere and that would make things much, much nicer.

oldergod:
We're still working on Okio. Wire will be able to get a migration really soon now.

PR #1874, opened 2021-01-04
Updating workflows for brew cask is dead

oldergod:
```
Error: Calling brew cask install is disabled! Use brew install [--cask] instead.
Error: Process completed with exit code 1.
```


Egorand (approved):

oldergod:
Failing because of #1875

Issue #1875, opened 2021-01-04
iOS CI is broken.


oldergod:
Example : https://github.com/square/wire/pull/1874/checks?check_run_id=1644532250
```
xcodebuild: error: SDK "/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk" cannot be located.
xcrun: error: unable to lookup item 'PlatformPath' in SDK '/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk'
> Task :wire-runtime-swift:compileTestSwift FAILED
```

Looking at
https://github.com/actions/virtual-environments/issues/2211
https://github.com/actions/virtual-environments/issues/2218

It seems that Xcode has been updated to 12.2 which doesn't include 10.15 SDK.

PR #1877, opened 2021-01-05
Implement JSON for boxed oneofs.

swankjesse:
Kotlin only.


oldergod:
All green but multiplatform because of #1875

oldergod (approved):

PR #1884, opened 2021-01-27
New protoLibrary option for the Gradle plugin

swankjesse:
Configuring a project as a protoLibrary will cause the generated
.jar file to include .proto sources.

Also automatically add a dependency when a protoPath or protoSource
depends on a project.

Also make protoPath and protoSource dependencies transitive by default.

### DIFF
--- a/wire-library/docs/wire_compiler.md
+++ b/wire-library/docs/wire_compiler.md
@@ -292,6 +292,21 @@ wire {
 }
 ```
 
+### Proto Libraries
+
+By default, `.proto` input files are not included in the generated `.jar` artifact. Use the
+`protoLibrary` option to include them:
+
+```groovy
+wire {
+  protoLibrary = true
+}
+```
+
+This is most useful when building `.jar` files for other `wire` tasks to use as dependencies. Note
+that only the true sources are included – proto messages that are pruned or not used are not
+included in the output artifact.
+
 Customizing Output
 ------------------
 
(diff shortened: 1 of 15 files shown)


wire-library/docs/wire_compiler.md
```diff
@@ -292,6 +292,21 @@ wire {
...
 ```
 
+### Proto Libraries
+
+By default, `.proto` input files are not included in the generated `.jar` artifact. Use the
+`protoLibrary` option to include them:
+
+```groovy
+wire {
+  protoLibrary = true
```
oldergod, line 302: Great idea

wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -90,6 +95,17 @@ class WirePlugin : Plugin<Project> {
       it.description = "Aggregation task which runs every generation task for every given source"
     }
 
+    if (extension.protoLibrary) {
+      val libraryProtoSources = File("${buildDir}/wire/proto-sources")
+      val sourceSets = project.extensions.getByType(SourceSetContainer::class.java)
+      sourceSets.getByName("main") { main: SourceSet ->
+        main.resources.srcDir(libraryProtoSources)
+      }
```
oldergod, line 103: I wonder if that would work on Android with variants, but I guess we don't expect Android modules to be proto libraries anyway.
I can check later to see what happens. We could just log a warning if `main` isn't found.

oldergod (approved):
Super nice thank you

jrodbx:
nice!

PR #1887, opened 2021-01-29
Fix Swift WriteBuffer memory leak.

pickens117:
When decoding proto data to Swift and the proto has an unknown field, the `WriteBuffer` class will leak. 

This leak is caused by this [`private var messageStack: UnsafeMutablePointer<MessageFrame>`](https://github.com/square/wire/blob/48e6c9534158004d6aca5a0680ba11aad685bb5a/wire-library/wire-runtime-swift/src/main/swift/ProtoReader.swift#L73) variable in `ProtoReader`  
 
The `messageStack's` `Pointee` is the [`MessageFrame`](https://github.com/square/wire/blob/48e6c9534158004d6aca5a0680ba11aad685bb5a/wire-library/wire-runtime-swift/src/main/swift/ProtoReader.swift#L46) struct.

`MessageFrame's` `WriteBuffer` property is a class/reference type.

`UnsafeMutablePointer’s` [`deallocate`](https://developer.apple.com/documentation/swift/unsafemutablepointer/2949213-deallocate) function is only able to deallocate memory block's of trivial types.

### DIFF
--- a/wire-library/wire-runtime-swift/src/main/swift/ProtoReader.swift
+++ b/wire-library/wire-runtime-swift/src/main/swift/ProtoReader.swift
@@ -103,6 +103,13 @@ public final class ProtoReader {
     }
 
     deinit {
+         //If decoding fails for any reason, there may be unknownFields to
+         //clean up that were not caught by the defer block in endMessage
+        if messageStackIndex > 0 {
+            for index in 0...messageStackIndex {
+                messageStack[index].unknownFields = nil
+            }
+        }
         messageStack.deallocate()
     }
 
@@ -220,6 +227,13 @@ public final class ProtoReader {
             throw ProtoDecoder.Error.invalidStructure(
                 message: "Expected to end message at at \(frame.messageEnd - buffer.start), but was at \(buffer.position)"
             )
+        }
+        
+        //nil out unknownFields to avoid memory leaks caused by
+        //UnsafeMutablePointer's inability to deallocate non-trivial types.
+        //This is caused by unknownFields type, WriteBuffer, being a class/reference type.
+        defer {
+            messageStack[token].unknownFields = nil
         }
 
         return frame.unknownFields.flatMap { Data($0, copyBytes: false) } ?? Data()


wire-library/wire-runtime-swift/src/main/swift/ProtoReader.swift
```diff
@@ -221,6 +228,13 @@ public final class ProtoReader {
                 message: "Expected to end message at at \(frame.messageEnd - buffer.start), but was at \(buffer.position)"
             )
         }
+        
+        //nil out unknownFields to avoid  memory leaks caused by
```
efirestone, line 232: typo: double space after "avoid"

wire-library/wire-runtime-swift/src/main/swift/ProtoReader.swift
```diff
@@ -221,6 +228,13 @@ public final class ProtoReader {
...
+        //UnsafeMutablePointer inability to deallocate non-trivial types.
```
efirestone, line 233: typo: `UnsafeMutablePointer*'s*`

efirestone (approved):

wire-library/wire-runtime-swift/src/main/swift/ProtoReader.swift
```diff
@@ -103,6 +103,13 @@ public final class ProtoReader {
     }
 
     deinit {
+         //If decoding fails for any reason, there may be unknownFields to
+         //clean up that were not caught by the defer block in endMessage
+        if messageStackIndex > -1 {
```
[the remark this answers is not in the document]
pickens117: I went down that route but most of the time `messageStackIndex` is `-1`. That would cause the range to throw a `Can't form Range with upperBound < lowerBound` exception.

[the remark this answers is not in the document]
efirestone: Ah, good to know. LGTM then.

wire-library/wire-runtime-swift/src/main/swift/ProtoReader.swift
pickens117: Fixed

PR #1893, opened 2021-02-10
Name boxed oneof keys with constant case.

oldergod:


wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -1970,11 +1973,31 @@ class KotlinGenerator private constructor(
     }
   }
 
+  /**
+   * Generates a class for this boxed oneof.
```
swankjesse, line 1977: Nice comment

wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/Internal.kt
```diff
@@ -203,10 +203,21 @@ fun sanitize(value: String): String {
 
 /** Return a string where `,[]{}\` are escaped with a `\`. */
 fun sanitize(values: List<String>): String {
-  return values.map(::sanitize).joinToString(prefix = "[", postfix = "]")
+  return values.joinToString(prefix = "[", postfix = "]", transform = ::sanitize)
```
swankjesse, line 206: neat

swankjesse (approved):

PR #1894, opened 2021-02-11
Some doc for dependencies between gradle modules

### LINKED ISSUE #1826: Document the proper way to use protoPath for protos defined in a module dependency
I have a [project](https://github.com/jdkoren/proto-deps) with two modules, `app` and `lib`. Both modules use Wire and define protos; moreover, `app` depends on `lib` and the proto in `app` wants to use definitions from a proto in `lib`.

Looking at the [documentation](https://square.github.io/wire/wire_compiler/#proto-path-for-libraries), `protoPath` does not appear to  support something like `project(':lib')`, as this is neither "a local directory, a local .jar file, [n]or an external artifact specified with Maven coordinates."

Currently I work around this by exploiting the fact that I know the directory structure of the project to located the other module's protos: [`app/build.gradle`](https://github.com/jdkoren/proto-deps/blob/master/app/build.gradle#L13).

What is the proper way to use `protoPath` to identify protos in another module of the same project?

oldergod:
fix #1826

### DIFF
--- a/wire-library/docs/wire_compiler.md
+++ b/wire-library/docs/wire_compiler.md
@@ -215,6 +215,46 @@ The source path and proto path are linked together but only types on the source 
 
 ![Library](images/gradle_library@2x.png)
 
+Dependencies between Gradle Modules
+-----------------------------
+
+Wire provides support to define dependencies between modules within the same project.
+
+A module can include its `.proto` files into the output resources. Use this when your `.jar` file
+can be used as a library for other proto or Wire projects. Note that only the `.proto` files used
+in the library will be included.
+
+```groovy
+wire {
+  protoLibrary = true
+}
+```
+
+Wire also creates two configurations, `protoPath` and `protoSource` you can use to define a
+dependency on another proto or Wire project.
+
+```groovy
+dependencies {
+  // The task `:common-protos:jar` will be added into the dependency
+  // graph of this module for the Wire generating tasks.
+  protoPath(project(':common-protos'))
+  implementation(project(':common-protos'))
+}
+
+wire {
+  kotlin {
+  }
+}
+```
+
+Note that `protoPath` and `protoSource` dependencies are not transitive by default. If needed, you
+can change it manually.
+
+```groovy
+configurations.protoPath {
+  transitive = true
+}
+```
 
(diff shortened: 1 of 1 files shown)


wire-library/docs/wire_compiler.md
```diff
@@ -215,6 +215,46 @@ The source path and proto path are linked together but only types on the source
 
 ![Library](images/gradle_library@2x.png)
 
+Dependencies between Gradle Modules
+-----------------------------
+
+Wire provides supports to define dependencies between modules within the same project.
```
swankjesse, line 221: supports –> support

wire-library/docs/wire_compiler.md
```diff
@@ -215,6 +215,46 @@ The source path and proto path are linked together but only types on the source
...
+
+A module can include its `.proto` files into the output resources. One would use this when their
```
swankjesse, line 223: One would use this with their –> Use this with your

wire-library/docs/wire_compiler.md
```diff
@@ -215,6 +215,46 @@ The source path and proto path are linked together but only types on the source
...
+`.jar` file can be used as a library for other proto or Wire projects. Note that only the `.proto`
+files used in the library will be included, and these files will have tree-shaking applied.
```
swankjesse, line 225: Maybe omit the “... and these” section, since tree shaking isn’t interesting for library dependencies

wire-library/docs/wire_compiler.md
```diff
@@ -215,6 +215,46 @@ The source path and proto path are linked together but only types on the source
...
+```
+
+Wire also creates two configurations, `protoPath` and `protoSource` one can use to define a
+dependency on another proto or Wire project.
+
+```groovy
+dependencies {
+  // The task `:common-protos:jar` will be added into the dependency
+  // graph of this module for the Wire generating tasks.
+  protoPath project(':common-protos')
```
swankjesse, line 240: wanna switch to `protoPath(project(":common-protos"))` which is syntax that works in both build.gradle and build.gradle.kts ?

swankjesse (approved):

PR #1897, opened 2021-02-12
Lenient parsing of bool fields

### LINKED ISSUE #1469: lenient parsing of bool
currently square fails if the bool is not 0 or 1:
```
  @Throws(IOException::class)
  override fun decode(reader: ProtoReader): Boolean = when (val value = reader.readVarint32()) {
    0 -> false
    1 -> true
    else -> throw IOException("Invalid boolean value 0x" + value.toString(16).padStart(2, '0'))
  }
```
which fails for a device I am talking to currently. I am switching from protobuf-java that does not throw in this case - see: https://github.com/protocolbuffers/protobuf/blob/5b232b8ecbce13286be09e703997e887ae0d464d/java/core/src/main/java/com/google/protobuf/BinaryReader.java#L205

I see 2 ways to approach this:
 - allow lenient reading e.g. with a flag
 - read bool lenient by default

oldergod:
fix #1469

Protoc considers any int32 but 0 to be true.
https://github.com/protocolbuffers/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/BinaryReader.java#L202-L206

What do you think?


ligi:
Thanks for the fix oldergod & for the merge @swankjesse

ligi:
Thanks for the fix oldergod & for the merge @swankjesse

PR #1899, opened 2021-02-25
Fully support Optional Int64 and UInt64 for JSONString

patmalt:
> Test Case '-[Wire_iOS_Unit_Tests.PerformanceTests testLargeMessages]' measured [Time, seconds] average: 187.346, relative standard deviation: 2.415%, values: [195.824060, 190.312459, 181.115462, 189.611396, 185.115898, 184.345728, 193.289135, 186.988711, 183.481014, 183.372329], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "Local Baseline", baselineAverage: 184.580, maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100

> Test Case '-[Wire_iOS_Unit_Tests.PerformanceTests testSmallMessages]' measured [Time, seconds] average: 0.908, relative standard deviation: 4.467%, values: [0.995099, 0.866084, 0.883802, 0.887528, 0.880648, 0.910776, 0.855597, 0.925192, 0.954589, 0.919320], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "Local Baseline", baselineAverage: 0.948, maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100

> Test Case '-[Wire_iOS_Unit_Tests.PerformanceTests testVaryingMessageSizes]' measured [Time, seconds] average: 1.593, relative standard deviation: 3.746%, values: [1.578649, 1.531343, 1.544598, 1.598733, 1.565962, 1.726319, 1.557558, 1.684333, 1.580753, 1.561848], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "Local Baseline", baselineAverage: 1.685, maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100

### DIFF
--- a/wire-library/wire-runtime-swift/src/main/swift/JSONString.swift
+++ b/wire-library/wire-runtime-swift/src/main/swift/JSONString.swift
@@ -24,20 +24,14 @@ public struct JSONString<T : Hashable> : Hashable, Codable {
 
     public init(from decoder: Decoder) throws {
         switch T.self {
-        case is Int64.Type, is Optional<Int64>.Type:
-            let container = try decoder.singleValueContainer()
-            if let value = try? container.decode(String.self) {
-                self.wrappedValue = Int64(value)! as! T
-            } else {
-                self.wrappedValue = Optional<Int64>.none as! T
-            }
-        case is UInt64.Type, is Optional<UInt64>.Type:
-            let container = try decoder.singleValueContainer()
-            if let value = try? container.decode(String.self) {
-                self.wrappedValue = UInt64(value)! as! T
-            } else {
-                self.wrappedValue = Optional<UInt64>.none as! T
-            }
+        case is Int64.Type:
+            self.wrappedValue = try decoder.decodeStringTo(Int64.self)
+        case is Optional<Int64>.Type:
+            self.wrappedValue = try decoder.decodeStringIfPresentTo(Int64.self)
+        case is UInt64.Type:
+            self.wrappedValue = try decoder.decodeStringTo(UInt64.self)
+        case is Optional<UInt64>.Type:
+            self.wrappedValue = try decoder.decodeStringIfPresentTo(UInt64.self)
         case is [Int64].Type:
             var container = try decoder.unkeyedContainer()
             var array: [Int64] = []
@@ -83,4 +77,22 @@ public struct JSONString<T : Hashable> : Hashable, Codable {
         }
     }
 
+}
+
+private extension Decoder {
+    func decodeStringTo<T, Integer: FixedWidthInteger>(_ type: Integer.Type) throws -> T {
+        let container = try singleValueContainer()
+        let value = try container.decode(String.self)
+        return Integer(value)! as! T
+    }
+
+    func decodeStringIfPresentTo<T, Integer: FixedWidthInteger>(_ type: Integer.Type) throws -> T {
+        let container = try singleValueContainer()
+        if container.decodeNil() {
+            return Optional<Integer>.none as! T
+        } else {
+            let value = try container.decode(String.self)
+            return Integer(value)! as! T
+        }
+    }
 }
--- a/wire-library/wire-runtime-swift/src/main/swift/JSONString.swift
+++ b/wire-library/wire-runtime-swift/src/main/swift/JSONString.swift
@@ -25,13 +25,15 @@ public struct JSONString<T : Hashable> : Hashable, Codable {
     public init(from decoder: Decoder) throws {
         switch T.self {
         case is Int64.Type:
-            self.wrappedValue = try decoder.decodeStringTo(Int64.self)
+            let container = try decoder.singleValueContainer()
+            self.wrappedValue = Int64(try container.decode(String.self))! as! T
         case is Optional<Int64>.Type:
-            self.wrappedValue = try decoder.decodeStringIfPresentTo(Int64.self)
+            self.wrappedValue = try decoder.decodeStringIfPresentToInt64()
         case is UInt64.Type:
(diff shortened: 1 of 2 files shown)


wire-library/wire-runtime-swift/src/main/swift/JSONString.swift
```diff
@@ -26,10 +26,18 @@ public struct JSONString<T : Hashable> : Hashable, Codable {
         switch T.self {
         case is Int64.Type, is Optional<Int64>.Type:
             let container = try decoder.singleValueContainer()
-            self.wrappedValue = Int64(try container.decode(String.self))! as! T
+            if let value = try? container.decode(String.self) {
```
efirestone, line 29: We're possibly swallowing real errors here now. Like if the value were actually (unexpectedly) an array, we would set the wrapped value to `.none` and carry on when we should throw the error. A few options I see:
1. Try explicitly decoding an optional if the type is optional, like `try container.decode(Optional<Int64>.self)`. Not totally sure if this works, but if it does that's ideal.
2. Figure out what error gets thrown for `null` and explicitly handle that error as a `.none` optional
3. Try `decodeNil` first, then if that fails use the old logic. Since `decodeNil` will fail in fewer ways we can be relatively sure it's failing because the value is not-null

Note that for a few of these it may make sense to split the optional and non-optional case statements out rather than grouping them like we were doing previously

wire-library/wire-runtime-swift/src/main/swift/JSONString.swift
```diff
@@ -66,8 +68,31 @@ public struct JSONString<T : Hashable> : Hashable, Codable {
...
+            } else {
+                fatalError("Unsupported type \(T.self)")
+            }
         }
     }
 
 }
+
+private extension Decoder {
+    func decodeStringTo<T, Integer: FixedWidthInteger>(_ type: Integer.Type) throws -> T {
```
efirestone, line 83: Performance nit: Speed matters in Wire and generics add additional function call overhead. Since these are both pretty small and we just have two types (`Int64` and `UInt64`), can we copy/paste these methods and have a variant for both types?

I know it seems crazy, but generics definitely added to execution time in my profiling.

wire-library/wire-runtime-swift/src/main/swift/JSONString.swift
patmalt: Option 1 does not work, and fails with the following error: `"Expected to decode Int64 but found a string/data instead.", underlyingError: nil))"` because the Ints are in actuality Strings. Opted for option 3 with extension functions on Decoder to make it look like option 1 at the call site.

wire-library/wire-runtime-swift/src/main/swift/JSONString.swift
efirestone, line 83: Naming nit: Move the "to" into the parameter name to be more Swifty, like `decodeString(to type: Int64.Type) throws -> Int64`

wire-library/wire-runtime-swift/src/main/swift/JSONString.swift
```diff
@@ -66,8 +68,31 @@ public struct JSONString<T : Hashable> : Hashable, Codable {
                 try container.encode(String(value))
             }
         default:
-            fatalError("Unsupported type \(T.self)")
+            if T.self is Optional<Int64>.Type || T.self is Optional<UInt64>.Type {
```
efirestone, line 71: nit: It's weird to have this in the default case. Can we add these as regular cases above, like:
```
case Optional<Int64>.none, Optional<UInt64>.none:
   var container = encoder.singleValueContainer()
   try container.encodeNil()
default:
  fatalError("Unsupported type \(T.self)")
```

efirestone (approved):
Some small cleanup, but this looks great. Thank you!

wire-library/wire-runtime-swift/src/main/swift/JSONString.swift
patmalt: Unforunately, this was the best place I could think to place this code. We cannot cast `T` as an optional due to the following error: `Cannot downcast from 'T' to a more optional type 'Optional<Int64>'`.

The logic flow is check if there is a non-nil value, then check the type to see if it is optional of the allowed int values. Otherwise, the type is not supported.

wire-library/wire-runtime-swift/src/main/swift/JSONString.swift
```diff
@@ -66,8 +71,35 @@ public struct JSONString<T : Hashable> : Hashable, Codable {
...
+            } else {
+                fatalError("Unsupported type \(T.self)")
+            }
         }
     }
 
 }
+
+private extension Decoder {
+    func decodeStringIfPresentToInt64<T>() throws -> T {
```
efirestone, line 86: These don't need to be generic anymore, right? They can just return `Int64`.

wire-library/wire-runtime-swift/src/main/swift/JSONString.swift
patmalt: Right

wire-library/wire-runtime-swift/src/main/swift/JSONString.swift
patmalt: Updated

PR #1900, opened 2021-02-26
Properly serialize enums to JSON in Swift implementation

efirestone:
Match the JSON spec for serialization of enums: https://developers.google.com/protocol-buffers/docs/proto3#json

### DIFF
--- a/wire-library/wire-runtime-swift/src/main/swift/JSONEnum.swift
+++ b/wire-library/wire-runtime-swift/src/main/swift/JSONEnum.swift
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2021 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+/**
+ Converts enums to/from their string equivalent when serializing to/from JSON.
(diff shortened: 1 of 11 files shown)


wire-library/wire-runtime-swift/src/main/swift/JSONEnum.swift
```diff
@@ -0,0 +1,54 @@
...
+    public init(from decoder: Decoder) throws {
+        // We support decoding from either the string value or the field index.
+        let container = try decoder.singleValueContainer()
+        do {
+            let string = try container.decode(String.self)
+            guard let value = T.allCases.first(where: { "\($0)" == string }) else {
+                throw ProtoDecoder.Error.unknownEnumString(type: T.self, string: string)
+            }
+            self.wrappedValue = value
+        } catch {
```
JakeWharton, line 40: Won't this catch the ProtoDecoder.Error that's thrown? Is this meant to catch only the decoding as string step?

wire-library/wire-runtime-swift/src/main/swift/JSONEnum.swift
```diff
@@ -0,0 +1,54 @@
...
+            let fieldNumber = try container.decode(UInt32.self)
+            guard let value = T.init(rawValue: fieldNumber) else {
+                throw ProtoDecoder.Error.unknownEnumCase(type: T.self, fieldNumber: fieldNumber)
+            }
+            self.wrappedValue = value
+        }
+    }
+
+    public func encode(to encoder: Encoder) throws {
+        try "\(wrappedValue)".encode(to: encoder)
```
JakeWharton, line 51: Is this efficient? I remember in Kotlin that things like `"$this"` are short but wind up creating more machinery than the slightly-more verbose `this.toString()`.

wire-library/wire-runtime-swift/src/main/swift/JSONEnum.swift
JakeWharton: Can we add test cases for these enum string mismatch and enum uint mismatch cases to validate the right error is propagated?

JakeWharton (approved):

oldergod:
@efirestone That'd be sweet to finish this one

PR #1902, opened 2021-03-03
Don't include all protos if we have a custom include

oldergod:
I'm really surprised it hasn't been a problem since we released the plugin so my confidence level isn't that high.
But _tests don't lie_.

image


swankjesse (approved):

wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
```diff
@@ -206,8 +206,11 @@ open class WireExtension(project: Project) {
       // map to SourceDirectorySet which does the work for us!
       val protoTree = objectFactory.sourceDirectorySet(name, "Wire proto sources for $name.")
       protoTree.srcDirs(protoRootSet.srcDirs)
-      protoTree.filter.include("**/*.proto")
-      protoTree.filter.include(protoRootSet.includes)
+      if (protoRootSet.includes.isEmpty()) {
+        protoTree.filter.include("**/*.proto")
+      } else {
+        protoTree.filter.include(protoRootSet.includes)
```
jrodbx, line 212: nice

PR #1905, opened 2021-03-10
reflector: Add initial gRPC reflection implementation

juliaogris:
Add initial implementation and basic unit test implementing the [grpc
Reflection API](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md). 
Outstanding work:

- a lot more tests
- generate reflection/v1alpha as part of build


juliaogris:
This has since been replaced.

PR #1908, opened 2021-03-12
Defer enforcement of source existence to execution time

shirlsgaw:
The configuration state of the WireInput does a sanity check for whether or not specified source exists.

Unfortunately, this strictly enforces dependency checks rather than the expected lazy configuration.  Because the Wire plugin is using Gradle's tree filtering, by the time the WireTask executes, `WireTask.{sourceInput, protoInput}` have been filtered to only the ones that exist.

This PR passes the state of inputs that the strict enforcement would have asserted on to the task execution time, which is effectively is a lazy enforcement.

Note: this still does not resolve the case where the source directory exists and the `include` does not, but I am not sure this is handled before anyways.

### DIFF
--- a/wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt
+++ b/wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt
@@ -43,6 +43,9 @@ internal class WireInput(var configuration: Provider<Configuration>) {
   val dependencies: DependencySet
     get() = configuration.get().dependencies
 
+  // Deferred dependency evaluation
+  val suspectFiles = mutableListOf<File>()
+
   fun addPaths(project: Project, paths: Set<String>) {
     for (path in paths) {
       val dependency = resolveDependency(project, path)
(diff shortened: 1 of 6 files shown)


swankjesse (approved):

wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt
```diff
@@ -43,6 +43,9 @@ internal class WireInput(var configuration: Provider<Configuration>) {
   val dependencies: DependencySet
     get() = configuration.get().dependencies
 
+  // Deferred dependency evaluation
+  val suspectFiles = mutableListOf<File>()
```
swankjesse, line 47: rename to input files ?

swankjesse (approved):

wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt
shirlsgaw: Done thanks!

swankjesse (approved):

swankjesse:
Nice. Thanks!

PR #1909, opened 2021-03-16
Support FieldOptions with a `.` qualified prefix

shirlsgaw:
https://developers.google.com/protocol-buffers/docs/overview?hl=en#packages_and_name_resolution

The `.foo.bar.Baz` resolution described was not handled for FieldOptions.  Fields were able to resolve.

### DIFF
--- a/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Options.kt
+++ b/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Options.kt
@@ -101,7 +101,6 @@ class Options(
         // Retry with one upper level package to resolve relative paths.
         if (path == null) {
           packageName = packageName.substringBeforeLast(".", missingDelimiterValue = "")
-
           if (packageName.isNullOrBlank() && !checkedExtensionFields) {
             checkedExtensionFields = true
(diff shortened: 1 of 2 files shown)


wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Options.kt
```diff
@@ -406,22 +405,30 @@ class Options(
...
-      fullyQualifiedNames: Set<String?>
+            name: String,
+            fullyQualifiedNames: Set<String?>
     ): Array<String>? { // Try to resolve a local name.
       var pos = 0
-      while (pos < name.length) {
-        pos = name.indexOf('.', pos)
-        if (pos == -1) pos = name.length
-        val candidate = name.substring(0, pos)
+      val chompedName = if (name.startsWith('.')) {
```
swankjesse, line 420: name.removePrefix(".")

swankjesse (approved):

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Options.kt
shirlsgaw: https://github.com/square/wire/pull/1911

PR #1911, opened 2021-03-18
replace startsWith string manipulation with removePrefix

shirlsgaw:
As title per suggestion from PR https://github.com/square/wire/pull/1909


swankjesse (approved):

swankjesse:
Thanks for following up

PR #1913, opened 2021-03-19
Introduce srcProject API

oldergod:


wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt
```diff
@@ -80,27 +90,32 @@ internal class WireInput(var configuration: Provider<Configuration>) {
...
-      inputFiles.add(file)
+      if (!path.mayBeProject) inputFiles.add(file)
 
       return when {
         file.isDirectory -> project.dependencies.create(project.files(path))
         file.isJar -> project.dependencies.create(project.files(file.path))
+        path.mayBeProject -> {
+          // Keys can be either `path` or `configuration`.
+          // Example: "[path: ':someProj', configuration: 'someConf']"
+          return project.dependencies.project(mutableMapOf("path" to path))
```
oldergod, line 101: This would fail with something like
`Project with path ':bad-name' could not be found in project ':mobile:android:investing-protos'`
so I think it's fine if we don't try/catch it.

wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt
```diff
@@ -159,4 +176,7 @@ internal class WireInput(var configuration: Provider<Configuration>) {
 
   private val File.isJar
     get() = path.endsWith(".jar")
+
+  private val String.mayBeProject: Boolean
+    get() = startsWith(":")
```
oldergod, line 181: Does it sound good enough? Could we ever have another thing that a project starting with `:` ?

swankjesse (approved):

wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt
swankjesse: Seems a bit clumsy. Lets go with it for now; maybe make it fancier later if there’s problems adopting this within Square.

PR #1915, opened 2021-03-23
Pull reflection out of RuntimeMessageAdapter

swankjesse:
Promote RuntimeMessageAdapter from jvmMain to commonMain. Leave the reflective
code behind in a new reflection.java file.

This requires a new interface, MessageBinding, that encapsulates the features
implemented by reflection.

This is in anticipation of a brand new implementation of MessageBinding that
uses a parsed schema to encode and decode values without any code generation
whatsoever.


wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/RuntimeMessageAdapter.kt
```diff
@@ -1,276 +0,0 @@
...
-      if (binding != null) {
-        append(", ")
-        append(fieldBinding.name)
-        append('=')
-        append(if (fieldBinding.redacted) REDACTED else binding)
-      }
-    }
-
-    // Replace leading comma with class name and opening brace.
-    replace(0, 2, messageType.simpleName + '{')
```
swankjesse, line 143: I think this code was incorrectly omitting the trailing `}` of the toString().

wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/RuntimeMessageAdapter.kt
```diff
@@ -0,0 +1,204 @@
...
+    for (field in fields.values) {
+      val binding = field[value] ?: continue
+      if (!first) append(", ")
+      first = false
+      append(field.name)
+      append('=')
+      append(if (field.redacted) REDACTED else binding)
+    }
+    append('}')
+  }
```
swankjesse, line 137: Note that I changed this method because it was missing the `}`, and because `StringBuilder.replace()` doesn’t exist on multiplatform.

wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/reflection.kt
```diff
@@ -0,0 +1,120 @@
+/*
+ * Copyright 2013 Square Inc.
```
oldergod, line 2:
```suggestion
 * Copyright 2021 Square Inc.
```

oldergod (approved):
So generated code wouldn't include their own adapter anymore?

PR #1918, opened 2021-03-24
Adopt Okio's FileSystem in Wire

swankjesse:
The biggest potential upside here is being able to perform file
loading in multiplatform code.

### DIFF
--- a/wire-library/wire-compiler/src/main/java/com/squareup/wire/WireCompiler.kt
+++ b/wire-library/wire-compiler/src/main/java/com/squareup/wire/WireCompiler.kt
@@ -25,15 +25,19 @@ import com.squareup.wire.schema.NullTarget
 import com.squareup.wire.schema.SwiftTarget
 import com.squareup.wire.schema.Target
 import com.squareup.wire.schema.WireRun
-import okio.buffer
-import okio.source
+import com.squareup.wire.schema.toOkioFileSystem
 import java.io.File
 import java.io.FileNotFoundException
 import java.io.IOException
-import java.nio.file.FileSystem
 import java.nio.file.FileSystems
 import java.nio.file.Files
-import java.nio.file.Path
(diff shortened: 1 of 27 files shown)


wire-library/wire-compiler/src/main/java/com/squareup/wire/WireCompiler.kt
```diff
@@ -240,11 +244,21 @@ class WireCompiler internal constructor(
       }
     }
 
+    @Throws(WireException::class)
+    @JvmStatic
+    fun forArgs(
```
oldergod, line 249: 👍

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/fileSystems.kt
```diff
@@ -0,0 +1,87 @@
+/*
```
oldergod, line 1: Just confirming you want the file name to start with a lower case

oldergod (approved):

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/fileSystems.kt
swankjesse: Yeah, I think the scheme I observed elsewhere is lowerCamel – no primary top-level class, UpperCamel – primary top-level class

oldergod (approved):

blacktooth:
Looks like this change prevents use of Jimfs for in-memory schema loading use-cases. Is it intentional? 

Can [FakeFileSystem](https://github.com/square/okio/blob/master/okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt) be reliably used for production use-cases where schemas will be loaded in-memory?

PR #1921, opened 2021-03-30
Don't generate @Nullable for .

oldergod:
For some reason, `emitAndroid` could override `emitAndroidAnnotations` when true.

Our API says
```kotlin
  /** True for emitted types to implement `android.os.Parcelable`. */
  val android: Boolean = false,

  /** True to enable the `androidx.annotation.Nullable` annotation where applicable. */
  val androidAnnotations: Boolean = false,
```

### DIFF
--- a/wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
+++ b/wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
@@ -257,7 +257,7 @@ public final class JavaGenerator {
     this.memberToJavaName = ImmutableMap.copyOf(memberToJavaName);
     this.profile = profile;
     this.emitAndroid = emitAndroid;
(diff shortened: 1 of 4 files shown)


wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -257,7 +257,7 @@ private JavaGenerator(Schema schema, Map<ProtoType, TypeName> typeToJavaName,
     this.memberToJavaName = ImmutableMap.copyOf(memberToJavaName);
     this.profile = profile;
     this.emitAndroid = emitAndroid;
-    this.emitAndroidAnnotations = emitAndroidAnnotations || emitAndroid;
```
Egorand, line 260: Ha, curious what the reasoning behind this was

Egorand (approved):

PR #1923, opened 2021-04-02
Use the version from README.md in podspecs

efirestone:
Rather than having to specify this manually, we can read it from the README.

### DIFF
--- a/Wire.podspec
+++ b/Wire.podspec
@@ -1,11 +1,15 @@
+require_relative 'wire-library/wire-runtime-swift/pod_helpers.rb'
+
 Pod::Spec.new do |s|
+  version = get_version
+
   s.name          = 'Wire'
-  s.version       = "#{ENV['POD_VERSION']}"
+  s.version       = version
   s.license       = { :type => 'apache2', :file => 'LICENSE.txt' }
   s.homepage      = 'https://github.com/square/wire'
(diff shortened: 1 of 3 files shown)


WireCompiler.podspec
```diff
@@ -1,11 +1,13 @@
 Pod::Spec.new do |s|
+  version = File.read('./README.md').match(/<version>([0-9.]+)<\/version>/).captures[0]
```
swankjesse, line 2: Wanna do the same technique, but build.gradle instead? That'll always be accurate. Sometimes for the readme we skip alpha releases.

Wire.podspec
```diff
@@ -1,11 +1,13 @@
 Pod::Spec.new do |s|
+  version = File.read('./README.md').match(/<version>([0-9.]+)<\/version>/).captures[0]
```
swankjesse, line 2: Wanna do the same technique, but gradle.properties instead? That'll always be accurate. Sometimes for the readme we skip alpha releases.

efirestone: Yup! And in fact, I was looking for places that were specifying 3.7.0 and missed gradle.properties because it had been bumped to 3.8.0. Will update.

JakeWharton (approved):

Issue #1925, opened 2021-04-05
wire-swift proto3 working document 


kevincianfarini:
Similar to #279, we'd like to track work that enables proto 3 support in wire-swift. This issue will start small and be updates as more is discovered. 

- [ ] Support for `Google.Protobuf.WellKnownTypes`
  - [ ] `Duration` code generation from `duration.proto`
    - [ ] `Duration` <-> `TimeInterval` lossy interop
  - [ ] Scalar wrapper types
    - [ ]  `BoolValue`
    - [ ]  `BytesValue`
    - [ ]  `DoubleValue`
    - [ ]  `FloatValue`
    - [ ]  `Int32Value`
    - [ ]  `Int64Value`
    - [ ]  `ListValue`
    - [ ]  `StringValue`
    - [ ]  `UInt32Value`
    - [ ]  `UInt64Value`

PR #1926, opened 2021-04-05
Add gradle task to generate WellKnownTypes

kevincianfarini:
This commit only adds the `duration.proto` definition
and the resulting generated type.

References #1925


wire-library/wire-runtime-swift/build.gradle
```diff
@@ -19,8 +19,28 @@ dependencies {
   wire project(':wire-compiler')
 }
 
+task generateSwiftProtos(type: JavaExec) {
```
kevincianfarini, line 22: so, I hate this. There's got to be a better way. Anyone with more knowledge, please help me level up my gradle-foo

wire-library/wire-runtime-swift/src/main/proto/duration.proto
```diff
@@ -0,0 +1,116 @@
+// Protocol Buffers - Google's data interchange format
```
efirestone, line 1: This exists in the repo already [here](https://github.com/square/wire/blob/master/wire-library/wire-schema/src/jvmMain/resources/google/protobuf/duration.proto). It seems reasonable to point at that rather than duplicating if we can. (But would like to hear others' opinions too, @oldergod)

efirestone (approved):
LGTM, but definitely want someone with more Gradle experience to look it over too.

wire-library/wire-runtime-swift/src/main/proto/duration.proto
kevincianfarini: Could we pull this out into a common folder?

efirestone: That seems reasonable too. I'm less familiar with Java file system organization conventions than all y'all, but seems like a reasonable approach. We will want a specific allow-list for iOS since we may not need all of them, but having the pool of .protos live somewhere common seems right.

wire-library/wire-runtime-swift/build.gradle
```diff
@@ -19,8 +19,28 @@ dependencies {
...
+  def swiftOut = 'src/main/swift/wellknowntypes'
```
JakeWharton, line 23: Should we be .gitignore-ing this folder? Otherwise we're not validating that the code which is committed is the same as is generated.

To put it another way, if I update the generator and forget to re-rerun this task the generated swift will be mismatched until someone fixes it.

I believe the test one is ignored.

wire-library/wire-runtime-swift/src/main/proto/duration.proto
JakeWharton: You should also be able to use the runtime jar as your input to the generation task and specify only the types you want. Wire can read inside jars.

wire-library/wire-runtime-swift/src/main/swift/wellknowntypes/Duration.swift
```diff
@@ -0,0 +1,138 @@
+// Code generated by Wire protocol buffer compiler, do not edit.
```
JakeWharton, line 1: See above comment about .gitignore

JakeWharton (approved):

wire-library/wire-runtime-swift/src/main/proto/duration.proto
kevincianfarini: Do you have any examples of this?

wire-library/wire-runtime-swift/build.gradle
efirestone: The generated Swift files for tests are [checked in](https://github.com/square/wire/blob/master/wire-library/wire-tests-swift/src/main/swift/FooBar.swift), so I figured we'd follow that pattern, but don't feel strongly at all. Generally I prefer not to check in generated code, although in this case since this project's main purpose is to do the generation it is nice to see any diffs that might arise.

WRT stale files, we're working on getting the Swift build into CI, which should hopefully alleviate that concern since CI would see dirty files and fail.

wire-library/wire-runtime-swift/build.gradle
```diff
@@ -19,8 +19,28 @@ dependencies {
...
+  def protoPath = 'src/main/proto'
```
efirestone, line 24: If we change this to like `../wire-schema/src/jvmMain/resources/google/protobuf/` then we can reuse the existing checked-in proto files. Then we can filter to only the messages we care about using something like:
```
--include=com.google.Duration,com.google.Instant
```
(messages are comma separated)

kevincianfarini: Awesome. Are there benefits of reading a JAR over just having a hardcoded path to the resources directory?

efirestone: Should be the same. It just depends on which one you have available to you, and I think in this case we have a raw directory already.

wire-library/wire-runtime-swift/build.gradle
kevincianfarini: [generated code in `wire-runtime-swift` is ignored. ](https://github.com/square/wire/blob/master/wire-library/wire-runtime-swift/.gitignore)

efirestone: I see lots of generated code that's [checked in](https://github.com/square/wire/blob/master/wire-library/wire-tests/src/commonTest/proto-kotlin/com/squareup/wire/protos/custom_options/FooBar.kt) too though. That feels like the larger precedent, but I really don't feel strongly personally. Should probably check with #wire-discuss

JakeWharton: That generated code that is checked in is validated against HEAD so that CI will not pass if what is generated is mismatched from what is committed.

https://github.com/square/wire/blob/85dc2120adfdf489d66ed425f2cbad0ecbf0184a/.github/workflows/build.yml#L62-L65

wire-library/wire-runtime-swift/build.gradle
JakeWharton: I would saying consuming via the jar is a little less bad since you are slightly less dependent on a bunch of hard-coded things that can change, but both still have tons of assumption in them. And both should fail in the same way if something moves so just go with the relative path I guess.

efirestone (approved):
Still LGTM

PR #1934, opened 2021-04-11
Generate @deprecated annotations for enum and messages

### LINKED ISSUE #1919: Type deprecation should add annotation on jvm generated classes
```
message MyType {
  option deprecated = true;
}
```

should deprecate the generated class.

oldergod:
fixes #1919

Split in 2 commits.

### DIFF
--- a/wire-library/wire-tests/src/jvmKotlinInteropTest/proto-kotlin/com/squareup/wire/protos/kotlin/NoFields.kt
+++ b/wire-library/wire-tests/src/jvmKotlinInteropTest/proto-kotlin/com/squareup/wire/protos/kotlin/NoFields.kt
@@ -10,6 +10,7 @@ import com.squareup.wire.ProtoWriter
 import com.squareup.wire.Syntax.PROTO_2
 import kotlin.Any
 import kotlin.Boolean
+import kotlin.Deprecated
 import kotlin.Int
 import kotlin.Long
 import kotlin.String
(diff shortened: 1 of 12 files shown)


wire-library/wire-tests/src/jvmKotlinInteropTest/proto-kotlin/com/squareup/wire/protos/kotlin/NoFields.kt
```diff
@@ -10,13 +10,15 @@ import com.squareup.wire.ProtoWriter
 import com.squareup.wire.Syntax.PROTO_2
 import kotlin.Any
 import kotlin.Boolean
+import kotlin.Deprecated
```
Egorand, line 13: Didn't know Kotlin has its own Deprecated now, nice!

Egorand (approved):

wire-library/wire-tests/src/jvmKotlinInteropTest/proto-kotlin/com/squareup/wire/protos/kotlin/NoFields.kt
oldergod: We already use it for fields. The downside with this one is that it requires a message.

Egorand: Oh, wait, I'm dumb, I confused it with `Nullable`... NVM

PR #1936, opened 2021-04-14
Fix up some quality of life things in the Wire Gradle Plugin

swankjesse:
We weren't doing the right thing with generated .java files when the
Kotlin plugin was installed.

We were using println() for logging instead of hooking in standard
Gradle logging stuff.


swankjesse (commented):
Benoît’s changes LGTM!

oldergod (approved):

oldergod:
Everything passes locally.

PR #1938, opened 2021-04-15
Convert wire-library to KTS

swankjesse:


wire-library/buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -0,0 +1,114 @@
+// If false - JS targets will not be configured in multiplatform projects.
```
swankjesse, line 1: I haven’t yet deleted the other `dependencies.gradle` file. I will do that once the enclosing project is converted to KTS.

oldergod (approved):

PR #1941, opened 2021-04-16
Automate removing redundant import in Swift WellKnownTypes

kevincianfarini:
References #1925.


wire-library/wire-runtime-swift/build.gradle.kts
```diff
@@ -59,6 +59,11 @@ val generateSwiftProtos by tasks.creating(JavaExec::class) {
     }
 
     actualFiles.forEach { File(it).delete() }
+
+    exec {
+      workingDir(".")
+      commandLine("sh", "-c", "find src/main/swift/wellknowntypes -type f | xargs sed -i '' 's/import Wire//g'")
```
JakeWharton, line 65: Does this work on Windows? I guess if you have WSL.

I would prefer we do this in Groovy/Gradle. Something like
```groovy
file(swiftOut).eachFileRecurse(FileType.FILES) { file ->
  file.text = file.text.replace("import Wire\n", "")
}
```
but I haven't tested that...

kevincianfarini: I think that's definitely a better approach. 

However, the current one _would_ still work because the build _always_ needs to run on macOS currently. Currently the `sed` command is actually bsd specific I believe. 

Will update, I like yours better.

JakeWharton (approved):

efirestone (approved):

PR #1942, opened 2021-04-19
Generate Swift Duration golden proto implementation

kevincianfarini:
References #1925


wire-library/wire-tests-swift/src/main/swift/proto3/AllTypes.swift
```diff
@@ -0,0 +1,2189 @@
...
+    public var any: Any? {
+        get {
+            storage.any
+        }
+        set {
+            copyStorage()
+            storage.any = newValue
+        }
+    }
+    public var duration: Duration? {
```
kevincianfarini, line 172: here's duration @efirestone :)

JakeWharton:
Is it safe to have two types with the same name in the same module (even though they're in different folders)? I would think that wouldn't work. Probably need to generate into actual Test2 and Test3 modules that can then be imported in the actual test.

kevincianfarini:
@JakeWharton it doesnt. Breaking the out into separate modules currently.

wire-library/wire-tests-swift/src/main/swift/proto3/AllTypes.swift
```diff
@@ -0,0 +1,2189 @@
...
+    public var `struct`: Struct? {
+        get {
+            storage.`struct`
+        }
+        set {
+            copyStorage()
+            storage.`struct` = newValue
+        }
+    }
+    public var list_value: ListValue? {
```
efirestone, line 190: This won't compile right now, right? Since we have things like `ListValue` and `Struct` in here? We would need to be importing the basic Google types for that to work (for now), but eventually we want to either be using all-included types, or native Swift types, with no dependency on the actual Google library.

So, options seem like:
1. It won't build, but we don't care for now and our goal is to get it building
2. Immediately add all of these Google types into Wire, like we did with Duration, then work to remove ones that can be modeled with native types instead
3. Set up the test to depend on the actual Google basic types protos (generated as a separate module) and slowly work to remove that dependency

Any of these seem fine to me, although if it's currently building/passing I'd hate to turn off all other tests in order to go with (1). If we can still run other tests for now then all seem like fine options. I'd lean mildly toward (1) or (3) just so that we're not technically temporarily adding API with (2) (like the Struct type), then removing it as we move toward using native Swift types.

kevincianfarini: Or we could: 

4. Generate a "golden" proto -- one for each WellKnownType we implement. Once we have every WellKnownType included, remove the fragmented golden protos in favor of `all_types.proto`.

efirestone: Also good with that, assuming other maintainers are cool with adding/removing golden protos. Could also make them iOS-specific golden protos in some new directory somewhere.

I leave it up to you.

kevincianfarini:
@efirestone requested review from @oldergod. The "new" golden proto isn't referenced anywhere outside of the swift build, so I'm hoping that it's good to go.

JakeWharton (approved):

PR #1943, opened 2021-04-20
Throwing an error when encountering an unexpected `ProtoReader.beginMessage` instead of calling `fatalError`.

cgarrett-square:
When developers attempt to decode invalid proto data, it's possible they'll hit a `fatalError` inside Wire's `ProtoReader`. Instead, we want to to allow them to catch the error and recover from it. This PR enables this by throwing an error in `ProtoReader.beginMessage` if `beginMessage` is called unexpectedly.

### DIFF
--- a/wire-library/wire-runtime-swift/src/main/swift/ProtoDecoder.swift
+++ b/wire-library/wire-runtime-swift/src/main/swift/ProtoDecoder.swift
@@ -49,9 +49,9 @@ public final class ProtoDecoder {
         case malformedVarint
         case mapEntryWithoutKey(value: Any?)
         case mapEntryWithoutValue(key: Any)
+        case messageWithoutLength
         case missingRequiredField(typeName: String, fieldName: String)
         case recursionLimitExceeded
-        case unexpectedCallToBeginMessage
         case unexpectedEndOfData
         case unexpectedEndGroupFieldNumber(expected: UInt32?, found: UInt32)
         case unexpectedFieldNumberInMap(_: UInt32)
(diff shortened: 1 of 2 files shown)


wire-library/wire-runtime-swift/src/main/swift/ProtoDecoder.swift
```diff
@@ -96,6 +97,8 @@ public final class ProtoDecoder {
                 return "Unknown case with value \(fieldNumber) found for enum of type \(String(describing: type))."
             case let .unterminatedGroup(fieldNumber):
                 return "The group with field number \(fieldNumber) has no matching end-group key."
+            case .unexpectedCallToBeginMessage:
+                return "Unexpected call to beginMessage()."
```
efirestone, line 101: Can we be more descriptive of what's wrong with the message data here (and in the error name). Since this is unlikely to be called from anything but Wire-generated code, it's more likely that the data is bad and not that someone accidentally called `beginMessage` unexpectedly.

In this case I think it's something like:

```
case .messageWithoutLength:
  return "Attempting to decode a message without first decoding the length of that message."
```

I'd say we should include the field number, but if we don't have the length then we almost certainly don't have the field number either, so this seems as good as we can do until we better track the actual byte offset.

efirestone (approved):
Awesome! Thanks for catching this.

wire-library/wire-runtime-swift/src/main/swift/ProtoDecoder.swift
cgarrett-square: Yeah, great point. That message sounds good to me -- I'll update.

efirestone (approved):

PR #1945, opened 2021-04-21
Remove tests against Java13

oldergod:


swankjesse (approved):

swankjesse:
What we really want to test are just the LTS ones + latest one: 1.8, 11, 16

JakeWharton:
LTS is a vendor-specific designation and not something intrinsic to the OpenJDK. Any vendor can choose to make any release an LTS release. As a topical example, Azul supports versions 13 and 15 longer than the normal 6-month period between OpenJDK releases. Unless we are saying that we only support certain OpenJDK vendors and their vendor-specific LTS designations, it's a dangerous precedent to consider those the only JDKs worth testing against.

Put another way, if 13 never failed would you have made such a change?

swankjesse:
I don’t think it’s useful for our test matrix to be completely exhaustive. We don’t test on multiple iOS or macOS versions, and we don’t test on multiple Linux operating system kernels. We don’t test on these things because Wire doesn’t really care! If Kotlin and the JVM are doing their jobs then all those platform differences are irrelevant to us.

So we instead test in clusters of configurations where the platform is meaningfully different to Wire. Android vs. OpenJDK is significant. OpenJDK 11 vs. 12 vs. 13? I can’t think of a bug or feature in Wire where that would ever be different between those. It’d almost be like testing on OpenJDK 11.0.1, 11.0.2, and 11.0.3.

JakeWharton:
That approach and framing is reasonable. It's just the LTS one I am strongly opposed to.

Issue #1948, opened 2021-04-24
Provide the possibility for clients to define the gRPC codec


k0shk0sh:
As of now the gRPC encoder is hardcoded to `gzip` and causes clients to have several issues when their backend doesn't support gziping. 

As per our discussion @swankjesse removing the gzip made it work for us and eliminates the STREAM RESET exception and I reckon many others will have the same issue if they are faced with what we encountered. 

Proposed solution would be: provide the client the possibility to set their own codec, if no codec passed  then Gzip by default will be used for backward compatibility reasons.

Issue #1777, opened 2020-09-15
RealGrpcCall/GrpcClient configurability for grpc-accept-encoding and grpc-encoding


buildbreaker:
Hello, I am currently using wire to talk to a back end grpc service which does not support any compression for requests/responses.

I did some spelunking to figure out if I can configure or work around this issue but it looks like the [`RealGrpcCall`](https://github.com/square/wire/blob/master/wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/grpc.kt#L51) set the request encoding to be `gzip` always. Additionally, the [`GrpcClient`](https://github.com/square/wire/blob/master/wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/GrpcClient.kt#L69-L70) also sets the headers `grpc-accept-encoding` and `grpc-encoding` to be `gzip` as well.

Are there immediate plans (or even ideas) on making this configurable?

I'm also happy to propose change (and help) for this but from a glance at the code, it seems like a little intrusive and non-trivial of a change (I still am happy to collaborate!)

Thank you,
Alan

swankjesse:
Yeah, good idea. It’s annoying that it isn’t negotiable!

I think we probably configure this in the same way as OkHttp’s WebSocket compression.
https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/-builder/min-web-socket-message-to-compress/

We’d do Long.MAX_VALUE for no compression ever.

buildbreaker:
Thanks for the quick response, @swankjesse!

I'll give that a shot and come back with the results!

buildbreaker:
Actually, this is a bit of a bummer but your suggestion of `Long.MAX_VALUE` will still flip the encryption bit to `1` ☹️ 

In the message preamble within the [`GrpcMessageSink`](https://github.com/square/wire/blob/master/wire-library/wire-grpc-client/src/commonMain/kotlin/com/squareup/wire/internal/GrpcMessageSink.kt#L35): since `grpcEncoding` is always `gzip` in the [grpc.kt](https://github.com/square/wire/blob/master/wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/grpc.kt#L51)

swankjesse:
I was intending to suggest we use MAX_VALUE as a special value that'd never flip that bit.

kam1sh:
[...]

PR #1949, opened 2021-04-24
provide clients to change the grpc codecs or create their own

k0shk0sh:
Provide the possibility for clients  to pass the codec for grpcClient to use or create their own that could be deflate, snappy or custom one. 

- Merged decoder and encoder into one class.
- Gzip is used still by default for backward compatibility.
- Not relying on the encoding header anymore but rather on what the client defines as in gRPC client and server has to always align on specific codec beforehand and therefore relying on the supplied codec is sufficient enough. 
- Updated GrpcDispatcher to use Identity by default.

Addressing: https://github.com/square/wire/issues/1948 and https://github.com/square/wire/issues/1777


swankjesse:
I don’t think we should make the codec a public API. I think it’s too low-level an abstraction; end users should just want to configure compression, and even that I think we can come up with better defaults.

Perhaps we can compress messages >= 1024 bytes by default, and let users configure something different if they prefer that.

k0shk0sh:
Yeah, that was also my thought at first too, but looking at gRPC documentation regards the encoding, client and server can have a custom codec and by making this as public API it allows this customization. But I'm open to suggestions how to make this better.

wire-library/wire-grpc-mockwebserver/src/main/java/com/squareup/wire/mockwebserver/GrpcDispatcher.kt
```diff
@@ -155,7 +156,7 @@ class GrpcDispatcher(
     val source = GrpcMessageSource(
         source = request.body,
         messageAdapter = protoAdapter,
-        grpcEncoding = request.headers["grpc-encoding"]
+        codec = GrpcCodec.IdentityGrpcCodec
```
oldergod, line 159: Why don't we read it anymore?

wire-library/wire-grpc-client/src/commonMain/kotlin/com/squareup/wire/GrpcCodec.kt
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2019 Square Inc.
```
oldergod, line 2: 2021

wire-grpc-sample/client/src/main/java/com/squareup/wire/whiteboard/GrpcClientProvider.kt
```diff
@@ -47,6 +48,10 @@ object GrpcClientProvider {
 
   val grpcClient = GrpcClient.Builder()
       .client(okHttpClient)
+      .apply {
+        // GzipGrpcCodec used by default, use Identity or define your own extending GrpcEncoder
```
oldergod, line 52: https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

wire-library/wire-grpc-mockwebserver/src/main/java/com/squareup/wire/mockwebserver/GrpcDispatcher.kt
k0shk0sh: The client supplies the codec for decoding and encoding as mentioned in my description.

k0shk0sh: From gRPC [documentation](https://www.google.com/url?sa=t&source=web&rct=j&url=https://github.com/grpc/grpc-go/blob/master/Documentation/encoding.md&ved=2ahUKEwielYGrnZvwAhVQhf0HHTVYBb4QFjAAegQIBRAC&usg=AOvVaw0SlY9CscYcxEOqSyk2lG0K)

> Codecs, by definition, must be symmetric, so the same desired Codec should be registered in both client and server binaries.

swankjesse:
We won’t add support for custom codecs until there are widely used ones that we can’t build in. It seems that’s unlikely as even Go doesn’t support the gzip codec.

k0shk0sh:
@swankjesse alright, im fine with this decision, what do you recommend would be the right way to allow clients to select which codec to use? enums, sealed classes?

swankjesse (changes requested):
I don’t think we want user-pluggable codecs. As far as I know the only widely-available codecs are identity and gzip, and Go’s implementation screwed up gzip by making it off by default.

swankjesse:
Obsoleted by https://github.com/square/wire/pull/1956

PR #1950, opened 2021-04-26
Only generate one task and bind it to all variants

oldergod:
Right now, we register `generate${Variant}Protos` although there is no different in what input they take, what output they produce, and where they produce it.
I wanna go back to a one task only world but I wanna bind this task properly to all variant sources and compilation tasks.
With this PR we will only have one task: `generateProtos`.

### DIFF
--- a/wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
+++ b/wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
@@ -94,11 +94,6 @@ class WirePlugin : Plugin<Project> {
       "Missing either the Java, Kotlin, or Android plugin"
     }
 
-    project.tasks.register(PARENT_TASK) {
-      it.group = GROUP
-      it.description = "Aggregation task which runs every generation task for every given source"
-    }
-
     if (extension.protoLibrary) {
       val libraryProtoSources = File(project.libraryProtoOutputPath())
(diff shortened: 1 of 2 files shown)


JakeWharton:
So you're going to remove the ability to have src/debug/protos/? What about on multiplatform can I have src/jvmMain/protos/ or src/jsMain/protos next to src/commonMain/protos?

I would rather fix up the inputs and output directories. The right solution to libraries which have no variation across variants is to disable all but a single variant: https://issuetracker.google.com/issues/62201427

JakeWharton:
SQLDelight had the same problem with output directory recently but Alec just added the corresponding variance to selecting the output directory (and also input directory set).

oldergod:
> you're going to remove the ability to have src/debug/protos/? What about on multiplatform can I have src/jvmMain/protos/ or src/jsMain/protos next to src/commonMain/protos?

The ability to have protos splitted per variants never worked for we always look for `src/main/protos`. That's why I took the approach to kill it, instead of faking it.

The thing I am still most puzzled with is on Android, say I have protos are in `src/main/protos` but I only get 2 variants `debug` and `release`. I don't wanna register the same task twice since there's only 1 source but none of the variant that I get are `main`.

JakeWharton:
Yeah I understand it never worked. My question is why this direction rather than enabling it to work?

oldergod:
Just a tradeoff between the use case relevancy and the cost to make it work. I'll check what SqlDelight did.

PR #1956, opened 2021-05-01
GrpcClient.Builder().minMessageToCompress()

swankjesse:
This allows users to have more control over our compression settings.
It's currently compress-everything, but only compressing messages
of a certain size is more efficient.

For interop with some GRPC servers that don't have compression,
using Long.MAX_VALUE to disable compression is useful.

Closes: https://github.com/square/wire/issues/1948

### DIFF
--- a/wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/GrpcClient.kt
+++ b/wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/GrpcClient.kt
@@ -17,13 +17,14 @@ package com.squareup.wire
 
 import com.squareup.wire.internal.RealGrpcCall
 import com.squareup.wire.internal.RealGrpcStreamingCall
+import kotlin.reflect.KClass
 import okhttp3.Call
 import okhttp3.OkHttpClient
-import kotlin.reflect.KClass
 
 actual class GrpcClient private constructor(
   internal val client: Call.Factory,
-  internal val baseUrl: GrpcHttpUrl
+  internal val baseUrl: GrpcHttpUrl,
+  internal val minMessageToCompress: Long
 ) {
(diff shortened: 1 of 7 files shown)


wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/GrpcClient.kt
```diff
@@ -61,21 +62,33 @@ actual class GrpcClient private constructor(
     return RealGrpcStreamingCall(this, method)
   }
 
+  fun newBuilder(): Builder {
```
swankjesse, line 65: # 🆕

wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/GrpcClient.kt
```diff
@@ -61,21 +62,33 @@ actual class GrpcClient private constructor(
...
+      }
+      .tag(GrpcMethod::class.java, method)
+      .method("POST", requestBody)
+      .build())
   }
 
   class Builder {
     private var client: Call.Factory? = null
     private var baseUrl: GrpcHttpUrl? = null
+    private var minMessageToCompress: Long = 0L
```
oldergod, line 91: No `MAX_VALUE` ?

oldergod (approved):

wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/GrpcClient.kt
swankjesse: I wanted to introduce the feature independent of changing our default behavior. For easier rollbacks!

PR #1959, opened 2021-05-04
Add sourceFile to WireRpc annotations

juliaogris:
Add sourceFile to WireRpc annotations, so that servers can
discover the schema and use it to implement grpc-reflection.


wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/WireRpc.kt
```diff
@@ -32,5 +32,6 @@ package com.squareup.wire
 annotation class WireRpc(
```
oldergod, line 32: We should add some doc to this class at some point.

oldergod (approved):

PR #1965, opened 2021-05-12
Add support for special float literals

ewolak-sq:
Both the [proto2 spec] and [proto3 spec] define floating-point literal values "inf", "-inf", and "nan" that may be used as default values for both float and double fields. This change adds support for such literal values in both the Java and Kotlin generators.

[proto2 spec]: https://developers.google.com/protocol-buffers/docs/reference/proto2-spec#floating-point-literals
[proto3 spec]: https://developers.google.com/protocol-buffers/docs/reference/proto3-spec#floating-point-literals

### DIFF
--- a/wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
+++ b/wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
@@ -1165,8 +1165,8 @@ class KotlinGenerator private constructor(
       typeName == BOOLEAN -> CodeBlock.of("%L", defaultValue)
       typeName == INT -> defaultValue.toIntFieldInitializer()
       typeName == LONG -> defaultValue.toLongFieldInitializer()
-      typeName == FLOAT -> CodeBlock.of("%Lf", defaultValue.toString())
-      typeName == DOUBLE -> CodeBlock.of("%L", defaultValue.toString().toDouble())
+      typeName == FLOAT -> defaultValue.toFloatFieldInitializer()
+      typeName == DOUBLE -> defaultValue.toDoubleFieldInitializer()
       typeName == STRING -> CodeBlock.of("%S", defaultValue)
       typeName == ByteString::class.asTypeName() -> CodeBlock.of("%S.%M()!!",
           defaultValue.toString().encode(charset = Charsets.ISO_8859_1).base64(),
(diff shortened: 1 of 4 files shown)


wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -1213,6 +1213,18 @@ class KotlinGenerator private constructor(
     else -> CodeBlock.of("%LL", long)
   }
 
+  private fun Any.toFloatFieldInitializer(): CodeBlock = when(this) {
```
swankjesse, line 1216: nan?

swankjesse (approved):
Wanna do NaN while you’re here?

swankjesse:
BTW, testing for NaN is annoying because it’s not reflexive!

ewolak-sq:
> Wanna do NaN while you’re here?

Yes, of course! Thanks for catching that.

> BTW, testing for NaN is annoying because it’s not reflexive!

Luckily we're only handling strings here, so equality should work fine

swankjesse (approved):

swankjesse:
This is great. Thanks!

oldergod:
`-nan` can actually happen too? If this can allow `-inf`, I think `-nan` is also valid syntax, right?

```
floatLit = ( decimals "." [ decimals ] [ exponent ] | decimals exponent | "."decimals [ exponent ] ) | "inf" | "nan"
decimals  = decimalDigit { decimalDigit }
exponent  = ( "e" | "E" ) [ "+" | "-" ] decimals 
```

swankjesse:
I don't think -NaN means anything different from NaN fortunately.

ewolak-sq:
@oldergod You make a good point. I think `-nan` is syntactically valid, but as @swankjesse points out, IEEE 754 says it's equivalent to "positive" NaN. I'll test this with the `protoc` compiler and ensure we do the same thing.

ewolak-sq:
Welp, compared to [over 9k occurrences of `default = nan`](https://github.com/search?q=%22default+%3D+nan%22+extension%3Aproto&type=code) on GitHub, there are only [13 results](https://github.com/search?q=%22default+%3D+-nan%22+extension%3Aproto&type=code) for `default = -nan` and they're all in `nanopb`. That being said, we might as well go for full compliance to the spec, as it's not much additional logic.

oldergod (approved):

PR #1966, opened 2021-05-13
Add 'nameSuffix' parameter for configuring generated service-class names

### LINKED ISSUE #1962: [Feature Request] Protocol-agnostic service client interfaces
I'm interested in using Wire to help our app talk to our backends, which speak Twirp.  (tl;dr: Twirp is basically "protobuf over HTTP 1.1").  The current implementation of `RpcRole.CLIENT` assumes gRPC.  Would you be open to adding an RpcRole that doesn't require a specific protocol?

`RpcRole.SERVER` is nearly ideal - the two drawbacks being that `WireRpc` is runtime-retained (not a dealbreaker), and also that the generated interface names have a `Server` suffix, which isn't great in a client context.  In my ideal world this hypothetical `RpcRole.GENERIC` would produce an interface only, with neither `Client` nor `Server` added to the name.

Is this something that you'd accept in a PR?

For clarity's sake, given this proto:

```proto
package tacotruck;

service TacoService {
  rpc FetchTacos(FetchTacosRequest) returns (FetchTacosResponse)
}
```

and this gradle plugin config:

```gradle
wire {
  kotlin {
    rpcRole = "generic"
  }
}
```

we'd get the following interface:

```kt
public interface TacoService : Service {
  @WireRpc(
      path = "/tacotruck.TacoService/FetchTacos",
      requestAdapter = "tacotruck.FetchTacosRequest#ADAPTER",
      responseAdapter = "tacotruck.FetchTacosResponse#ADAPTER")
  public suspend fun FetchTacos(request: FetchTacosRequest): FetchTacosResponse
}
```

(I guess we could get by with typealiasing `TacoServiceServer` to `TacoService`, or just living with the naming incongruity)

benjamin-bader:
This PR introduces a `nameSuffix` option to `KotlinGenerator`.  If specified, it will override the suffix that is currently computed from `rpcRole`.  It also adds an eponymous option to `KotlinOutput` in the Gradle plugin, along with the plumbing to get it from point A to point B.

Fixes #1962

### DIFF
--- a/wire-library/docs/wire_compiler.md
+++ b/wire-library/docs/wire_compiler.md
@@ -450,6 +450,12 @@ wire {
     // `none` to not generate services.
     rpcRole = 'server'
 
+    // When null, generated service types will have a suffix appended to
+    // their names, inferred from rpcRole - 'Server' for 'server',
+    // 'Client' for 'client', etc. When non-null, generated service types
+    //  will have the given value for a suffix.
+    nameSuffix = null
+
     // True for emitted services to implement one interface per RPC.
     singleMethodServices = true
   }
--- a/wire-library/docs/wire_compiler.md
+++ b/wire-library/docs/wire_compiler.md
@@ -450,11 +450,9 @@ wire {
     // `none` to not generate services.
     rpcRole = 'server'
 
-    // When null, generated service types will have a suffix appended to
-    // their names, inferred from rpcRole - 'Server' for 'server',
-    // 'Client' for 'client', etc. When non-null, generated service types
-    //  will have the given value for a suffix.
-    nameSuffix = null
+    // If set, the value will be appended to generated service type names. If null, their rpcRole
+    // will be used as a suffix instead.
+    nameSuffix = "Suffix"
 
     // True for emitted services to implement one interface per RPC.
     singleMethodServices = true
--- a/wire-library/wire-kotlin-generator/src/test/java/com/squareup/wire/kotlin/KotlinGeneratorTest.kt
+++ b/wire-library/wire-kotlin-generator/src/test/java/com/squareup/wire/kotlin/KotlinGeneratorTest.kt
@@ -1006,7 +1006,7 @@ class KotlinGeneratorTest {
           |/**
           | * RouteGuide service interface.
           | */
-          |public interface RouteGuideBlockingThing : Service {
+          |public interface RouteGuideThing : Service {
           |  /**
           |   * Returns the \[Feature\] for a \[Point\].
           |   */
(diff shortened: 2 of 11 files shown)


oldergod (approved):
Thanks. Could you update the doc here as well? `wire-library/docs/wire_compiler.md`

wire-library/docs/wire_compiler.md
```diff
@@ -450,6 +450,12 @@ wire {
     // `none` to not generate services.
     rpcRole = 'server'
 
+    // When null, generated service types will have a suffix appended to
+    // their names, inferred from rpcRole - 'Server' for 'server',
+    // 'Client' for 'client', etc. When non-null, generated service types
+    //  will have the given value for a suffix.
+    nameSuffix = null
```
oldergod, line 457: The list has only non-default values so let's use one.

```suggestion
    // If set, the value will be appended to generated service type names. If null, their rpcRole
    // will be used as a suffix instead.
    nameSuffix = "Suffix"
```

oldergod (approved):
I think `rpcNameSuffix` might be a better option name than `nameSuffix`.
What do you think @swankjesse ?.

wire-library/wire-kotlin-generator/src/test/java/com/squareup/wire/kotlin/KotlinGeneratorTest.kt
```diff
@@ -971,6 +971,87 @@ class KotlinGeneratorTest {
...
+    val blockingInterface = """
+          |package routeguide
+          |
+          |import com.squareup.wire.Service
+          |import com.squareup.wire.WireRpc
+          |
+          |/**
+          | * RouteGuide service interface.
+          | */
+          |public interface RouteGuideBlockingThing : Service {
```
swankjesse, line 1009: So the word `Blocking` shows up here and I think it shouldn’t? Like because the user’s requested suffix is `Client` and not `BlockingClient` I think this interface should be `RouteGuideClient` ?

swankjesse (commented):
Thanks for doing this. I really like it! One behavior question!

wire-library/wire-kotlin-generator/src/test/java/com/squareup/wire/kotlin/KotlinGeneratorTest.kt
benjamin-bader: So you're thinking that the presence of `nameSuffix` overrides not just "Client"/"Server", but also the other suffix components like "Blocking"?  I hadn't considered that, but it seems makes intuitive sense.  Let me know if I understand you properly - it should be an easy fix.

swankjesse: Yep exactly!

benjamin-bader: Cool; done!

benjamin-bader:
Looks like one build timed out at the "verify gradle wrapper" step.  I don't have permissions to retry it, but I think that's all it'll take.

oldergod:
Thank you!

PR #1967, opened 2021-05-14
Drop support for emitKotlinxSerialization

swankjesse:
This was an experimental feature I added. It's not useful to me anymore,
and I don't want to invest the effort to promote it into a proper API.


oldergod (approved):

iTanChi:
hi @swankjesse With this change, we cannot use Kotlin Serialization anymore, right? Currently we use this feature in a multi platform module and request / response objects for ktor are generated from proto files.

iTanChi:
hi @swankjesse With this change, we cannot use Kotlin Serialization anymore, right? Currently we use this feature in a multi platform module and request / response objects for ktor are generated from proto files.

swankjesse:
@iTanChi I apologize for experimenting in public; we never really did the work to make this work properly. In particular the JSON was not consistent with the protobuf spec. If you wanna do JSON with Wire I recommend switching to Moshi which does all the tricky things required by the spec.

iTanChi:
I wish I could. We are using Kotlin Multiplatform so the only choice ATM is Kotlin Serialization. I think I have to stick with the current version then

swankjesse:
Wanna open a new feature request for this? You're right that we should have something. The code you're using was a quick experiment that I hoped nobody would find or use.

iTanChi:
It will be great @swankjesse. I will open a new feature request. Thank you so much.

PR #1968, opened 2021-05-17
Detail the AssertionError thrown on newBuilder

quad:
When a developer, despite the best guidance of `@Deprecated`, accidentally calls the `newBuilder` method on generated classes, they get an exception that looks something like:

```java
java.lang.AssertionError
	at com.namespace.protos.MessageType.newBuilder(MessageType.kt:60)
	at com.namespace.protos.MessageType.newBuilder(MessageType.kt:28)
	at com.namespace.code.Thing.doSomething(Thing.kt:42)
```

This PR intends to make that first line instead say something like:

```java
java.lang.AssertionError: newBuilder has been deprecated in Kotlin
```

### DIFF
--- a/wire-grpc-tests/src/test/proto-grpc/routeguide/Feature.kt
+++ b/wire-grpc-tests/src/test/proto-grpc/routeguide/Feature.kt
@@ -52,7 +52,8 @@ public class Feature(
     message = "Shouldn't be used in Kotlin",
     level = DeprecationLevel.HIDDEN
   )
-  public override fun newBuilder(): Nothing = throw AssertionError()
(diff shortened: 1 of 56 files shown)


swankjesse:
Or perhaps, advice to set javaInterop=true for this?

quad:
How about:

> Builders are deprecated and only available in a javaInterop build; see https://square.github.io/wire/wire_compiler/#kotlin"

oldergod:
You might have to run `./gradlew generateTests` to update our golden files, and push what you locally get.

oldergod (approved):

quad:
@oldergod I pushed the updated golden files.

Issue #1955, opened 2021-04-29
Wire equivalent to io.grpc.Metadata / MetadataUtils?


carterhudson:
https://stackoverflow.com/questions/67325214/does-wire-have-an-equivalent-to-io-grpc-metadata-and-metadatautils-for-attaching

Basically what I outlined there - is there some mechanism for attaching headers to a request for only a single call in a client? I have concerns about how Channel reuse happens in cases where I need to modify headers with OkHttp interceptors...

swankjesse:
Commented on the Stack Overflow issue.

I'm imagining what the API for this should look like. Maybe something on GrpcCall?

```
interface GrpcCall {
  var metadata: Headers
  ...
}
```

carterhudson:
Good to know that using `OkHttpClient.newBuilder()` enables connection reuse; I think that might unblock me.

`GrpcCall` was where I started looking for that functionality when the need came up, so I think that would be a decent place to add headers. Something like `someClient.SomeGrpcMethod().apply { requestMetadata = "SomeKey" to "SomeValue" }.execute()` seems fairly simple and clean.

carterhudson:
I also find myself needing to track _all_ grpc calls. Previously, we did this through an `Interceptor` added to an `OkHttpChannelBuilder` via `.intercept`:

```kotlin
private val META_TRACE = Metadata.Key.of("x-emb-trace-id", Metadata.ASCII_STRING_MARSHALLER)

class EmbraceInterceptor : ClientInterceptor {
    override fun <ReqT, RespT> interceptCall(
        method: MethodDescriptor<ReqT, RespT>,
        options: CallOptions,
        channel: Channel
    ) = object : SimpleForwardingClientCall<ReqT, RespT>(channel.newCall(method, options)) {

        override fun start(responseListener: Listener<RespT>, headers: Metadata) {
            headers.put(META_TRACE, Embrace.getInstance().traceIdHeader)

            val listener = object : ForwardingClientCallListener<RespT>() {

                override fun delegate(): Listener<RespT> = responseListener

                override fun onMessage(message: RespT) {
                    Embrace.getInstance().logBreadcrumb("gRPC RECV ${method.serviceName} ${method.fullMethodName}")
                    super.onMessage(message)
                }
            }

            super.start(listener, headers)
        }

        override fun sendMessage(message: ReqT) {
            Embrace.getInstance().logBreadcrumb("gRPC SEND ${method.serviceName} ${method.fullMethodName}")
[...]

PR #1969, opened 2021-05-20
Initial implementation of gRPC metadata

swankjesse:
It's a bit unfortunate to use mutable fields on the Call classes for these.
But it's a stateful thread-confined object already (timeout, isExecuted).

Most of the difficulty here was setting up the test infrastructure.

Closes: https://github.com/square/wire/issues/1955


wire-library/wire-grpc-client/src/commonMain/kotlin/com/squareup/wire/GrpcCall.kt
```diff
@@ -34,6 +34,21 @@ interface GrpcCall<S : Any, R : Any> {
   /** Configures how long the call can take to complete before it is automatically canceled. */
   val timeout: Timeout
 
+  /**
+   * A map containing request metadata. This is initially empty; it can be assigned to any other map
+   * of metadata before the call is executed. It is an error to set this value after the call is
+   * executed.
```
oldergod, line 40: We can crash for the error on followup

oldergod (approved):
kudos

PR #1970, opened 2021-05-20
emitDeclaredOptions / emitAppliedOptions default values

quad:
According to [wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt:2192](https://github.com/square/wire/blob/6cc58cda6285a1d6fb48259d6ecb234b176c5803/wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt#L2192):
```kotlin
emitDeclaredOptions: Boolean = true,
emitAppliedOptions: Boolean = false,
```

### DIFF
--- a/wire-library/docs/wire_compiler.md
+++ b/wire-library/docs/wire_compiler.md
@@ -390,10 +390,10 @@ wire {
     compact = true
 
     // True to emit types for options declared on messages, fields, etc.
-    emitDeclaredOptions = false,
+    emitDeclaredOptions = true,
 
     // True to emit annotations for options applied on messages, fields, etc.
-    emitAppliedOptions = true
+    emitAppliedOptions = false
   }
 }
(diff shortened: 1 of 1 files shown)


oldergod:
You can see the line above saying

> Here's an exhaustive Java configuration. Booleans are shown with their *non-default behavior*.

quad:
OK, an exhaustive list options with the opposite of their default values surprised me.

swankjesse:
My (foolish?) idea was that you paste in what you want and don’t need to toggle the booleans. But if you’re surprised by that I doubt you’re alone, and we could flip it so every value is with its default. That’s also move obvious for fields that have more than one option!

quad:
Closed in favor of #1971

PR #1977, opened 2021-05-27
Redacted Kotlin scalars respect nullability

dhenry-stripe:
Prior to this PR  `KotlinGenerator` was generating invalid code for redacted proto3 scalars. For example:

```
public override fun redact(value: MyExample): MyExample = value.copy(
    my_string = null, // build fails because type is String not String?
    my_int32 = null, // build fails because type is Int and not Int?
    unknownFields = ByteString.EMPTY
)
```

With this change, non-nullable Kotlin scalar values are redacted as the identity value for the given type.

### DIFF
--- a/wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
+++ b/wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
@@ -1476,10 +1476,9 @@ class KotlinGenerator private constructor(
     }
 
     val redactedFields = mutableListOf<CodeBlock>()
-    val flattenedOneOfFields = message.flatOneOfs().flatMap { it.fields }
     for (field in message.fieldsAndFlatOneOfFields()) {
       val fieldName = nameAllocator[field]
-      val redactedField = field.redact(fieldName, flattenedOneOfFields.contains(field))
+      val redactedField = field.redact(fieldName)
       if (redactedField != null) {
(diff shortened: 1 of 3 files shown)


wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -1495,11 +1496,13 @@ class KotlinGenerator private constructor(
         .build()
   }
 
-  private fun Field.redact(fieldName: String): CodeBlock? {
+  private fun Field.redact(fieldName: String, isOneOf: Boolean): CodeBlock? {
     if (isRedacted) {
       return when {
         isRepeated -> CodeBlock.of("emptyList()")
         isMap -> CodeBlock.of("emptyMap()")
+        isOptional || isOneOf -> CodeBlock.of("null")
```
oldergod, line 1504: Is there an existing test case for one ofs?

oldergod: Also could we look at `Field.EncodeMode.NULL_IF_ABSENT` instead?

dhenry-stripe: Hrmm not for this use case. I've updated the included test to cover redacted one ofs and switched over to `Field.EncodeMode.NULL_IF_ABSENT` as well, thanks.

oldergod (approved):
Sweet, thank you

PR #1980, opened 2021-05-31
Update to KotlinPoet 1.8.0 and address few API changes

oldergod:
Our 2 problems are value being escaped all the times. We were using name allocator for the enum's constructor param but we cannot do that. Otherwise it would override nothing if the name changed.
The other problem is `kotlin.hashcode()` being inlined so I inlined its content manually instead. see https://github.com/square/kotlinpoet/issues/1089

2 commits

### DIFF
--- a/wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
+++ b/wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
@@ -219,7 +219,6 @@ public final class JavaGenerator {
         }
 
       } else if (type instanceof EnumType) {
-        nameAllocator.newName("value", "value");
         nameAllocator.newName("i", "i");
         nameAllocator.newName("reader", "reader");
         nameAllocator.newName("writer", "writer");
@@ -513,7 +512,7 @@ public final class JavaGenerator {
 
   private TypeSpec generateEnum(EnumType type) {
     NameAllocator nameAllocator = nameAllocators.getUnchecked(type);
-    String value = nameAllocator.get("value");
+    String value = "value";
     ClassName javaType = (ClassName) typeName(type.getType());
 
     TypeSpec.Builder builder = TypeSpec.enumBuilder(javaType.simpleName())
@@ -823,7 +822,7 @@ public final class JavaGenerator {
    */
   private TypeSpec enumAdapter(NameAllocator nameAllocator, EnumType type, ClassName javaType,
       ClassName adapterJavaType) {
-    String value = nameAllocator.get("value");
+    String value = "value";
     String i = nameAllocator.get("i");
     String reader = nameAllocator.get("reader");
     String writer = nameAllocator.get("writer");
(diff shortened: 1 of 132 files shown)


wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -640,21 +639,20 @@ class KotlinGenerator private constructor(
...
-      val hashCode = MemberName("kotlin", "hashCode")
       addStatement("%N = unknownFields.hashCode()", resultName)
       for (field in type.fieldsAndFlatOneOfFields()) {
         val fieldName = localNameAllocator[field]
         add("%1N = %1N * 37 + ", resultName)
         if (field.isRepeated || field.isRequired || field.isMap) {
           addStatement("%L.hashCode()", fieldName)
         } else {
-          addStatement("%L.%M()", fieldName, hashCode)
+          addStatement("(%L?.hashCode() ?: 0)", fieldName)
```
Egorand, line 649: If this is all `kotlin.hashCode` is doing then I totally agree with inlining! FWIW I'm planning to take a look at the KotlinPoet bug and find a solution.

oldergod: Yep that's all

```kotlin
package kotlin

import kotlin.internal.InlineOnly


/**
 * Returns a hash code value for the object or zero if the object is `null`.
 *
 * @see Any.hashCode
 */
@SinceKotlin("1.3")
@InlineOnly
public inline fun Any?.hashCode(): Int = this?.hashCode() ?: 0
```

wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -513,7 +512,7 @@ public TypeSpec generateType(Type type) {
 
   private TypeSpec generateEnum(EnumType type) {
     NameAllocator nameAllocator = nameAllocators.getUnchecked(type);
-    String value = nameAllocator.get("value");
+    String value = "value";
```
swankjesse, line 515: This looks like a regression?

swankjesse (approved):
Why not use NameAllocator for `value` ?

wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
oldergod: because we cannot rename it.

```kotlin
  public enum class PhoneType(
    public override val value: Int
  ) : WireEnum {
```

If we rename `value`, it won't be overriding anything and the generated code won't compile.

oldergod: It might actually be ok for Java but not Kotlin. I'll confirm it works fine or not on Java before merging

oldergod: And I'll leave a comment on the Kotlin code

oldergod: I don't need to change Java of course... I bugged.

oldergod: merging on green

PR #1982, opened 2021-06-01
Add instructions for publishing the CocoaPods

efirestone:


wire-library/wire-runtime-swift/README.md
```diff
@@ -26,6 +26,43 @@ To build the runtime and run tests from the command line using Gradle, run the f
...
+```
+
+If publishing a release version (like `4.0.0` rather than `4.0.0-alpha1`) then setting the `POD_VERSION` is optional and it will be pulled automatically from `wire-library/gradle.properties`.
+
+### Publishing the Podspecs
+
+After setting the version as described above, you can publish the two Podspecs by doing:
+
+```
+# Tests are currently failing, thus --skip-tests is required
```
efirestone, line 56: Going to take a look at these shortly and try to get them into CI

JakeWharton:
These look great. I would just move them to RELEASING.md which has the other steps

efirestone:
> These look great. I would just move them to RELEASING.md which has the other steps

Makes sense. Will move.

JakeWharton (approved):

oldergod:
I'll give it a try tomorrow probably

PR #1993, opened 2021-06-10
Exhaustively load imports if said so

oldergod:


wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/SchemaLoader.kt
```diff
@@ -47,6 +47,12 @@ class SchemaLoader : Closeable, Loader, ProfileLoader {
   /** Strict by default. Note that golang cannot build protos with package cycles. */
   var permitPackageCycles = false
 
+  /**
+   * If true, the schema loader will load the whole graph, including files and types not used by
+   * anything in the source path.
+   */
+  var loadExhaustively = false
```
oldergod, line 54: Looking for a better name

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/FileLinker.kt
```diff
@@ -52,6 +52,7 @@ internal class FileLinker(
     for (path in paths) {
       if (sink.add(path)) {
         val fileLinker = linker.getFileLinker(path)
+        addImportsRecursive(sink, fileLinker.protoFile.imports)
```
oldergod, line 55: Without this, we would not recursively load "yellow" here. I don't think it would add any substantial work even when we are not exhaustively loading

oldergod: Not sure

swankjesse: I think this is the place to check loadExhaustively

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Linker.kt
```diff
@@ -135,15 +143,25 @@ class Linker {
         continue
       }
 
+      if (loadExhaustively) {
```
swankjesse, line 146: I think we can skip the entire retain() step completely if loadExhaustively is used

swankjesse: ie. on line 141 do, ` if (loadExhaustively || sourceFiles.contains(fileLinker))`

oldergod: Fixed!

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/FileLinker.kt
oldergod: Fixed!

PR #1994, opened 2021-06-17
Improve WireTask cacheability by storing project-relative paths instead of absolute ones in the task inputs

### LINKED ISSUE #1859: WireTask uses non-normalized paths and breaks Gradle build caching

staktrace:
Fixes #1859

### DIFF
--- a/wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
+++ b/wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
@@ -60,8 +60,21 @@ sealed class Target : Serializable {
 
   /**
    * Directory where this target will write its output.
+   *
+   * In Gradle, when this class is serialized, this is relative to the project to improve build
+   * cacheability. Callers must use [copyTarget] to resolve it to real path prior to use.
    */
   abstract val outDirectory: String
+
+  /**
+   * Returns a new Target object that is a copy of this one, but with the given fields updated.
+   */
+  abstract fun copyTarget(
+    includes: List<String> = this.includes,
+    excludes: List<String> = this.excludes,
+    exclusive: Boolean = this.exclusive,
+    outDirectory: String = this.outDirectory,
+  ): Target
 
   /**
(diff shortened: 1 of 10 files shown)


staktrace:
(I was unable to run `./gradlew clean check` locally because it needed an android SDK or something, so I'm making a PR and hoping the CI will tell me if I screwed something up)

staktrace:
I'll try to write a test as well.

staktrace:
I discussed this further with @swankjesse. One idea he had which we tried implementing was to just always have the `Target.outDirectory` field be relative to `$CWD`. However the problem with that is that the relative path may still contain things like the project name which breaks cacheability. In particular the test I added still failed in that case because the `Target.outDirectory` was `wire-gradle-plugin/src/test/projects/cache-relocation-1/src/main/proto` for one project and `wire-gradle-plugin/src/test/projects/cache-relocation-2/src/main/proto` for the other project, breaking cacheability.

Therefore the `Target.outDirectory` does need to be relative to the project directory, and to make the aesthetics of the API nicer he suggested replace the `resolveOutDirectory` function in my original implementation with a more generic `copyTarget` function. Updated patches with that change coming up shortly.

swankjesse (approved):
Great PR. Thank you!

staktrace:
Not really sure why github actions is just cancelling checks instead of running them.. it was green earlier so I'm going to merge and hope for the best!

PR #1997, opened 2021-06-22
Kotlin generated message's contructor respect fields declaration order

oldergod:
I hesitated using a sealed class something like `sealed class FieldOrOneOf` but didn't; could be convinced otherwise...
Split in 2 commits.

Here is the before after of how fields were ordered.

Before
------
fields
extensions
flatOneOfs
boxedOneOfs

After
-----
(everything).sortBy { it.location.line }


swankjesse (approved):
Yes, this is rad.

PR #1998, opened 2021-06-23
Fix exhaustive loading to load everything

swankjesse:
We had a bug where we were loading all of the types, but not linking
all the symbols within those types.


oldergod (approved):
Thank you

PR #2000, opened 2021-06-23
Kotlin redact support for boxedOneOf

oldergod:


wire-library/wire-tests-swift/src/main/swift/Form.swift
```diff
@@ -207,6 +207,16 @@ extension Form.Decision : Hashable {
 }
 #endif
 
+#if !WIRE_REMOVE_REDACTABLE
```
swankjesse, line 210: ?

swankjesse (approved):

wire-library/wire-tests-swift/src/main/swift/Form.swift
oldergod: That's generated automatically by `generateTests`, it's already in a few places.

Issue #2006, opened 2021-06-29
`WireTask` should use FileCollection instead of `List<String>`


runningcode:
The `WireTask` should use `FileCollection` instead of `List<String>` as a property so that the relative path of the file can be used instead of the full directory as a string. This would allow for more cache hits. The property should be annotated with RelativePath normalization. This would be the case for the `targets` property and the `inputFiles` property.

See the cache miss below.
<img width="992" alt="Screen Shot 2021-06-29 at 10 45 38 AM" src="https://user-images.githubusercontent.com/332597/123826665-44a7b100-d900-11eb-91af-f92d036a2c83.png">

Egorand:
Looks like this issue is fixed based on the linked PR - I'm gonna close it, but please feel free to re-open if there are still any outstanding tasks.

runningcode:
Hi there, I know it is confusing but the PR only fixed the `inputFiles` issue not the `targets` property. The `targets` property is a bit trickier since it has nested properties with fully qualified paths but is located in a library that does not depend on the Gradle API. A refactoring will be necessary in order to fix it. I recommend keeping this ticket open.

Egorand:
Got it, thanks for the extra context Nelson!

oldergod:
I think @ZacSweers fixed it https://github.com/square/wire/commit/aed27d91d8a513f23a3a440bcc7a84b3f7e29ea3

runningcode:
I don't see how that fixes the issue. The `Target` object still contains the fully qualified path to the output directory: https://github.com/square/wire/blob/master/wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt#L68

runningcode:
The `Target` object needs to be annotated with proper Gradle input/outputs. This is currently not possible since the `Target` object is not in a module that depends on the Gradle API.
One solution is to turn `Target` in to an interface and have a separate implementation in the Gradle module that has the proper input/output annotations.
Another possibility is to go through the properties of the `Target` object and add them as inputs/outputs on the `WireTask`.

I'm not so familiar with the implications of the first solution in regards to the rest of the wire project. Feel free to chime in.

PR #2007, opened 2021-06-30
Use relative path sensitivity and file collection

runningcode:
This will make it easier to debug cache misses when they happen due to this property since each file with a difference will be tracked individually instead of as a single property.

ref #2006


swankjesse (approved):
Thanks Nelson! 🏃

oldergod (approved):
Thank you

Issue #1903, opened 2021-03-05
GRPC: 'Nothing' return type needs to be specified explicitly


RdeWilde:
When I generate my rpc-services from .proto-file using Wire, I get this error in the `GreeterWireGrpc`-file.

This is an example:
```
  public abstract class GreeterImplBase : BindableService {
    public open fun sayHello(request: HelloRequest, response: StreamObserver<HelloReply>) = throw
        UnsupportedOperationException()
```
Produces this error:
```
'Nothing' return type needs to be specified explicitly
```
Should this have a return type (Unit) added?
https://github.com/square/wire/blob/ce5ff17d6176f85cc4084d8b271a7cae0f3f03c3/wire-library/wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/ImplBaseGenerator.kt#L60

Also it is missing this class: `GreetersayHelloBlockingServer` used in the `GreeterImplLegacyAdapter` like this:
```
private val sayHello: () -> GreetersayHelloBlockingServer
```
Should we leave `${rpc.name}` out? Or should it have created a server per `${rpc.name}` although I did not set the option in gradle?
https://github.com/square/wire/blob/ce5ff17d6176f85cc4084d8b271a7cae0f3f03c3/wire-library/wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/LegacyAdapterGenerator.kt#L80

Three lines below this, there's another error about `Unit` as return type

This is in my gradle file:
```
wire {
    kotlin {
        rpcCallStyle = "blocking" // suspending
        rpcRole = "server"
        grpcServerCompatible = true
        singleMethodServices = false
    }

    java {}
}
```
Proto:
```
// The greeting service definition.
service Greeter {
  // Sends a greeting
  rpc sayHello (HelloRequest) returns (HelloReply) {}
}
```
Why does it even give me a StreamObserver for HelloReply?

What am I missing? Can this be resolved?

PS. Is there a Slack or Discord or something for this community?

RdeWilde:
Maybe this is relevant for the `Nothing` and Unit `errors`
https://github.com/square/wire/pull/1587

RdeWilde:
It seems like adding a return type to the `${rpc.name}()` / `sayHello()` method in the `GreeterImplBase` (extending `BindableService`)

So changing this:
```
  public abstract class GreeterImplBase : BindableService {
[...]

PR #2013, opened 2021-07-11
Add Unit return type to unary rpc calls.

aarnaut:
This PR addresses #1903. 

### Overview
When specifying unary RPCs, `wire-grpc-server-generator` generates classes that do not compile. Generated methods throw an exception, but do not specify the return type, which is mandatory in this case.

### How to reproduce
1. Add a unary RPC to `samples/wire-grpc-sample/protos/../whiteboard.proto`. For example:
    ```
    service Test {
      rpc Test(WhiteboardCommand) returns (WhiteboardUpdate) { }
    }
    ```
2. Try to compile `samples/wire-grpc-sample/server-plain`. The compilation will fail with "'Nothing' return type needs to be specified explicitly"

### DIFF
--- a/wire-library/wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/ImplBaseGenerator.kt
+++ b/wire-library/wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/ImplBaseGenerator.kt
@@ -15,14 +15,8 @@
  */
 package com.squareup.wire.kotlin.grpcserver
 
-import com.squareup.kotlinpoet.ClassName
-import com.squareup.kotlinpoet.CodeBlock
-import com.squareup.kotlinpoet.FunSpec
-import com.squareup.kotlinpoet.KModifier
-import com.squareup.kotlinpoet.MemberName
-import com.squareup.kotlinpoet.ParameterSpec
+import com.squareup.kotlinpoet.*
 import com.squareup.kotlinpoet.ParameterizedTypeName.Companion.parameterizedBy
-import com.squareup.kotlinpoet.TypeSpec
 import com.squareup.wire.schema.Rpc
 import com.squareup.wire.schema.Service
 import java.io.InputStream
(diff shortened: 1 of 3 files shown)


aarnaut:
There are a few other issues I encountered while attempting to fix this:
1. `wire-grpc-server-generator` does not support `rcsStyle="suspending"`. It assumes the blocking method and will always generate blocking stubs.
2. In case of unary RPC, the response is always a a stream i.e. `StreamObserver<T>`.
3. `wire-grpc-server-generator` tests define classes that are expected to be generated. These classes are not compiled. This makes it hard to spot compile time issues. For example, if these were compiled, it would be very easy to spot the issue that this PR addresses.
4. There's a general lack of documentation around this area.

I'm not sure what your plan when it comes to support of plain grpc server. That's why I didn't create separate issues for things listed above. I would be happy to do so and even tackle them if this is something you see on your roadmap.

wire-library/wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/ImplBaseGenerator.kt
```diff
@@ -15,14 +15,8 @@
  */
 package com.squareup.wire.kotlin.grpcserver
 
-import com.squareup.kotlinpoet.ClassName
-import com.squareup.kotlinpoet.CodeBlock
-import com.squareup.kotlinpoet.FunSpec
-import com.squareup.kotlinpoet.KModifier
-import com.squareup.kotlinpoet.MemberName
-import com.squareup.kotlinpoet.ParameterSpec
+import com.squareup.kotlinpoet.*
```
oldergod, line 18: No stars please

aarnaut: Fixed: b21c3ba

r3mariano (approved):

r3mariano:
@aarnaut plain grpc server support is currently experimental. Happy to review PRs, but I haven't really had time to flesh it out more. I'll put up some annotations to indicate this.

PR #2020, opened 2021-07-17
Always add generated directories to commons

oldergod:


swankjesse:
Upon 2nd inspection I think this is probably a mistake? In particular I don’t think we can safely make any assumptions about custom targets. They might not be generating Kotlin!

PR #2022, opened 2021-07-22
Change field order sorting condition

juliaogris:
We had the fields and oneofs sorted by tag which lead to a subtle FileDescriptor validation error:

	[libprotobuf ERROR external/com_google_protobuf/src/google/protobuf/descriptor.cc:3623]   [redacted].Element.id: Fields in the same oneof must be defined consecutively. "id" cannot be defined before the completion of the "element" oneof definition.
	
Let's sort by location (file and column in stread).


swankjesse (approved):

swankjesse:
Nice fix & great test. Thank you!

oldergod (commented):
Super nice

PR #2025, opened 2021-07-23
ReverseProtoWriter: like ProtoWriter but without measurement

swankjesse:
In performance measurement I observed Wire's encodedSize() showing up too much in
profiles. This is likely due to the fact that this method must be invoked repeatedly
for deeply-nested datastructures.

Attempt to improve things by changing encoding from front-to-back to back-to-front,
which allows us to capture the data size for free as a side-effect.


wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -886,17 +887,19 @@ private TypeSpec enumAdapter(NameAllocator nameAllocator, EnumType type, ClassNa
...
-        .addModifiers(PUBLIC)
-        .addParameter(ProtoWriter.class, writer)
-        .addParameter(javaType, value)
-        .addException(IOException.class)
-        .addStatement("int $N = toValue($N)", i, value)
-        .addStatement("if ($N == $L) throw new $T($S + $N)",
-            i, -1, ProtocolException.class, "Unexpected enum constant: ", value)
-        .addStatement("$N.writeVarint32($N)", writer, i)
-        .build());
+    for (boolean reverse : Arrays.asList(false, true)) {
```
oldergod, line 890: maybe we could locally define a method and call it twice in a row with a named parameter to make it more reasonable

oldergod: hmmmm Java™

oldergod (approved):
great tests

wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
swankjesse: Fixed!

PR #2038, opened 2021-08-06
Handle writing/reading exceptions for duplex calls

oldergod:
For reading, an exception would be thrown when we try to read a message
when the stream had already been closed. We now properly handle all
exceptions and pass them along into our sending channel.

For sending, even though flushing the message would have failed, the
thrown exception would be eaten up and we would not realize the writing
has failed from upstream. We now catch it and pass it along into our
receive channe.

### DIFF
--- a/wire-grpc-tests/src/test/java/com/squareup/wire/GrpcClientTest.kt
+++ b/wire-grpc-tests/src/test/java/com/squareup/wire/GrpcClientTest.kt
@@ -24,12 +24,15 @@ import com.squareup.wire.MockRouteGuideService.Action.ReceiveError
 import com.squareup.wire.MockRouteGuideService.Action.SendCompleted
 import com.squareup.wire.MockRouteGuideService.Action.SendMessage
 import io.grpc.Status
+import io.grpc.StatusException
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.ObsoleteCoroutinesApi
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import okhttp3.Call
 import okhttp3.Interceptor
 import okhttp3.Interceptor.Chain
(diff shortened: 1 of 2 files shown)


wire-grpc-tests/src/test/java/com/squareup/wire/GrpcClientTest.kt
```diff
@@ -352,59 +355,110 @@ class GrpcClientTest {
   }
 
   @Test
-  fun duplexSuspend_errorsAreBubbledUpWhenSendingFailed() {
+  fun duplexSuspend_requestChannelThrowsWhenResponseChannelExhausted() {
```
swankjesse, line 358: I think we need one more test, which is this but the server sends an error instead of SendCompleted at the end

oldergod: I added one.

wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/grpc.kt
```diff
@@ -115,25 +129,40 @@ internal fun <R : Any> SendChannel<R>.readFromResponseBodyCallback(
...
         consumeEach { message ->
+          channelReadFailed = false
           requestWriter.write(message)
+          channelReadFailed = true
         }
-        success = true
+        channelReadFailed = false
       } finally {
-        if (!success) requestWriter.cancel()
+        if (channelReadFailed) requestWriter.cancel()
```
oldergod, line 157: @swankjesse You mean `channelWriteFailed`, yeah?

wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/grpc.kt
```diff
@@ -115,25 +129,40 @@ internal fun <R : Any> SendChannel<R>.readFromResponseBodyCallback(
...
-      var success = false
+  val requestWriter = requestBody.messageSink(minMessageToCompress, requestAdapter, callForCancel)
+  try {
+    requestWriter.use {
+      var channelReadFailed = true
       try {
         consumeEach { message ->
+          channelReadFailed = false
           requestWriter.write(message)
+          channelReadFailed = true
```
oldergod, line 153: I don't get this line. `consumeEach` is throwing after the _whole_ block has been executed? So if it doesn't throw, the line below override `channelReadFailed` to `false` so we're good?

wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/grpc.kt
oldergod: No, you meant `channelReadFailed` because we're reading from the receivechannel of course

wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/grpc.kt
swankjesse: If we were to inline and desugar consumeEach(), it’d looks something like this:

```
  while (true) {
    val message = receiveOrNull()
    if (message == null) break
   requestWriter.write(message)
  }
```

Given that behavior, I want to detect exceptions only thrown by receiveOrNull():

```
  var channelReadFailed: Boolean
  while (true) {
    channelReadFailed = true
    val message = receiveOrNull()
    channelReadFailed = false
    if (message == null) break
    requestWriter.write(message)
  }
```

But unfortunately due to the way consumeEach works, I can’t slot my code in there. So instead I rewrite the above to this:

```
  var channelReadFailed: Boolean = true
  while (true) {
    val message = receiveOrNull()
    if (message == null) break
    channelReadFailed = false
    requestWriter.write(message)
    channelReadFailed = true
  }
  channelReadFailed = false
```

I hope you agree that this is equivalent?

swankjesse (approved):

wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/grpc.kt
oldergod: Perfect yes, thank you

PR #2050, opened 2021-08-24
wire-gradle-plugin: revert default jar filter behaviour

quad:
This PR reverts to the exact behaviour of the original implementation.

* Source directories are, by default, filtered for `**/*.proto`
* Jars, by default, are passed directly to Wire

The thing is that [the wire compiler _already_ filters into `.proto`](https://github.com/square/wire/blob/08928e8068834201581a4555210879ec319bf5b6/wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Root.kt#L160-L160), so in the interests of conciseness I'd _rather_ make the default `filter` be `**/*.proto` in`ProtoRootSet#filters` and then remove the special casing in [`WireInput#toLocation`](https://github.com/square/wire/blob/08928e8068834201581a4555210879ec319bf5b6/wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt#L168-L168) and [`WireExtension#populateRootSets`](https://github.com/square/wire/blob/08928e8068834201581a4555210879ec319bf5b6/wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt#L222-L222).

However, that refactor wouldn't be valid if there's _another_ reason to passthrough `Location.get(jarFile)` to the wire compiler. Anyway, I'll propose it in a follow-up PR.


swankjesse:
Works for me. One hangup I just remembered is we also read files ending in `.wire`. It's an oddball feature we need to either invest in or rollback!

quad:
Haha, wow, that's *exactly* the kind of context that made me hesitate on #2051! I'll close it now.

PR #2051, opened 2021-08-24
wire-gradle-plugin: simplify filtering

quad:
A refactor I'd like to make, but am not sure if it's OK.


quad:
I gave up on this.

PR #2053, opened 2021-08-25
(WIP) This PR adds a configuration to the Wire gradle plugin to generate message descriptors

kalfonso:
Java protobuf implementation encodes message descriptors in the generated Java code. We need similar functionality in Wire to be able to use tools like Kafka serialization/deserialization using Schema Registry which uses the encoded descriptors in the Java protobufs generated messages.

This PR adds:
- [X] to ability to generate message descriptors via the Wire plugin. The encoded message descriptors are written to a separate file and added to the resources.
- [X] field to ProtoAdapter to hold the path to the message's encoded file descriptor.
- [X] ```FileDescriptorGenerator``` in its own module with the logic to generate encoded file descriptors.
- [X] ```ProtoTypeDescriptor``` utility class to retrieve a message descriptor.
- [X] a sample project to demonstrate file descriptor generation and retrieving the descriptor for a given message

TODO:
- [] generate encoded file descriptors for Google protos: any.proto, descriptor.proto, duration.proto, etc in ```wire-schema```
- [] add test using above Google protos dependencies.
- [] Generate ```sourceFile``` field also for ```EnumAdapter```?
- [] Split out ```KafkaWireSerializer``` to a separate PR.

Checks:
- [] Code compiles correctly
- [] Created tests which fail without the change (if possible)
- [] All tests passing

### DIFF
--- a/samples/wire-descriptorgen-sample/build.gradle.kts
+++ b/samples/wire-descriptorgen-sample/build.gradle.kts
@@ -1,0 +1,20 @@
+plugins {
+  kotlin("jvm")
+  id("com.squareup.wire")
+}
+
+val jar by tasks.getting(Jar::class) {
+  manifest {
+    attributes("Automatic-Module-Name" to "wire-descriptorgen-sample")
+  }
+}
+
+dependencies {
+  implementation(deps.wire.descriptor)
+  implementation(project(":samples:wire-descriptorgen-sample:protos"))
+
+  testImplementation(deps.junit)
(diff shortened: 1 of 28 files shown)


swankjesse:
```
e: /home/runner/work/wire/wire/wire-library/wire-descriptor/src/main/kotlin/com/squareup/wire/protos/ProtoTypeDescriptor.kt: (17, 75): Unresolved reference: packageName
```

PR #2056, opened 2021-08-30
Don't require -multiplatform in artifact names

swankjesse:


swankjesse:
This may require consumers to make a build configuration change. Good to get that in for 4.0.

JakeWharton (approved):
Okio next!

swankjesse:
https://github.com/square/okio/pull/985

PR #2057, opened 2021-08-31
Fix some bugs that were blocking release

swankjesse:


JakeWharton:
Are you using https://github.com/vanniktech/gradle-maven-publish-plugin? Neither of these are a problem on any of our other projects.

JakeWharton (approved):

swankjesse:
I dropped down to the built-in publish plugin in the KTS migration. I (foolishly?) want the built in plugin to be sufficient!

PR #2061, opened 2021-09-03
Fix publishing for JVM artifacts

swankjesse:
I'm doing my best to get consistency between JVM-only and Kotlin
Multiplatform publishing, but unfortunately they start out quite
different.

### DIFF
--- a/wire-library/internal-plugins/src/main/kotlin/com/squareup/gradle/InternalPublishingPlugin.kt
+++ b/wire-library/internal-plugins/src/main/kotlin/com/squareup/gradle/InternalPublishingPlugin.kt
@@ -67,6 +67,16 @@ class InternalPublishingPlugin : Plugin<Project> {
 
     project.extensions.getByType<PublishingExtension>().apply {
       publications {
+        project.afterEvaluate {
+          // Multiplatform projects get publications automatically, but JVM-only do not. Add 'em.
+          if (this@publications.isEmpty()) {
+            create("maven", MavenPublication::class.java) {
+              artifact(project.tasks.named("jar"))
+              artifact(project.tasks.named("kotlinSourcesJar"))
+            }
+          }
+        }
+
         all {
           if (this !is MavenPublication) return@all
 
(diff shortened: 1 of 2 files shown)


JakeWharton (approved):
You know there's a plugin which does all of this for you that we use on just about every other cashapp/square open source project.

wire-library/internal-plugins/src/main/kotlin/com/squareup/gradle/InternalPublishingPlugin.kt
```diff
@@ -67,6 +67,16 @@ class InternalPublishingPlugin : Plugin<Project> {
 
     project.extensions.getByType<PublishingExtension>().apply {
       publications {
+        project.afterEvaluate {
+          // Multiplatform projects get publications automatically, but JVM-only do not. Add 'em.
+          if (this@publications.isEmpty()) {
```
JakeWharton, line 72: This is kinda backwards. You should be reacting to the application of the 'java-library' plugin or something and then creating the publications. And then in afterEvaluate you can query the presence of a Kotlin plugin as to whether to call `java { withSourcesJar() }` or manually attach the Kotlin sources jar.

But... see thing about using the plugin.

swankjesse: Good call.

swankjesse:
Yeah I’m grumpy about all of this.

The vanniktech plugin is good stuff. I switched because under `.kts` it’s much nicer to work in `build.gradle.kts` files than `gradle.properties` files. It’s discoverable! But I’ve discovered that the defaults are a mess and there’s a ton of boilerplate to keep things working.

JakeWharton:
I requested a proper DSL as the properties do suck. I'm not sure what happened with that...

JakeWharton:
https://github.com/vanniktech/gradle-maven-publish-plugin#base-plugin

JakeWharton:
I haven't upgraded past 0.14.2 because of this pain point: https://github.com/vanniktech/gradle-maven-publish-plugin/issues/287. Although it's fixed in the next release!

swankjesse:
Yep, the base plugin is exactly what I want. Hooray for the next release.

PR #2064, opened 2021-09-10
Remove more build boilerplate

swankjesse:


wire-library/build.gradle.kts
```diff
@@ -80,6 +80,14 @@ subprojects {
 }
 
 allprojects {
+  tasks.withType<Jar>().configureEach {
+    if (name == "jar") {
```
jrodbx, line 84: Curious: what else could the name be?

jrodbx (approved):

wire-library/build.gradle.kts
swankjesse: I think maybe sourcesJar?

PR #2065, opened 2021-09-11
Runtime module to integrate kotlinx.serialization

swankjesse:
This might be better as a sample?


wire-library/wire-kotlinx-serialization/src/commonMain/kotlin/com/squareup/wire/kotlinxserialization/wireKotlinxSerialization.kt
```diff
@@ -0,0 +1,48 @@
...
+  override fun serialize(encoder: Encoder, value: E) {
+    val byteArray = protoAdapter.encode(value)
+    encoder.encodeSerializableValue(delegate, byteArray)
+  }
+
+  override fun deserialize(decoder: Decoder): E {
+    val byteArray = decoder.decodeSerializableValue(delegate)
+    return protoAdapter.decode(byteArray)
+  }
+}
```
swankjesse, line 48: If the lines of Gradle build stuff exceeds the lines of behavior, is it a bad module?

oldergod: I don't follow

oldergod (approved):
Did you just bring back what you deleted a while ago?

PR #2072, opened 2021-09-18
[wire-kotlin-generator] `hashCode()` improvement

chrimaeon:
Improve `hashCode()` code generation for Kotlin by removing unnecessary null checks which result in `Unnecessary safe call on a non-null receiver` and `Elvis operator (?:) always returns the left operand of non-nullable type`


swankjesse (approved):

swankjesse:
Thanks!

PR #2080, opened 2021-09-28
Also set the display name on the Gradle plugin

swankjesse:
Without this releasing fails.


lkerford (approved):

oldergod (approved):

jrodbx (commented):
Is this only required in kts world?

jrodbx (approved):

swankjesse:
Maven Central rejected the publish without this.

PR #2084, opened 2021-10-07
Swift: Memory fixes found with Address Sanitizer

efirestone:


wire-library/wire-runtime-swift/src/main/swift/WriteBuffer.swift
```diff
@@ -98,7 +98,8 @@ final class WriteBuffer {
     }
 
     func remove(count: Int, at offset: Int) {
-        memmove(storage.advanced(by: offset), storage.advanced(by: offset + count), self.count - offset)
+        let moveCount = self.count - (offset + count)
+        memmove(storage.advanced(by: offset), storage.advanced(by: offset + count), moveCount)
```
NickEntin, line 102: This isn't a functional change, is it?

NickEntin (approved):

wire-library/wire-runtime-swift/src/main/swift/WriteBuffer.swift
efirestone: Yeah, it is. Previous value was `self.count - offset`. New value is `self.count - (offset + count)`.

NickEntin: Ahh missed that. LGTM

PR #2088, opened 2021-10-16
Modernize WireTask APIs to use more modern Gradle APIs and be configuration caching friendly

### LINKED ISSUE #2086: Compiler not compatible with configuration cache
When building protos using wire `4.0.0-alpha.12` and configuration cache enabled, gradle will complain about incompatibility:

```
- Task `:my-project:generateStagingGlobalDebugProtos` of type `com.squareup.wire.gradle.WireTask`: invocation of 'Task.project' at execution time is unsupported.
  See https://docs.gradle.org/7.2/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution
```

ZacSweers:
This leans heavily on [managed properties](https://docs.gradle.org/current/userguide/custom_gradle_types.html#managed_properties), proper `Property` types, and fixes `project` accessor issues that break configuration caching in the `WireTask`. It's a breaking change but since 4.0 is in alpha it seemed like now is a good time, and it would be hard to do this without breaking anything.

Breaking in the following ways:
- Binary breaking change (obviously)
- Kotlin source breaking change (also obviously)
- _not_ a Groovy source breaking change (possibly not obvious)

Resolves #2086

This also cleans up `WirePluginTest` a bit and enables configuration caching on it.

This ensures individual tests are properly isolated from each other by defining and clearing their project and gradle home dirs between runs. Before, they would reuse caches and possibly return false positives on caches. This also adds `--configuration-cache` to the test arguments to catch any regressions to that support in the future. It does run more slowly, but that's the cost of full isolation.

Note that this also inlines the test runner initializer to the property, because it's needless to put it directly in the setup method here unless you're expecting it to possibly throw an exception and want a cleaner error message.

### DIFF
--- a/wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
+++ b/wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
@@ -191,20 +191,23 @@ class WirePlugin : Plugin<Project> {
           protoSourceInput.debug(task.logger)
           protoPathInput.debug(task.logger)
         }
-        task.outputDirectories = targets.map { target -> project.file(target.outDirectory) }
+        task.outputDirectories.setFrom(targets.map { it.outDirectory })
         task.sourceInput.set(protoSourceInput.toLocations(project))
         task.protoInput.set(protoPathInput.toLocations(project))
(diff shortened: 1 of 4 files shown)


oldergod:
That's dope, thank you

oldergod (approved):

PR #2093, opened 2021-10-19
Add support for serializing kotliny properties to JSON

jingibus:
I defer to better reflection wizards than I as to whether `trySetAccessible()` is a major issue here. I don't think it's a correctness problem — the fields should be no more or less safe than they are in java-interop classes.


wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/FieldBinding.kt
```diff
@@ -100,6 +103,15 @@ class FieldBinding<M : Message<M, B>, B : Message.Builder<M, B>> internal constr
     }
   }
 
+  private fun getInstanceGetter(messageType: Class<M>): (M)->Any? =
+    if (Modifier.isPrivate(messageField.modifiers)) {
+      val getterName = "get" + messageField.name.replaceFirstChar { it.uppercase() }
+      val getter: Method = messageType.getMethod(getterName)
+      ({ instance -> getter.invoke(instance) })
```
oldergod, line 110:
```suggestion
      { instance -> getter.invoke(instance) }
```

oldergod: might as well be able to fancy return the reference....

jingibus: Yeah, I don't get it either

PR #2097, opened 2021-10-28
Upgrade to Okio 3.0.0

swankjesse:
Use the BOM for both Okio and OkHttp.


oldergod:
Not sure what's broken so I'm redoing the work splitting things among PRs

PR #2098, opened 2021-10-28
Skunkworks impl for KMP Ktor based GRPClient

steve-the-edwards:
Async (ktor) vs. sync (okio/okhttp) apis clash but just testing to see what was possible.


steve-the-edwards:
> I believe that Ktor cannot do gRPC.

Hey sorry I had not seen this until I came back here to share the link with someone. Yes - it cannot to gRPC with the bidirectional connections but it can do the simple single-direction calls (it was hacky anyway!).

PR #2099, opened 2021-10-31
Fix enum constant name conflict with kotlin.Enum

scaraher-sc:
Given an enum such as
```proto
enum ConflictingEnumConstants {
  hello = 0;
  name = 1;
  ordinal = 2;
  open = 3;
}
```
currently the resulting generated class looks like:
```kotlin
public enum class ConflictingEnumConstants(
  public override val `value`: Int
) : WireEnum {
  hello(0),
  name(1),
  ordinal(2),
  @WireEnumConstant(declaredName = "open")
  open_(3),
  ;
// ...
```
which fails to compile with:
```
Conflicting declarations: public final val name: String, enum entry name
Conflicting declarations: public final val ordinal: Int, enum entry ordinal
```
This is a similar issue to square/kotlinpoet#991 (which resulted in the fix for `open` seen above), and I'm not sure if it would be better fixed here or in that repository. Additionally, I'm not sure where to put/how to name `safeEnumConstantName`.

### DIFF
--- a/wire-library/wire-gson-support/src/test/java/com/squareup/wire/GsonTest.java
+++ b/wire-library/wire-gson-support/src/test/java/com/squareup/wire/GsonTest.java
@@ -81,20 +81,23 @@ public class GsonTest {
                 KeywordKotlinEnum.when_,
                 KeywordKotlinEnum.fun_,
                 KeywordKotlinEnum.return_,
-                KeywordKotlinEnum.open_
+                KeywordKotlinEnum.open_,
+                KeywordKotlinEnum.name_,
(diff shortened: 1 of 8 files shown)


oldergod:
Nice find, thank you. I'll have a look this week to see if we can find a more robust approach.

oldergod:
I created a ticket in KotlinPoet https://github.com/square/kotlinpoet/issues/1183
Ideally KotlinPoet should not generate that code but even if it doesn't, Wire needs to handle it as well because `NameAllocation` would not tag them.

oldergod:
I inlined the method and added some tests. Thanks again.

PR #2108, opened 2021-11-02
Prepend package name to instance variable names clashing with their t…

mellster2012:
…ype names to avoid compilation errors

This is similar to https://github.com/square/wire/issues/299
and the fix provided and merged at https://github.com/square/wire/pull/1252
It looks like that code from the 2019 PR got lost or was removed on purpose.
The added test would lead to "error: non-static variable CommonMessage cannot be referenced from a static context" when compiling the generated Java source without this PR.


oldergod:
Why do you think the code was lost? Isn't this the test you wrote back then?
https://github.com/square/wire/blob/master/wire-library/wire-java-generator/src/test/java/com/squareup/wire/java/JavaGeneratorTest.java#L483-L497

mellster2012:
> Why do you think the code was lost? Isn't this the test you wrote back then? https://github.com/square/wire/blob/master/wire-library/wire-java-generator/src/test/java/com/squareup/wire/java/JavaGeneratorTest.java#L483-L497

Correct. When looking at the JavaGenerator code it seemed like it was the same fix for the same method but it could be a variation (as a lot of code has change since then). This also excludes scalar types which have built in adapters so there is never a name clash. The one edge case this won't work is if there are no package scopes and everything sits at the root level (as the qualified name would equal the simple name), but that is likely very rare.

wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -214,6 +214,8 @@
         for (Field field : fieldsAndOneOfFields) {
           String suggestion = collidingNames.contains(field.getName())
             || schema.getType(field.getQualifiedName()) != null
+            || (field.getName().equals(field.getType().getSimpleName())
+              && !field.getType().isScalar())
```
oldergod, line 218: Let's match what the KotlinGenerator does:
```suggestion
            || field.getName().equals(field.getType().getSimpleName())
            || schema.getType(field.getQualifiedName()) != null
```

wire-library/wire-java-generator/src/test/java/com/squareup/wire/java/JavaGeneratorTest.java
```diff
@@ -609,6 +609,27 @@ public void generateTypeUsesPackageNameOnFieldAndClassNameClashWithinPackage() t
...
+                            + "message CommonMessage {\n"
+                            + "	required string First = 1;\n"
+                            + "}\n")
+            .add("example.proto",
+                    "package a;\n"
+                            + "import \"common/common_message.proto\";\n"
+                            + "\n"
+                            + "message Example {\n"
+                            + "	required Common.CommonMessage CommonMessage = 1;\n"
+                            + "}\n");
```
oldergod, line 626: I locally formatted it for better reading.
```suggestion
        .add("common/common_message.proto",
            "package a.Common;\n"
          + "option java_package = \"a.common\";"
          + "message CommonMessage {\n"
          + "   required string First = 1;\n"
          + "}\n")
        .add("example.proto",
            "package a;\n"
          + "import \"common/common_message.proto\";\n"
          + "\n"
          + "message Example {\n"
          + "   required Common.CommonMessage CommonMessage = 1;\n"
          + "}\n");
```

oldergod (approved):
Thank you

mellster2012:
I accepted the changes, but it looks like there are issues in the build pipeline. I don't believe I need to do anything at this point, correct? Thank you!

wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -213,6 +213,7 @@
         Set<String> collidingNames = collidingFieldNames(fieldsAndOneOfFields);
         for (Field field : fieldsAndOneOfFields) {
           String suggestion = collidingNames.contains(field.getName())
+            || field.getName().equals(field.getType().getSimpleName())
```
oldergod, line 216: My apologies, the CI failure makes me realize that we didn't care about scalar types in Java because we use builders. So let's go back to your original take.
```suggestion
            || (field.getName().equals(field.getType().getSimpleName())
              && !field.getType().isScalar())
```

oldergod:
If you could lastly rebased on master, I think that'd make CI happy.

mellster2012:
Hmm, still failing..

oldergod:
It's good then. I'll fix if this happens on master. Thanks again.

PR #2111, opened 2021-11-04
Fix Swift test build

efirestone:
The build requirements changed in Xcode 12.5 such that we need to include a new framework. This change adds the /usr/lib directory that contains that framework into the build arguments so that the Swift build can find it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,7 +45,7 @@ jobs:
           ./gradlew build
 
   multiplatform:
-    runs-on: macOS-latest
+    runs-on: macos-latest
 
     steps:
       - name: Checkout
@@ -59,6 +59,9 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: 14
+
+      - name: Setup Xcode
+        run: sudo xcode-select -switch "/Applications/Xcode_12.app"
 
       - name: Check test files
         run: |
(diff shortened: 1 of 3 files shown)


oldergod (approved):
It's green now

.github/workflows/build.yml
```diff
@@ -60,6 +60,9 @@ jobs:
           distribution: 'zulu'
           java-version: 14
 
+      - name: Setup XCode
+        run: sudo xcode-select -switch "/Applications/Xcode_12.app"
```
oldergod, line 64: There was error in multiplatform about xcode not found and stuff so I added it and now it's green.

efirestone: That makes sense. Are those paths (like `Xcode_12.app`) documented from GitHub?

Also, nit: It's `Xcode` with a lower-case "C"

efirestone:
@oldergod Thanks for taking this the extra mile. I'm going to squash the commits and re-push so we can merge.

oldergod:
You know GitHub can squash it for us, right? Is that different?

efirestone:
> You know GitHub can squash it for us, right? Is that different?

Yeah...the automatic squash creates ugly commit comments by combining all the WiP comments. It's a small thing, but I usually manually squash so that I can create one nice cohesive comment.

PR #2122, opened 2021-11-10
Validate enum constant uniqueness for the entire package

swankjesse:


wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
```diff
@@ -162,10 +162,6 @@ data class ProtoFile(
     return toElement().toSchema()
   }
 
-  fun validate(linker: Linker) {
```
swankjesse, line 165: This was too narrow, they must be unique across all files in the package.

JakeWharton (approved):
C++, the gift that keeps on giving

oldergod (commented):
LGTM

gen-tests.gradle.kts
```diff
@@ -99,8 +99,8 @@ val generateGsonAdapterJavaTests by tasks.creating(JavaExec::class) {
   classpath = wire
   main = "com.squareup.wire.WireCompiler"
   args = listOf(
-          "--proto_path=wire-library/wire-tests/src/commonTest/shared/proto",
           "--proto_path=wire-library/wire-tests/src/commonTest/shared/proto/proto2",
+          "--proto_path=wire-library/wire-tests/src/commonTest/shared/proto/proto3",
```
swankjesse, line 103: This caused some files to be double-included, which had the unfortunate side-effect of triggering our enums warning. Gotta fix that too!

https://github.com/square/wire/issues/2124

PR #2125, opened 2021-11-12
Modernize WireCompiler tests and fix symlinks

oldergod:
- Renamed `FileSystem.find` -> `FileSystem.findFiles`
  - For some reason, the order we would get the list of files differed with the previous implementation so I switch to `containsExactlyInAnyOrder` for I don't think it matters.
- Renamed `FileSystem.get` -> `FileSystem.readUtf8`
- Removed `@ignore` tags on symlink tests.
- Fixed symlink implementation when loading roots.

### DIFF
--- a/wire-library/wire-test-utils/src/main/java/com/squareup/wire/testing/files.kt
+++ b/wire-library/wire-test-utils/src/main/java/com/squareup/wire/testing/files.kt
@@ -15,7 +15,6 @@
  */
 package com.squareup.wire.testing
 
-import java.io.IOException
 import java.nio.charset.Charset
 import java.util.zip.ZipEntry
 import java.util.zip.ZipOutputStream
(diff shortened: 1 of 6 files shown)


wire-library/wire-test-utils/src/main/java/com/squareup/wire/testing/files.kt
```diff
@@ -42,45 +41,18 @@ fun FileSystem.add(
   }
 }
 
-fun FileSystem.symlink(linkPathString: String, targetPathString: String) {
-  throw UnsupportedOperationException("symlinks are not yet implemented in okio.FileSystem")
```
oldergod, line 46: Now we do mate

swankjesse: # 🕺🏻

swankjesse (approved):

PR #2145, opened 2021-11-24
Don't use cache if sourceInput or protoInput changed

### LINKED ISSUE #2142: Wire task caching doesn't account for include directives for jar file
If you have a project with a wire plugin block like this:

```
wire {
  sourcePath {
    srcJar("src/main/proto/protos.jar")
    include("squareup/geology/period.proto")
  }
}
```

and then you add another `include` directive, so that it looks like this:

```
wire {
  sourcePath {
    srcJar("src/main/proto/protos.jar")
    include("squareup/geology/period.proto")
    include("squareup/dinosaurs/dinosaur.proto")
  }
}
```

this does NOT invalidate the gradle cache for the wire task, even though it should. This means adding include directives can produce erroneous output. In the example above, building the jar file with the original source would create and cache the jar file with just the Period classfile, and then after adding the dinosaur proto it wouldn't regenerate the jar file.

As far as I can tell from poking around the code, this is because when the jar file is added to the `WireInput` [here](https://github.com/square/wire/blob/de1651027879d285bdc41f0647a7fa7ceb604e58/wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt#L57-L65), it doesn't add the include directives, and so the `WireInput` (which appears to be part of the cache key) doesn't incorporate those include directives and they end up excluded from the cache key.

I wrote a test case already to convince myself of this bug, you can find it at https://github.com/square/wire/commit/5a19ba013b9b475bff1a6312122e3d930fe7a07b

oldergod:
Fixes #2142


wire-library/wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
```diff
@@ -1130,6 +1130,85 @@ class WirePluginTest {
...
+    assertThat(result.task(":generateMainProtos")?.outcome).isEqualTo(TaskOutcome.SUCCESS)
+    assertThat(result.task(":generateProtos")?.outcome).isEqualTo(TaskOutcome.SUCCESS)
+    assertThat(File(fixtureRoot, generatedPeriodProto)).exists()
+    assertThat(File(fixtureRoot, generatedDinosaurProto)).doesNotExist()
+
+    // The task has been cached.
+    assertThat(buildCacheDir.exists()).isTrue()
+
+    // Even though the task is now cached, the configuration of the sourcePath has now changed. We
+    // expect the new task to run againt, without using the cache.
```
staktrace, line 1195:
```suggestion
    // expect the new task to run again, without using the cache.
```

staktrace (approved):

wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireTask.kt
```diff
@@ -139,8 +140,8 @@ abstract class WireTask @Inject constructor(objects: ObjectFactory) : SourceTask
 
     val allTargets = targets.get()
     val wireRun = WireRun(
-      sourcePath = sourceInput.get(),
-      protoPath = protoInput.get(),
+      sourcePath = sourceInput.get().map { it.resolveWith(projectDirProperty.asFile.get()) },
```
swankjesse, line 143: extract a variable for `projectDirProperty.asFile.get()` ?

wire-library/wire-gradle-plugin/src/test/projects/cache-include-paths-1/build.gradle
```diff
@@ -0,0 +1,17 @@
+plugins {
+  id 'org.jetbrains.kotlin.jvm'
+  id 'com.squareup.wire'
+}
+
+// This file should be kept in sync with `cache-include-paths-2/build.gradle` with the exception of
```
swankjesse, line 6: Great comment!

swankjesse (approved):

oldergod:
I'll fix windows in a follow up

PR #2156, opened 2021-11-26
Handle out of order proto fields when initializing Kotlin constructor

### LINKED ISSUE #2153: JSON serializers fail on fields declared after repeated fields in Kotlin types without Java interop
To reproduce:

1. Add a new field to the `wire-library/wire-tests/src/commonTest/shared/proto/proto2/person_kotlin.proto` message:

```diff
  message Person {
    ...
    repeated PhoneNumber phone = 4;    
+
+   optional bool canadian = 5;
  }    
```

**Note**: the name of the new field should not start with "is", otherwise you'll hit #2152.

2. Regenerate test protos by running `./gradlew generateTests`.
3. Run the `wire-moshi-adapter/src/test/.../MoshiPersonTest.kotlinWithoutBuilderFromJson` test.
4. Test will fail with the following exception:

```
argument type mismatch
java.lang.IllegalArgumentException: argument type mismatch
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
	at com.squareup.wire.KotlinConstructorBuilder.build(KotlinConstructorBuilder.kt:70)
	at com.squareup.wire.MessageJsonAdapter.fromJson(MessageJsonAdapter.kt:84)
	at com.squareup.wire.MessageJsonAdapter.fromJson(MessageJsonAdapter.kt:24)
	at com.squareup.moshi.internal.NullSafeJsonAdapter.fromJson(NullSafeJsonAdapter.java:41)
	at com.squareup.moshi.JsonAdapter.fromJson(JsonAdapter.java:70)
	at com.squareup.wire.MoshiPersonTest.kotlinWithoutBuilderFromJson(MoshiPersonTest.kt:41)
```

The source of the problem seems to be the ordering of properties annotated with `WireField` in the generated code:

```kotlin
public class Person(
  @field:WireField(tag = 1)
  public val name: String,
  @field:WireField(tag = 2)
  public val id: Int,
  ...
  phone: List<PhoneNumber> = emptyList(),
  @field:WireField(tag = 5)
  public val canadian: Boolean? = null,
) : Message<..., ...> {
  @field:WireField(tag = 4)
  public val phone: List<PhoneNumber>
}
```

The following code inside `KotlinConstructorBuilder` retrieves all properties annotated with `WireField` in the order of declaration in the file, gets their values, locates the constructor and calls it passing the property values:

```kotlin
@Suppress("UNCHECKED_CAST")
override fun build(): M {
  val args = messageType.declaredWireFields() 
      .map(::get) 
      .plus(buildUnknownFields())
      .toTypedArray()

  val constructor = messageType.declaredConstructors.first()

  return constructor.newInstance(*args) as M
}

private fun Class<M>.declaredWireFields() = declaredFields
    .mapNotNull {
      it.declaredAnnotations.filterIsInstance(WireField::class.java).firstOrNull()
    }
```

But because the order in which `WireField` properties are declared differs from the order in which they appear in the constructor, arguments get passed in the wrong order leading to the type mismatch:

```
order of WireFields: (name, id, canadian, phone)
order of constructor params: (name, id, phone, canadian)
```

Working on a fix!

Egorand:
Fixes #2153

### DIFF
--- a/wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/KotlinConstructorBuilder.kt
+++ b/wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/KotlinConstructorBuilder.kt
@@ -15,6 +15,8 @@
  */
 package com.squareup.wire
 
+import com.squareup.wire.WireField.Label.REPEATED
+
 internal class KotlinConstructorBuilder<M : Message<M, B>, B : Message.Builder<M, B>>(
   private val messageType: Class<M>,
 ) : Message.Builder<M, B>() {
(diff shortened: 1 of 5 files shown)


oldergod (approved):

wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/KotlinConstructorBuilder.kt
```diff
@@ -61,17 +63,44 @@ internal class KotlinConstructorBuilder<M : Message<M, B>, B : Message.Builder<M
...
+    // - For certain types (group A), a primary constructor property is created
+    // - For other types (group B), a primary constructor parameter and a simple property are
+    //   created
+    //
+    // This leads to a mismatch between the order in which proto fields annotated with WireField
+    // appear in generated code (which matches the order in which they appear in the proto file) and
+    // the order of matching constructor parameters (https://github.com/square/wire/issues/2153).
+    // Therefore, we'll partition fields from groups A and B separately, where fields in each group
+    // will appear in the right order. This will allow us to iterate over the constructor parameters
+    // and retrieve the matching proto field based on the parameter type.
```
swankjesse, line 79: Love this

swankjesse (approved):

Egorand:
I'll merge on red, the only failures I'm seeing are windows, which @oldergod mentioned are expected at this time, and the `wrapper-validation-action` randomly failing on various jobs - there doesn't seem to be a fix for it yet.

PR #2157, opened 2021-11-29
gradle/wrapper-validation-action 1.0.4

Egorand:
This action frequently fails with "connect ETIMEDOUT". v1.0.4
seems to address this issue:

https://github.com/gradle/wrapper-validation-action/releases/tag/v1.0.4

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Validate Gradle Wrapper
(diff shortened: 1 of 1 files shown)


jrodbx (commented):
Having tried this in paparazzi, I can confirm that this doesn't work 😞

I think there's an open issue on the wrapper action repo tracker.

Egorand:
Ahh dang, thanks for the info. Actually seeing one of the jobs already failed with the same timeout error I was seeing previously. Closing.

PR #2159, opened 2021-11-29
Add flag to custom output to opt into included the default target

lkerford:
Custom Output typically set a custom target when generating proto classes.
Allow for custom outputs to opt into included the default target if no other target is supplied


wire-library/wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
```diff
@@ -855,6 +855,17 @@ class WirePluginTest {
...
 
+  @Test
+  fun customOutputDefaultTarget() {
+    val fixtureRoot = File("src/test/projects/custom-output-use-java-default-target")
+
+    val result = gradleRunner.runFixture(fixtureRoot) { buildAndFail() }
+    assertThat(result.task(":generateMainProtos")).isNotNull
+    assertThat(result.output)
+      .contains("Writing com.squareup.geology.Period")
+      .contains("src/test/projects/custom-output-use-java-default-target/build/generated/source/wire")
```
oldergod, line 866:
```suggestion
      .contains("src/test/projects/custom-output-use-java-default-target/build/generated/source/wire".withPlatformSlashes())
```

oldergod (commented):
I don't get the use case.
Why not add `java {}` along side the custom handler?

wire-library/wire-gradle-plugin/src/test/projects/custom-output-use-java-default-target/gradle.properties
```diff
@@ -0,0 +1 @@
+VERSION_NAME=4.0.0-SNAPSHOT
```
oldergod, line 1: Let's delete this file.
Same for settings.gradle below

wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireOutput.kt
```diff
@@ -121,6 +121,7 @@ open class CustomOutput @Inject constructor() : WireOutput() {
   var includes: List<String>? = null
   var excludes: List<String>? = null
   var exclusive: Boolean = true
+  var applyDefaultTargetIfNoTargetAssigned = false
```
oldergod, line 124: We need doc, or a better name. By the way, how can a custom output _not_ providing a target? Isn't it a target itself?

lkerford:
> I don't get the use case. Why not add `java {}` along side the custom handler?

We only want to add  `java {}` if no other target has been specified. We apply the custom handler in a plugin at which time we cannot check if any target have been assigned.

wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireOutput.kt
lkerford: I'm not a fan of the name, let me give it some more thought. 
So it is assumed that the customer handler generate proto classes. Our custom handler doesn't, it purely meant as addition action to be taken, after/during the generating of the proto classes. In most causes the `kotlin` or `java` target is specified but in some causes we depend on the defaults. When the rely on the defaults, the wire plugin runs the custom handler but doesn't generate any proto classes, which isn't surprising.

PR #2162, opened 2021-11-30
Require a target to be set.

lkerford:
Force a target to be specified instead of defaulting to using java.

### DIFF
--- a/wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
+++ b/wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
@@ -109,7 +109,10 @@ class WirePlugin : Plugin<Project> {
       }
     }
 
-    val outputs = extension.outputs.ifEmpty { listOf(JavaOutput()) }
+    val outputs = extension.outputs
+    check(outputs.isNotEmpty()) {
+      "At least one target must be provided for project '${project.path}"
+    }
     val hasJavaOutput = outputs.any { it is JavaOutput }
     val hasKotlinOutput = outputs.any { it is KotlinOutput }
     check(!hasKotlinOutput || kotlin.get()) {
--- a/wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
+++ b/wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
@@ -111,7 +111,8 @@ class WirePlugin : Plugin<Project> {
 
     val outputs = extension.outputs
     check(outputs.isNotEmpty()) {
-      "At least one target must be provided for project '${project.path}"
+      "At least one target must be provided for project '${project.path}\n" +
+          "See our documentation for details: https://square.github.io/wire/wire_compiler/#customizing-output"
     }
     val hasJavaOutput = outputs.any { it is JavaOutput }
     val hasKotlinOutput = outputs.any { it is KotlinOutput }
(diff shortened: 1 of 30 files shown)


wire-library/wire-gradle-plugin/src/test/projects/custom-output/build.gradle
```diff
@@ -30,4 +30,7 @@ wire {
   custom {
     customHandler = new MyCustomHandler()
   }
+
+  java{
+  }
```
oldergod, line 35: That's a change in behavior here, no?
I don't think Java was created here

wire-library/wire-gradle-plugin/src/test/projects/custom-output-no-such-class/build.gradle
```diff
@@ -9,4 +9,7 @@ wire {
   custom {
     customHandlerClass = "NoSuchClass"
   }
+
+  java{
+  }
```
oldergod, line 14: Likewise, I don't think the JavaTarget was defined here

wire-library/wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
```diff
@@ -103,6 +103,18 @@ class WirePluginTest {
...
   }
 
+  @Test
+  fun requireTarget() {
+    val fixtureRoot = File("src/test/projects/require-target")
+
+    val result = gradleRunner.runFixture(fixtureRoot) { buildAndFail() }
+
+    val task = result.task(":generateProtos")
+    assertThat(task).isNull()
```
oldergod, line 113: I wonder why it would be null, isn't it set but failed? Do you know?

I was expecting something like this
```suggestion
    assertThat(result.task(":generateProtos")?.outcome).isEqualTo(TaskOutcome.FAILED)
```

oldergod (approved):

wire-library/wire-gradle-plugin/src/test/projects/custom-output-no-such-class/build.gradle
lkerford: Good catch will fix

wire-library/wire-gradle-plugin/src/test/projects/custom-output/build.gradle
lkerford: Good catch will fix

wire-library/wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
lkerford: It should be null because we failed to apply the plugin and haven't started running the `generateProtos` task

wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -109,7 +109,10 @@ class WirePlugin : Plugin<Project> {
       }
     }
 
-    val outputs = extension.outputs.ifEmpty { listOf(JavaOutput()) }
+    val outputs = extension.outputs
+    check(outputs.isNotEmpty()) {
+      "At least one target must be provided for project '${project.path}"
```
oldergod, line 114: I wonder if we're gonna get a lot of failing consumers with this.
How about we link our doc https://square.github.io/wire/wire_compiler/#customizing-output
What do you think? I'm not sure it's obvious what a target is.


```suggestion
      "At least one target must be provided for project '${project.path}\n" +
          "See our documentation for details: https://square.github.io/wire/wire_compiler/#customizing-output"
```

lkerford: I think that is a great idea :fire

oldergod:
windows/mpp failing because of some 502 on nodejs

PR #2163, opened 2021-12-01
Expose version in wire-runtime

oldergod:


wire-library/wire-runtime/build.gradle.kts
```diff
@@ -138,3 +139,26 @@ configure<MavenPublishBaseExtension> {
...
+  outputs.dir(outputDir)
+
+  doLast {
+    val versionFile = file("${outputDir}/com/squareup/wire/Version.kt")
+    versionFile.parentFile.mkdirs()
+    versionFile.writeText(
+        """
+      |// Generated file. Do not edit!
+      |package com.squareup.wire
+      |const val VERSION = "${project.version}"
```
swankjesse, line 156: I don’t think this should be const.
https://kotlinlang.org/docs/whatsnew11.html#constant-inlining
https://kotlinlang.slack.com/archives/C0922A726/p1526398431000786

swankjesse (approved):

wire-library/wire-runtime/build.gradle.kts
swankjesse: https://github.com/square/okhttp/pull/6947

quad:
Thank you!

wire-library/wire-runtime/build.gradle.kts
```diff
@@ -43,6 +43,7 @@ kotlin {
       languageSettings.useExperimentalAnnotation("kotlin.Experimental")
     }
     val commonMain by getting {
+      kotlin.srcDir("src/generated/kotlin")
```
JakeWharton, line 46: The right way to do this is to create a task which has an output (like [this](https://github.com/cashapp/zipline/blob/85116a14c55dd2ed28138e1f483b43224943978f/zipline/build.gradle.kts#L17-L37)) and add the task to the source set (like [this](https://github.com/cashapp/zipline/blob/85116a14c55dd2ed28138e1f483b43224943978f/zipline/build.gradle.kts#L84)).

PR #2165, opened 2021-12-03
Properly inject version generating task

oldergod:


wire-library/buildSrc/src/main/kotlin/VersionWriterClass.kt
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (C) 2021 Block, Inc.
```
oldergod, line 2: 👀

JakeWharton: Thanks, I hate it.

JakeWharton (approved):

PR #2166, opened 2021-12-03
Rename WireLogger to EventListener, modifies methods and add some doc

oldergod:


wire-library/wire-compiler/src/main/java/com/squareup/wire/WireLogger.kt
```diff
@@ -15,18 +15,38 @@
...
 import okio.Path
-import io.outfoxx.swiftpoet.FileSpec as SwiftFileSpec
 
+/**
+ * Logger class used by [WireRun][com.squareup.wire.schema.WireRun] and
+ * [SchemaHandlers][com.squareup.wire.schema.Target.SchemaHandler] to log information related to
+ * processing the protobuf [Schema][com.squareup.wire.schema.Schema].
+ */
 interface WireLogger {
+  // TODO(Benoit) I think we should remove this one.
```
swankjesse, line 27: 👍🏻

swankjesse: At a minimum it belongs on the implementation, not the interface

wire-library/wire-compiler/src/main/java/com/squareup/wire/WireLogger.kt
```diff
@@ -15,18 +15,38 @@
...
+   * This is called when an artifact has been passed down to a
+   * [SchemaHandler][com.squareup.wire.schema.Target.SchemaHandler] but has been skipped. This is
+   * useful for dry-runs.
+   * @param type is the unique identifier for the skipped type.
+   */
   fun artifactSkipped(type: ProtoType)
+
+  /**
+   * This is called when [WireRun][com.squareup.wire.schema.WireRun] found unusual configurations.
+   * For instance when types marked as roots or prunes are not used.
```
swankjesse, line 49: I think the ultimate form of this is to include specific calls for each such case . . . have a method like `unusedInclude(...)` instead of a generic warn?

wire-library/wire-compiler/src/main/java/com/squareup/wire/WireLogger.kt
```diff
@@ -15,18 +15,38 @@
...
   fun setQuiet(quiet: Boolean)
-  fun artifact(outputPath: Path, filePath: String)
-  fun artifact(outputPath: Path, javaFile: JavaFile)
-  fun artifact(outputPath: Path, kotlinFile: FileSpec)
-  fun artifact(outputPath: Path, type: ProtoType, swiftFile: SwiftFileSpec)
+  /**
+   * This is called when an artifact is handled by a
+   * [SchemaHandler][com.squareup.wire.schema.Target.SchemaHandler].
+   * @param outputPath is the path where the artifact is written on disk.
+   * @param qualifiedName is the file path when generating a `.proto` file, the type or service
```
swankjesse, line 33: This is not a file path, it’s a type name?

wire-library/wire-compiler/src/main/java/com/squareup/wire/WireLogger.kt
```diff
@@ -15,18 +15,38 @@
...
+   *   name prefixed with its package name when generating a `.java` or `.kt` file, and the type
+   *   name when generating a `.swift` file.
+   */
+  fun artifactHandled(outputPath: Path, qualifiedName: String)
```
swankjesse, line 37: add a parameter with the target type? Like "Java" or "Kotlin" or "Swift" ?

swankjesse (approved):
This is good forward progress!

wire-library/wire-compiler/src/main/java/com/squareup/wire/WireLogger.kt
oldergod: I'm not sure how we do that without leaking implementations? I would assume the implementation to do that instead if needed?

wire-library/wire-compiler/src/main/java/com/squareup/wire/WireLogger.kt
oldergod: Not for protos. Since we copy/generate 1 to 1 between files.

Here's an example of what's currently logged

> Writing squareup/cash/money/MoneyElement.proto to /Users/bquenaudon/workspace/project/build/wire/proto-sources/squareup/cash/money

PR #2171, opened 2021-12-07
Don't fail the compile when an enum is in a oneof

swankjesse:


wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -404,10 +406,27 @@ class SwiftGenerator private constructor(
...
             oneOf.fields.forEach { field ->
-              addStatement(
-                  "case %L: %N = .%N(try $reader.decode(%T.self))", field.tag,
-                  oneOf.name, field.name, field.typeName.makeNonOptional()
-              )
+              when {
+                // ProtoReader.decode() return optional for enums. Handle that specially.
+                field.type!!.isEnum -> {
+                  addStatement(
+                    "case %1L: %2N = (try $reader.decode(%4T.self)).flatMap { .%3N(\$0) }",
```
bradfol, line 413: does `%3N` mean it pulls the third parameter even though it is in the fourth position?

bradfol (approved):
Thank you! 🙏

wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
swankjesse: Yes. I wanted to put the parameters in the same order for both cases.

PR #2175, opened 2021-12-10
Download WireCompiler jar in podspec instead of building from scratch

### LINKED ISSUE #2173: WireCompiler podspec should not cause Gradle to run
From @swankjesse :
Gradle leaves behind a daemon, and this pollutes the host machine.
If there’s a Gradle daemon already running, the two tend to fight and cause the build to fail
This is too slow

Maybe we could instead ship wire-compiler.jar in the Podspec?
Or even better, build a GraalVM native image so there’s no JVM at all polluting our Macs.

sethfri:
We were previously building the compiler jar, which significantly slows
down `pod install` times and pollutes the dev machine environment with a
Gradle process. Instead, we'll now download the jar from Maven Central.

Used the Maven repo URL documented here, which uses a CDN for
distribution: https://maven.apache.org/guides/mini/guide-mirror-settings.html

For testing, I pointed the Cash iOS repo at the local version of the podspec, successfully
ran `pod install`, and then used the `WireCompiler` jar.

Fixes #2173

### DIFF
--- a/WireCompiler.podspec
+++ b/WireCompiler.podspec
@@ -13,10 +13,7 @@ Pod::Spec.new do |s|
   s.module_name   = 'WireCompiler'
 
   s.prepare_command = <<-CMD
-    JAVA_HOME="$(/usr/libexec/java_home -v 1.8)"
-    ./gradlew -p wire-library :wire-compiler:shadowJar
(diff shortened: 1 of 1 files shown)


swankjesse (approved):

bradfol:
🙏

PR #2182, opened 2021-12-29
Inline WireLogger#warn into each scenarios

oldergod:
`skippedForSyntax` wasn't ever used anymore so I removed it.

I don't really like reference to `treeShakingRoots` and `treeShakingPrunes` in the logger.

### DIFF
--- a/wire-library/wire-compiler/src/main/java/com/squareup/wire/WireLogger.kt
+++ b/wire-library/wire-compiler/src/main/java/com/squareup/wire/WireLogger.kt
@@ -43,8 +43,44 @@ interface WireLogger {
   fun artifactSkipped(type: ProtoType)
 
   /**
-   * This is called when [WireRun][com.squareup.wire.schema.WireRun] found unusual configurations.
-   * For instance when types marked as roots or prunes are not used.
+   * This is called if some `root` values have not been used when Wire pruned the schema model.
+   * Note that `root` should contain package names (suffixed with `.*`), type names, and member
+   * names only. It should not contain file paths. Unused roots can happen if the referenced type
+   * or service isn't part of any `.proto` files defined in either
+   * [sourcePath][com.squareup.wire.schema.WireRun.sourcePath] or
+   * [protoPath][com.squareup.wire.schema.WireRun.protoPath], or if a broader root value is already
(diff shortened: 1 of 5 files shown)


wire-library/wire-compiler/src/main/java/com/squareup/wire/WireLogger.kt
```diff
@@ -45,8 +46,44 @@ interface WireLogger {
...
-   * For instance when types marked as roots or prunes are not used.
+   * This is called if some `root` values have not been used when Wire pruned the schema model.
+   * Note that `root` should contain package names (suffixed with `.*`), type names, and member
+   * names only. It should not contain file paths. Unused roots can happen if the referenced type
+   * or service isn't part of any `.proto` files defined in either
+   * [sourcePath][com.squareup.wire.schema.WireRun.sourcePath] or
+   * [protoPath][com.squareup.wire.schema.WireRun.protoPath], or if a broader root value is already
+   * defined.
+   */
+  fun unusedRoots(unusedRoots: Set<String>)
```
swankjesse, line 57: Love it

wire-library/wire-compiler/src/main/java/com/squareup/wire/WireLogger.kt
```diff
@@ -45,8 +46,44 @@ interface WireLogger {
...
+  fun unusedPrunes(unusedPrunes: Set<String>)
+
+  /**
+   * This is called if some `includes` values have not been used by the target they were defined in.
+   * Note that `includes` should contain package names (suffixed with `.*`) and type names only. It
+   * should not contain member names, nor file paths. Unused includes can happen if the referenced
+   * type or service isn't part of the parsed and pruned schema model, or has already been consumed
+   * by another preceding target.
+   */
+  // TODO(Benoit) We could pass the target name or something which makes it identifiable.
```
swankjesse, line 77: Yeah that’d be great.

swankjesse: (Would require us to actually have names on the targets.)

swankjesse (approved):

PR #2184, opened 2021-12-29
Adds targetname to WireLogger artifact handled/skipped methods

oldergod:


wire-library/wire-compiler/src/main/java/com/squareup/wire/WireLogger.kt
```diff
@@ -33,16 +33,26 @@ interface WireLogger {
    * @param qualifiedName is the file path when generating a `.proto` file, the type or service
    *   name prefixed with its package name when generating a `.java` or `.kt` file, and the type
    *   name when generating a `.swift` file.
+   * @param targetName is used to identify the concerned target. For
+   * [JavaTarget][com.squareup.wire.schema.JavaTarget], the name will be "Java". For
+   * [KotlinTarget][com.squareup.wire.schema.KotlinTarget], the name will be "Kotlin". For
+   * [SwiftTarget][com.squareup.wire.schema.SwiftTarget], the name will be "Swift". For
+   * [ProtoTarget][com.squareup.wire.schema.ProtoTarget], the name will be "Proto".
```
quad, line 40: Would you be open to a change that broke `Java`, `Kotlin`, `Swift` and `Proto` out into an enum?

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -255,7 +255,7 @@ data class JavaTarget(
             javaFile.packageName.replace(".", "/") /
             "${javaTypeName.simpleName()}.java"
 
-        logger.artifactHandled(modulePath, "${javaFile.packageName}.${javaFile.typeSpec.name}")
+        logger.artifactHandled(modulePath, "${javaFile.packageName}.${javaFile.typeSpec.name}", "Java")
```
quad, line 258: Could we pass the two strings into `artifactHandled` as `packageName` and `typeName` and let the logging close itself handle the interpolation / output?

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -612,12 +612,12 @@ data class NullTarget(
   ): SchemaHandler {
     return object : SchemaHandler {
       override fun handle(type: Type): Path? {
-        logger.artifactSkipped(type.type)
+        logger.artifactSkipped(type.type, "Dry-run")
```
quad, line 615: nit: `Dry-run` isn't in the documented set of targets. Would you be open any of the following ideas:

1. Refactor in `WireCompiler` to use a specialised `DryRunLogger`
2. Refactor `ConsoleWireLogger` to add a `isDryRun` field to match the `quiet` field

quad (approved):

wire-library/wire-compiler/src/main/java/com/squareup/wire/WireLogger.kt
quad: (or sealed classes, for that sweet-sweet exhaustive `when`s)

oldergod: I didn't wanna block a consumer from putting whatever they wanted here. Like custom handler should be able to use the logger and call those methods too.

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
oldergod: How about proto generation? And swift ( `"${swiftFile.moduleName}.${typeName.canonicalName}",` ) ?

My main thinking is something like:

- Wire has some built in targets (Java, Kotlin, Proto, Swift).
- Here is what to expect from Wire's built-in targets' logs.
- But Wire doesn't wanna force an API that would require gymnastic or awkward naming from its consumers.

Broadly, if something is generated, there's a location where it was generated, and its id (for java that's qualified name, for protos just the file name, etc)

It could be good to renamed `artifactHandled` to `artifactGenerated` and another `artifactProcessed` or something for side effect which don't generate code.

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
oldergod: Or maybe kill the dry-run feature... Wire Gradle plugin doesn't provide it.

How would the `DryRunLogger` work? Dry-runs use the target `NullTarget`, we'd need a way to know which target was replaced by this one

wire-library/wire-compiler/src/main/java/com/squareup/wire/ConsoleWireLogger.kt
```diff
@@ -31,15 +31,15 @@ internal class ConsoleWireLogger : WireLogger {
     }
   }
 
-  override fun artifactHandled(outputPath: Path, qualifiedName: String) {
+  override fun artifactHandled(outputPath: Path, qualifiedName: String, targetName: String) {
     if (quiet) {
       println(qualifiedName)
     } else {
-      println("Writing $qualifiedName to $outputPath")
+      println("target $targetName writing $qualifiedName to $outputPath")
```
swankjesse, line 38: I think in this specific case it’s redundant? The output path implies the file type? (We can include this information for other event listeners without using it in the console logger.)

wire-library/wire-compiler/src/main/java/com/squareup/wire/ConsoleWireLogger.kt
```diff
@@ -31,15 +31,15 @@ internal class ConsoleWireLogger : WireLogger {
...
     }
   }
 
-  override fun artifactSkipped(type: ProtoType) {
-    println("Skipping $type")
+  override fun artifactSkipped(type: ProtoType, targetName: String) {
+    println("target $targetName skipping $type")
```
swankjesse, line 43: This will print as

```
target kotlin skipping squareup.common.CurrencyCode
```

I think I’d like something else?

```
Skipping squareup.common.CurrencyCode (target=kotlin)
```
?

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
swankjesse: I think we kill it from the core project. It’s trivial to do as a custom handler now.

swankjesse (approved):

wire-library/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
oldergod: I'll drop it in follow up

PR #2189, opened 2022-01-20
Implement writeIdentityValues in JsonFactories

### LINKED ISSUE #2187: JSON option to emit identity values unconditionally
Wire follows proto3’s specified behavior and omits `0` and `false` etc. in both JSON and binary encoding.

This can be really annoying for JSON clients that aren’t aware that protobuf is on the other side of their calls. It’s unconventional in JSON to omit 0s and false values, and recovering from absent values is a hassle.

What if we added an option to `WireJsonAdapterFactory` like this?
```
  val adapter = WireJsonAdapterFactory(
    alwaysWriteIdentityValues = true,
  )
```

oldergod:
By default proto3 doesn't write identity on JSON. Let's opt that in.

fix #2187


swankjesse (approved):
Fix the binary compatibility; great PR.

PR #2191, opened 2022-02-02
Field and Constant names cannot start with digits.

oldergod:
This matches what protoc already does.

### DIFF
--- a/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/internal/parser/SyntaxReader.kt
+++ b/wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/internal/parser/SyntaxReader.kt
@@ -140,23 +140,23 @@ class SyntaxReader(
   }
 
   /** Reads a (paren-wrapped), [square-wrapped] or naked symbol name. */
-  fun readName(): String {
+  fun readName(allowLeadingDigit: Boolean = true): String {
     return when (peekChar()) {
       '(' -> {
         pos++
-        readWord().also {
+        readWord(allowLeadingDigit).also {
           expect(readChar() == ')') { "expected ')'" }
         }
       }
 
       '[' -> {
         pos++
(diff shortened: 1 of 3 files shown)


wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/internal/parser/SyntaxReader.kt
```diff
@@ -195,6 +195,11 @@ class SyntaxReader(
       }
     }
     expect(start < pos) { "expected a word" }
+    if (!allowLeadingDigit) {
+      expect(!data[start].isDigit(), location().copy(column = start - lineStart)) {
+        "field and constant names cannot start with a digit"
```
oldergod, line 200: Maybe I could rephrase it so that it doesn't use `field` or `constant` here... This method is used in many places and doesn't really know why `allowLeadingDigit` is set

Egorand: Maybe
1) Throw an exception with a more generic message, indicating that "`allowLeadingDigit` is false but name started with digit",
2) Catch at the call site that expects names to not start with digits,
3) Rethrow with a more specific message?

Egorand (approved):

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/internal/parser/SyntaxReader.kt
oldergod: `allowLeadingDigit` isn't public API so I didn't write it in the message.
The trick is to get the proper location as well.

Maybe throwing internal exception and catching it might be nice. I'll think about it.

swankjesse (commented):
Nice

PR #2192, opened 2022-02-08
Update to the latest version of swiftpoet 

pickens117:
The latest version includes the escaped reserved type names changes.

### DIFF
--- a/wire-library/wire-runtime-swift/src/test/proto/type.proto
+++ b/wire-library/wire-runtime-swift/src/test/proto/type.proto
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2021 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


JakeWharton (approved):

oldergod:
Why the draft? Let's merge?

wire-library/wire-runtime-swift/src/test/proto/type.proto
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+syntax = "proto2";
+
+enum Type {
+  UNKNOWN = 0;
+  VALUE = 1;
+}
```
oldergod, line 21: What's that for?

pickens117: Prior to bumping swiftpoet to the latest version, the `:wire-runtime-swift:generateSwiftProtos` gradle task would  fail when attempting to generate the swift version of that proto. I left it in as a "pseudo" integration test. Happy to remove it if you think it's uanessacry or add a unit test specifically tested the generated swift file from the proto.

oldergod: Then please add a comment to the proto file on the type explaining why it's there. And it's good!

pickens117: Done!

Issue #2198, opened 2022-02-11
Add support for generating repeated extension fields in annotations.


oldergod:
Related to #2195

Right now, we would generate something like
```
public annotation class SomeOption(
  public val `value`: Some
)
```
but the call site would be
```
  @SomeOption(listOf(
    LOCATION
  ))
```

We need to use arrays here instead.

PR #2199, opened 2022-02-11
Do not generate repeated options

### LINKED ISSUE #2195: Generated annotations for repeated fields doesn't compile.
Hi folks, the following compile time error occurs after updating from `4.0.0-alpha.11` to `4.0.0`:

extension.proto:
```
syntax = "proto2";
package ext;

enum PII {
  LOCATION,
  // additional values
}

extend google.protobuf.FieldOptions {
  repeated PII pii = 1;
}
```
message.proto:
```
message Address {
  string line1 = 1 [json_name = "line1", (.ext.pii) = LOCATION];
  // additional fields 
}
```

Wire `4.0.0-alpha11`:
```
public class Address(
  @field:WireField(
    tag = 1,
    adapter = "com.squareup.wire.ProtoAdapter#STRING",
    label = WireField.Label.OMIT_IDENTITY
  )
  // additional fields
)
```

Wire `4.0.0`:
```
public class Address(
  @PiiOption(listOf(
    PII.LOCATION
  ))
  @field:WireField(
    tag = 1,
    adapter = "com.squareup.wire.ProtoAdapter#STRING",
    label = WireField.Label.OMIT_IDENTITY
  )
 // additional fields
)
```
The code generated by Wire `4.0.0` fails to compile with the following errors:
`An annotation argument must be a compile-time constant`
`Type mismatch: inferred type is List<PII> but PII was expected`

So the generated `PiiOption` annotation is unaware that its value is repeated. The generated `Address` class is aware of the repeated value, but the generated code is invalid.

What is the expected behavior here? Is this a regression?

oldergod:
Fixes #2195
Related to #2198

There are no tests because the project would not compile anyway if we were to generate the repeated options in annotations.

### DIFF
--- a/wire-library/wire-schema/src/jvmMain/kotlin/com/squareup/wire/schema/internal/JvmLanguages.kt
+++ b/wire-library/wire-schema/src/jvmMain/kotlin/com/squareup/wire/schema/internal/JvmLanguages.kt
@@ -62,6 +62,10 @@ fun builtInAdapterString(type: ProtoType): String? {
 fun eligibleAsAnnotationMember(schema: Schema, field: Field): Boolean {
   val type = field.type!!
 
+  if (field.label == Field.Label.REPEATED) {
+    return false
+  }
+
   if (type == ProtoType.BYTES) {
     return false
   }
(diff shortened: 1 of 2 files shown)


oldergod:
LGTM

PR #2216, opened 2022-03-07
Gradle to 7.4, Kotlin to 1.6.10, AGP to 7.1.2, and more!

oldergod:


wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/kotlin/SourceSets.kt
```diff
@@ -23,7 +23,7 @@ import org.jetbrains.kotlin.gradle.plugin.KOTLIN_DSL_NAME
 import org.jetbrains.kotlin.gradle.plugin.KOTLIN_JS_DSL_NAME
 import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet
 
-internal val AndroidSourceSet.kotlin: SourceDirectorySet?
+internal val AndroidSourceSet.kotlinSourceDirectorySet: SourceDirectorySet?
```
oldergod, line 26: Android itself added a property called `kotlin` so we have to pick another name

.github/workflows/build.yml
```diff
@@ -13,9 +13,9 @@ jobs:
       fail-fast: false
       matrix:
         java-version:
-          - 8
-          - 9
-          - 10
+          #- 8
+          #- 9
+          #- 10
```
oldergod, line 18: I'll address that in follow up

build.gradle.kts
```diff
@@ -9,7 +9,6 @@ buildscript {
     classpath(deps.animalSniffer.gradle)
     classpath(deps.plugins.android)
     classpath(deps.protobuf.gradlePlugin)
-    classpath(deps.plugins.jmh)
```
oldergod, line 12: this is server via the portal only now 😬

wire-benchmarks/build.gradle.kts
```diff
@@ -25,9 +25,9 @@ wire {
 }
 
 jmh {
-  jvmArgs = listOf("-Djmh.separateClasspathJAR=true")
-  include = listOf("""com\.squareup\.wire\.benchmarks\.EncodeBenchmark.*""")
-  duplicateClassesStrategy = DuplicatesStrategy.WARN
+  jvmArgs.addAll(listOf("-Djmh.separateClasspathJAR=true"))
+  includes.addAll(listOf("""com\.squareup\.wire\.benchmarks\.EncodeBenchmark.*"""))
+  duplicateClassesStrategy.set(DuplicatesStrategy.WARN)
```
oldergod, line 30: I had a lot of errors like "val cannot be set more than once" and "Received a String but expected a Propery<String>", etc...

.github/workflows/build.yml
oldergod: Even though I don't have a clear plan right now.

They are commented out because the android version we updated to doesn't wanna run on java 10 and below

wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/kotlin/SourceSets.kt
swankjesse: Awww

swankjesse (approved):

PR #2218, opened 2022-03-08
Change GrpcCall.isCanceled to honor OkHttp cancelations

swankjesse:


frojasg (approved):
Thanks for doing this change. It would make it easier for application to detect when timeout happened.

oldergod (approved):

PR #2226, opened 2022-03-18
Spotless config for kotlin

oldergod:
Excluding all golden files.


build.gradle.kts
```diff
@@ -39,6 +41,17 @@ allprojects {
 }
 
 subprojects {
+  apply(plugin = "com.diffplug.spotless")
+  configure<SpotlessExtension> {
+    setEnforceCheck(false)
```
oldergod, line 46: So that CI doesn't break when formatting is off.

.editorconfig
```diff
@@ -0,0 +1,10 @@
+root = true
```
swankjesse, line 1: 👍🏻👍🏻

build.gradle.kts
swankjesse: Thanks.

wire-library/.editorconfig
```diff
@@ -0,0 +1,10 @@
+root = true
```
swankjesse, line 1: Is this file necessary?

oldergod: Hmmm, maybe not? I thought I needed one because wire-library is its own project. I'm not sure how spotless look for editorconfig
I did a quick test locally and it seems to apply properly without this.

I'll remove for now

PR #2230, opened 2022-03-24
Use CoreLoader's classLoader asResourceFileSystem instead of FileSystem.RESOURCES

oldergod:


swankjesse (approved):

wire-library/wire-schema/src/jvmMain/kotlin/com/squareup/wire/schema/CoreLoader.kt
```diff
@@ -44,7 +44,7 @@ actual object CoreLoader : Loader {
 
   override fun load(path: String): ProtoFile {
     if (isWireRuntimeProto(path)) {
-      FileSystem.RESOURCES.read("/".toPath() / path) {
+      CoreLoader::class.java.classLoader.asResourceFileSystem().read("/".toPath() / path) {
```
swankjesse, line 47: Can you cache the result of `asResourceFileSystem()` ? It has to do a non-zero amount of work!

oldergod: I put it in a `lazy` delegate

swankjesse (approved):

Issue #2231, opened 2022-03-24
Maps of scalar key/value with missing key/value cause key/value null check fails


zsoltvilagos:
### TL;DR;

ScalaPB encodes `"some" -> ""` for a field of type `map<string, string>` differently than Wire and `protoc`. While decoding with `protos` works, Wire decode fails.

### Detailed

We have a proto schema with a `map<string, string>` field like https://github.com/square/wire/blob/4.3.0/wire-library/wire-tests/src/commonTest/shared/proto/proto3/map_types.proto or https://github.com/square/wire/blob/4.3.0/wire-protoc-compatibility-tests/src/main/proto/squareup/proto3/kotlin/map_types.proto.

We use this with multiple techs including protoc, ScalaPB and Wire. One of our Scala code produces such a proto message with a `"some" -> ""` mapping.

Decoding that message with protoc works but Wire fails with:
```
Map entry with null value
java.lang.IllegalStateException: Map entry with null value
	at com.squareup.wire.MapProtoAdapter.decode(ProtoAdapter.kt:517)
	at com.squareup.wire.MapProtoAdapter.decode(ProtoAdapter.kt:440)
	at squareup.proto3.kotlin.MapTypes$Companion$ADAPTER$1.decode(MapTypes.kt:446)
	at squareup.proto3.kotlin.MapTypes$Companion$ADAPTER$1.decode(MapTypes.kt:345)
	at com.squareup.wire.ProtoAdapter.decode(ProtoAdapter.kt:374)
	at com.squareup.wire.ProtoAdapter.decode(ProtoAdapter.kt:372)
```

Producing the same message with protoc or Wire and then decoding either with protoc or Wire has no issue.

I stepped through how Wire decodes the ScalaPB produced message and identified that the map entry has key tag but not value tag.

For such, https://github.com/square/wire/blob/4.3.0/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoAdapter.kt#L490 leaves `var value` on null and then
`check(value != null) { "Map entry with null value" }` produces the above failure.

Checked protoc and Wire encoding, they produce both tags explicitly (value tag length 0).

Looked into finding clues in the official Proto Language Guide. From https://developers.google.com/protocol-buffers/docs/proto3#maps, the below seems applicable:
> If you provide a key but no value for a map field, the behavior when the field is serialized is language-dependent. In C++, Java, Kotlin, and Python the default value for the type is serialized, while in other languages nothing is serialized

[...]

PR #2232, opened 2022-03-25
Deserialize missing map scalar key/value as identity

zsoltvilagos:
Properly deal with maps of scalar types, deserializing missing scala key/value into identity.

For https://github.com/square/wire/issues/2231

### DIFF
--- a/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoAdapter.kt
+++ b/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoAdapter.kt
@@ -488,8 +488,11 @@ internal class MapProtoAdapter<K, V> internal constructor(
 
   @Throws(IOException::class)
   override fun decode(reader: ProtoReader): Map<K, V> {
-    var key: K? = null
-    var value: V? = null
+    var (key: K?, value: V?) = when (syntax){
+      Syntax.PROTO_2 -> null to null
+      // Default to identity for the sake of scalars.
+      Syntax.PROTO_3 -> entryAdapter.keyAdapter.identity to entryAdapter.valueAdapter.identity
+    }
 
     val token = reader.beginMessage()
     while (true) {
@@ -502,10 +505,6 @@ internal class MapProtoAdapter<K, V> internal constructor(
       }
     }
     reader.endMessageAndGetUnknownFields(token)
-
-    // Default to identity for the sake of scalars.
-    key = key ?: entryAdapter.keyAdapter.identity
-    value = value ?: entryAdapter.valueAdapter.identity
 
     check(key != null) { "Map entry with null key" }
     check(value != null) { "Map entry with null value" }
--- a/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoAdapter.kt
+++ b/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoAdapter.kt
@@ -488,11 +488,8 @@ internal class MapProtoAdapter<K, V> internal constructor(
 
   @Throws(IOException::class)
   override fun decode(reader: ProtoReader): Map<K, V> {
-    var (key: K?, value: V?) = when (syntax) {
-      Syntax.PROTO_2 -> null to null
-      // Default to identity for the sake of scalars.
-      Syntax.PROTO_3 -> entryAdapter.keyAdapter.identity to entryAdapter.valueAdapter.identity
-    }
+    var key: K? = entryAdapter.keyAdapter.identity
+    var value: V? = entryAdapter.valueAdapter.identity
 
     val token = reader.beginMessage()
     while (true) {
--- a/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoAdapter.kt
+++ b/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoAdapter.kt
@@ -503,6 +503,10 @@ internal class MapProtoAdapter<K, V> internal constructor(
     }
     reader.endMessageAndGetUnknownFields(token)
 
+    // Default to identity for the sake of scalars.
+    key = key ?: entryAdapter.keyAdapter.identity
+    value = value ?: entryAdapter.valueAdapter.identity
+
     check(key != null) { "Map entry with null key" }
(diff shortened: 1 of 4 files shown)


wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/Proto3WireProtocCompatibilityTests.kt
```diff
@@ -573,6 +574,33 @@ class Proto3WireProtocCompatibilityTests {
     assertThat(parsed).isEqualTo(value)
   }
 
+  @Test fun validateMapTypesWithScalarKeyValueDeserializeOmittedKeyValue() {
+    val bytes = listOf(
```
zsoltvilagos, line 578: 🤔 Hm. Actually, I think it would be best to comment why bytes are handcrafted here.

Reason: like noted in the related issue: both `protoc` & `Wire` encodes empty string in map explicity having a value field. So can't do a roundrobin test like encoding with `protoc` and decoding with `Wire` but need to have the bytes like allowed by standard (and done by ScalaPB) then parse and verify `protoc` & `Wire` decodes equally.

zsoltvilagos: I was also looking for getting the field tag (maybe in tag + wire type) format from `ADAPTER` instead of hard-coding here based on manually looking up the proto.

wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -503,6 +503,10 @@ internal class MapProtoAdapter<K, V> internal constructor(
     }
     reader.endMessageAndGetUnknownFields(token)
 
+    // Default to identity for the sake of scalars.
+    key = key ?: entryAdapter.keyAdapter.identity
+    value = value ?: entryAdapter.valueAdapter.identity
```
oldergod, line 508: 2 things
I think we can set those default values when we create both `key` and `value` above.
We also wanna to do that **only for proto3**.

So that'd give us something like
```kotlin
    var (key: K?, value: V?) = when (syntax){
      Syntax.PROTO_2 -> null to null
      Syntax.PROTO_3 -> entryAdapter.keyAdapter.identity to entryAdapter.valueAdapter.identity
    }
```

wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/Proto3WireProtocCompatibilityTests.kt
oldergod: I think that having them like this is nice

wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/Proto3WireProtocCompatibilityTests.kt
```diff
@@ -573,6 +574,33 @@ class Proto3WireProtocCompatibilityTests {
     assertThat(parsed).isEqualTo(value)
   }
 
+  @Test fun validateMapTypesWithScalarKeyValueDeserializeOmittedKeyValue() {
```
oldergod, line 577:
```suggestion
  @Test fun mapKeysAndValuesDefaultsToTheirRespectiveIdentityValue() {
```

wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/Proto3WireProtocCompatibilityTests.kt
oldergod: As a comment on the `bytes` field, could you explain what you're building, specially what is missing, like there's an entry with a key but no value, viceversa.

oldergod (approved):
👍

zsoltvilagos:
@oldergod Did the adjustments based on the comments. Could you have a look again, please?

zsoltvilagos:
> @oldergod Did the adjustments based on the comments. Could you have a look again, please?

Local test failed. Because `ProtoAdapter.newMapAdapter` via `commonNewMapAdapter` & `MapProtoAdapter` constructor takes syntax from `valueAdapter` syntax. In this case, `stringAdapter`, and that specifies Proto2 syntax.

Shall we remove the switch on syntax?

oldergod:
I see. We might have to pass the syntax of the wrapping message... I'll think about it and let you know.

oldergod:
I've checked what protoc does in proto2 and the behavior seems the same. I've added a test in proto2. The only thing missing is the misleading comment.

wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/Proto3WireProtocCompatibilityTests.kt
```diff
@@ -573,6 +574,35 @@ class Proto3WireProtocCompatibilityTests {
...
+    // Both protoc and Wire encodes such with explicit zero-length fields, so had to be hand-crafted.
+    val bytes = listOf(
+      0x0a, // MapType.map_string_string tag -> 1|010 -> 10 -> 0a
+      0x04, // length
+      0x0a, // map key tag 1 -> 1|010 -> 10 -> 0a
+      0x02, // length
+      0x64, 0x65, // de
+      0x0a, // MapType.map_string_string tag -> 1|010 -> 10 -> 0a
+      0x04, // length
+      0x12, // map key tag 1 -> 1|010 -> 10 -> 0a
```
oldergod, line 589: Could you update this one? The comment is for the key, not the value. Same for the other test in proto2.

oldergod (approved):

wire-protoc-compatibility-tests/src/test/java/com/squareup/wire/Proto3WireProtocCompatibilityTests.kt
zsoltvilagos: Sure thing.

zsoltvilagos: Done.

oldergod: Thank you, I'll merge on green

Issue #2234, opened 2022-03-28
Option annotations for Kotlin are awkward to fetch from Java reflect APIs.


oldergod:
We currently generate option on fields like the following:
```kotlin
  @MyFieldOptionOneOption(17)
  public val foo: Int? = null,
```

Kotlin reflect can access it easily but Java has to do the following:
```java
FooBar::class.java.getMethod("getFoo\$annotations").annotations
```

I'm thinking:
1. We could leave it as is.
2. We could use the use-target `field`, like we do for the `@WireField` annotation.
3. We could apply the same annotation twice, one on the property, one on the field.

`2` breaks compatibility so I don't like it much.
`3` could work but it feels wrong.

oldergod:
Won't fix

PR #2239, opened 2022-03-29
Adds generated option annotation tests for Kotlin

oldergod:
Related to #2234

No behavior change, just adding test against Kotlin generated which we didn't have, and moving the current code testing Java in a mirrored test class.
It's probably redundant to use both Kotlin and Java reflect APIs but I'd like to leave it for now as examples of accessing annotations both ways.


swankjesse (commented):
LGTM

PR #2262, opened 2022-05-03
Move SchemaHandler and dependencies into wire-schema's commonMain

oldergod:
Also inlined wire-profiles into wire-schema to avoid circular deps.


wire-library/wire-schema/build.gradle.kts
```diff
@@ -39,6 +39,8 @@ kotlin {
       dependencies {
         implementation(deps.okio.core)
         api(deps.guava)
+        api(deps.javapoet)
+        api(deps.kotlinpoet)
```
oldergod, line 43: Hmm yuck?

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ClaimedPaths.kt
```diff
@@ -27,22 +27,26 @@ class ClaimedPaths {
 
   /** Tracks that [type] has been generated to [path]. */
   fun claim(path: Path, type: Type) {
-    val existingEntry = paths.putIfAbsent(path, type.asErrorMessage())
+    val errorMessage = type.asErrorMessage()
```
oldergod, line 30: `putIfAbsent` isn't in native

PR #2269, opened 2022-05-04
MPP'ed wire-test-utils

oldergod:


oldergod:
`wire-schema-tests` isn't MPPeable as of now so this is blocked.

PR #2270, opened 2022-05-04
MPP'ed some tests for SchemaLoader and kept some JVM one as well

oldergod:


oldergod:
wire-test-utils isn't MPPeable right now so this is blocked.

PR #2272, opened 2022-05-16
Fix a crash where ProtoMember was populated with the wrong data

swankjesse:
We were mixing up the declaring class with the value class. This only
caused harm in tracking double claims.


staktrace (approved):
This is much cleaner, thanks!

oldergod (commented):
LGTM!

PR #2273, opened 2022-05-20
Use reflection for KotlinCompile.source() invocations

### LINKED ISSUE #2271: Gradle plugin is incompatible with Kotlin 1.7.0-Beta
It appears an API changed in Kotlin 1.7.0-Beta

image

ZacSweers:
This allows it to safely support Kotlin 1.7.x, where the signature of this function changed the return type. Resolves #2271


wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -159,17 +160,22 @@ class WirePlugin : Plugin<Project> {
       // Even though we add the Wire output directories into the corresponding sourceSets, the
       // compilation tasks won't know about them so we fix that here.
       if (hasJavaOutput) {
-        project.tasks.matching { it.name == "compileJava" }.configureEach {
-          (it as JavaCompile).source(generatedSourcesDirectories)
-        }
+        project.tasks
+          .withType(JavaCompile::class.java)
```
ZacSweers, line 164: Opportunistically added these to save the cast below. You may even be able to nix the name matching entirely but I wasn't sure how precise those were intended to be

gavra0 (approved):
LGTM

PR #2288, opened 2022-06-09
Properly define api dependency for okio in wire-schema

oldergod:


wire-library/wire-schema/build.gradle.kts
```diff
@@ -37,7 +36,7 @@ kotlin {
     }
     val jvmMain by getting {
       dependencies {
-        implementation(libs.okio.core)
+        api(libs.okio.core)
```
oldergod, line 39: we expose filesystem in the `SchemaHandler.Context`

Egorand (approved):

PR #2290, opened 2022-06-10
Check HTTP2 protocol is set for passed client to GrpcClient

### LINKED ISSUE #2284: Wire should confirm the pass http client is set to use HTTP2

oldergod:
fixes #2284

Should we allow `QUIC` or `SPDY_3` ?

### DIFF
--- a/wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/GrpcClient.kt
+++ b/wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/GrpcClient.kt
@@ -19,6 +19,8 @@ import com.squareup.wire.internal.RealGrpcCall
 import com.squareup.wire.internal.RealGrpcStreamingCall
 import okhttp3.Call
 import okhttp3.OkHttpClient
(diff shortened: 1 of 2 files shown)


wire-library/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/GrpcClient.kt
```diff
@@ -99,7 +101,12 @@ actual class GrpcClient private constructor(
...
 
-    fun client(client: OkHttpClient): Builder = callFactory(client)
+    fun client(client: OkHttpClient): Builder {
+      require(client.protocols.contains(HTTP_2) || client.protocols.contains(H2_PRIOR_KNOWLEDGE)) {
+        "OkHttpClient is not configured with a HTTP/2 protocol which is required for gRPC connections."
+      }
+      return callFactory(client)
+    }
 
     fun callFactory(client: Call.Factory): Builder = apply {
```
oldergod, line 111: I don't believe there's a way to check protocols on the call factory?

PR #2291, opened 2022-06-20
KotlinPoet 1.12.0

Egorand:
Note that this will also bump Kotlin to 1.7.0, so perhaps need to do the Kotlin upgrade first.


oldergod:
`./gradlew generateTests`

oldergod:
🚀

PR #2310, opened 2022-07-29
Avoid NPEs with kotlin 1.7.20

staktrace:
With kotlin 1.7.20, some of the source sets (that are nullable properties)
are actually null. The wire gradle plugin code was treating these cases as
exceptional and was throwing exceptions. With this fix, the cases are handled
more gracefully by creating an empty source set and using that instead.

No test is added, since the wire build itself suffers from this problem if
kotlin 1.7.20 is used to build, and so a successful build with 1.7.20 is itself
a test of this code.


oldergod:
We are not using `1.7.20` yet though, am I wrong?

staktrace:
> We are not using `1.7.20` yet though, am I wrong?

You are correct, wire itself is not using 1.7.20 yet. However the NPE also occurs if *consumers* of wire use 1.7.20 in their build. So it's good to fix this sooner rather than later, as it will block wire consumers (e.g. the cash polyrepo) from upgrading to kotlin 1.7.20.

swankjesse (approved):

PR #2313, opened 2022-08-03
Add buildersOnly option for Java and Kotlin target

oldergod:
I'll follow up with one golden file, and documentations.


ean5533 (approved):
❤️

PR #2321, opened 2022-08-08
Use KotlinProjectExtension to access source sets

ZacSweers:
Ported from a similar fix here: https://github.com/cashapp/sqldelight/pull/3419. This is the more correct way to hook into Kotlin source sets (https://youtrack.jetbrains.com/issue/KT-47047/Migrate-Kotlin-Gradle-Plugin-from-using-Gradle-conventions#focus=Comments-27-6224481.0-0)

This is also backward compatible to older KGP versions while fully fixing Kotlin 1.7.20+. CC @AlecStrong


oldergod:
Very sweet, thank you

PR #2322, opened 2022-08-08
Use Node 18 to support building on arm64 macOS machines

ZacSweers:
The default is 14.x, but the earliest version with apple silicon support is 16. Raising to the latest stable 18, more details here: https://youtrack.jetbrains.com/issue/KT-49109#focus=Comments-27-5259190.0-0


autonomousapps:
I ran into this issue just now as well.

oldergod:
Addressed in #2378

PR #2324, opened 2022-08-11
Introduce WireDescriptor

oldergod:


wire-library/wire-descriptor/src/main/java/com/squareup/wire/WireDescriptor.kt
```diff
@@ -0,0 +1,29 @@
...
+  private val schema: Schema,
+) {
+  // TODO(Benoit) keep lazy execution but cache it to add `getDescriptorForType`, etc?
+  fun fileDescriptorSet(): FileDescriptorSet {
+    val schemaEncoder = SchemaEncoder(schema)
+
+    return FileDescriptorSet.newBuilder()
+      .addAllFile(schema.protoFiles.map {
+        FileDescriptorProto.parseFrom(
+          schemaEncoder.encode(it).toByteArray()
```
jhump, line 24: I'm a little confused. So the schema already can emit the binary representation of a descriptor? And it does so via manually written exporter, instead of first converting to a descriptor proto and _then_ marshalling that to bytes?

Will it do the right thing for proto3 optional fields and maps? Those aspects seem far simpler to implement by first converting to the descriptor model.

Also, a couple of remarks:
1. `FileDescriptorSet` should be topologically ordered, so that a file appears _before_ it is referenced as a dependency (by later files).
2. A `List<FileDescriptor>` might be a more useful return value. Things interested in interacting with descriptors will usually prefer the `Descriptors` types over the raw `DescriptorProtos` types.

PR #2325, opened 2022-08-11
Support any ZIP archive, not just JAR

damianw:
If a dependency has a filename with an extension other than `.jar` (such as `.zip`, `.aar`, etc), Wire should still be able to ingest the proto files. This can happen even with Gradle dependencies where the filename is kind of irrelevant as well.

This PR changes the check for `isJar` to `isZip`, which is implementing by checking the magic number at the start of the file. A new test covers the case of adding sources from a file ending in `.zip`.


wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt
```diff
@@ -185,8 +186,15 @@ internal class WireInput(var configuration: Configuration) {
...
-  private val File.isJar
-    get() = path.endsWith(".jar")
+  private val File.isZip: Boolean
+    get() {
+      if (!exists()) {
+        return false
+      }
+      // See "magic numbers": https://en.wikipedia.org/wiki/ZIP_(file_format)
+      val signature = RandomAccessFile(this, "r").use { it.readInt() }
+      return signature == 0x504B0304 || signature == 0x504B0506 || signature == 0x504B0708
```
oldergod, line 196: The wiki page says the magic number could be absent as well right?
I would not mind still having a quick extension check at the beginning of the method
```
if (path.endsWith(".jar") || path.endsWith(".zip")) return true
// then what you added.
```
What do you think?

wire-library/wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
```diff
@@ -380,6 +380,21 @@ class WirePluginTest {
       )
   }
 
+  @Test
+  fun sourceZipLocalOneZipWithProtoPath() {
```
oldergod, line 384: Nice

oldergod (approved):

wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt
JakeWharton: Yeah I think the general mechanism is to scan backwards and look for the central directory at the end of the file.

Okio does this: https://github.com/square/okio/blob/14dbe665f297749df76100896f7506eed3df28fa/okio/src/jvmMain/kotlin/okio/internal/zip.kt#L68-L71

damianw: I'm fine with either of these approaches. The latter one seems a bit heavy handed for such a check, but if we want to make it as general as possible, I can implement that. @oldergod's solution probably covers 99% of the use cases though?

oldergod: good enough!

oldergod:
Thank you

PR #2330, opened 2022-08-24
Fixup test protobuf option usage to be protoc compatible

buildbreaker:
Fixed some protobuf syntax by running `protoc *.proto -o /dev/null`

Some violations:
 - String values can't be `/X`
 - Repeated values needs to be multiple assignments rather than a list
 - Options need to be surrounded with `()` before accessing it's fields

### DIFF
--- a/wire-library/wire-tests/src/commonTest/proto/java/all_types.proto
+++ b/wire-library/wire-tests/src/commonTest/proto/java/all_types.proto
@@ -114,8 +114,8 @@ message AllTypes {
   optional float default_float = 412 [default = 123.456e7 ];
   optional double default_double = 413 [default = 123.456e78 ];
   // Note: protoc doesn't allow some characters of the default value.
-  optional string default_string = 414 [default = "çok\a\b\f\n\r\t\v\1\01\001\17\017\176\x1\x01\x11\X1\X01\X11güzel" ];
-  optional bytes default_bytes = 415 [default = "çok\a\b\f\n\r\t\v\1\01\001\17\017\176\x1\x01\x11\X1\X01\X11güzel" ];
+  optional string default_string = 414 [default = "çok\a\b\f\n\r\t\v\1\01\001\17\017\176\x1\x01\x11\x01\x01\x11güzel" ];
+  optional bytes default_bytes = 415 [default = "çok\a\b\f\n\r\t\v\1\01\001\17\017\176\x1\x01\x11\x1\x01\x11güzel" ];
   optional NestedEnum default_nested_enum = 416 [default = A ];
 
   map<int32, int32> map_int32_int32 = 501;
--- a/wire-library/wire-tests/src/commonTest/proto/java/custom_options.proto
+++ b/wire-library/wire-tests/src/commonTest/proto/java/custom_options.proto
@@ -30,8 +30,10 @@ message FooBar {
   optional uint64 qux = 4 [(squareup.protos.custom_options.my_field_option_one) = 18,
                            (squareup.protos.custom_options.my_field_option_two) = 34.5,
                            (squareup.protos.custom_options.my_field_option_five) = 3,
-                           (squareup.protos.custom_options.my_field_option_six) = ["a", "b"],
-                           (squareup.protos.custom_options.my_field_option_seven) = [BAV, BAX]
+                           (squareup.protos.custom_options.my_field_option_six) = "a",
+                           (squareup.protos.custom_options.my_field_option_six) = "b",
+                           (squareup.protos.custom_options.my_field_option_seven) = BAV,
+                           (squareup.protos.custom_options.my_field_option_seven) = BAX
                            ];
   repeated float fred = 5 [(squareup.protos.custom_options.my_field_option_four) = {
       foo: 11, bar: "22", baz: { value: BAR }, fred : [444.0, 555.0],
@@ -57,7 +59,9 @@ message FooBar {
              (repeated_enum_value_option_one)=3];
     BAZ = 3 [(squareup.protos.custom_options.enum_value_option)=18,
             (squareup.protos.foreign.foreign_enum_value_option)=false,
-            (repeated_enum_value_option_two) = ["c", "d"]];
+            (repeated_enum_value_option_two) = "c",
+            (repeated_enum_value_option_two) = "d"
+    ];
   }
 }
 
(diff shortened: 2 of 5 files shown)


wire-library/wire-tests/src/commonTest/proto/java/all_types.proto
```diff
@@ -114,8 +114,8 @@ message AllTypes {
   optional float default_float = 412 [default = 123.456e7 ];
   optional double default_double = 413 [default = 123.456e78 ];
   // Note: protoc doesn't allow some characters of the default value.
-  optional string default_string = 414 [default = "çok\a\b\f\n\r\t\v\1\01\001\17\017\176\x1\x01\x11\X1\X01\X11güzel" ];
-  optional bytes default_bytes = 415 [default = "çok\a\b\f\n\r\t\v\1\01\001\17\017\176\x1\x01\x11\X1\X01\X11güzel" ];
+  optional string default_string = 414 [default = "çok\a\b\f\n\r\t\v\1\01\001\17\017\176\x1\x01\x11\x01\x01\x11güzel" ];
+  optional bytes default_bytes = 415 [default = "çok\a\b\f\n\r\t\v\1\01\001\17\017\176\x1\x01\x11\x1\x01\x11güzel" ];
```
jhump, line 118: Without this fix, `protoc` complains about syntax errors:
```
all_types.proto:117:102: Invalid escape sequence in string literal.
all_types.proto:117:105: Invalid escape sequence in string literal.
all_types.proto:117:109: Invalid escape sequence in string literal.
all_types.proto:118:100: Invalid escape sequence in string literal.
all_types.proto:118:103: Invalid escape sequence in string literal.
all_types.proto:118:107: Invalid escape sequence in string literal.
```

The "official" [spec](https://developers.google.com/protocol-buffers/docs/reference/proto2-spec#string_literals) is just wrong here. Hexadecimal literals can use either lower-case or upper-case "x", but hex byte escapes must use lower-case.

wire-library/wire-tests/src/commonTest/proto/java/custom_options.proto
```diff
@@ -30,8 +30,10 @@ message FooBar {
   optional uint64 qux = 4 [(squareup.protos.custom_options.my_field_option_one) = 18,
                            (squareup.protos.custom_options.my_field_option_two) = 34.5,
                            (squareup.protos.custom_options.my_field_option_five) = 3,
-                           (squareup.protos.custom_options.my_field_option_six) = ["a", "b"],
-                           (squareup.protos.custom_options.my_field_option_seven) = [BAV, BAX]
+                           (squareup.protos.custom_options.my_field_option_six) = "a",
+                           (squareup.protos.custom_options.my_field_option_six) = "b",
+                           (squareup.protos.custom_options.my_field_option_seven) = BAV,
+                           (squareup.protos.custom_options.my_field_option_seven) = BAX
```
jhump, line 36: Without this change (and the similar two changes below), `protoc` complains of syntax errors:
```
custom_options.proto:33:83: Expected option value.
custom_options.proto:60:48: Expected option value.
custom_options.proto:142:38: Expected option value.
custom_options.proto:143:37: Expected option value.
```
The `protoc` parser does not really understand message literals or list literals. This list literal syntax is only supported _inside_ a message literal. `protoc` treats everything between the `{` and `}` of a message literal as an opaque blob which is later unmarshalled via the proto text format, when interpreting options.

wire-library/wire-tests/src/commonTest/proto/java/custom_options.proto
```diff
@@ -115,8 +119,8 @@ extend FooBar {
 
 message MessageWithOptions {
   option (my_message_option_one).foo = 1234;
-  option (my_message_option_one.bar) = "5678";
-  option (my_message_option_one.baz.value) = BAZ;
+  option (my_message_option_one).bar = "5678";
```
jhump, line 122: When interpreting the options, everything inside parenthesis is the name of an extension. Subsequent field accesses must be outside the parentheses. Subsequent field access can actually even refer to an extension, if it also has parentheses:
```protobuf
option (cusotm.option).msg_field.(extension.for.msg_field) = "abc";
```
The actual error from `protoc`:
```
custom_options.proto:118:10: Option "(my_message_option_one.bar)" unknown. Ensure that your proto definition file imports the proto which defines the option.
```

wire-library/wire-tests/src/commonTest/proto3/java/all_types.proto
```diff
@@ -141,7 +141,7 @@ message AllTypes {
     google.protobuf.Duration oneof_duration = 619;
     google.protobuf.Struct oneof_struct = 620;
     google.protobuf.ListValue oneof_list_value = 621;
-    // google.protobuf.Value oneof_value = 622;
+//     google.protobuf.Value oneof_value = 622;
```
jhump, line 144: @buildbreaker, this looks like an accident.

jhump (commented):
I dropped some comments explaining the issues in more detail, including `protoc` error messages that arise from these sources.

wire-library/wire-tests/src/commonTest/proto/java/custom_options.proto
swankjesse: Ooooh I had no idea how it worked, and I couldn't find a good spec. This is rad!

wire-library/wire-tests/src/commonTest/proto/java/custom_options.proto
swankjesse: We've been reverse engineering but hastily. Love that you know how it's supposed to work.

swankjesse (approved):

wire-library/wire-tests/src/commonTest/proto/java/custom_options.proto
jhump: Yep, I understand. I think a lot of the issues I see in Wire were also present in my earlier attempts in [protoparse](https://pkg.go.dev/github.com/jhump/protoreflect/desc/protoparse).

At this point, I've made dozens of changes over the past two years (some non-trivial) to make it nearly perfectly match the output of `protoc`. A lot of the fine-tuning had to be based on actually reading all of the `protoc` code -- especially for some of the weirder stuff it does, like resolving relative references (in type and extension names), interpreting and serializing custom options, and the areas where the actual parser varies from the published grammars on the developer docs web sites. I've only got a handful of remaining gaps (7 to be precise, and 2 of those are actually due to bugs in `protoc` that I'm hoping will be accepted as fixes over there.)

I'm actually in the process of documenting the language, as a real spec (instead of the protoc C++ source code being the de-facto spec). Stay tuned!

PR #2332, opened 2022-08-25
provide supported options back to protoc/buf

jhump:
This also has a tiny fix so we only generate code for the files indicated in the request, instead of for the entire transitive closure of dependencies as well.


jhump:
Oops, didn't mean to open against the square repo

PR #2334, opened 2022-08-26
support nested extensions in message

jhump:
NOTE: This is a work in-progress. I've got the compiler and schema updated, and they seem to work correctly. But the code generators are not updated yet (so they don't emit extensions/annotations nested inside messages).

Extensions are allowed to be nested inside a message, just like types (messages and enums) can be.

But Wire was incorrectly treating these extensions as if they were defined at the top-level of the file. This is not correct. Like nested types, nested extensions have a fully-qualified name that includes the enclosing message(s).

I've updated the `custom_options.proto` test file with some examples and added usage of some of these as custom options to also test resolving option names to nested extensions. This revealed another issue in the way that option names were parsed.

An option name like `(foo.bar).baz.buzz.(frob.nitz).fizzle` would cause a parse failure. The parser would only read up to `(foo.bar).bar.buzz` and then compain that the next character was not an equals.

And a name like `(foo.bar).(baz.buzz)` would be successfully parsed but result in a malformed `OptionElement`: it would be represented as if it instead were `(foo.bar).baz.buzz`.

It isn't well-specified (yet), but the correct way to interpret option names is like so:

1. A name component inside of parentheses is an extension name. This component can be a compound name and have interior dots. The actual name can be a relative reference (omit some shared prefix of current context) if it does not begin with a dot.
2. A name component without parentheses is a normal field name. This component must be a simple name.

For example, `(foo.bar).baz.buzz.(frob.nitz).fizzle` is interpreted like so:
* `(foo.bar)`: This names a custom option, an extension field named `bar` in some context (enclosing message, package, etc) named `foo`.
*  `baz`: A field inside  of the type of the above extension.
* `buzz`: A field inside of the type of the above field.
* `(frob.nitz)`: An extension field named `nitz` in some context named `frob`. The extendee is the type of the above field.
* `fizzle`: A field inside of the type of the above extension.

This PR aims to resolve both -- modeling of nested extensions and resolving references thereto AND parsing complex option names in a way that they can be correctly resolved after linking.


wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Linker.kt
```diff
@@ -274,11 +274,16 @@ class Linker {
   }
 
   /** Returns the current package name from the context stack. */
-  fun packageName(): String? {
+  fun fullyQualifiedNameOfContext(): String {
```
swankjesse, line 277: Oooooh I don’t like this function name as much! Might be necessary if it’s more accurate though. And this isn’t really facing so it’s fine!

swankjesse (approved):
Looks reasonable enough! I’m happy as-is, or you can tidy it up as much as you like.

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Linker.kt
jhump: I'm happy to accept proposals for other names. Maybe `packageOrEnclosingTypeName`?

jhump:
@swankjesse, I see CI failed. I think you had to approve something for those to run, right? Now that it's approved, will it re-run on every push, so I can iterate and get everything working independently? Or would someone from Block need to approve the workflow each time I push a change?

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Linker.kt
swankjesse: Yeah that's much better. Let's use that.

swankjesse:
@jhump welcome to Wire committers. You should be able to trigger rebuilds as necessary. The committer rules are the ones you remember from Block: code changes need a code review from a committer! That might mean I approve your changes or vice versa!

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Linker.kt
jhump: So... I think we can actually just _remove_ this method and use the existing `resolveContext` (but make it non-private). It looks like it already does the right thing.

Also, I was thinking of changing this in a few places where it's named `packageName` to instead be called `namespace` and adding doc comments to clarify that it's either a package name or the name of the enclosing message. For this I am referring to the eponymous field on the `Field` type, and parameters of the same name in `Extend.fromElements`, `Field.fromElements`, and `OneOf.fromElements`.

wire-library/wire-java-generator/src/test/java/com/squareup/wire/java/JavaGeneratorTest.java
```diff
@@ -514,7 +514,7 @@ public void generateTypeUsesPackageNameOnFieldAndClassNameClash() throws Excepti
       .build();
     assertThat(new JavaWithProfilesGenerator(schema)
       .generateJava("common.proto.Person"))
-        .contains("public final Gender common_proto_Gender;");
+        .contains("public final Gender common_proto_Person_Gender;");
```
jhump, line 517: @swankjesse, could you provide some insight on why we were doing conditional use of fully-qualified names?

I think I have to change these expectations back, since these would potentially result in backwards incompatibility issues with users of Wire.

But I'm not clear on _why_ it was behaving the way it was. In general, this is now "fixed" -- in that the code was trying to use a "qualified name", but was not actually using the real qualified name for the field. The condition that does this is in `KotlinGenerator.kt`, [here](https://github.com/square/wire/blob/e259584f901b24214be1841fb86fced38e848e33/wire-library/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt#L442-L448).

The actual conditions are not clear to me. I _think_ this might be something about naming conflicts, but it's really unclear how the checks in the condition map precisely to name conflicts in the generated code.

So not only am I unclear on the premise of this name-mangling, I am also looking for input as to whether we can call this change a "fix" or we need to do something else in the kotlin generator to preserve the old naming behavior.

wire-library/wire-java-generator/src/test/java/com/squareup/wire/java/JavaGeneratorTest.java
```diff
@@ -563,7 +563,7 @@ public void generateTypeUsesPackageNameOnFieldAndClassNameClashWithinPackage() t
       .build();
     assertThat(new JavaWithProfilesGenerator(schema)
       .generateJava("common.proto.A"))
-        .contains("public final AnotherStatus common_proto_Status;");
+        .contains("public final AnotherStatus Status;");
```
jhump, line 566: I think this is from the _second_ check in the `if` condition linked in previous comment. Now that we have the correct fully-qualified name of the field (`common.proto.A.Status`), it no longer conflicts with the name of the enum (`common.proto.Status`).

Like above, I don't understand _why_ we were doing this, so I'd like guidance on whether this change is a "fix" or if we need to change the generator to preserve the previous behavior.

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/internal/parser/SyntaxReader.kt
```diff
@@ -140,20 +140,22 @@ class SyntaxReader(
   }
 
   /** Reads a (paren-wrapped), [square-wrapped] or naked symbol name. */
-  fun readName(allowLeadingDigit: Boolean = true): String {
+  fun readName(allowLeadingDigit: Boolean = true, retainWrap: Boolean = false): String {
```
oldergod, line 143: Could you add doc for `retainWrap` please?

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Field.kt
```diff
@@ -22,7 +22,12 @@ import com.squareup.wire.schema.internal.parser.OptionElement.Companion.PACKED_O
 import kotlin.jvm.JvmStatic
 
 data class Field(
-  val packageName: String?,
+  // The namespace in which the field is defined. For top-level extensions,
+  // the namespace is the package of the file in which the extension is
+  // declared or null if the file has no package. For other fields and
+  // extensions defined inside a message, this is the fully-qualified name
+  // of the enclosing message.
```
oldergod, line 29: Could we remove `The namespace in which the field is defined.` and make the comment a documentation one with `**` ?

wire-library/wire-java-generator/src/test/java/com/squareup/wire/java/JavaGeneratorTest.java
oldergod: I think that's a regression. The test case is for this specific scenario. Here is the original PR: https://github.com/square/wire/pull/1318

Maybe we could create a golden file of this and make sure it compiles.

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/MessageType.kt
```diff
@@ -246,19 +254,25 @@ data class MessageType(
       }
       val nestedTypes =
         messageElement.nestedTypes.map { get(packageName, protoType.nestedType(it.name), it, syntax) }
+
+      // namespace for nested extensions is fully-qualified name of enclosing message
```
oldergod, line 258: Please make sentences out of comments.

https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

wire-library/wire-tests/src/commonTest/proto/java/custom_options.proto
```diff
@@ -49,10 +49,22 @@ message FooBar {
 
   message More {
     repeated int32 serial = 1;
+
+    extend FooBar {
+      optional string more_string = 150;
+    }
+
+    option (my_message_option_six).(More.more_string) = "foobar";
```
oldergod, line 57: Can options actually be written as so? I thought the parens were used for custom options but what does it mean in this case? An option nested within another?

oldergod: Does protoc like this?

wire-library/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Field.kt
jhump: Doh, my bad. I've been writing too much Go lately (where doc comments use `//`). Will fix.

wire-library/wire-tests/src/commonTest/proto/java/custom_options.proto
jhump: Yes, this is valid. The path is a sequence of dot-separated names, and each name component can either be:
1. simple `field_name`
2. parenthesized `(optionally.qualified.extension.name)`

Examples:
* `option deprecated = true;`
   The field name is not parenthesized, so it's a normal field on the relevant options message.
* `option (my.message.option) = {};`
   The field name is parenthesized, so the whole thing (including all dot-separated components) is a reference to an extension (which might be qualified, hence the dots).
* `option (my.message.option).foo.bar.(Bar.baz) = 123;`
   In this case, `my.message.option` is an extension on an options message (a custom option). That extension's type is some message (let's call it `Foo`). The `foo` component is a normal field on `Foo` (let's say its type is a message named `Bar`). The `bar` component is a normal field on `Bar` (let's say its type is a message named `Baz`). Finally, the `Bar.baz` component refers to an _extension_ for `Baz`.

You can feel free to compile this file using `protoc custom_options.proto -o /dev/null` to see it is valid. If you want to see more details in the resulting descriptor, try the following:
```shell
protoc custom_options.proto -o /dev/stdout \
    | protoc --decode google.protobuf.FileDescriptorSet \
        google/protobuf/descriptor.proto custom_options.proto
```

Also see the docs for `google.protobuf.UninterpretedOption`: https://github.com/protocolbuffers/protobuf/blob/v21.5/src/google/protobuf/descriptor.proto#L740-L749

wire-library/wire-java-generator/src/test/java/com/squareup/wire/java/JavaGeneratorTest.java
jhump: Hmmm. That PR has no description 😢 

If this is a compile error, then the check that was originally here was wrong. The proto language already prevents a field from having the same name as a nested type, so there should not be an issue there. (In this case, the `Status` type is defined at a different lexical scope in the file.)

Perhaps the issue is just when generating Java code, types are unconditionally referred to without qualifiers? In the example here, inside the scope of class `A`, references to the enum would need to use the qualified name `common.proto.Status`, to disambiguate from the field named `Status`. (Seems weird to be renaming the field instead.)

oldergod: Hmm, I'm not really sure. I can add a golden file from the test once this PR merges and I'll ping you if this doesn't compile. Or we can add the golden file right now and add it somewhere in `gen-tests.gradle.kts`.

wire-library/wire-tests/src/commonTest/proto/java/custom_options.proto
oldergod: Super cool, thank you

wire-library/wire-java-generator/src/test/java/com/squareup/wire/java/JavaGeneratorTest.java
jhump: How about I change this so the old behavior is still used, but add a TODO that it smells incomplete -- the logic that is there may have solved the issue at hand, but I strongly suspect there are still edge cases that could cause compilation issues. And I also think there are alternative ways around it that are more user-friendly than mangling their field names (IMO, that should only be necessary when field names conflict with language keywords).

wire-library/wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt
```diff
@@ -1266,7 +1266,7 @@ class WireRunTest {
         |import "squareup/options/documentation.proto";
         |
         |message Octagon {
-        |  option (documentation_url) = "https://en.wikipedia.org/wiki/Octagon";
+        |  option (options.documentation_url) = "https://en.wikipedia.org/wiki/Octagon";
```
jhump, line 1269: This example was invalid. The actual protobuf compiler, `protoc` does not allow incorrectly unqualified names like this. With these example sources, `protoc` emits the following error:

```
squareup/polygons/octagon.proto:6:10: Option "(documentation_url)" unknown. Ensure that your proto definition file imports the proto which defines the option.
```
The fix here, to include the partial package name `options`, fixes it to be valid proto source.

wire-library/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -218,8 +222,8 @@ public static boolean builtInType(ProtoType protoType) {
           String suggestion = collidingNames.contains(field.getName())
             || (field.getName().equals(field.getType().getSimpleName())
               && !field.getType().isScalar())
-            || schema.getType(field.getQualifiedName()) != null
-              ? field.getQualifiedName()
+            || hasEponymousType(schema, field)
+              ? legacyQualifiedFieldName(field)
```
jhump, line 226: This effectively reverts the prior changes in this PR that resulted in differences in generated code for when a field is decided to conflict with a type name.

wire-library/wire-java-generator/src/test/java/com/squareup/wire/java/JavaGeneratorTest.java
jhump: I reverted these assertions in the tests and updated the generators to preserve prior behavior here.

jhump:
@oldergod, I think I addressed your feedback. PTAL.

wire-library/wire-java-generator/src/test/java/com/squareup/wire/java/JavaGeneratorTest.java
oldergod: 👍

oldergod:
Super sweet, thank you

PR #2345, opened 2022-10-31
gRPC Adapter Fixes

jvmakine:
Fixes two issues in the `LegacyAdapterGenerator`:
- The generated code did not take `java_package` option into account when generating a references to the request and response classes. This means it worked when the proto and kotlin packages matched, but failed otherwise. Now, the kotlin package is correctly used
- Now supports services generated with `singleMethodServices = false`. Previously, only single method service with just one method was supported. Now multi method service works with multiple services. When using `singleMethodServices = true` only one method is still supported. This needs to be fixed later.

Added integration tests for `singleMethodServices = false` and `singleMethodServices = true` checking that the generated code is what is expected.


oldergod:
a rebase on master should fix CI.

oldergod (approved):

PR #2347, opened 2022-11-03
[Swift] Add support for `google.protobuf.Any`

rebello95:
Today, adding a `google.protobuf.Any` field to a message and invoking the Wire Swift generator yields references to `Swift.Any`. This subsequently fails compilation since `Swift.Any` is not a proto-compatible type.

This PR adds support for `google.protobuf.Any` in a way that is very similar to Wire Kotlin's `AnyMessage` [implementation](https://github.com/square/wire/blob/85dc2120adfdf489d66ed425f2cbad0ecbf0184a/wire-library/docs/wire_vs_protoc.md#any):

- Adds an `AnyMessage` implementation to the Wire runtime
- Updates the generator to yield `Wire.AnyMessage` instead of `Swift.Any`
- Adds a protocol `ProtoMessage` to the Wire Runtime and alters the generator to make every outputted message conform to this protocol by adding a generated function that returns the protobuf type URL. (Note: Prior to this PR, Wire did not generate the type URLs for any messages in Swift. However, the Kotlin generator did support this through `Adapter` implementations generated with each message.)
- Adds tests for validating the `AnyMessage` serialization, deserialization, and type URL validation.


rebello95:
Not really sure about the CI failures on this

wire-library/wire-runtime-swift/src/main/swift/AnyMessage.swift
```diff
@@ -0,0 +1,118 @@
+/*
+ * Copyright 2020 Square Inc.
```
oldergod, line 2:
```suggestion
 * Copyright 2022 Block Inc.
```

wire-library/wire-runtime-swift/src/main/swift/AnyMessage.swift
```diff
@@ -0,0 +1,118 @@
...
+    public init(typeUrl: String, value: Data) {
+        self.typeUrl = typeUrl
+        self.value = value
+    }
+
+    public enum DecodingError: Swift.Error, Equatable {
+        case typeUrlMismatch
+    }
+
+    public func unpack<Output: ProtoMessage & ProtoDecodable>(
```
oldergod, line 45: For the JVM, we also added `unpackOrNull` which would returns null in case nothing was found (instead of throwing).
Could we mirror it here? https://github.com/square/wire/blob/master/wire-library/wire-runtime/src/commonMain/kotlin/com/squareup/wire/AnyMessage.kt#L40-L49

wire-library/wire-runtime-swift/src/test/swift/AnyMessageTests.swift
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright 2020 Square Inc.
```
oldergod, line 2:
```suggestion
 * Copyright 2022 Block Inc.
```

oldergod (approved):
LGTM! I don't have much experience in Swift but the logic looks good.

wire-library/wire-runtime-swift/src/test/swift/AnyMessageTests.swift
rebello95: Ah yea I saw a few variants of these around the codebase - I'll change to this one though

wire-library/wire-runtime-swift/src/main/swift/AnyMessage.swift
rebello95: Yea I saw that one in the Kotlin file, but I think it's more idiomatic to achieve the same thing in Swift by simply using an optional try at the callsite - `let unpacked = try? anyMessage.unpack(something)`. If you feel strongly I can add that though

wire-library/wire-runtime-swift/src/main/swift/AnyMessage.swift
```diff
@@ -0,0 +1,118 @@
...
+        _ type: Output.Type,
+        enumDecodingStrategy: ProtoDecoder.UnknownEnumValueDecodingStrategy = .returnNil
```
efirestone, line 47: The default strategy on `ProtoDecoder` is `.throwError`, so at the least we should match that for consistency. But perhaps as a better alternative, we should pass in the full decoder (with a default value). It allows for reuse of the same decoder if desired, and also means that if we have other customizations available on the decoder in the future, then this code won't need to change.

As an optimization, we can make the default a statically allocated, reused value, like `static let defaultDecoder: Decoder = .init()`.

wire-library/wire-runtime-swift/src/main/swift/AnyMessage.swift
```diff
@@ -0,0 +1,118 @@
...
+ * ```
+ * let anyMessage = try AnyMessage.pack(person)
+ * let person = try anyMessage.unpack(Person.self)
+ * ```
+ */
+public struct AnyMessage {
+    public let typeUrl: String
+    public let value: Data
+
+    public init(typeUrl: String, value: Data) {
```
efirestone, line 36:
```suggestion
    public init(typeURL: String, value: Data) {
```

efirestone (approved):
This looks great! Thank you for doing it!

Two small changes:
- `typeUrl` should be `typeURL` in all of the Swift APIs (which is different than Kotlin)
- Let's pass the full decoder into the `unpack` method. See my comment for more.

wire-library/wire-runtime-swift/src/main/swift/AnyMessage.swift
rebello95: Sounds good regarding passing a full `ProtoDecoder` - updated.

Regarding having a `static` default, it feels a bit weird to have that live on `AnyMessage` since decoders are used for all messages. Maybe we make that universal as a separate change if that's something we want to do?

oldergod:
Very nice, let's merge on green!

PR #2348, opened 2022-11-07
Add *.DS_Store to .gitignore

rebello95:
These files appear quite often when developing for iOS on macOS.


JakeWharton:
Please add these to your global ignores. There's no reason each project should have to do this.

Instructions are here: https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

PR #2351, opened 2022-11-11
feat: implement ProtoFileDescriptorSupplier in legacy adapter's schemaDescriptor

jvmakine:
This adds a `schemaDescriptor` to the legacy gRPC adapter, implementing `io.grpc.protobuf.ProtoFileDescriptorSupplier`
This then returns the FileDescriptor for the file where the service is defined.

This is something expected by Armeria DocService to provide a dynamic interface to the gRPC service generated from these FileDescriptors.

For now, we insert the full base64 encoded FDS to each service adapter. This can lead to duplication if multiple services are referring to the same file, but for now this is done for simplicity.

### DIFF
--- a/wire-library/wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/FileDescriptorGenerator.kt
+++ b/wire-library/wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/FileDescriptorGenerator.kt
@@ -1,0 +1,130 @@
+/*
+ * Copyright (C) 2022 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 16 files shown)


wire-library/wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/FileDescriptorGenerator.kt
```diff
@@ -0,0 +1,110 @@
+package com.squareup.wire.kotlin.grpcserver
```
r3mariano, line 1: nit: copy-paste license comment here

wire-library/wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/FileDescriptorGenerator.kt
```diff
@@ -0,0 +1,110 @@
...
+
+import com.google.protobuf.DescriptorProtos
+import com.google.protobuf.Descriptors
+import com.squareup.kotlinpoet.*
```
r3mariano, line 5: nit: no star imports (we should probably add .editorconfig, but maybe not in this PR)

PR #2352, opened 2022-11-17
Decode grpc-status-details-bin and add it to GrpcException

### LINKED ISSUE #2343: Include grpc-status-details-bin in GrpcException
Would be great to copy the structure of `misk.web.exceptions.GrpcErrorResponse` to the client side:

https://github.com/cashapp/misk/blob/af8742192729e65c22b598c5ee04a3f22513ce69/misk/src/main/kotlin/misk/web/exceptions/GrpcErrorResponse.kt#L9

Hexcles:
Only attempt to base64-decode it when there's no transport error. An invalid base64 value will be treated as IOException.

It is up to the user to unmarshal it into a string if desired.

This is a de-facto standard used by grpc:
https://groups.google.com/g/grpc-io/c/p_gCk1bn2JE

Note that this is *not* included into `responseMetadata` because gRPC metadata must not include reserved names `grpc-*` per spec: https://grpc.io/docs/what-is-grpc/core-concepts/#metadata In other words, by definition this is not "metadata".

Fixes #2343.

### DIFF
--- a/wire-library/wire-grpc-client/src/commonMain/kotlin/com/squareup/wire/GrpcException.kt
+++ b/wire-library/wire-grpc-client/src/commonMain/kotlin/com/squareup/wire/GrpcException.kt
@@ -21,4 +21,5 @@ import okio.IOException
 class GrpcException(
   val grpcStatus: GrpcStatus,
   val grpcMessage: String?
+  val grpcStatusDetails: ByteArray? = null,
 ) : IOException("grpc-status=${grpcStatus.code}, grpc-status-name=${grpcStatus.name}, grpc-message=$grpcMessage")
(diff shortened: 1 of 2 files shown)


oldergod:
Thanks @Hexcles , you think you could write a test for this?

oldergod:
We do not have a dev setup doc unfortunately. It's been a pain to build Wire for a while specially since we went MPP + included projects. We'll flatten the project soon which should help some of those hopefully.
CI says there's a compilation error right now:
``` here: `/home/runner/work/wire/wire/wire-library/wire-grpc-client/src/commonMain/kotlin/com/squareup/wire/GrpcException.kt`

Hexcles:
Replaced by https://github.com/square/wire/pull/2408

PR #2353, opened 2022-11-17
[Swift] Fix JSON incorrectly encoding `unknownFields`

rebello95:
For empty messages, the current implementation encodes `{"unknownFields":""}` as the JSON representation instead of `{}`.

This is because `CodingKeys` are not generated at all for empty messages, and iOS instead synthesizes the `unknownFields` key.

To fix the issue, we should generate empty `CodingKeys` for empty messages.

### DIFF
--- a/wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
+++ b/wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
@@ -569,6 +569,15 @@ class SwiftGenerator private constructor(
               }
               .build()
           )
+        } else {
+          // Coding keys still need to be specified in order to prevent `unknownFields` from getting serialized
+          // in the JSON/Codable scenario.
+          addType(
+            TypeSpec.enumBuilder(codingKeys)
+              .addModifiers(PUBLIC)
+              .addSuperType(codingKey)
+              .build()
+          )
         }
 
         // If there are any oneofs we cannot rely on the built-in Codable support since the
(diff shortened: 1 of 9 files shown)


wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -569,6 +569,15 @@ class SwiftGenerator private constructor(
               }
               .build()
           )
+        } else {
```
JakeWharton, line 572: Can we remove the `if` altogether and get the same result? Is there any harm in having String as the super type of an empty enum?

JakeWharton (approved):

wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
rebello95: Unfortunately that doesn't compile: 
Screen Shot 2022-11-17 at 4 33 03 PM

JakeWharton: Ah, got it.

Can we do then instead
```kotlin
.apply {
  if (theStuff.isNotEmpty()) {
    addSupertype(string)
    for (stuff in theStuff) {
      addEnumConstant(...)
    }
  }
}
```
so we can at least reuse the majority of the type definition?

rebello95: Sure, updated.

PR #2354, opened 2022-11-25
Add interfaces to legacy adapter Service and Marshaller

jvmakine:
So that we can detect at runtime we are using Wire implementations.

This will be useful in projects like Armeria, where we can use Wire specific JSON transcoder with Wire services

### DIFF
--- a/wire-library/wire-grpc-server/src/main/java/com/squareup/wire/kotlin/grpcserver/WireMethodMarshaller.kt
+++ b/wire-library/wire-grpc-server/src/main/java/com/squareup/wire/kotlin/grpcserver/WireMethodMarshaller.kt
@@ -1,0 +1,14 @@
+package com.squareup.wire.kotlin.grpcserver
+
+import io.grpc.MethodDescriptor
+
+/**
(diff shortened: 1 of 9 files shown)


wire-library/wire-grpc-server/src/main/java/com/squareup/wire/kotlin/grpcserver/WireMethodMarshaller.kt
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Interface for generated method marshallers used in Google style grpc stubs.
+ *
+ * Adds a method for fetching the underlying class at run time to support
+ * generating other marshallers (like JSON encoding) from an instance of
+ * this class.
+ */
+interface WireMethodMarshaller<T>: MethodDescriptor.Marshaller<T> {
+  fun clazz(): Class<T>
```
mpeyper, line 13: Feel free to ignore, but I find `clazz` to be not very useful in identifying what the class represents.  Perhaps something like `marshalledClass` or something like that.

mpeyper (approved):

PR #2357, opened 2022-12-08
feat: support suspending APIs in the protoc adapter

jvmakine:
Adds support for 
```
rpcCallStyle = "suspending"
```
servers in the protoc adapter, making it possible to use Wire generated suspending services with Armeria.

Does not yet support generating stubs for the coroutine based services, this will be a further PR


wire-library/wire-grpc-server/src/main/java/com/squareup/wire/kotlin/grpcserver/FlowAdapter.kt
```diff
@@ -0,0 +1,62 @@
+package com.squareup.wire.kotlin.grpcserver
```
oldergod, line 1: copyright

r3mariano (approved):

PR #2366, opened 2023-01-05
Swift: Sendable conformance

lickel:
This adds `Sendable` conformance to all codegened types.
Structs that need to use `@Heap` are `@unchecked`. 
All other types are checked.

This follows the existing patterns of guarding with `#if` checks as well as checking for `swift(>=5.5)`

### DIFF
--- a/wire-library/wire-runtime-swift/src/main/swift/AnyMessage.swift
+++ b/wire-library/wire-runtime-swift/src/main/swift/AnyMessage.swift
@@ -73,7 +73,7 @@ extension AnyMessage: Hashable {
 }
 #endif
 
-#if swift(>=5.5) && !WIRE_REMOVE_HASHABLE
+#if swift(>=5.5) && !WIRE_REMOVE_SENDABLE
 extension AnyMessage: Sendable {
 }
 #endif
--- a/wire-library/wire-runtime-swift/src/main/swift/AnyMessage.swift
+++ b/wire-library/wire-runtime-swift/src/main/swift/AnyMessage.swift
@@ -73,7 +73,7 @@ extension AnyMessage: Hashable {
 }
 #endif
 
-#if swift(>=5.5) && !WIRE_REMOVE_SENDABLE
+#if swift(>=5.5)
 extension AnyMessage: Sendable {
 }
 #endif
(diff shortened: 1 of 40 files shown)


wire-library/wire-runtime-swift/src/main/swift/AnyMessage.swift
```diff
@@ -64,11 +64,18 @@ public struct AnyMessage {
 }
 
 #if !WIRE_REMOVE_EQUATABLE
-extension AnyMessage: Equatable {}
+extension AnyMessage: Equatable {
```
lickel, line 67: Just matching the style of other extensions here

wire-library/wire-tests-swift/src/main/swift/AllTypes.swift
```diff
@@ -1943,6 +1948,11 @@ extension AllTypes : Hashable {
 }
 #endif
 
+#if swift(>=5.5) && !WIRE_REMOVE_SENDABLE
+extension AllTypes : @unchecked Sendable {
```
lickel, line 1952: I explicitly did not make `Heap<>` conform to sendable since that would necessarily be @unchecked.
Due to that, this need @unchecked conformance.

In theory, we could change things to be:
```swift
struct HeapCow<T> {
    private var storage: Heap<T>
    public var wrappedValue {
        get { storage.wrappedValue }
        set { 
            copyStorage()
            storage.wrappedValue = newValue
        }
    }
}
```

and then remove `copyStorage()` from the heap-allocated protos

lickel: If we did this, then `HeapCow` could be `@unchecked Sendable`

efirestone (approved):

wire-library/wire-tests-swift/src/main/swift/AllTypes.swift
efirestone: I agree with what you have. `Heap` itself is not `Sendable` and marking it as such isn't correct. Put another way, the thing that makes these (unchecked) `Sendable` is the calls to `copyStorage`, which are in the proto object code.

PR #2369, opened 2023-01-13
Swift: Codable JSON Enums

lickel:
This started as an attempt to fix all feedback from https://github.com/square/wire/pull/1900

Once I started compiling everything I noticed that it _only_ worked for non-optional Enums.
I then added support for `Enum?` and `Array<Enum>`.

Additionally, I added encoding and decoding strategies via `CodingUserInfoKey`

Of note: I _explicitly_ did not add support for decoding a non-existent value into `Array<Enum>`. 
I intend to add another property wrapper that does this and apply it to all arrays.


wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -732,6 +732,10 @@ class SwiftGenerator private constructor(
       if (!forStorageType && field.isDeprecated) {
         property.addAttribute(deprecated)
       }
+      val prototype = field.type
+      if (prototype != null && schema.getType(prototype) is EnumType) {
+        property.addAttribute("JSONEnum")
```
lickel, line 737: Does Proto version matter here?

wire-library/wire-runtime-swift/src/main/swift/JSONEnum.swift
```diff
@@ -0,0 +1,61 @@
...
+            try self.init(string: string)
+        } else {
+            // If the value wasn't a string, then look for the field index instead.
+            let fieldNumber = try container.decode(UInt32.self)
+            try self.init(fieldNumber: fieldNumber)
+        }
+    }
+
+    public func encode(to encoder: Encoder) throws {
+        try String(describing: wrappedValue).encode(to: encoder)
```
lickel, line 45: Would it make sense to configure this to encode integers vs strings via [Encoder.userInfo](https://developer.apple.com/documentation/swift/encoder/userinfo)?

lickel:
I think now I want a Wire lossy enum conversion flag that won't throw and just skip

Package.swift
```diff
@@ -21,6 +21,23 @@ let package = Package(
             name: "Wire",
             path: "wire-library/wire-runtime-swift/src/main/swift"
         ),
+        // The tests depend on valid protos via gradle
```
lickel, line 24: By defining these test targets most if not all Swift development can be done / validated in Xcode

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/JSONEnumEncoding.swift
```diff
@@ -0,0 +1,56 @@
...
+/// Defaults to .string
+public enum EnumEncodingStrategy {
+    case string
+    case integer
+}
+
+/// The decoding strategy to use for JSONEnum types
+/// Defaults to .shouldThrow
+/// - Note: @JSONEnum will _always_ throw
+public enum EnumDecodingStrategy {
```
lickel, line 29: I figured an enum is the most flexible thing here so that we could change the implementation without having a source-breaking change.

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/JSONEnum.swift
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+import Foundation
+
+/**
+ Converts enums to/from their string equivalent when serializing to/from JSON.
+ This matches the Proto3 JSON spec: https://developers.google.com/protocol-buffers/docs/proto3#json
+ */
+@propertyWrapper
+public struct JSONEnum<T : CaseIterable & Hashable & RawRepresentable> : Codable, Hashable where T.RawValue == UInt32 {
```
lickel, line 24: I will be applying this pattern of having 3 property wrappers to the current JSONString.
Of note, I will mark `JSONString` as deprecated and create new property wrappers.

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/JSONOptionalEnum.swift
```diff
@@ -0,0 +1,77 @@
...
+extension JSONOptionalEnum : Sendable where T : Sendable {
+}
+#endif
+
+public extension KeyedDecodingContainer {
+    func decode<T: CaseIterable & Hashable & RawRepresentable>(
+        _: JSONOptionalEnum<T>.Type,
+        forKey key: Key
+    ) throws -> JSONOptionalEnum<T> {
+        if let value = try decodeIfPresent(JSONOptionalEnum<T>.self, forKey: key) {
```
lickel, line 71: This allows for non-existent values to encode as a fundamentally non-optional `JSONOptionalEnum<T>` type

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/JSONEnum.swift
lickel: I'm tempted to say a better name for this would be `ProtoEnumEncoded`.

I'm also tempted to suggest that we add `LosslessStringConvertible` to proto enums

PR #2371, opened 2023-01-13
Swift: Codable JSON Enums

lickel:
Match the JSON spec for serialization of enums: https://developers.google.com/protocol-buffers/docs/proto3#json

This started as an attempt to fix all feedback from https://github.com/square/wire/pull/1900
Given the level of changes, I am just targeting `master` directly.

Once I started compiling everything I noticed that it _only_ worked for non-optional Enums.
I then added support for `Enum?` and `Array<Enum>`.

Additionally, I added encoding and decoding strategies via `CodingUserInfoKey`

Of note: I _explicitly_ did not add support for decoding a non-existent value into `Array<Enum>`. 
I implemented this in https://github.com/square/wire/pull/2372 via a property wrapper that I applied to all arrays and maps.


Package.swift
```diff
@@ -21,6 +21,23 @@ let package = Package(
             name: "Wire",
             path: "wire-library/wire-runtime-swift/src/main/swift"
         ),
+        // The tests depend on valid protos via gradle
+        // ./gradlew -p wire-library :wire-runtime-swift:generateTestProtos
+        .testTarget(
```
lickel, line 26: By defining these test targets, most swift development can be done seamlessly in Xcode, including test validation

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/JSONEnum.swift
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+import Foundation
+
+/**
+ Converts enums to/from their string equivalent when serializing to/from JSON.
+ This matches the Proto3 JSON spec: https://developers.google.com/protocol-buffers/docs/proto3#json
+ */
+@propertyWrapper
+public struct JSONEnum<T : CaseIterable & Hashable & RawRepresentable> : Codable, Hashable where T.RawValue == UInt32 {
```
lickel, line 24: I will be applying this pattern of having 3 property wrappers to the current JSONString.
Of note, I will mark `JSONString` as deprecated and create new property wrappers.

- [ ] Rename this to (say) `ProtoEnumEncoded`.
- [ ] Add `LosslessStringConvertible` conformance to proto enums and simplify this wrapper

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/JSONEnumEncoding.swift
```diff
@@ -0,0 +1,56 @@
...
+/// Defaults to .string
+public enum EnumEncodingStrategy {
+    case string
+    case integer
+}
+
+/// The decoding strategy to use for JSONEnum types
+/// Defaults to .shouldThrow
+/// - Note: @JSONEnum will _always_ throw
+public enum EnumDecodingStrategy {
```
lickel, line 29: I figured an enum is the most flexible thing here so that we could change the implementation without having a source-breaking change.

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/JSONOptionalEnum.swift
```diff
@@ -0,0 +1,77 @@
...
+        try storage.encode(to: encoder)
+    }
+}
+
+#if swift(>=5.5)
+extension JSONOptionalEnum : Sendable where T : Sendable {
+}
+#endif
+
+public extension KeyedDecodingContainer {
```
lickel, line 66: This extension allows for non-existent values to encode as a fundamentally non-optional `JSONOptionalEnum<T>` type

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/JSONEnum.swift
lickel: https://github.com/square/wire/pull/2373

lickel: That said, I was able to drop the property wrappers entirely for Enums:
https://github.com/square/wire/pull/2374

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/JSONEnumEncoding.swift
efirestone: I like the strategy enum. Can we make the naming (both the type and the cases) closer to the existing [UnknownEnumValueDecodingStrategy](https://github.com/square/wire/blob/master/wire-library/wire-runtime-swift/src/main/swift/ProtoDecoder.swift#L34)? Or possible refactor to use that one, but I think they're conceptually separate and having two probably makes sense.

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/JSONEnumEncoding.swift
```diff
@@ -0,0 +1,56 @@
...
+    case shouldThrow
+    case shouldSkip
+}
+
+extension CodingUserInfoKey {
```
efirestone, line 34: This gets replaced with the enum box type later, right?

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/JSONOptionalEnum.swift
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+import Foundation
+
+/**
+ Converts enums to/from their string equivalent when serializing to/from JSON.
+ This matches the Proto3 JSON spec: https://developers.google.com/protocol-buffers/docs/proto3#json
+ */
+@propertyWrapper
+public struct JSONOptionalEnum<T : CaseIterable & Hashable & RawRepresentable> : Codable, Hashable where T.RawValue == UInt32 {
```
efirestone, line 24: This goes away in the stacked PR, right?

PR #2374, opened 2023-01-13
Swift: Enum encoding without wrappers

lickel:
This enables (de)serialization of Proto Enums with the exact same semantics as property wrappers without use of a property wrapper.

I needed to use a private Box type in order to maintain support for `decoder.enumDecodingStrategy`.
The reason for this is that the container does not have a (public) reference to `decoder`


wire-library/wire-runtime-swift/src/main/swift/ProtoEnum.swift
```diff
@@ -27,4 +29,120 @@ extension ProtoEnum {
...
+            let box = try container.decode(BoxedEnum<T>.self)
+            if let value = box.value {
+                results.append(value)
+            }
+        }
+
+        return results
+    }
+
+    func decode<T: ProtoEnum>(
```
lickel, line 128: It should be possible to do something like this for Dictionary; however, it's a bit more work and currently Wire doesn't handle them.

The only marginally hard part is getting access to `decoder.enumDecodingStrategy` to support lossy dictionary conversion.

lickel: Broadly: 
* make another Box type for Dictionary<Enum, Decodable>
* make a hack coding key like `_DictionaryCodingKey`
* fetch a keyed container with the hack coding key
* `container.allKeys.reduce(into: [:])`
    * potentially lossily convert hack coding key -> enum
    * insert into memo

iOS 15.4 adds support for `CodingKeyRepresentable` but I don't know if that makes anything easier while still attempting to maintain lossy conversion.

lickel: If you want me to implement this, let me know

wire-library/wire-runtime-swift/src/test/swift/ProtoEnumCodableTests.swift
```diff
@@ -0,0 +1,393 @@
...
+
+        let jsonData = try! encoder.encode(expectedStruct)
+        let actualJson = String(data: jsonData, encoding: .utf8)!
+        XCTAssertEqual(expectedJson, actualJson)
+
+        let actualStruct = try! JSONDecoder().decode(OptionalTypes.self, from: jsonData)
+        XCTAssertEqual(expectedStruct, actualStruct)
+    }
+
+
```
efirestone, line 217: nit: extra newline

efirestone (approved):

wire-library/wire-runtime-swift/src/main/swift/CodingUserInfoKey+Wire.swift
```diff
@@ -0,0 +1,57 @@
...
+    /// The encoding strategy to use for ProtoEnum types
+    /// Defaults to .string
+    public enum CodableEnumEncodingStrategy {
+        case string
+        case integer
+    }
+}
+
+extension CodingUserInfoKey {
+    /// Control the encoding of ProtoEnum types
```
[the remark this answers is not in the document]
lickel: I would prefer not to use `JSONEncoder` since _technically_ any Encoder can use this.
That said, I do like making them properties (and debated doing that before)

wire-library/wire-runtime-swift/src/main/swift/ProtoEnum.swift
```diff
@@ -0,0 +1,148 @@
...
+    ) throws -> Set<T> {
+        var container = try nestedUnkeyedContainer(forKey: key)
+
+        var results: Set<T> = []
+        if let count = container.count {
+            results.reserveCapacity(count)
+        }
+
+        while !container.isAtEnd {
+            let box = try container.decode(BoxedEnum<T>.self)
```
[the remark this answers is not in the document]
lickel: (weird, I thought I already posted this comment)

I could decode `Array<BoxedEnum<T>>.self`, etc, but that would create an extra copy / iteration.
I'd prefer to just do manual iteration since this nominally benefits from a fast path.

wire-library/wire-runtime-swift/src/main/swift/CodingUserInfoKey+Wire.swift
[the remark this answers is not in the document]
lickel: _gah_ I can't make it settable on all (en/de)coders... OK JSON namespacing it is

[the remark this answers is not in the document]
lickel: OK I rooted it off of the relevant JSON coder, but kept the user info keys public.

wire-library/wire-runtime-swift/src/main/swift/ProtoEnum.swift
lickel: I ended up needing a property wrapper 🤷‍♂️

PR #2376, opened 2023-01-17
Swift: string encoded maps

lickel:
This correctly encodes/decodes maps per the [Proto3 spec](https://developers.google.com/protocol-buffers/docs/proto3#json):

```
proto3   | JSON   | JSON example | Notes
map<K,V> | object | {"k": v, …}  | All keys are converted to strings.
```

# Note

Per the spec, this should _by default_ be serializing and deserializing `camelCase` keys.
Further, the .proto file can customize the key name via `json_name`.

I think the only way that could feasibly be done is via custom Codable implementations.
That said, that's kind of idempotent from being able to encode/decode all properties.


wire-library/wire-tests/src/commonTest/proto/kotlin/map.proto
```diff
@@ -21,6 +21,19 @@ message Mappy {
   map<string, Thing> things = 1;
 }
 
+message MappyTwo {
```
lickel, line 24: This may technically be proto syntax 3 ?
That said, iOS refuses to emit proto3 maps for whatever reason (I don't see anything in SwiftGenerator)

wire-library/wire-runtime-swift/src/test/swift/CodableTests.swift
```diff
@@ -29,10 +29,10 @@ final class CodableTests: XCTestCase {
           "repeated_int32":[1,2,3],
           "opt_uint64":"4",
           "opt_double":6,
-          "map_int32_string":[
```
lickel, line 32: I'd be shocked if anyone actually depended upon this Codable representation since it's not remotely idiomatic JSON

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/ProtoMap.swift
```diff
@@ -0,0 +1,179 @@
+//
+//  ProtoMap.swift
+//
+//  Created by Adam Lickel on 1/16/23.
```
efirestone, line 4: Update this

wire-library/wire-runtime-swift/src/main/swift/ProtoEnum.swift
```diff
@@ -64,7 +64,7 @@ extension ProtoEnum {
 }
 
 /// This type gives access to `decoder.enumDecodingStrategy`
-private struct BoxedEnum<T: ProtoEnum> : Decodable {
+struct BoxedEnum<T: ProtoEnum> : Decodable {
```
efirestone, line 67: Can this be `internal`?

wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -83,6 +83,11 @@ class SwiftGenerator private constructor(
     get() = schema.getType(this) is EnumType
   private val Type.typeName
     get() = type.typeName
+  private val ProtoType.isStringEncoded
```
efirestone, line 86: What about `SFIXED64` and `UINT64`?

efirestone (approved):

wire-library/wire-runtime-swift/src/main/swift/ProtoEnum.swift
lickel: It's implicitly internal

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/ProtoMap.swift
lickel: Whoops!

wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
lickel: Hah I do not remember where I got this list from

lickel: Or why it's different from `fun TypeName.needsStringEncoded()` which is _only_ Int types...
back to the docs

wire-library/wire-runtime-swift/src/main/swift/ProtoEnum.swift
efirestone: Oh I swear it said `public` when I looked at it. Maybe that was before the rebase.

wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
lickel: OK unified and fixed up

PR #2377, opened 2023-01-19
Swift: Custom codable implementation

lickel:
This does a few things:
* Remove most of the property wrappers from Proto structs
    * They are still being used under the hood for encoding/decoding
* Removes effectively all `CodingKeys` enums
    * I'm using a struct that conforms to `ExpressibleByStringLiteral` so I can use "strings"
    * The alternative would be to have additional cases like `case .camelCase`
* Always implements Encodable / Decodable methods
    * This is required because we have two valid keys for all encoding / decoding operations
    * `camelCase` is the default / preferred encoding for all key names
    * Empty / nil values are not conditionally emitted into JSON (default false)


wire-library/wire-runtime-swift/src/main/swift/CodingUserInfoKey+Wire.swift
```diff
@@ -29,16 +29,6 @@ extension JSONDecoder {
         /// - Note: For dictionaries, it is necessary to wrap it in `@ProtoMapEnumValues`
         case returnNil
     }
-
-    /// The decoding strategy to use for StringEncoded types that are themselves Decodable
-    /// Defaults to .disallowRawDecoding
-    /// - Note: ProtoMap Dictionary keys are always decoded as strings
-    public enum StringEncodedDecodingStrategy {
```
lickel, line 36: I noticed this while re-reading the proto3<->json document.
Raw numbers are supposed to be accepted

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/StringEncoded.swift
```diff
@@ -36,33 +36,14 @@ extension StringEncoded : Decodable {
...
-
-        default:
-            let stringValue = try container.decode(String.self)
+        if let stringValue = try? container.decode(String.self) {
             let value = try Self.create(optionalEncodedValue: stringValue)
             self.init(wrappedValue: value)
+            return
         }
+
+        let value = try container.decode(Value.self)
```
lickel, line 45: Since a) all types we care about are actually codable, and b) Raw decoding is expected, I figured I might as well just require `Value` to conform to Codable

wire-library/wire-runtime-swift/src/main/swift/wellknowntypes/Duration.swift
```diff
@@ -119,10 +118,38 @@ extension Duration : Proto3Codable {
 
 #if !WIRE_REMOVE_CODABLE
 extension Duration : Codable {
-    public enum CodingKeys : String, CodingKey {
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: Duration.CodingKeys.self)
```
lickel, line 122: @efirestone this is incorrect. Since this is generated code, I'm not sure how you want to proceed.

Timestamp | string | "1972-01-01T10:00:20.021Z" | Uses RFC 3339, where generated output will always be Z-normalized and uses 0, 3, 6 or 9 fractional digits. Offsets other than "Z" are also accepted.
-- | -- | -- | --
Duration | string | "1.000340012s", "1s" | Generated output always contains 0, 3, 6, or 9 fractional digits, depending on required precision, followed by the suffix "s". Accepted are any fractional digits (also none) as long as they fit into nano-seconds precision and the suffix "s" is required.

wire-library/wire-runtime-swift/src/main/swift/wellknowntypes/Timestamp.swift
```diff
@@ -131,10 +130,38 @@ extension Timestamp : Proto3Codable {
 
 #if !WIRE_REMOVE_CODABLE
 extension Timestamp : Codable {
-    public enum CodingKeys : String, CodingKey {
+    public init(from decoder: Decoder) throws {
```
lickel, line 133: @efirestone this is also incorrect:

proto3 | JSON | JSON example | Notes
-- | -- | -- | --
Timestamp | string | "1972-01-01T10:00:20.021Z" | Uses RFC 3339, where generated output will always be Z-normalized and uses 0, 3, 6 or 9 fractional digits. Offsets other than "Z" are also accepted.
Duration | string | "1.000340012s", "1s" | Generated output always contains 0, 3, 6, or 9 fractional digits, depending on required precision, followed by the suffix "s". Accepted are any fractional digits (also none) as long as they fit into nano-seconds precision and the suffix "s" is required.

wire-library/wire-runtime-swift/src/main/swift/CodingUserInfoKey+Wire.swift
```diff
@@ -60,6 +50,25 @@ extension JSONEncoder {
         /// Encodes the raw Encodable value, like `"myValue": 1`
         case raw
     }
+
+    /// The encoding strategy to use for key names
```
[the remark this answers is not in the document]
lickel: FWIW it's not just JSON. This would be used in _any_ Codable implementation

wire-library/wire-tests-swift/src/main/swift/RedactedOneOf.swift
```diff
@@ -103,11 +103,9 @@ extension RedactedOneOf : Proto2Codable {
 extension RedactedOneOf : Codable {
     public init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: RedactedOneOf.CodingKeys.self)
-        if container.contains(.b) {
```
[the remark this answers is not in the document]
lickel: `decodeIfPresent()` is functionally equivalent to `if contains(key) { decode() }` per the documentation

wire-library/wire-runtime-swift/src/main/swift/wellknowntypes/Duration.swift
lickel: Resolved in https://github.com/square/wire/pull/2380

wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -201,7 +213,7 @@ class SwiftGenerator private constructor(
                   storageParams += CodeBlock.of("%1N: %1N", oneOf.name)
                 }
                 addStatement(
-                  "_storage = %T(value: %T(%L))", heap, storageType,
```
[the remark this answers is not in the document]
lickel: > It's possible this is slightly slower due to the willSet overhead and additional method calls

That's incorrect. 
The compiler will convert `self.storage = Storage()` into `_storage = Heap(wrapperValue: xyz)` unless `Heap` has a plain `init()` method.
Further willSet/didSet is _never_ called in `init()`

wire-library/wire-tests-swift/src/main/swift/AllTypes.swift
```diff
@@ -1489,125 +1479,66 @@ fileprivate struct _AllTypes {
     public var req_bytes: Data
     public var req_nested_enum: AllTypes.NestedEnum
     public var req_nested_message: AllTypes.NestedMessage
-    @DefaultEmpty
```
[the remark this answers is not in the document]
lickel: I am using it for a bunch of tests with non-codegenned types, so I'd prefer to keep it to be honest.

wire-library/wire-runtime-swift/src/main/swift/CodingUserInfoKey+Wire.swift
```diff
@@ -52,61 +44,98 @@ extension JSONEncoder {
...
+    }
+
+    /// The encoding strategy to use for optional values and collections
+    /// - Note: Defaults to .skip
+    /// - SeeAlso: [Proto3 JSON Mapping](https://developers.google.com/protocol-buffers/docs/proto3#json)
+    public enum DefaultValuesEncodingStrategy {
+        // Skip "default" values
+        case skip
+        // Include "default" values
+        case include
```
lickel, line 74: > A proto3 JSON implementation may provide the following options:
> 
> Emit fields with default values: Fields with default values are omitted by default in proto3 JSON output. An implementation may provide an option to override this behavior and output fields with their default values.

That's why I used `emit` before and why I have the name that I did.
Open to suggestions @efirestone

wire-library/wire-tests-swift/src/main/swift/Person.swift
```diff
@@ -210,13 +238,50 @@ extension Person : Proto2Codable {
...
-        case id
-        case email
-        case phone
-        case aliases
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: Person.CodingKeys.self)
+
+        try container.encode(self.name, forKey: "name")
+        try container.encode(self.id, forKey: "id")
+        if encoder.protoDefaultValuesEncodingStrategy == .emit || self.email != nil {
```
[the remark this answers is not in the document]
lickel: it's actually kinda weird...

> When generating JSON-encoded output from a protocol buffer, if a protobuf field has the default value and if the field doesn't support field presence, it will be omitted from the output by default. An implementation may provide options to include fields with default values in the output.
> 
> A proto3 field that is defined with the optional keyword supports field presence. Fields that have a value set and that support field presence always include the field value in the JSON-encoded output, even if it is the default value.

I think checking `nil` is actually correct? Weirdly that means the `required` fields would should be doing additional checking?

But yeah, I can ask where to look in the channel

wire-library/wire-runtime-swift/src/main/swift/wellknowntypes/Duration.swift
```diff
@@ -119,10 +118,38 @@ extension Duration : Proto3Codable {
...
+        self.seconds = try container.decode(StringEncoded<Int64>.self, forKey: "seconds").wrappedValue
```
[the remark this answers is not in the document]
lickel: I'm honestly not sure... I'd have to play around with it. But yeah, I can take a look in a separate PR

wire-library/wire-runtime-swift/src/main/swift/CodingUserInfoKey+Wire.swift
```diff
@@ -52,61 +44,98 @@ extension JSONEncoder {
     }
 
     /// The encoding strategy to use for StringEncoded types that are themselves Encodable
-    /// Defaults to .string
+    /// - Note: Defaults to .string
     /// - Note: ProtoMap Dictionary keys are always encoded as strings
+    /// - SeeAlso: [Proto3 JSON Mapping](https://developers.google.com/protocol-buffers/docs/proto3#json)
     public enum StringEncodedEncodingStrategy {
```
efirestone, line 50: For a follow up: I'm not totally sure we need this. At least for numbers, the spec is pretty clear that the 64-bit numbers are encoded as strings, and the 32-bits are encoded as numbers. And also that both are allowable when decoding.

wire-library/wire-runtime-swift/src/main/swift/CodingUserInfoKey+Wire.swift
efirestone: Google's JVM library calls it "output" (`alwaysOutputDefaultValueFields`), which I find confusing. Some of the other Wire code uses emitter (`EmitterRules`), so I'm down with that. `include` probably makes for the nicest variable name (`includeDefaults`), so it seems good/fine too. I defer to you.

wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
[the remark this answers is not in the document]
efirestone: Nice. Improvement all around then

wire-library/wire-tests-swift/src/main/swift/AllTypes.swift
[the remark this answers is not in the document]
efirestone: Can we move it into the tests so that it's not compiled into the main library?

efirestone (approved):

wire-library/wire-runtime-swift/src/main/swift/CodingUserInfoKey+Wire.swift
lickel: I'll look into this in my Strip-String-Encodable branch tomorrow

wire-library/wire-runtime-swift/src/main/swift/CodingUserInfoKey+Wire.swift
lickel: OK I'll leave it as is

wire-library/wire-runtime-swift/src/main/swift/CodingUserInfoKey+Wire.swift
lickel: Implemented on the Data branch: https://github.com/square/wire/pull/2379/commits/2d20ea1ee39382e81457c4baa6bd666212df84e4

PR #2379, opened 2023-01-24
Swift: Correct Base64 suport

lickel:
Protos are supposed to accept base64 _and_ base64url encodings.
Apple doesn't have any native base64url support, but it's trivial to convert between the two.


wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/StringEncoded+Data.swift
```diff
@@ -0,0 +1,85 @@
...
+        self.init(base64urlEncoded: base64urlString, options: options)
+    }
+
+    init?(
+        base64urlEncoded base64urlString: String,
+        options: Data.Base64DecodingOptions = []
+    ) {
+        let count = base64urlString.count
+        let remainder = 8 - (count % 8)
+        let desiredLength = count + remainder
```
lickel, line 76: - [ ] verify padding length

lickel:
- [ ] support "empty" values in encoding and decoding

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/StringEncoded+Data.swift
```diff
@@ -0,0 +1,86 @@
...
+    func base64urlEncodedString(options: Data.Base64EncodingOptions = []) -> String {
+        let base64 = base64EncodedString(options: options)
+        return base64
+            .replacingOccurrences(of: "/", with: "_")
+            .replacingOccurrences(of: "+", with: "-")
+            .replacingOccurrences(of: "=", with: "")
+    }
+
+    init?(
+        base64urlEncoded base64urlData: Data,
```
efirestone, line 61: nit: I think this should technically be named `base64URLEncoded base64URLData:`

lickel: Yeah I went back and forth on this a few times.
I went with this originally because of the [RFC](https://www.rfc-editor.org/rfc/rfc4648#section-5)

>  This encoding may be referred to as "base64url".  This encoding
   should not be regarded as the same as the "base64" encoding and
   should not be referred to as only "base64".  Unless clarified
   otherwise, "base64" refers to the base 64 in the previous section.

but semantically I agree, it feels like it should be `URL`

lickel: (renamed)

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/StringEncoded+Data.swift
```diff
@@ -0,0 +1,86 @@
...
+            throw ProtoDecoder.Error.unparsableString(
+                type: Data.self,
+                value: encodedValue
+            )
+        }
+        self = value
+    }
+}
+
+public extension Data {
```
efirestone, line 38: nit: make the methods public rather than the extension

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/StringEncoded+Data.swift
```diff
@@ -0,0 +1,86 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+extension Data : StringCodable {
+    public func stringEncodedValue(in encoder: Encoder) throws -> String {
```
efirestone, line 20: The encoder and decoder in these methods isn't used anymore.

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/StringEncoded+Data.swift
```diff
@@ -0,0 +1,86 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+extension Data : StringCodable {
```
efirestone, line 19: The methods in this extension could be `internal`. Maybe that'll change if we get rid of the `StringEncoded` property wrapper, but feels like we should start with them being internal and expose them later if needed.

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/StringEncoded.swift
```diff
@@ -46,11 +46,14 @@ extension StringEncoded : Decodable {
         self.init(wrappedValue: value)
     }
 
-    private static func create(optionalEncodedValue: String?) throws -> Value {
+    private static func create(
```
efirestone, line 49: These methods can lose the encoder and decoder arguments too now.

wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -585,6 +584,36 @@ class SwiftGenerator private constructor(
       PROTO_3 -> proto3Codable
     }
 
+  private fun heapCodableExtension(
```
efirestone, line 587: Good catch on this. Did we not have a test for heap type encode/decode?

efirestone (approved):

wire-library/wire-runtime-swift/src/main/swift/propertyWrappers/StringEncoded+Data.swift
lickel: In either world, I need `StringCodable` to exist, and thus these methods need to be `public`.

wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
lickel: Nope... there's only a _single_ heap type right now and it would not be feasible to write a test for it.

lickel: New test: https://github.com/square/wire/pull/2379/commits/110661eb7d992757ca00a91306f500cd0e6db18b

PR #2380, opened 2023-01-24
Swift: Duration and Timestamp need custom Codable implementations

lickel:
Implement two well known types Codable implementation manually

I also moved a _bunch_ of files around for readability (with no code changes).

proto3 | JSON | JSON example | Notes
-- | -- | -- | --
Timestamp | string | "1972-01-01T10:00:20.021Z" | Uses RFC 3339, where generated output will always be Z-normalized and uses 0, 3, 6 or 9 fractional digits. Offsets other than "Z" are also accepted.
Duration | string | "1.000340012s", "1s" | Generated output always contains 0, 3, 6, or 9 fractional digits, depending on required precision, followed by the suffix "s". Accepted are any fractional digits (also none) as long as they fit into nano-seconds precision and the suffix "s" is required.


wire-library/wire-runtime-swift/src/main/swift/wellknowntypes/Duration.swift
```diff
@@ -115,42 +115,3 @@ extension Duration : Proto3Codable {
         try writer.writeUnknownFields(unknownFields)
     }
 }
-
-#if !WIRE_REMOVE_CODABLE
```
lickel, line 119: I don't actually know how to codegen something that uses these

wire-library/wire-runtime-swift/src/main/swift/wellknowntypes+Custom/Duration+Codable.swift
```diff
@@ -0,0 +1,56 @@
...
+import Foundation
+
+#if !WIRE_REMOVE_CODABLE
+
+extension Wire.Duration : Codable {
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.singleValueContainer()
+
+        let encoded: String = {
+            if nanos == 0 {
```
lickel, line 26: This implementation was cribbed from [the Kotlin implementation](https://github.com/square/wire/blob/master/wire-library/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/DurationJsonFormatter.kt#L48)

wire-library/wire-runtime-swift/src/main/swift/wellknowntypes+Custom/Duration+Codable.swift
```diff
@@ -0,0 +1,56 @@
...
+        try container.encode(encoded)
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.singleValueContainer()
+        var string = try container.decode(String.self)
+        guard let last = string.popLast(), last == "s" else {
+            throw DecodingError.dataCorruptedError(in: container, debugDescription: "Invalid duration format \(string)")
+        }
+        guard let interval = TimeInterval(string) else {
```
efirestone, line 46: This is going to be lossy, and decode->encode round trips aren't going to be identical. `TimeInterval` is a `Double` (64-bits total), but `Duration` is (64-bits + 32-bits for seconds + nanos). Specifically, the significand precision of a `Double` is 53-bits, so anything over 2^53 would be lossy, and the exponent is 11-bits, so anything some of those 32-bits of nanos would be lossy as well.

I think you can split on "." and then use the `String` -> Int64 and Int32 conversions.

wire-library/wire-runtime-swift/src/main/swift/wellknowntypes+Custom/Duration+Codable.swift
```diff
@@ -0,0 +1,103 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+#if swift(>=5.7)
+
+@available(macOS 13, iOS 16, watchOS 9, tvOS 16, *)
+private extension Swift.Duration {
```
efirestone, line 22: nit: put the `private` on the field instead of the extension. It might have to be `fileprivate`, but I noticed that Swift has gotten more lax about accessing private things if they're on the same type, in the same file, even if they're in different extensions.

wire-library/wire-runtime-swift/src/main/swift/wellknowntypes+Custom/Duration+Codable.swift
```diff
@@ -0,0 +1,103 @@
...
+    var nanos: Int32 {
+        let nanos = components.attoseconds * 1_000_000_000
+        return Int32(truncatingIfNeeded: nanos)
+    }
+}
+
+@available(macOS 13, iOS 16, watchOS 9, tvOS 16, *)
+extension Wire.Duration {
+    public var attos: Int64 {
+        Int64(nanos) / 1_000_000_000
```
efirestone, line 32: Shouldn't this be the other way (multiplied)? 1 nano is 1,000,000,000 attos

wire-library/wire-runtime-swift/src/main/swift/wellknowntypes+Custom/Timestamp+Codable.swift
```diff
@@ -0,0 +1,76 @@
...
+        let seconds = Int64(self.rounded())
+        let remainder = self.remainder(dividingBy: 1)
+        let nanos = Int32(remainder * pow(10, 9))
+
+        return (seconds, nanos)
+    }
+
+    init(seconds: Int64, nanos: Int32) {
+        self.init(seconds)
+        self += TimeInterval(nanos) / pow(10, 9)
```
efirestone, line 30: nit: can we use a constant here for `pow(10, 9)`. For speeeeeed. L23 as well. Could be a private static constant like `nanosecondsPerSecond` too.

wire-library/wire-runtime-swift/src/test/swift/CodableTests.swift
```diff
@@ -309,6 +309,32 @@ extension CodableTests {
...
 
+// MARK: - Duration and Timestamp
+
+extension CodableTests {
+    struct DurationAndTimestamp: Equatable, Codable {
+        var duration: Wire.Duration
+        var timestamp: Wire.Timestamp
+    }
+
+    func testDurationRoundtrip() throws {
```
efirestone, line 320: Can we add a test for a duration larger than 2^53 seconds, and one for `0.1s` and make sure they round trip?

We could add similar ones for Timestamp and comment them out too, just to illustrate that that doesn't work right now (with a comment)

efirestone (approved):

wire-library/wire-runtime-swift/src/test/swift/CodableTests.swift
lickel: https://github.com/square/wire/pull/2380/commits/e12a65fbe024131a4962b84e95575bad0130c20c

wire-library/wire-runtime-swift/src/test/swift/CodableTests.swift
```diff
@@ -333,6 +333,53 @@ extension CodableTests {
...
+          "duration":"18014398509481984s",
+          "timestamp":"4001-01-01T00:00:00Z"
+        }
+        """
+        let proto = DurationAndTimestamp(
+            duration: Duration(seconds: 1 << 54, nanos: 0),
+            timestamp: Timestamp(date: .distantFuture)
+        )
+
+        // In theory, we should use a much further distant future; however, there woudl be rounding errors at play
```
efirestone, line 385: typo: `woudl`

Also, maybe phrase it as something like `Swift's Date has less precision than Timestamp, so ideally we'd support something further in the future than Date.distantFuture, but our current implementation makes use of Date, and so that's as far as we currently support without encountering rounding errors."

wire-library/wire-runtime-swift/src/test/swift/CodableTests.swift
```diff
@@ -333,6 +333,53 @@ extension CodableTests {
...
+          "duration":"1.000000010s",
+          "timestamp":"0001-01-01T00:00:00Z"
+        }
+        """
+        let proto = DurationAndTimestamp(
+            duration: Duration(seconds: 1, nanos: 10),
+            timestamp: Timestamp(date: .distantPast)
+        )
+
+        // In theory, we should use a much further distant future; however, there woudl be rounding errors at play
```
efirestone, line 402: typo: This is distant past, not distant future, plus see comment above.

wire-library/wire-runtime-swift/src/test/swift/CodableTests.swift
```diff
@@ -333,6 +333,53 @@ extension CodableTests {
...
+        try assertDecode(json: json, expected: proto)
+        try assertEncode(proto: proto, expected: json)
+    }
+
+    func testSmallDurationRoundtrip() throws {
+        let json = """
+        {
+          "duration":"1.000000010s",
```
efirestone, line 393: This might be an additional test, but I would test `0.1s` specifically, as it can't be represented by a `Float` or `Double`, so it would expose any floating point rounding that we're unintentionally doing.

PR #2382, opened 2023-01-26
Swift codable refactor two

lickel:
This embeds the idea of supporting multiple keys for Decoding into Wire.

It supports var args to try to fetch keys N times.
I explicitly require a `firstKey` and then can take N keys after that.


wire-library/wire-runtime-swift/src/main/swift/Codable/Codable+Wire.swift
```diff
@@ -0,0 +1,75 @@
...
+
+        return try seq.lazy.compactMap { key in
+            try decodeIfPresent(type, forKey: key)
+        }.first
+    }
+
+    public func decodeFirst<T>(
+        _ type: T.Type,
+        forKeys firstKey: Key,
+        _ additionalKeys: Key...
```
efirestone, line 56: Sorry to be annoying, but since this is in one of the critical paths it's going to be performance-sensitive. We only ever use this with two keys, so can we remove the array manipulation and hardcode for that case? Basically, this function:
```
public func decodeFirst<T>(_ type: T.Type, forKeys firstKey: String, _ secondKey: String) -> throws T ...
```
If we ever have a third key (like due to a custom JSON key annotation on the field), we can create a three-key overload. There shouldn't ever need to be more than that though.

wire-library/wire-runtime-swift/src/main/swift/Codable/Codable+Wire.swift
```diff
@@ -0,0 +1,75 @@
...
+        let value = try seq.lazy.compactMap { key in
+            try decodeIfPresent(type, forKey: key)
+        }.first
+
+        guard let value else {
+            throw DecodingError.keyNotFound(
+                firstKey,
+                DecodingError.Context(
+                    codingPath: codingPath,
+                    debugDescription: "decodeFirst() could find a valid key"
```
efirestone, line 69: typo: `could *not*`

efirestone (approved):

PR #2383, opened 2023-01-26
Swift codable refactor two: StringEncoded flavors

lickel:
This shows what would be required to remove (or make private) all of the property wrappers.


wire-library/wire-runtime-swift/src/main/swift/Codable/Codable+Wire.swift
```diff
@@ -72,4 +49,114 @@ extension KeyedDecodingContainer {
...
+
+extension KeyedDecodingContainer {
+    public func decodeStringEncodedValues<T : StringCodable & Codable>(
+        _ type: T.Type,
+        forKey key: Key
+    ) throws -> [T] {
+        try decode(StringEncodedValues<Array<T>>.self, forKey: key).wrappedValue
+    }
+
+    public func decodeStringEncodedValuesIfPresent<T : StringCodable & Codable>(
```
lickel, line 117: _in theory_ we don't need both IfPresent and not IfPresent versions of this method.
That said, if we ever wanted to model a `required` repeated value, we'd need to be able to differentiate.

efirestone: Yeah, no such thing. `optional`, `required`, and `repeated` are all mutually exclusive values for the optionality modifier. There's no such concept as "a repeated field that contains at least one value"; put another way, an empty `repeated` field is the same as an unset `optional`.

In the encoded bytes, `repeated` fields aren't even represented as a collection. They're represented as the same single field encoded over and over, and if they never show up then the client is supposed to synthesize an empty array. (There's an optimized repeated packing format, but it's optional.)

efirestone (approved):
I like it. Can `StringEncoded` and `StringEncodedValues` be `internal` now?

efirestone (approved):

PR #2386, opened 2023-01-27
Swift: Don't encode default values in JSON

lickel:
This creates a new protocol and makes a bunch of things conform to it.

I intentionally did not make `Array`, `Dictionary` or `Optional` conform to it.
The reason being that all three of these things should ideally have _conditional_ conformance to the protocol and I don't want to break that in the future.

The big problem is we don't have a good way for `ProtoMessage` to conform answer `isDefaultProtoValue` right now.

Of note, this appears to be kind of an unsolved problem:
>When a message is parsed, if the encoded message does not contain a particular singular element, the corresponding field in the parsed object is set to the default value for that field. These defaults are type-specific:

*snip*

> For message fields, the field is not set. Its exact value is language-dependent. See the generated code guide for details.


wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -103,13 +103,35 @@ class SwiftGenerator private constructor(
...
 
+  private val Field.isCollection: Boolean
+    get() = when (typeName.makeNonOptional()) {
+      DATA, STRING -> true
+      else -> isMap || isRepeated
+    }
+
+  private val Field.codableDefaultValue: String?
+    get() = default?.let {
+      return it
```
efirestone, line 127: I'm not sure we'll ever hit this since we shouldn't be calling this for optionals, and optionals are the only fields that can have defaults. Doesn't hurt, but wanted to call it out.

wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -735,7 +757,7 @@ class SwiftGenerator private constructor(
                 if (type.fieldsAndOneOfFields.any { it.codableName != null }) {
                   addStatement("let preferCamelCase = encoder.protoKeyNameEncodingStrategy == .camelCase")
                 }
-                if (type.fields.any { it.typeName.optional || it.isRepeated || it.isMap }) {
+                if (type.fields.any { it.isOptional || it.isCollection || it.codableDefaultValue != null }) {
```
efirestone, line 760: I think you need `|| it.isEnum` in here as well.

wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -768,15 +790,29 @@ class SwiftGenerator private constructor(
...
+                  } else if (field.isEnum) {
+                    beginControlFlow("if", "includeDefaults || self.%N.rawValue != 0", field.name)
+                    addEncode()
+                    endControlFlow("if")
+                  } else if (defaultValue != null) {
+                    beginControlFlow("if", "includeDefaults || self.%N != $defaultValue", field.name)
+                    addEncode()
+                    endControlFlow("if")
                   } else {
+                    // Message is fundamentally broken right now when it comes to evaluating "isDefault"
```
efirestone, line 815: This comment should probably explain how it's broken and what needs to be added.

wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -768,15 +790,29 @@ class SwiftGenerator private constructor(
...
                   }
 
-                  if (field.typeName.optional) {
+                  val defaultValue = field.codableDefaultValue
+
+                  if (field.isOptional) {
+                    // A proto3 field that is defined with the optional keyword supports field presence.
+                    // Fields that have a value set and that support field presence always include the field value
+                    // in the JSON-encoded output, even if it is the default value.
                     beginControlFlow("if", "includeDefaults || self.%N != nil", field.name)
```
efirestone, line 799: We should verify the `protoc` behavior for this. The comment above from the docs doesn't actually clarify what the behavior is for *unset* optional fields. Do we encode `null`? Do we omit it? Do we do the former if `includeDefaults == true` and the latter if not (which is what's currently happening)?

The last option (the current option) seems reasonable for now, but wanted to call it out.

efirestone (approved):

wire-library/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
lickel: Yup, I'll take a look.
JSON definitely supports an explicit `null` for any value so it's definitely _valid_.

PR #2387, opened 2023-01-27
Swift: Add references to appropriate READMEs

lickel:
Some sections are cribbed heavily from https://code.cash.app/wire-support-for-swift-part-1


wire-library/wire-runtime-swift/build.gradle.kts
```diff
@@ -69,6 +69,24 @@ val generateSwiftProtos by tasks.creating(JavaExec::class) {
   }
 }
 
+val generateSampleProtos by tasks.creating(JavaExec::class) {
```
lickel, line 72: There is _almost certainly_ a better way to do this, but I don't know what I'm doing in Gradle

wire-library/wire-runtime-swift/src/test/swift/sample/Dinosaur.swift
```diff
@@ -0,0 +1,121 @@
...
+#endif
+
+extension Dinosaur : ProtoMessage {
+    public static func protoMessageTypeURL() -> String {
+        return "type.googleapis.com/squareup.dinosaurs.Dinosaur"
+    }
+}
+
+extension Dinosaur : Proto2Codable {
+    public init(from reader: ProtoReader) throws {
```
lickel, line 59: For the purposes of this sample code, should we `/* snip */` out the Proto/Foundation codable implementations?

Package.swift
```diff
@@ -23,6 +23,7 @@ let package = Package(
         ),
         // The tests depend on valid protos via gradle
         // ./gradlew -p wire-library :wire-runtime-swift:generateTestProtos
+        // ./gradlew -p wire-library :wire-runtime-swift:generateSampleProtos
```
oldergod, line 26: The comment makes it sound like the tests would not pass without having run `generateSampleProtos` but is that really the case? If not, no need for the comment I think ?

wire-library/wire-runtime-swift/build.gradle.kts
oldergod: Why not add those files in the existing `generateTestProtos` directly?

oldergod (approved):

oldergod:
@lickel I've made a big change in Wire and rebased your PR. the comments are gone but still valid, let's sync next week.

wire-runtime-swift/build.gradle.kts
```diff
@@ -69,6 +69,24 @@ val generateSwiftProtos by tasks.creating(JavaExec::class) {
   }
 }
 
+val generateSampleProtos by tasks.creating(JavaExec::class) {
```
lickel, line 72: There's almost certainly a better way to do this but I don't know gradle

Package.swift
```diff
@@ -22,21 +22,22 @@ let package = Package(
...
             name: "WireTests",
             dependencies: ["Wire"],
-            path: "wire-library/wire-tests-swift/src/main/swift"
+            path: "wire-tests-swift/src/main/swift"
         ),
         .testTarget(
             name: "WireCompilerTests",
             dependencies: ["WireTests"],
-            path: "wire-library/wire-tests-swift/src/test/swift"
+            path: "wire-tests-swift/src/test/swift"
```
lickel, line 40: This rebase change implies we should have an SPM GH check

lickel:
Might be worth linking to the 3 Dinosaur.source files instead of large code dumps?

Package.swift
oldergod: What's that?

lickel: Eg we have a github action to verify Cocoapods. We probably should have something for Seift Package Manager too

wire-runtime-swift/build.gradle.kts
lickel: @oldergod 

> Why not add those files in the existing generateTestProtos directly?

I think there's value in a different subdirectory, tho it for sure should just be happening as a dependency or something. 

I am very unfamiliar with Gradle; my only experience with it was XML _years_ ago.

Package.swift
```diff
@@ -22,21 +22,22 @@ let package = Package(
             path: "wire-runtime-swift/src/main/swift"
         ),
         // The tests depend on valid protos via gradle
-        // ./gradlew -p wire-library :wire-runtime-swift:generateTestProtos
+        // ./gradlew :wire-runtime-swift:generateTestProtos
```
lickel, line 25: @oldergod 

> The comment makes it sound like the tests would not pass without having run generateSampleProtos but is that really the case? If not, no need for the comment I think ?

True, the samples _currently_ are not a dependency, and if they were we'd need the same codegen in the Cocoapods file. 

That said, I still would prefer we codegen the dinosaur sample in the repo somewhere.

oldergod: Got it

wire-library/wire-runtime-swift/src/test/swift/sample/Dinosaur.swift
oldergod: Yes, let's do that and then merge 🎸

wire-runtime-swift/build.gradle.kts
lickel: Added as a dependency

wire-library/wire-runtime-swift/build.gradle.kts
lickel: Added as a dependency

wire-library/wire-runtime-swift/src/test/swift/sample/Dinosaur.swift
lickel: I went a different route and used some \<details\> tags for all three languages.

PR #2388, opened 2023-01-29
ProtoAdapter and FieldBinding should use rawType's classloader.

autonomousapps:
It is arguably a bug that `ProtoAdapter` implicitly uses the classLoader from itself when calling `Class.forName(String)`. Instead, we should use `Class.forName(String, boolean, ClassLoader)`, and the `ClassLoader` argument should be supplied by `rawType`.

## Original description
This draft PR is intended first for discussion purposes, and to get feedback on what it would take to land something like this on trunk. We need this in our project because the `wire-runtime` library is loaded in a different classloader than the main app classes, leading to serialization errors when wire tries to instantiate a class instance via `Class.forName(String)`. Arguably, best practice in cases like this indicates that an overload ought to be provided such that a library can use `Class.forName(String, boolean, ClassLoader)`, because in general libraries cannot guarantee they'll be loaded in the same classloaders as application classes.

### DIFF
--- a/wire-gson-support/src/main/java/com/squareup/wire/WireTypeAdapterFactory.kt
+++ b/wire-gson-support/src/main/java/com/squareup/wire/WireTypeAdapterFactory.kt
@@ -21,7 +21,6 @@ import com.google.gson.TypeAdapterFactory
 import com.google.gson.reflect.TypeToken
 import com.squareup.wire.internal.EnumJsonFormatter
 import com.squareup.wire.internal.createRuntimeMessageAdapter
-import java.util.Optional
 
 /**
  * A [TypeAdapterFactory] that allows Wire messages to be serialized and deserialized
@@ -49,7 +48,6 @@ import java.util.Optional
 class WireTypeAdapterFactory @JvmOverloads constructor(
   private val typeUrlToAdapter: Map<String, ProtoAdapter<*>> = mapOf(),
   private val writeIdentityValues: Boolean = false,
-  private val loader: Optional<ClassLoader> = Optional.empty()
 ) : TypeAdapterFactory {
   /**
    * Returns a new WireJsonAdapterFactory that can encode the messages for [adapters] if they're
@@ -63,7 +61,7 @@ class WireTypeAdapterFactory @JvmOverloads constructor(
       )
       newMap[key] = adapter
     }
-    return WireTypeAdapterFactory(newMap, writeIdentityValues, loader)
+    return WireTypeAdapterFactory(newMap, writeIdentityValues)
   }
 
   /**
@@ -83,7 +81,7 @@ class WireTypeAdapterFactory @JvmOverloads constructor(
         val messageAdapter = createRuntimeMessageAdapter<Nothing, Nothing>(
           rawType as Class<Nothing>,
           writeIdentityValues,
-          loader.orElseGet { rawType.classLoader },
+          rawType.classLoader,
         )
         val jsonAdapters = GsonJsonIntegration.jsonAdapters(messageAdapter, gson)
         MessageTypeAdapter(messageAdapter, jsonAdapters).nullSafe() as TypeAdapter<T>
--- a/wire-gson-support/src/main/java/com/squareup/wire/WireTypeAdapterFactory.kt
+++ b/wire-gson-support/src/main/java/com/squareup/wire/WireTypeAdapterFactory.kt
@@ -76,7 +76,6 @@ class WireTypeAdapterFactory @JvmOverloads constructor(
 
   override fun <T> create(gson: Gson, type: TypeToken<T>): TypeAdapter<T>? {
     val rawType = type.rawType
-    val loader = if (loader.isPresent) loader.get() else rawType.classLoader
 
     return when {
       rawType == AnyMessage::class.java -> AnyMessageTypeAdapter(gson, typeUrlToAdapter) as TypeAdapter<T>
@@ -84,7 +83,7 @@ class WireTypeAdapterFactory @JvmOverloads constructor(
         val messageAdapter = createRuntimeMessageAdapter<Nothing, Nothing>(
           rawType as Class<Nothing>,
           writeIdentityValues,
-          loader,
+          loader.orElseGet { rawType.classLoader },
         )
         val jsonAdapters = GsonJsonIntegration.jsonAdapters(messageAdapter, gson)
         MessageTypeAdapter(messageAdapter, jsonAdapters).nullSafe() as TypeAdapter<T>
--- a/wire-gson-support/src/main/java/com/squareup/wire/WireTypeAdapterFactory.kt
+++ b/wire-gson-support/src/main/java/com/squareup/wire/WireTypeAdapterFactory.kt
@@ -21,6 +21,7 @@ import com.google.gson.TypeAdapterFactory
 import com.google.gson.reflect.TypeToken
 import com.squareup.wire.internal.EnumJsonFormatter
 import com.squareup.wire.internal.createRuntimeMessageAdapter
+import java.util.Optional
 
 /**
  * A [TypeAdapterFactory] that allows Wire messages to be serialized and deserialized
@@ -48,6 +49,7 @@ import com.squareup.wire.internal.createRuntimeMessageAdapter
 class WireTypeAdapterFactory @JvmOverloads constructor(
   private val typeUrlToAdapter: Map<String, ProtoAdapter<*>> = mapOf(),
   private val writeIdentityValues: Boolean = false,
+  private val loader: Optional<ClassLoader> = Optional.empty()
 ) : TypeAdapterFactory {
   /**
    * Returns a new WireJsonAdapterFactory that can encode the messages for [adapters] if they're
@@ -61,7 +63,7 @@ class WireTypeAdapterFactory @JvmOverloads constructor(
       )
       newMap[key] = adapter
     }
-    return WireTypeAdapterFactory(newMap, writeIdentityValues)
+    return WireTypeAdapterFactory(newMap, writeIdentityValues, loader)
   }
 
   /**
@@ -74,7 +76,7 @@ class WireTypeAdapterFactory @JvmOverloads constructor(
 
   override fun <T> create(gson: Gson, type: TypeToken<T>): TypeAdapter<T>? {
     val rawType = type.rawType
-    val loader = rawType.classLoader
+    val loader = if (loader.isPresent) loader.get() else rawType.classLoader
 
     return when {
       rawType == AnyMessage::class.java -> AnyMessageTypeAdapter(gson, typeUrlToAdapter) as TypeAdapter<T>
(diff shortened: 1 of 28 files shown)


wire-runtime/src/jvmMain/kotlin/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -201,26 +201,29 @@ actual abstract class ProtoAdapter<E> actual constructor(
 
     // Obsolete; for Java classes generated before typeUrl and syntax were added.
     @JvmStatic fun <M : Message<M, B>, B : Message.Builder<M, B>> newMessageAdapter(
```
oldergod, line 203: As the comment pointed out, some methods are here because some old generated classes might use the more recent runtime. We need to keep the signatures as is.
We wanna keep all current `newMessageAdapter`s and add one with the new signature.

```kotlin
    @JvmStatic fun <M : Message<M, B>, B : Message.Builder<M, B>> newMessageAdapter(
      type: Class<M>,
      typeUrl: String,
      syntax: Syntax,
      loader: ClassLoader,
    )
```

wire-runtime/src/jvmMain/kotlin/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -201,26 +201,29 @@ actual abstract class ProtoAdapter<E> actual constructor(
...
-      typeUrl: String
+      typeUrl: String,
+      loader: ClassLoader,
     ): ProtoAdapter<M> {
-      return createRuntimeMessageAdapter(type, typeUrl, Syntax.PROTO_2)
+      return createRuntimeMessageAdapter(type, typeUrl, Syntax.PROTO_2, loader)
     }
 
     /** Creates a new proto adapter for `type`. */
     @JvmStatic fun <M : Message<M, B>, B : Message.Builder<M, B>> newMessageAdapter(
```
oldergod, line 220: We keep this method, update the comment saying it's obsolete but we keep it for classes generated before `loader` was added. and call the newly created method with the default loader (I think `ProtoAdapter::class.java.classLoader`?)

wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/reflection.kt
```diff
@@ -30,6 +30,7 @@ fun <M : Message<M, B>, B : Message.Builder<M, B>> createRuntimeMessageAdapter(
   messageType: Class<M>,
   typeUrl: String?,
   syntax: Syntax,
+  loader: ClassLoader,
```
oldergod, line 33: can we put a default param here? I'm not 100% we don't need one (even though it's within internal)

I think we need one

wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/reflection.kt
```diff
@@ -79,12 +87,14 @@ private fun <M : Message<M, B>, B : Message.Builder<M, B>> getKeys(
 fun <M : Message<M, B>, B : Message.Builder<M, B>> createRuntimeMessageAdapter(
   messageType: Class<M>,
   writeIdentityValues: Boolean,
+  loader: ClassLoader,
```
oldergod, line 90: I think we need a default param here as well. For instance, `WireTypeAdapterFactory` is using it

wire-gson-support/src/main/java/com/squareup/wire/WireTypeAdapterFactory.kt
```diff
@@ -74,11 +74,16 @@ class WireTypeAdapterFactory @JvmOverloads constructor(
 
   override fun <T> create(gson: Gson, type: TypeToken<T>): TypeAdapter<T>? {
     val rawType = type.rawType
+    val loader = rawType.classLoader
```
oldergod, line 77: let's inline

wire-moshi-adapter/src/main/java/com/squareup/wire/WireJsonAdapterFactory.kt
```diff
@@ -44,6 +44,7 @@ import java.lang.reflect.Type
 class WireJsonAdapterFactory @JvmOverloads constructor(
   private val typeUrlToAdapter: Map<String, ProtoAdapter<*>> = mapOf(),
   private val writeIdentityValues: Boolean = false,
+  private val loader: ClassLoader,
```
oldergod, line 47: can we have a default param here?

wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/reflection.kt
bendotli: @autonomousapps is it important to be able to pass a classloader that's different from `messageType`'s classloader? You do have some examples in the test changes where you pass the unit test class's classloader, rather than the message class's classloader. In those cases they are the same, but generally it seems more correct to pass the message class's classloader?

And if `loader` is always going to be the message class's classloader, maybe we can avoid adding the `ClassLoader` argument here and in a few other places? Or is the problem that, when a class is cast into `Class<Nothing>` (as done in wire-gson-support) we lose the ability to grab its classloader?

autonomousapps: Great questions @bendotli . The truth is that I haven't quite settled on the best approach yet, which is why I left this PR in draft. I first wanted to ensure that wire's maintainers were ok with the general idea before going deeper.

I think having a fully injectable classloader would be useful, _however_ it is complicated by a couple of factors:
1. Our path to `Class.forName(String)` actually goes through Retrofit and a gson adapter, so we'd have to modify at least two other libraries for that level of control (I think).
2. If we provided an overloaded constructor in a few places, the primary constructor would need to handle the current case (no classloader). We can't let `null` be the default value, since that actually has special meaning on the JVM: it means "use the boot classloader." So we'd need some sentinel value, or maybe `Optional<ClassLoader>` (which I think is still a nonstarter in Android), or grab the classloader from another of the values passed into that constructor. The latter is what I've gone with in this draft.

autonomousapps:
@oldergod @bendotli I've updated this PR to use default arguments everywhere (I think), and also to use `Optional<ClassLoader> = Optional.empty()` in two places to handle the fact that a `null` classloader has a special value on the JVM. Let me know what you think.

autonomousapps:
With a test failure like the following, is it literally saying that it doesn't like that I made the declaration take multiple lines?
```
 public final class Pizza extends Message<Pizza, Pizza.Builder> {
-  public static final ProtoAdapter<Pizza> ADAPTER = ProtoAdapter.newMessageAdapter(
-    Pizza.class,
-    "type.googleapis.com/squareup.proto3.Pizza",
-    Syntax.PROTO_3,
-    Pizza.class.getClassLoader()
-  );
+  public static final ProtoAdapter<Pizza> ADAPTER = ProtoAdapter.newMessageAdapter(Pizza.class, "type.googleapis.com/squareup.proto3.Pizza", Syntax.PROTO_3);
```

wire-gson-support/src/main/java/com/squareup/wire/WireTypeAdapterFactory.kt
```diff
@@ -48,6 +49,7 @@ import com.squareup.wire.internal.createRuntimeMessageAdapter
 class WireTypeAdapterFactory @JvmOverloads constructor(
   private val typeUrlToAdapter: Map<String, ProtoAdapter<*>> = mapOf(),
   private val writeIdentityValues: Boolean = false,
+  private val loader: Optional<ClassLoader> = Optional.empty()
```
oldergod, line 52: You said we cannot use `ClassLoader?`, what's the special meaning for it?

oldergod: That'd be really nice if we could not use `Optional` and pass `ClassLoader?` directly

wire-runtime/src/jvmMain/kotlin/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -218,11 +218,21 @@ actual abstract class ProtoAdapter<E> actual constructor(
...
```
oldergod, line 218:
```suggestion
    // Obsolete; for Java classes generated before `classLoader` was added.
    @JvmStatic fun <M : Message<M, B>, B : Message.Builder<M, B>> newMessageAdapter(
```

wire-runtime/src/jvmMain/kotlin/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -218,11 +218,21 @@ actual abstract class ProtoAdapter<E> actual constructor(
...
     ): ProtoAdapter<M> {
       return createRuntimeMessageAdapter(type, typeUrl, syntax)
     }
 
+    /** Creates a new proto adapter for `type`. */
+    @JvmStatic fun <M : Message<M, B>, B : Message.Builder<M, B>> newMessageAdapter(
+      type: Class<M>,
+      typeUrl: String,
+      syntax: Syntax,
+      loader: ClassLoader?
```
oldergod, line 231: Let's keep `classLoader` for the param name

wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/FieldBinding.kt
```diff
@@ -32,6 +32,7 @@ class FieldBinding<M : Message<M, B>, B : Message.Builder<M, B>> internal constr
   private val messageField: Field,
   builderType: Class<B>,
   override val writeIdentityValues: Boolean,
+  private val loader: ClassLoader? = messageType.classLoader,
```
oldergod, line 35: `classLoader` please

oldergod:
@autonomousapps for the build fail, it says "Test files changed. Did you run `./gradlew generateTests`?"

autonomousapps:
> for the build fail, it says "Test files changed. Did you run `./gradlew generateTests`?"

a-ha, I did not! Thanks for the tip.

wire-gson-support/src/main/java/com/squareup/wire/WireTypeAdapterFactory.kt
autonomousapps: A `null` classloader has special meaning on the JVM. It means "use the boot classloader" (which basically contains the java.lang package). If that's not a use-case you want to support, we can drop the `Optional`, forbid the usage of `null` (except in the two cases like this), and away we go! I was just trying to be maximally semantically correct.

autonomousapps:
Fixed the tests (`./gradlew test` succeeds), removed the injectable classLoader (we only need `rawType`'s classLoader), and also changed all parameter names from `loader` to `classLoader`. @oldergod @swankjesse

swankjesse (approved):

oldergod (approved):
Let's go

PR #2393, opened 2023-01-31
Add ability to supply custom schema handler factory class from command line

dnkoutso:
Sample invocation:

```
./wire-compiler/build/install/wire-compiler/bin/wire-compiler --custom_out=/Users/dnkoutso/Desktop --schema_handler_factory_class="com.squareup.wire.schema.MyCustomSchemaHandlerFactory" --proto_path=...
```

At least when compiling a temporary `MyCustomSchemaHandlerFactory` into the jar it all worked end to end.

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/WireCompiler.kt
+++ b/wire-compiler/src/main/java/com/squareup/wire/WireCompiler.kt
@@ -130,10 +130,13 @@ class WireCompiler internal constructor(
       targets += SwiftTarget(
         outDirectory = swiftOut
       )
-    } else if (customOut != null) {
+    } else if (customOut != null || schemaHandlerFactoryClass != null) {
+      if (customOut == null || schemaHandlerFactoryClass == null) {
+        throw IllegalArgumentException("Both custom_out and schema_handler_factory_class need to be set")
+      }
       targets += CustomTarget(
         outDirectory = customOut,
-        schemaHandlerFactory = newSchemaHandler(schemaHandlerFactoryClass!!)
+        schemaHandlerFactory = newSchemaHandler(schemaHandlerFactoryClass)
       )
     }
 
--- a/wire-compiler/src/main/java/com/squareup/wire/WireCompiler.kt
+++ b/wire-compiler/src/main/java/com/squareup/wire/WireCompiler.kt
@@ -15,6 +15,7 @@
  */
 package com.squareup.wire
 
+import com.squareup.wire.schema.CustomTarget
 import com.squareup.wire.schema.JavaTarget
 import com.squareup.wire.schema.KotlinTarget
 import com.squareup.wire.schema.Location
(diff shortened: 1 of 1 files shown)


wire-compiler/src/main/java/com/squareup/wire/WireCompiler.kt
```diff
@@ -188,6 +197,8 @@ class WireCompiler internal constructor(
     private const val JAVA_OUT_FLAG = "--java_out="
     private const val KOTLIN_OUT_FLAG = "--kotlin_out="
     private const val SWIFT_OUT_FLAG = "--swift_out="
+    private const val CUSTOM_OUT_FLAG = "--custom_out="
```
dnkoutso, line 200: let me know if you want me to have one parameter instead of two

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.kt
```diff
@@ -272,6 +285,14 @@ class WireCompiler internal constructor(
             swiftOut = arg.substring(SWIFT_OUT_FLAG.length)
           }
 
+          arg.startsWith(CUSTOM_OUT_FLAG) -> {
+            customOut = arg.substring(CUSTOM_OUT_FLAG.length)
+          }
+
+          arg.startsWith(SCHEMA_HANDLER_FACTORY_CLASS_FLAG) -> {
```
dnkoutso, line 292: I was hoping to check that if one is set the other must be set too

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.kt
dnkoutso: I guess we dont really need a `custom_out` param? Any cases you can think of that this can be useful?

oldergod: Not ideal but let's keep it

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.kt
```diff
@@ -126,6 +130,11 @@ class WireCompiler internal constructor(
       targets += SwiftTarget(
         outDirectory = swiftOut
       )
+    } else if (customOut != null) {
+      targets += CustomTarget(
+        outDirectory = customOut,
+        schemaHandlerFactory = newSchemaHandler(schemaHandlerFactoryClass!!)
+      )
```
oldergod, line 137: Maybe we do the check here like this:

```suggestion
    } else if (customOut != null || schemaHandlerFactoryClass != null) {
      if (customOut == null || schemaHandlerFactoryClass == null) {
        throw IllegalArgumentException("Both custom_out and schema_handler_factory_class need to be set")
      }
      targets += CustomTarget(
        outDirectory = customOut,
        schemaHandlerFactory = newSchemaHandler(schemaHandlerFactoryClass)
      )
```

oldergod (approved):

PR #2396, opened 2023-02-01
Update WireCompiler docs to include custom_out

dnkoutso:


oldergod (approved):
Anything we need to explain about proving the factory to the classpath?

dnkoutso:
> Anything we need to explain about proving the factory to the classpath?

I think we can add to https://github.com/square/wire/blob/master/docs/wire_compiler.md#custom-handlers perhaps? I just dont know the steps to describe it yet :)

PR #2402, opened 2023-02-07
Support deserialization of camelCased fields in proto2

oldergod:
Proto3 support was made by following the spec and setting `jsonName` to camel case https://github.com/square/wire/blob/master/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt#L146

For proto2, instead of falsely setting up the `jsonName`, we add the alternative for deserialization if it makes sense.

### DIFF
--- a/wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/RuntimeMessageAdapter.kt
+++ b/wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/RuntimeMessageAdapter.kt
@@ -45,17 +45,32 @@ class RuntimeMessageAdapter<M : Any, B : Any>(
    * name.
    *
    * For reserved keywords in Kotlin or Java, a field like `public` are written as so although the
-   * generated field name is `public_`. We wanna read in either form. As well, in proto3 fields
+   * generated field name is `public_`. We want to read in either form. As well, in proto3 fields
    * declared in snake_case like `customer_id` are written in camelCase like `customerId`, but can
    * be read in either form. We can use exclusive logic between the two cases because there's no
    * keyword defined in snake_case. The alternate name will be absent if the field name isn't a
    * keyword or if the snake and camel cases are the same, such as in single-word identifiers.
+   * Lastly, in order to deserialize camelCased field from proto2, we add a camelCase alternative
+   * if the `jsonName` doesn't already match it.
    */
-  val jsonAlternateNames: List<String?> = fieldBindingsArray.map {
-    when {
-      it.jsonName != it.declaredName -> it.declaredName
-      it.jsonName != it.name -> it.name
-      else -> null
+  val jsonAlternateNames: List<String?> = run {
+    val jsonNames = fieldBindingsArray.map { it.jsonName }
+    return@run fieldBindingsArray.map {
+      when {
+        it.jsonName != it.declaredName -> it.declaredName
+        it.jsonName != it.name -> it.name
+        else -> {
+          val camelCaseDeclaredName = camelCase(it.declaredName)
+          if (it.jsonName != camelCaseDeclaredName &&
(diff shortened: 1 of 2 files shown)


swankjesse (approved):
Add a Moshi and a Gson regression test for this?

oldergod:
@swankjesse Our interop tests already run Gson and Moshi round trips!

wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/RuntimeMessageAdapter.kt
```diff
@@ -45,17 +45,32 @@ class RuntimeMessageAdapter<M : Any, B : Any>(
    * name.
    *
    * For reserved keywords in Kotlin or Java, a field like `public` are written as so although the
-   * generated field name is `public_`. We wanna read in either form. As well, in proto3 fields
+   * generated field name is `public_`. We want to read in either form. As well, in proto3 fields
    * declared in snake_case like `customer_id` are written in camelCase like `customerId`, but can
    * be read in either form. We can use exclusive logic between the two cases because there's no
    * keyword defined in snake_case. The alternate name will be absent if the field name isn't a
    * keyword or if the snake and camel cases are the same, such as in single-word identifiers.
+   * Lastly, in order to deserialize camelCased field from proto2, we add a camelCase alternative
```
lickel, line 53: What's specific about proto2 here?

Since both versions are supposed to behave the same, is the proto2 comment effectively an implementation detail?

lickel: Oh sorry, I think I follow now. Would it make sense to reword things in so far as the json_name value of the annotation is only provided for explicitly defined json names?

In other words, default/implicit value will not be present in the annotation. 

Or am I still misreading?

wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/RuntimeMessageAdapter.kt
```diff
@@ -45,17 +45,32 @@ class RuntimeMessageAdapter<M : Any, B : Any>(
...
-      else -> null
+  val jsonAlternateNames: List<String?> = run {
+    val jsonNames = fieldBindingsArray.map { it.jsonName }
+    return@run fieldBindingsArray.map {
+      when {
+        it.jsonName != it.declaredName -> it.declaredName
+        it.jsonName != it.name -> it.name
+        else -> {
+          val camelCaseDeclaredName = camelCase(it.declaredName)
+          if (it.jsonName != camelCaseDeclaredName &&
```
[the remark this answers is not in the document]
lickel: (Tho perhaps that's true in practice?)

lickel:
If the client serializes a proto2 message, will it be camelCase or snake_case?

I know it shouldn't matter, just curious about the round trip

oldergod:
> If the client serializes a proto2 message, will it be camelCase or snake_case?

It will be serialized as defined in the `.proto` we don't manipulate.

wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/RuntimeMessageAdapter.kt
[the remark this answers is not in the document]
oldergod: jsonName is never null here

wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/RuntimeMessageAdapter.kt
oldergod: Yes. Because proto2 has no spec around JSON mapping.

PR #2403, opened 2023-02-09
Wire Runtime changes to enable protoc integration

buildbreaker:
Protoc integration requires overriding the file writing logic in the `SchemaHandler.Context` object. The change here is to abstract out the file writing capabilities in the `SchemaHandler.Context` object into an interface called `FileWriter` which will have the file system based writer which is currently used called `FileSystemWriter`.

The intention here is to break up the protoc integration into two PRs and have a more focused review of changes:
1. Changes needed to make the protoc integration work
2. The additive changes for the protoc integration module

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
+++ b/wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
@@ -260,8 +260,10 @@ data class WireRun(
             target.outDirectory.toPath() / moduleName
           }
         val context = SchemaHandler.Context(
-          fileSystem = fs,
-          outDirectory = outDirectory,
+          fileWriter = FileSystemWriter(
+            fs,
+            outDirectory,
+          ),
           logger = logger,
           errorCollector = errorCollector,
           emittingRules = targetToEmittingRules.getValue(target),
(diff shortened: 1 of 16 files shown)


wire-compiler/src/main/java/com/squareup/wire/schema/WireRun.kt
```diff
@@ -271,6 +273,8 @@ data class WireRun(
           module = module,
           profileLoader = schemaLoader,
         )
+        // Ensure the out directory is created.
+        fs.createDirectories(outDirectory)
```
buildbreaker, line 277: This is a bit of a strange change but it seems like there exists a test which ensures this behavior: https://github.com/square/wire/blob/cad71b5754551cf541386b73c88c51e97f4dea75/wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt#L1004-L1023

oldergod (commented):
If I got it right, `FileWriter` is created so that we can do that instead of writing to disk.
```kotlin
class ProtoFileWriter(
  private val response: Plugin.Response
) : FileWriter {
  override fun write(file: Path, content: String): Path {
    response.addFile(file.toString(), content)
    return file
  }
}
```

I'll try to think of a solution which doesn't break Wire's API.

PR #2404, opened 2023-02-10
Ensure linking descriptor.proto when provided

buildbreaker:
Changing the Linker to accept a `descriptor.proto` from the input and add the internal `descriptor.proto` when the input omits the proto.


oldergod (approved):

buildbreaker:
Thank you @oldergod !

PR #2405, opened 2023-02-10
Swift:  Fix broken Copy-on-Write behavior

lickel:
🐮


dnkoutso (approved):
Does the same test fail if the fix is not present? I usually do that to double check.

lickel:
> Does the same test fail if the fix is not present? I usually do that to double check.

Yeah, it would.
The problem is that `Heap` is a class and it wasn't getting re-inited at mutation time.

PR #2406, opened 2023-02-10
Add protoc and buf integration for Wire Java/Kotlin generation

buildbreaker:
Adds `protoc` integration via the `wire-protoc-integration` module.

The executables `protoc-gen-wire-java` and `protoc-gen-wire-kotlin` are called by `protoc`. The executables have a dependency on the jar output from the `wire-protoc-integration` module. The `protoc-java.jar` and `protoc-kotlin.jar` outputs generated code via the integration the protoc request and response API.

# Protoc integration and execution
To execute `protoc-gen-wire-kotlin`, you will need to install `protoc`. For more information on using the `protoc` tool, check out the [`protobuf.dev`](https://protobuf.dev/getting-started/javatutorial/#compiling-protocol-buffers) website.

To use the `protoc-gen-wire-kotlin` plugin, the following command would generate code from the `src/main/testproto/simple.proto` file:

```
protoc \
  --proto_path=./src/main/testproto \
  --plugin=$HOME/wire/wire-protoc-integration/protoc-gen-wire-kotlin \
  --wire-kotlin_out=./gen  \
  ./src/main/testproto/simple.proto
```

`--plugin` accepts the fully qualified path for the executable (`protoc-gen-wire-kotlin`).
`--wire-kotlin_out` specifies the output directory for the plugin.
(Note: if you're using `protoc-gen-wire-java` this will change to `--wire-java_out`)

# How the plugin integration works
Wire handles the compilation and linking of protobuf files along with outputting ergonomic and idiomatic code. Google's`protoc` handles compilation and linking of protobuf files along with the extensibility to provide custom generators. The module here primarily integrates `wire`'s Schema and code generation with the compilation and linking within `protoc`.

The `WireGenerator` is the primary class which mostly does data mapping from the `protoc` file descriptor to `wire`'s Schema. The input (via `System.in`) from `WireGenerator` is a `PluginProtos.CodeGeneratorRequest` which is modeled after the [`descriptor.proto`](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/descriptor.proto) file. The output (via `System.out`) from `WireGenerator` is the `Plugin.Response`.


wire-protoc-integration/src/main/java/com/squareup/wire/protocwire/CodeGenerator.kt
```diff
@@ -0,0 +1,19 @@
...
+import com.squareup.wire.protocwire.Plugin.DescriptorSource
+import com.google.protobuf.compiler.PluginProtos
+
+/**
+ * Interface to be implemented by a code generator.
+ */
+interface CodeGenerator {
+    /**
+     * Generates code for the given proto file, generating one or more files to
+     * the given response.
```
oldergod, line 12: Where is the "proto file" described here? What are the roles of the 3 parameters? Is the generation to happen only in memory, can the `CodeGenerator` write on disk as well?

wire-protoc-integration/src/main/java/com/squareup/wire/protocwire/JavaTargetConfig.kt
```diff
@@ -0,0 +1,36 @@
...
+ * Translates the protoc options into known options for the
+ * JavaTarget via the gradle plugin JavaOutput.
+ */
+internal class JavaTargetConfig {
+  companion object {
+    /**
+     * @param parameters the map of string to string from the protoc options.
+     * @return the translated JavaTarget.
+     */
+    internal fun parse(parameters: Map<String, String>): JavaTarget {
```
oldergod, line 23: protoc doesn't have a type wrapping this map? That'd be sweet if we could do something like
`JavaTarget.of(optionMap)`. I'm not sure how we can add this API without touching `JavaTarget` directly.

If we add an empty companion on `JavaTarget`, we might be able to do so. I don't know how to set it `JvmStatic` so the java call site would not be nice.
image

wire-protoc-integration/src/main/java/com/squareup/wire/protocwire/JavaTargetConfig.kt
```diff
@@ -0,0 +1,36 @@
...
+private const val BUILDERS_ONLY = "buildersOnly"
+
+/**
+ * Translates the protoc options into known options for the
+ * JavaTarget via the gradle plugin JavaOutput.
+ */
+internal class JavaTargetConfig {
+  companion object {
+    /**
+     * @param parameters the map of string to string from the protoc options.
```
oldergod, line 20: Is there a spec as to how they're defined?

wire-protoc-integration/src/main/java/com/squareup/wire/protocwire/JavaTargetConfig.kt
```diff
@@ -0,0 +1,36 @@
...
+     * @return the translated JavaTarget.
+     */
```
oldergod, line 22: We can keep it but we should `@JvmStatic` for java callers.

wire-protoc-integration/src/main/java/com/squareup/wire/protocwire/Plugin.kt
```diff
@@ -0,0 +1,260 @@
...
+import com.google.protobuf.Descriptors
+import com.google.protobuf.DynamicMessage
+import com.google.protobuf.ExtensionRegistry
+import com.google.protobuf.compiler.PluginProtos
+import java.io.IOException
+import java.io.InputStream
+import java.io.OutputStream
+
+/**
+ * Front-end for protoc code generator plugins written in Kotlin (translated from Java).
```
oldergod, line 15: From which Java class has it been translated? Can we link it here?

wire-protoc-integration/src/main/java/com/squareup/wire/protocwire/Plugin.kt
```diff
@@ -0,0 +1,260 @@
...
+            throw PluginException("protoc sent unparseable request to plugin.", e)
+        }
+        files = asDescriptors(request.protoFileList)
+        val output = CodedOutputStream.newInstance(
+            environment.outputStream()
+        )
+        try {
+            // go ahead and write response preamble
+            PluginProtos.CodeGeneratorResponse
+                .newBuilder() // add more here as more features are introduced and then supported
```
oldergod, line 98: Is it a TODO?, and let's keep comments in their own line, that helps readability

wire-protoc-integration/src/main/java/com/squareup/wire/protocwire/Plugin.kt
```diff
@@ -0,0 +1,260 @@
...
+            rawRequest = ByteString.readFrom(environment.inputStream())
+            request = PluginProtos.CodeGeneratorRequest.parseFrom(rawRequest)
+        } catch (e: IOException) {
+            throw PluginException("protoc sent unparseable request to plugin.", e)
+        }
+        var files = asDescriptors(request.protoFileList)
+        val reg = createExtensionRegistry(files.values)
+
+        // now we must *re-parse* the request, but this time we can properly parse any
+        // custom options therein
```
oldergod, line 85: Let's keep comments as sentences.

wire-protoc-integration/src/main/java/com/squareup/wire/protocwire/ProtocFileSystem.kt
```diff
@@ -0,0 +1,45 @@
+package com.squareup.wire.protocwire
+
+import okio.Buffer
+import okio.ForwardingFileSystem
+import okio.Path
+import okio.Sink
+import okio.Timeout
+
+internal class ProtocFileSystem(private val response: Plugin.Response): ForwardingFileSystem(SYSTEM) {
```
oldergod, line 9: Let's add a comment as to why we need this file system, namely that we don't wanna write on disk but do memory manipulation on the response

oldergod (approved):
Great PR!!!
How hard would it be to write some tests around file descriptor -> schema conversion, and for the few CodeGenerators?

wire-protoc-integration/src/main/java/com/squareup/wire/protocwire/JavaTargetConfig.kt
buildbreaker: Yeah, I was considering that but I was erroring on keeping the changes in an isolated module here. I can make the change to move it to the `JavaTarget` and `KotlinTarget` if that is preferred

wire-protoc-integration/src/main/java/com/squareup/wire/protocwire/CodeGenerator.kt
```diff
@@ -0,0 +1,19 @@
+package com.squareup.wire.protocwire
+
+import com.squareup.wire.protocwire.Plugin.DescriptorSource
+import com.google.protobuf.compiler.PluginProtos
+
+/**
+ * Interface to be implemented by a code generator.
```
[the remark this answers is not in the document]
buildbreaker: Ah. I did copy this code from somewhere and then translated it to Kotlin -- I was previously hesitant to make further modifications but I feel much more confident that we can simplify this and inline it as you said :)

wire-protoc-integration/src/main/java/com/squareup/wire/protocwire/ProtocFileSystem.kt
buildbreaker: I added a comment: ```/**
 * The primary responsibility of the protoc plugin is to do some memory
 * manipulation and deferring to protoc to manage the creation of the generated
 * files through the [Plugin.Response].
 *
 * The [ProtocFileSystem] helps with replacing Wire's file writing behavior with
 * one that defers to protoc to handle generated files.
 */``` wdyt?

wire-protoc-integration/src/main/java/com/squareup/wire/protocwire/Plugin.kt
buildbreaker: Removed

wire-protoc-integration/src/main/java/com/squareup/wire/protocwire/Plugin.kt
```diff
@@ -0,0 +1,260 @@
...
+import java.io.OutputStream
+
+/**
+ * Front-end for protoc code generator plugins written in Kotlin (translated from Java).
+ *
+ *
+ * To implement a protoc plugin in Kotlin, simply write an implementation of
+ * [CodeGenerator], then create a main() method like:
+ *
+ * <pre>`
```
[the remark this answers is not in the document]
buildbreaker: I cleaned these tags up and made sure the javadoc views are correct!

wire-protoc-integration/src/main/java/com/squareup/wire/protocwire/Plugin.kt
buildbreaker: Added!

wire-protoc-integration/src/main/java/com/squareup/wire/protocwire/JavaTargetConfig.kt
buildbreaker: I've linked the top comment to https://github.com/buildbreaker/wire/blob/achiu/protoc-plugin/wire-protoc-integration/src/main/java/com/squareup/wire/protocwire/ParameterParser.kt#L21-L28

wire-protoc-integration/src/main/java/com/squareup/wire/protocwire/CodeGenerator.kt
buildbreaker: I reworded the comment a bit calling out that it takes in a `.proto` file. Wdyt?

oldergod:
Alan, I'll close the PR for now. And I think it'll be best to put that into its own repo if we ever take it to completion.

PR #2409, opened 2023-02-13
Fix base64url decoding when string length is a multiple of 4

jgilbert002:
## Issue

Fixed an issue with `base64url` -> `base64` conversion when string length is a multiple of 4.

4 `"="` were added as padding, no padding should be added in this case.

e.g:
`CiHSZFeg4VTJRJO5B9vXbTmlOqQSjwe9CAZG6jSj2RxDt/AVAQAAABgWIg1NTFZCQkU0TVo3WVE0====`
vs
`CiHSZFeg4VTJRJO5B9vXbTmlOqQSjwe9CAZG6jSj2RxDt/AVAQAAABgWIg1NTFZCQkU0TVo3WVE0`

## Testing

Moved the `base64`->`base64url` conversion under a `String` extension so they can be tested.

Added unit test to cover `base64`->`base64url` & `base64url`->`base64`


lickel (approved):
Thanks!

PR #2413, opened 2023-03-09
Swift: fix proto3 messages not always omitting fields with default values

jszumski:
Fix issue where proto3 messages with omitted values were still encoding those fields when at the top level of a message.

----

When encoding a message like: 
```
syntax = "proto3";

message EmptyOmitted {
    int32 numeric_value = 1;
}
``` 

`numeric_value` should be omitted when set to `0`.  In the current runtime however `isProto3` defaults to `false`, so the first pass through `encode(tag:value:encoding:)` would always fail the `if value == 0 && isProto3 { return }` test.

To fix that, I set the initial `isProto3` value to that syntax of the parent message.

### DIFF
--- a/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoEncoder.swift
+++ b/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoEncoder.swift
@@ -80,7 +80,11 @@ public final class ProtoEncoder {
         // Use the size of the struct as an initial estimate for the space needed.
         let structSize = MemoryLayout.size(ofValue: value)
 
-        let writer = ProtoWriter(data: .init(capacity: structSize))
+        let writer = ProtoWriter(
+            data: .init(capacity: structSize),
+            outputFormatting: [],
+            rootMessageProtoSyntax: T.self.protoSyntax ?? .proto2
+        )
         writer.outputFormatting = outputFormatting
         try value.encode(to: writer)
 
--- a/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoWriter.swift
+++ b/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoWriter.swift
@@ -41,6 +41,11 @@ public final class ProtoWriter {
     private(set) var buffer: WriteBuffer
 
     /**
+     The syntax used by the root message, which can be different than child messages, which will individually set `isProto3`.
+     */
+    private let rootIsProto3: Bool
+
+    /**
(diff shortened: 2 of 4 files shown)


wire-runtime-swift/src/main/swift/ProtoCodable/ProtoWriter.swift
```diff
@@ -59,10 +64,15 @@ public final class ProtoWriter {
 
     init(
         data: WriteBuffer = .init(),
-        outputFormatting: ProtoEncoder.OutputFormatting = []
+        outputFormatting: ProtoEncoder.OutputFormatting = [],
+        protoSyntax: ProtoSyntax = .proto2
```
efirestone, line 68: nit: can we name this `rootMessageSyntax` or `rootMessageProtoSyntax` so that it's clear this isn't the syntax used for all messages that are written out?

wire-runtime-swift/src/main/swift/ProtoCodable/ProtoEncoder.swift
```diff
@@ -80,7 +80,9 @@ public final class ProtoEncoder {
         // Use the size of the struct as an initial estimate for the space needed.
         let structSize = MemoryLayout.size(ofValue: value)
 
-        let writer = ProtoWriter(data: .init(capacity: structSize))
+        let writer = ProtoWriter(data: .init(capacity: structSize),
```
efirestone, line 83: nit: style
```suggestion
        let writer = ProtoWriter(
            data: .init(capacity: structSize),
            protoSyntax: T.self.protoSyntax ?? .proto2
        )
```

wire-runtime-swift/src/main/swift/ProtoCodable/ProtoEncoder.swift
```diff
@@ -80,7 +80,9 @@ public final class ProtoEncoder {
...
+                                 outputFormatting: [],
+                                 protoSyntax: T.self.protoSyntax ?? .proto2)
```
efirestone, line 85: I wonder why we made this optional...can't remember now. This is right though, as "no syntax definition" means proto2.

efirestone (approved):

wire-runtime-swift/src/main/swift/ProtoCodable/ProtoWriter.swift
jszumski: I'll update to `rootMessageProtoSyntax`

PR #2415, opened 2023-03-14
Swift: Improve generated code for proto3 messages

jszumski:
- Inserts proto3 default values into generated Swift code where applicable
- Adds an `encode(tag:value:)` pair for `Bool` so that it can be correctly omitted when `false`

-------

```diff
@@ -88,7 +88,7 @@ extension Person3.PhoneNumber : ProtoMessage {
 
 extension Person3.PhoneNumber : Proto3Codable {
     public init(from reader: ProtoReader) throws {
-        var number: String? = nil
+        var number: String = ""
         var type: Person3.PhoneType? = nil
 
         let token = try reader.beginMessage()
@@ -101,8 +101,8 @@ extension Person3.PhoneNumber : Proto3Codable {
         }
         self.unknownFields = try reader.endMessage(token: token)
 
-        self.number = try Person3.PhoneNumber.checkIfMissing(number, "number")
-        self.type = try Person3.PhoneNumber.checkIfMissing(type, "type")
+        self.number = number
+        self.type = try Person3.PhoneType.defaultIfMissing(type)
     }
 
     public func encode(to writer: ProtoWriter) throws {
@@ -157,8 +157,8 @@ extension Person3 : ProtoMessage {
 
 extension Person3 : Proto3Codable {
     public init(from reader: ProtoReader) throws {
-        var name: String? = nil
-        var id: Int32? = nil
+        var name: String = ""
+        var id: Int32 = 0
         var email: String? = nil
         var phone: [Person3.PhoneNumber] = []
         var aliases: [String] = []
@@ -178,8 +178,8 @@ extension Person3 : Proto3Codable {
         }
         self.unknownFields = try reader.endMessage(token: token)
 
-        self.name = try Person3.checkIfMissing(name, "name")
-        self.id = try Person3.checkIfMissing(id, "id")
+        self.name = name
+        self.id = id
         self.email = email
         self.phone = phone
         self.aliases = aliases
```

### DIFF
--- a/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
+++ b/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
@@ -132,6 +132,22 @@ class SwiftGenerator private constructor(
       else -> null
     }
 
+  // see https://protobuf.dev/programming-guides/proto3/#default
+  private val Field.proto3InitialValue: String
+    get() = when {
+      isMap -> "[:]"
+      isRepeated -> "[]"
+      isOptional -> "nil"
+      else -> when (typeName.makeNonOptional()) {
+        BOOL -> "false"
+        DOUBLE, FLOAT -> "0"
+        INT32, UINT32, INT64, UINT64 -> "0"
+        STRING -> """""""" // evaluates to the empty string
+        DATA -> ".init()"
+        else -> "nil"
+      }
+    }
+
   private val Field.codableName: String?
     get() = jsonName?.takeIf { it != name } ?: camelCase(name).takeIf { it != name }
 
@@ -192,6 +208,27 @@ class SwiftGenerator private constructor(
 
   private val MessageType.isHeapAllocated get() = fields.size + oneOfs.size >= 16
 
+  /**
+   * Checks that every enum in a proto3 message contains a value with tag 0.
+   *
+   * @throws NoSuchElementException if the case doesn't exist
+   */
+  @Throws(NoSuchElementException::class)
+  private fun validateProto3DefaultsExist(type: MessageType) {
+    if (type.syntax == PROTO_2) { return }
+
+    // validate each enum field
+    type
+      .fields
+      .mapNotNull { schema.getType(it.type!!) as? EnumType }
+      .forEach { enum ->
+         // ensure that a 0 case exists
+         if (enum.constants.filter { it.tag == 0 }.isEmpty()) {
+           throw NoSuchElementException("Missing a zero value for ${enum.name}")
+         }
+      }
+  }
+
   @OptIn(ExperimentalStdlibApi::class) // TODO move to build flag
   private fun generateMessage(
     type: MessageType,
@@ -207,6 +244,8 @@ class SwiftGenerator private constructor(
     val storageName = if ("storage" in propertyNames) "_storage" else "storage"
 
     val typeSpecs = mutableListOf<TypeSpec>()
+
+    validateProto3DefaultsExist(type)
 
     typeSpecs += TypeSpec.structBuilder(structType)
       .addModifiers(PUBLIC)
@@ -449,18 +488,30 @@ class SwiftGenerator private constructor(
         .addParameter("from", reader, protoReader)
         .throws(true)
         .apply {
-          // Declare locals into which everything is writen before promoting to members.
+          // Declare locals into which everything is written before promoting to members.
           type.fields.forEach { field ->
-            val localType = if (field.isRepeated || field.isMap) {
-              field.typeName
-            } else {
-              field.typeName.makeOptional()
+            val localType = when (type.syntax) {
+              PROTO_2 -> if (field.isRepeated || field.isMap) {
+                  field.typeName
+                } else {
+                  field.typeName.makeOptional()
+                }
+              PROTO_3 -> if (field.isOptional || (field.isEnum && !field.isRepeated)) {
+                  field.typeName.makeOptional()
+                } else {
+                  field.typeName
+                }
             }
-            val initializer = when {
-              field.isMap -> "[:]"
-              field.isRepeated -> "[]"
-              else -> "nil"
+
+            val initializer = when (type.syntax) {
+              PROTO_2 -> when {
+                  field.isMap -> "[:]"
+                  field.isRepeated -> "[]"
+                  else -> "nil"
+                }
+              PROTO_3 -> field.proto3InitialValue
             }
+
             addStatement("var %N: %T = %L", field.name, localType, initializer)
           }
           type.oneOfs.forEach { oneOf ->
@@ -533,10 +584,17 @@ class SwiftGenerator private constructor(
           // Check required and bind members.
           addStatement("")
           type.fields.forEach { field ->
-            val initializer = if (field.isOptional || field.isRepeated || field.isMap) {
-              CodeBlock.of("%N", field.name)
-            } else {
-              CodeBlock.of("try %1T.checkIfMissing(%2N, %2S)", structType, field.name)
+            val initializer = when(type.syntax) {
+              PROTO_2 -> if (field.isOptional || field.isRepeated || field.isMap) {
+                CodeBlock.of("%N", field.name)
+              } else {
+                CodeBlock.of("try %1T.checkIfMissing(%2N, %2S)", structType, field.name)
+              }
+              PROTO_3 -> if (field.isEnum && !field.isRepeated) {
+                CodeBlock.of("try %1T.defaultIfMissing(%2N)", field.typeName.makeNonOptional(), field.name)
+              } else {
+                CodeBlock.of("%N", field.name)
+              }
             }
             addStatement("self.%N = %L", field.name, initializer)
           }
--- a/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoCodable.swift
+++ b/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoCodable.swift
@@ -105,3 +105,18 @@ extension ProtoDecodable {
     }
 
 }
+
+extension ProtoEnum where Self : RawRepresentable, RawValue == UInt32 {
+
+    /**
+     A convenience function used with enum fields that throws an error if the field is null
+     and its default value can't be used instead.
+     */
+    public static func defaultIfMissing(_ value: Self?) throws -> Self {
+        guard let value = value ?? Self(rawValue: 0) else {
+            throw ProtoDecoder.Error.missingEnumDefaultValue(type: Self.self)
+        }
+        return value
+    }
+
+}
--- a/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoDecoder.swift
+++ b/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoDecoder.swift
@@ -51,6 +51,7 @@ public final class ProtoDecoder {
         case mapEntryWithoutKey(value: Any?)
         case mapEntryWithoutValue(key: Any)
         case messageWithoutLength
+        case missingEnumDefaultValue(type: Any.Type)
         case missingRequiredField(typeName: String, fieldName: String)
         case recursionLimitExceeded
         case unexpectedEndOfData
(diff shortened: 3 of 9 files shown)


jszumski (commented):
This is my first time writing Kotlin so there may be better ways to do some of the code gen

lickel:
Something @efirestone and I had talked about was removing the initializer defaults entirely since it likely is the cause of code bloat

efirestone:
> Something @efirestone and I had talked about was removing the `init()` defaults entirely since it likely is the cause of code bloat

I mentioned this too, but I think that's probably a separate initiative. Worth fixing the proto3 bugs first, then we can do some profiling to see if the `init`s actually cause a lot of the bulk.

lickel:
In that vein would it make sense to not touch `init()` and just touch the proto inflation bits?

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -88,6 +89,8 @@ class SwiftGenerator private constructor(
     get() = schema.getType(this) is MessageType
   private val ProtoType.isEnum
     get() = schema.getType(this) is EnumType
+  private val ProtoType.asEnum
```
lickel, line 92: This should probably be `as? EnumType`

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -115,6 +118,8 @@ class SwiftGenerator private constructor(
     get() = type!!.keyType!!
   internal val Field.valueType: ProtoType
     get() = type!!.valueType!!
+  internal val Field.enumType: EnumType
```
lickel, line 121: And similarly here, `EnumType?`

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -132,6 +137,23 @@ class SwiftGenerator private constructor(
       else -> null
     }
 
+  // see https://protobuf.dev/programming-guides/proto3/#default
+  private val Field.proto3InitialValue: String
+    get() = when {
+      isMap -> "[:]"
+      isRepeated -> "[]"
+      isOptional -> "nil"
+      isEnum -> "." + enumType.constants.first().name
```
lickel, line 146: This is kind of dangerous. The correct value is `EnumType(rawValue: 0)`.

You can always prune an enum value which also makes things a little funky

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -449,19 +471,31 @@ class SwiftGenerator private constructor(
...
-            } else {
-              field.typeName.makeOptional()
-            }
-            val initializer = when {
-              field.isMap -> "[:]"
-              field.isRepeated -> "[]"
-              else -> "nil"
+            when (type.syntax) {
+              PROTO_2 -> {
+                val localType = if (field.isRepeated || field.isMap) {
```
lickel, line 478: Maybe make a `proto2InitialValue` ?

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -844,11 +887,14 @@ class SwiftGenerator private constructor(
...
-      field.isRepeated -> defaultValue("[]")
-      field.isOptional -> defaultValue("nil")
+  private fun ParameterSpec.Builder.withFieldDefault(field: Field, syntax: Syntax) = apply {
+    when (syntax) {
+      PROTO_2 -> when {
+        field.isMap -> defaultValue("[:]")
+        field.isRepeated -> defaultValue("[]")
+        field.isOptional -> defaultValue("nil")
+      }
+      PROTO_3 -> defaultValue(field.proto3InitialValue)
```
lickel, line 897: Hmm... maybe you should just delete `proto3InitialValue` and unwind it here?

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -132,6 +137,23 @@ class SwiftGenerator private constructor(
...
+      isRepeated -> "[]"
+      isOptional -> "nil"
+      isEnum -> "." + enumType.constants.first().name
+      else -> when (typeName.makeNonOptional()) {
+        BOOL -> "false"
+        DOUBLE, FLOAT -> "0"
+        INT32, UINT32, INT64, UINT64 -> "0"
+        STRING -> "\"\""
+        DATA -> ".init()"
+        else -> "nil"
```
lickel, line 153: Definitely not saying it's wrong, but I feel a little itchy that you're deriving a lot of logic on this "nil".

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -498,6 +532,12 @@ class SwiftGenerator private constructor(
               }
             }
             decoder.add(")")
+
+            // local proto3 enum fields are initialized to a default value, so ensure we keep that if the decode returns nil
+            if (type.syntax == PROTO_3 && field.isEnum && !field.isRepeated) {
+              decoder.add(" ?? %N", field.name)
```
efirestone, line 538: This feels dangerous. IIRC, this is for the case where we return `nil` for an unknown enum case, but in this case we're going to fill that in with the default value. If that's the case, then I think this is still an error (at least IMO it is). If you want the "return nil" behavior, then you can/should mark the field as optional (which still works in proto3).

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
jszumski: Is it valid to prune the default value? If no, is there currently any protection against it?

I chose this route because it makes the generated code _much_ nicer to read, but I can see why `rawValue` is probably where this should end up.

efirestone (approved):

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -132,6 +137,23 @@ class SwiftGenerator private constructor(
...
+    get() = when {
+      isMap -> "[:]"
+      isRepeated -> "[]"
+      isOptional -> "nil"
+      isEnum -> "." + enumType.constants.first().name
+      else -> when (typeName.makeNonOptional()) {
+        BOOL -> "false"
+        DOUBLE, FLOAT -> "0"
+        INT32, UINT32, INT64, UINT64 -> "0"
+        STRING -> "\"\""
```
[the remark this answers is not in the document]
JakeWharton: It's actually `""""""""`. Bonus points if you can figure out WTF is going on there...

https://pl.kotl.in/niP51h4np

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
efirestone: Yeah, was kind of feeling the same. There should be an `isOptional` property on most fields I believe.

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
efirestone: In all cases we want to conform to the [proto spec](https://protobuf.dev/programming-guides/proto3/#enum), so `EnumType(rawValue: 0)!` is right. If you do prune it then this force unwrap would fail.

This could also be the right place to guard against pruning it...if we can't find the zero-value here, then throw an exception and have the compiler bail out. I believe the prunes have already happened at this point (but we should verify).

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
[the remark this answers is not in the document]
lickel: Oh no!

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
lickel: Personally I have been pruning my Proto2 enums' 0 value since they're all optional, but yeah if it's required then you're kinda stuck with rawValue: 0

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
jszumski: The only alternative is to force unwrap the `decode` and crash if the enum decoding strategy is set to `.returnNil`.  The local `type` here isn't optional:

```diff
extension Person3.PhoneNumber : Proto3Codable {
    public init(from reader: ProtoReader) throws {
        var number: String = ""
-        var type: Person3.PhoneType = .MOBILE
+        var type: Person3.PhoneType = Person3.PhoneType(rawValue: 0)!

        let token = try reader.beginMessage()
        while let tag = try reader.nextTag(token: token) {
            switch tag {
            case 1: number = try reader.decode(String.self)
-            case 2: type = try reader.decode(Person3.PhoneType.self) ?? type
+            case 2: type = try reader.decode(Person3.PhoneType.self)!
            default: try reader.readUnknownField(tag: tag)
            }
        }
        self.unknownFields = try reader.endMessage(token: token)

        self.number = number
        self.type = type
    }
```

Some much force unwrapping makes me think it's the wrong approach.  If folks are worried about correctness, wouldn't they use the `throwError` strategy?

lickel: Why not: 
```swift
var type: Person3.PhoneType?
// ...
case 2: type = try reader.decode(Person3.PhoneType.self)
//
self.type = try {
    guard let type = type ?? Person3.PhoneType(rawValue: 0) else {
        throw SomeError()
    }
    return type
}()
```

jszumski: I like that approach, I'll add an extension that can be used similarly to `checkIfMissing()`

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
jszumski: Moved control flow to rely on `isOptional` instead

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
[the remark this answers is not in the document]
jszumski: Updated to `""""""""`

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
jszumski: I made this change.  It's now the only place where `proto3InitialValue` is read, so we could inline that if it would be more understandable

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
jszumski: I removed this and now use `rawValue:0` inside of the new `defaultIfMissing`

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
efirestone: Can't we check this at codegen time (and refuse to generate if it's wrong)? We know at that point whether or not case 0 exists.

jszumski: The proto structure itself is checked in https://github.com/square/wire/blob/master/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SyntaxRules.kt#L104, but I added a runtime check that will protect against pruning the 0th value

lickel (approved):

PR #2421, opened 2023-03-20
Clearing some warnings

oldergod:


wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/InternalJvm.kt
```diff
@@ -23,15 +23,15 @@ import com.squareup.wire.WireEnum
...
     list[i] = adapter.redact(list[i])
   }
 }
 
-fun <T> redactElements(map: java.util.Map<*, T>, adapter: ProtoAdapter<T>) {
-  for (entry in map.entrySet()) {
-    entry.setValue(adapter.redact(entry.value))
+fun <T> redactElements(map: MutableMap<*, T>, adapter: ProtoAdapter<T>) {
+  for ((key, value) in map as MutableMap<Any, T>) {
+    map[key] = adapter.redact(value)
```
JakeWharton, line 34: This is not safe to do. You can't mutate a map separately from iterating. A MutableMap iterator returns mutable entries on which you can still call setValue.

wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/Internal.kt
```diff
@@ -209,17 +209,17 @@ fun sanitize(values: List<String>): String {
 }
 
 fun boxedOneOfClassName(oneOfName: String): String {
-  return oneOfName.capitalize()
+  return oneOfName.replaceFirstChar { if (it.isLowerCase()) it.titlecase() else it.toString() }
```
JakeWharton, line 212: You don't need the if/else here and can unconditionally call titlecase().

PR #2422, opened 2023-03-20
Refactoring some internal implementation

oldergod:


wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/InternalJvm.kt
```diff
@@ -30,8 +30,8 @@ fun <T> redactElements(list: MutableList<T>, adapter: ProtoAdapter<T>) {
 }
 
 fun <T> redactElements(map: MutableMap<*, T>, adapter: ProtoAdapter<T>) {
-  for ((key, value) in map as MutableMap<Any, T>) {
-    map[key] = adapter.redact(value)
+  for (entry in map.entries) {
+    entry.setValue(adapter.redact(entry.value))
   }
```
oldergod, line 35: Why is it dangerous if I'm not adding/removing entries?

wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/Internal.kt
```diff
@@ -209,7 +209,7 @@ fun sanitize(values: List<String>): String {
 }
 
 fun boxedOneOfClassName(oneOfName: String): String {
-  return oneOfName.replaceFirstChar { if (it.isLowerCase()) it.titlecase() else it.toString() }
```
oldergod, line 212: btw that is what the kotlin lib would `replaceWith` if one uses `capitalize()`

JakeWharton: Yeah it has to maintain the old behavior and its quirks so that the replacement works unconditionally. Here we're safe to always transform the first character though.

JakeWharton (approved):

PR #2423, opened 2023-03-20
Support Builders for Native/JS

oldergod:
I'm not really happy with `jvmOnly` but that's what I came up with which would not introduce breaking changes.


wire-runtime/src/jsMain/kotlin/com/squareup/wire/Message.kt
```diff
@@ -53,5 +54,73 @@ protected actual constructor(
   /**
    * Superclass for protocol buffer message builders.
    */
-  actual abstract class Builder<M : Message<M, B>, B : Builder<M, B>>
+  actual abstract class Builder<M : Message<M, B>, B : Builder<M, B>> protected constructor() {
+    /**
```
Egorand, line 58: These comments aren't inherited from the `expect Builder`, right?

wire-runtime/src/jsMain/kotlin/com/squareup/wire/Message.kt
```diff
@@ -53,5 +54,73 @@ protected actual constructor(
...
+     * [Buffer] of the message's unknown fields that is lazily instantiated between calls to
+     * [buildUnknownFields]. It's automatically cleared in [buildUnknownFields], and can also be
+     * manually cleared by calling [clearUnknownFields].
+     */
+    internal var unknownFieldsBuffer: Buffer? = null
+    internal var unknownFieldsWriter: ProtoWriter? = null
+
+    fun addUnknownFields(unknownFields: ByteString): Builder<M, B> = apply {
+      if (unknownFields.size > 0) {
```
Egorand, line 74: There's probably no good way to extract this code to avoid duplication 😢

Egorand (approved):

wire-runtime/src/jsMain/kotlin/com/squareup/wire/Message.kt
oldergod: But you're right, there are things I could move onto the `expect`

oldergod: Done!

oldergod:
I could have otherwise created typealiases for `kotlin.jvm.*` annotations...

PR #2424, opened 2023-03-29
Swift: encode & decode size-delimited messages

jszumski:
Adds the ability to encode and decode a size-delimited message collection in Swift.  

The intended use case is to decode a stream of messages recorded to disk (e.g., [Bazel's Build Event Protocol](https://bazel.build/remote/bep#consuming-bep-binary)). Similar implementations exist in Java ([writeDelimitedTo](https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/MessageLite.html#writeDelimitedTo-java.io.OutputStream-) and [parseDelimitedFrom](https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/AbstractParser.html#parseDelimitedFrom-java.io.InputStream-)) and other languages in https://github.com/protocolbuffers/protobuf/issues/10229.

### DIFF
--- a/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoDecoder.swift
+++ b/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoDecoder.swift
@@ -279,4 +279,52 @@ public final class ProtoDecoder {
         }
         return value
     }
+
+    /** Decodes the provided size-delimited data into instances of the requested type.
+     *
+     * A size-delimited collection of messages is a sequence of varint + message pairs
+     * where the varint indicates the size of the subsequent message.
+     *
+     * - Parameters:
+     *   - type: the type to decode
+     *   - data: the serialized size-delimited data for the messages
+     * - Returns: an array of the decoded messages
+     */
+    public func decodeSizeDelimited<T: ProtoDecodable>(_ type: T.Type, from data: Foundation.Data) throws -> [T] {
+        var values: [T] = []
+
+        try data.withUnsafeBytes { buffer in
+            // Handle the empty-data case.
+            guard let baseAddress = buffer.baseAddress, buffer.count > 0 else {
+                return
+            }
+
+            let fullBuffer = ReadBuffer(
+                storage: baseAddress.bindMemory(to: UInt8.self, capacity: buffer.count),
+                count: buffer.count
+            )
+
+            while fullBuffer.isDataRemaining, let size = try? fullBuffer.readVarint() {
+                if size == 0 { break }
+
+                let messageBuffer = ReadBuffer(
+                    storage: fullBuffer.pointer,
+                    count: Int(size)
+                )
+
+                let reader = ProtoReader(
+                    buffer: messageBuffer,
+                    enumDecodingStrategy: enumDecodingStrategy
+                )
+
+                values.append(try reader.decode(type))
+
+                // Advance the buffer before reading the next item in the stream.
+                _ = try fullBuffer.readBuffer(count: Int(size))
+            }
+        }
+
+        return values
+    }
+
 }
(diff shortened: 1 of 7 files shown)


wire-runtime-swift/src/main/swift/ProtoCodable/ProtoEncoder.swift
```diff
@@ -82,13 +82,40 @@ public final class ProtoEncoder {
...
         try value.encode(to: writer)
 
         return Data(writer.buffer, copyBytes: false)
     }
 
+    public func encodeSizeDelimited<T: ProtoEncodable>(_ values: [T]) throws -> Data {
+        // Use the size of the struct as an initial estimate for the space needed.
+        let structSize = MemoryLayout.size(ofValue: T.self)
+
+        let fullBuffer = WriteBuffer(capacity: structSize * values.count)
```
jszumski, line 98: Should this calculation attempt to reserve some capacity for the varints or just let the buffer expand naturally?

wire-runtime-swift/src/main/swift/ProtoCodable/WriteBuffer.swift
```diff
@@ -36,7 +36,7 @@ final class WriteBuffer {
     init(capacity: Int = 0) {
         self.capacity = 0
 
-        if capacity > 0 {
+        if capacity >= 0 {
```
jszumski, line 39: This was required to ensure `storage` was allocated in all cases (see `testAppendEmptyFirst`)

efirestone: I'm a little wary of calling `malloc(0)` in `expand()`. In C at least, the behavior for this [can vary](https://stackoverflow.com/questions/2132273/what-does-malloc0-return) from runtime to runtime.

Can we achieve the same thing by adding something like this into `append`:
```
guard !data.isEmpty else { return }
```

wire-runtime-swift/src/main/swift/ProtoCodable/ProtoEncoder.swift
jszumski: Updated to reserve space 👍

wire-runtime-swift/src/main/swift/ProtoCodable/WriteBuffer.swift
jszumski: Added `guard`s where appropriate

wire-runtime-swift/src/main/swift/ProtoCodable/ProtoDecoder.swift
```diff
@@ -148,5 +154,51 @@ public final class ProtoDecoder {
...
+    ///   - type: the type to decode
+    ///   - data: the serialized size-delimited data for the messages
+    /// - Returns: an array of the decoded messages
+    public func decodeSizeDelimited<T: ProtoDecodable>(_ type: T.Type, from data: Data) throws -> [T] {
+        var values: [T] = []
+
+        try data.withUnsafeBytes { buffer in
+            // Handle the empty-data case.
+            guard let baseAddress = buffer.baseAddress, buffer.count > 0 else {
+                values = [try T(from: .empty)]
```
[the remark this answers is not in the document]
jszumski: This should have been an empty array

oldergod (commented):
Could you make sure all your comments are sentences?, they all look good but some are missing a period.

dnkoutso (approved):

wire-runtime-swift/src/main/swift/ProtoCodable/ProtoDecoder.swift
```diff
@@ -279,4 +279,51 @@ public final class ProtoDecoder {
...
+
+    /// Decodes the provided size-delimited data into instances of the requested type.
+    ///
+    /// A size-delimited collection of messages is a sequence of varint + message pairs
+    /// where the varint indicates the size of the subsequent message.
+    ///
+    /// - Parameters:
+    ///   - type: the type to decode
+    ///   - data: the serialized size-delimited data for the messages
+    /// - Returns: an array of the decoded messages
```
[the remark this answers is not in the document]
jszumski: Updated

efirestone (approved):

PR #2429, opened 2023-04-06
Swift: Preserve oneof capitalization from the original message

jszumski:
Preserve `oneof` capitalization from the original message instead of always capitalizing it.

This matches the behavior for message and enum types, but is a breaking change.

-----

```
message OneOfs {
  oneof choice {
    NestedEnum enum_option = 1;
    NestedEnum similar_enum_option = 2;
    string string_option = 3;
  }
}
```

```diff
public struct OneOfs {

    public var standalone_enum: NestedEnum?
-    public var choice: Choice?
+    public var choice: choice?
    public var unknownFields: Data = .init()

-    public init(standalone_enum: NestedEnum? = nil, choice: Choice? = nil) {
+    public init(standalone_enum: NestedEnum? = nil, choice: choice? = nil) {
        self.standalone_enum = standalone_enum
        self.choice = choice
    }

-    public enum Choice {
+    public enum choice {

        case enum_option(NestedEnum)
        case similar_enum_option(NestedEnum)
        case string_option(String)

        ...
    }
}
```


jszumski:
After discussing with @efirestone directly, I'm going to close this and pursue a refactor that let's the compiler disambiguate these cases regardless of capitalization.  

As written this change would accomodate messages like:
```
message File {
  oneof file {
    string val1 = 1;
    bytes val2 = 2;
  }
}
```

but not 

```
message File {
  oneof File {
    string val1 = 1;
    bytes val2 = 2;
  }
}
```

where the generated code would need to know its enclosing module to use the fully-qualified type `{module}.File.File`

PR #2431, opened 2023-04-10
Optimize packed scalar list sizes in initial allocation

JGulbronson:
On the JVM, large `repeated` fields can result in excessive memory allocations. `mutableListOf<T>` returns an `ArrayList`, which must do an `Arrays.copyOf` when the list runs out of space. If the number of elements in the `repeated` field is large enough, it can call `Arrays.copy` of multiple times for a single `decode` call.

We can improve this if we are using `packed`. The length of the packed field is known when decoding, and some types have a fixed encoded length (either FIXED32 or FIXED64). In those cases, we're able to calculate the number of elements we're about to decode, and initialize the `ArrayList` with the proper capacity. If we are using `packed` but the field wasn't encoded as `packed` (which must be allowed to be backwards compatible), we will simply initialize the list to a capacity of 1, and it will get resized like normal as elements are added. The default capacity is normally 10, so we'd end up resizing from 1 -> 2 -> 4 -> 8, but I feel that is a small penalty for what amounts to an edge case.

The next step will be to see if we can use the Array types (`FloatArray`, `IntArray`, etc), which will give us the same benefits with the addition of fixing autoboxing, as all Kotlin `Lists` autobox primitives.

### DIFF
--- a/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
+++ b/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
@@ -1712,7 +1712,7 @@ class KotlinGenerator private constructor(
           buildCodeBlock {
             beginControlFlow("if (%L == null)", fieldName)
             addStatement("val minimumByteSize = ${field.getMinimumByteSize()}")
-            addStatement("val initialCapacity = (reader.nextFieldLengthInBytes() / minimumByteSize)")
+            addStatement("val initialCapacity = (reader.nextFieldMinLengthInBytes() / minimumByteSize)")
             addStatement("⇥.coerceAtMost(Int.MAX_VALUE.toLong())")
             addStatement(".toInt()")
             addStatement("⇤%L = ArrayList(initialCapacity)", fieldName)
@@ -2186,7 +2186,7 @@ class KotlinGenerator private constructor(
   }
 
   private fun Field.getDeclaration(allocatedName: String) = when {
-    isPacked && isScalar -> CodeBlock.of("var $allocatedName: MutableList<%T>? = null", type!!.typeName)
+    isPacked && isScalar -> CodeBlock.of("var %N: MutableList<%T>? = null", allocatedName, type!!.typeName)
     isRepeated -> CodeBlock.of("val $allocatedName = mutableListOf<%T>()", type!!.typeName)
     isMap -> CodeBlock.of(
       "val $allocatedName = mutableMapOf<%T, %T>()",
--- a/wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoReader.kt
+++ b/wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoReader.kt
@@ -452,7 +452,7 @@ class ProtoReader(private val source: BufferedSource) {
    * have a variable length. LENGTH_DELIMITED fields have a known variable length, while VARINT fields
    * could be as small as a single byte.
    */
-  fun nextFieldLengthInBytes(): Long {
+  fun nextFieldMinLengthInBytes(): Long {
     return when (nextFieldEncoding) {
       FieldEncoding.LENGTH_DELIMITED -> limit - pos
       FieldEncoding.FIXED32 -> 4
--- a/wire-tests/src/commonTest/proto-kotlin/squareup/protos/packed_encoding/EmbeddedMessage.kt
+++ b/wire-tests/src/commonTest/proto-kotlin/squareup/protos/packed_encoding/EmbeddedMessage.kt
@@ -120,7 +120,7 @@ public class EmbeddedMessage(
             1 -> {
               if (inner_repeated_number == null) {
                 val minimumByteSize = 1
-                val initialCapacity = (reader.nextFieldLengthInBytes() / minimumByteSize)
+                val initialCapacity = (reader.nextFieldMinLengthInBytes() / minimumByteSize)
                   .coerceAtMost(Int.MAX_VALUE.toLong())
                   .toInt()
                 inner_repeated_number = ArrayList(initialCapacity)
--- a/wire-gson-support/src/test/java/com/squareup/wire/proto2/person/kotlin/Person.kt
+++ b/wire-gson-support/src/test/java/com/squareup/wire/proto2/person/kotlin/Person.kt
@@ -227,7 +227,7 @@ public class Person(
             5 -> {
               if (favorite_numbers == null) {
                 val minimumByteSize = 1
-                val initialCapacity = (reader.nextFieldLengthInBytes() / minimumByteSize)
+                val initialCapacity = (reader.nextFieldMinLengthInBytes() / minimumByteSize)
                   .coerceAtMost(Int.MAX_VALUE.toLong())
                   .toInt()
                 favorite_numbers = ArrayList(initialCapacity)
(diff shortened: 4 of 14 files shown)


wire-tests/src/commonTest/proto-kotlin/squareup/protos/packed_encoding/EmbeddedMessage.kt
```diff
@@ -113,17 +113,27 @@ public class EmbeddedMessage(
...
+                          if (inner_repeated_number == null) {
+                            val minimumByteSize = 1
+                            val initialCapacity = (reader.nextFieldLengthInBytes() /
+                    minimumByteSize)
+                              .coerceAtMost(Int.MAX_VALUE.toLong())
+                              .toInt()
+                            inner_repeated_number = ArrayList(initialCapacity)
+                          }
+                          inner_repeated_number!!.add(ProtoAdapter.INT32.decode(reader))
+                          }
```
JGulbronson, line 130: This is the aforementioned ugly indentation

JakeWharton:
It's implementation detail, but if you allocate 1 the first expansion will go directly to 10. You will not do 2-->4-->etc. Also the default despite being 10 is actually 0 and only the first insertion will switch to 10.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -1696,6 +1700,18 @@ class KotlinGenerator private constructor(
...
+          """{
+          if (%1L == null) {
+            val minimumByteSize = ${field.getMinimumByteSize()}
+            val initialCapacity = (reader.nextFieldLengthInBytes() / minimumByteSize)
+              .coerceAtMost(Int.MAX_VALUE.toLong())
+              .toInt()
+            %1L = ArrayList(initialCapacity)
+          }
+          %1L!!.add(%2L)
+          }
```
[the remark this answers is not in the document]
JakeWharton: Each line should be a call to a CodeBlock API. The conditional should be built with `beginControlFlow`. There should be other examples in the file.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -1704,6 +1720,21 @@ class KotlinGenerator private constructor(
...
+    require(isScalar && isPacked) { "Field $name is not a scalar" }
+    return when (type) {
+      ProtoType.FLOAT,
+      ProtoType.FIXED32,
+      ProtoType.SFIXED32 -> 4
+      ProtoType.DOUBLE,
+      ProtoType.FIXED64,
+      ProtoType.SFIXED64 -> 8
+      // If we aren't 100% confident in the size of the field, we assume the worst case scenario of 1 byte which gives
+      // us the maximum possible number of elements in the list.
```
oldergod, line 1733: sweet comment

wire-kotlin-generator/src/test/java/com/squareup/wire/kotlin/KotlinGeneratorTest.kt
```diff
@@ -2067,6 +2067,21 @@ class KotlinGeneratorTest {
...
+        "message.proto".toPath(),
+        """
+        |message Embedding {
+        |	repeated float values = 1 [packed = true];
+        |}""".trimMargin()
+      )
+    }
+    val code = KotlinWithProfilesGenerator(schema).generateKotlin("Embedding")
+    assertContains(code, "val values: MutableList<Float>? = null")
+    assertContains(code, "values = ArrayList((reader.nextFieldLengthInBytes() / 4).coerceAtMost(Int.MAX_VALUE.toLong().toInt()))")
```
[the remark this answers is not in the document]
oldergod: That might be a sign we are not testing the right thing? Those are always questionnable. Golden files are great, and we have them. When they change though, we don't know whether the change is expected so still meaningful to have stuff here I guess.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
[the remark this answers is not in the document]
oldergod: What Jake said. The idea is to let KotlinPoet you're opening a "block", closing one etc

oldergod (approved):

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -1647,17 +1651,25 @@ class KotlinGenerator private constructor(
...
+              beginControlFlow("%L -> try", field.tag)
+              addStatement("%L", decodeAndAssign(field, fieldName, adapterName))
+              nextControlFlow("catch (e: %T)", ProtoAdapter.EnumConstantNotFoundException::class)
+              addStatement(
+                "reader.addUnknownField(%L, %T.VARINT, e.value.toLong())", tag,
+                FieldEncoding::class
+              )
+              endControlFlow()
+            }
+            field.isPacked && field.isScalar -> {
```
swankjesse, line 1665: Today I learned enums can be packed!

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -2148,6 +2186,7 @@ class KotlinGenerator private constructor(
   }
 
   private fun Field.getDeclaration(allocatedName: String) = when {
+    isPacked && isScalar -> CodeBlock.of("var $allocatedName: MutableList<%T>? = null", type!!.typeName)
```
swankjesse, line 2189:
```suggestion
    isPacked && isScalar -> CodeBlock.of("var %N: MutableList<%T>? = null", allocatedName, type!!.typeName)
```

(avoid two kinds of string interpolation in one line)

wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoReader.kt
```diff
@@ -447,6 +447,21 @@ class ProtoReader(private val source: BufferedSource) {
...
+   * Returns the min length of the next field in bytes. Some encodings have a fixed length, while others
+   * have a variable length. LENGTH_DELIMITED fields have a known variable length, while VARINT fields
+   * could be as small as a single byte.
+   */
+  fun nextFieldLengthInBytes(): Long {
+    return when (nextFieldEncoding) {
+      FieldEncoding.LENGTH_DELIMITED -> limit - pos
+      FieldEncoding.FIXED32 -> 4
+      FieldEncoding.FIXED64 -> 8
+      FieldEncoding.VARINT -> 1
```
swankjesse, line 460: It could be as little as 1, but it might be more? The var in varint is variable-length

Fix by renaming the function to contain say, _minimum_ or something?

swankjesse (approved):
Love it

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
JGulbronson: Yeah I was surprised, but I guess it makes sense

wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoReader.kt
JGulbronson: Yep you're right, I have it in the comment and then totally forgot to update the name

JGulbronson:
I think everything is addressed, with the exception of @JakeWharton's comment. Not quite sure what the concrete suggestion is there.

PR #2433, opened 2023-04-12
Add option to use primitive arrays for packed scalars

JGulbronson:
This will allow us to generate repeated, packed, scalar fields to use the appropriate Array type. This has huge benefits for the JVM, because it will prevent autoboxing of values. (For example, `float` -> `Float`).

This is still a WIP, I've largely hardcoded things as `FloatArray` while I play around with it, but it's starting to take shape. Unfortunately we can't use generics and avoid the autoboxing, hence the 4 different `*ArrayList` classes.


JGulbronson:
Note that this doesn't prevent autoboxing, because of the use of generics in the ProtoAdapter abstract class

JGulbronson:
Okay this should prevent autoboxing now

JGulbronson:
The multiplatform test failed because of an HTTP error
> Caused by: org.gradle.internal.resource.transport.http.HttpErrorStatusCodeException: Could not HEAD 'https://nodejs.org/dist/v16.13.1/node-v16.13.1-darwin-x64.tar.gz'. Received status code 500 from server: Internal Server Error

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -659,7 +704,11 @@ class KotlinGenerator private constructor(
         when (fieldOrOneOf) {
           is Field -> {
             val fieldName = localNameAllocator[fieldOrOneOf]
-            addStatement("if (%1L != %2N.%1L) return·false", fieldName, otherName)
+            if (fieldOrOneOf.useArray) {
+              addStatement("if (!%1L.contentEquals(%2N.%1L)) return·false", fieldName, otherName)
```
swankjesse, line 708: Good catch

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -1536,12 +1602,23 @@ class KotlinGenerator private constructor(
...
           add(fieldEqualsIdentityBlock(field, fieldName))
         }
-        addStatement(
-          "%L.encodeWithTag(writer, %L, value.%L)",
-          adapterFor(field),
-          field.tag,
-          fieldName
-        )
+        if (field.useArray && reverse) {
+          val encodeArray = MemberName("com.squareup.wire.internal", "encodeArray")
```
swankjesse, line 1606: 👍🏻

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -1620,7 +1697,11 @@ class KotlinGenerator private constructor(
               }
 
             if (fieldOrOneOf.isPacked && fieldOrOneOf.isScalar) {
-              addStatement("%1L = %1L ?: listOf(),", fieldName)
+              if (fieldOrOneOf.useArray) {
+                addStatement("%1L = %1L?.getTruncatedArray() ?: %2L,", fieldName, fieldOrOneOf.emptyPrimitiveArrayForType)
```
swankjesse, line 1701: Should this be just `toArray()` ?

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -2216,8 +2305,14 @@ class KotlinGenerator private constructor(
     val type = type!!
     val baseClass = type.typeName
     return when (encodeMode!!) {
-      EncodeMode.REPEATED,
-      EncodeMode.PACKED -> List::class.asClassName().parameterizedBy(baseClass)
+      EncodeMode.REPEATED -> List::class.asClassName().parameterizedBy(baseClass)
+      EncodeMode.PACKED -> {
+        if (useArray) {
+          primitiveArrayClassForType.asTypeName()
```
swankjesse, line 2311: 👍🏻

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -2235,8 +2330,13 @@ class KotlinGenerator private constructor(
       return when (encodeMode!!) {
         EncodeMode.MAP ->
           Map::class.asTypeName().parameterizedBy(keyType.typeName, valueType.typeName)
-        EncodeMode.REPEATED,
-        EncodeMode.PACKED -> List::class.asClassName().parameterizedBy(type.typeName)
+        EncodeMode.PACKED -> {
+          when {
+            useArray -> primitiveArrayClassForType.asTypeName()
```
swankjesse, line 2335: I think there’s some places where we check `useArray` without checking encodeMode? I think we probably gotta assert elsewhere that you can’t have `useArray` without PACKED, otherwise I think we might generate an inconsistent file?

wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -437,6 +451,226 @@ internal class RepeatedProtoAdapter<E>(
...
+  override fun encodeWithTag(writer: ReverseProtoWriter, tag: Int, value: DoubleArray?) {
+    if (value != null && value.isNotEmpty()) {
+      super.encodeWithTag(writer, tag, value)
+    }
+  }
+
+  override fun encodedSize(value: DoubleArray): Int {
+    var size = 0
+    for (i in 0 until value.size) {
+      size += originalAdapter.encodedSize(value[i])
```
swankjesse, line 480: I don’t know if I prefer this or not:

```
return originalAdapter.encodedSize(0.0) * value.size
```

swankjesse: Possibly not worth the deviation

wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/DoubleArrayList.kt
```diff
@@ -0,0 +1,42 @@
+package com.squareup.wire.internal
```
swankjesse, line 1: Please paste the copyright header from org.jetbrains.kotlin.utils.IntArrayList

wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/FloatArrayList.kt
```diff
@@ -0,0 +1,43 @@
+package com.squareup.wire.internal
```
swankjesse, line 1: same

wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/IntArrayList.kt
```diff
@@ -0,0 +1,42 @@
+package com.squareup.wire.internal
```
swankjesse, line 1: same

wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/Internal.kt
```diff
@@ -223,3 +229,67 @@ fun boxedOneOfKeyFieldName(oneOfName: String, fieldName: String): String {
 fun boxedOneOfKeysFieldName(oneOfName: String): String {
   return "${oneOfName}_keys".uppercase()
 }
+
+fun encodeArray(
```
swankjesse, line 233: Yay

wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/Internal.kt
```diff
@@ -223,3 +229,67 @@ fun boxedOneOfKeyFieldName(oneOfName: String, fieldName: String): String {
...
+    val byteCountBefore = writer.byteCount
+    for (i in (array.size - 1) downTo 0) {
+      protoAdapter.encodePrimitive(writer, array[i])
+    }
+    writer.writeVarint32(writer.byteCount - byteCountBefore)
+    writer.writeTag(tag, FieldEncoding.LENGTH_DELIMITED)
+  }
+}
+
+fun encodeArray(
```
swankjesse, line 249: yay

wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/LongArrayList.kt
```diff
@@ -0,0 +1,42 @@
+package com.squareup.wire.internal
```
swankjesse, line 1: Same

wire-runtime/src/commonTest/kotlin/com/squareup/wire/internal/LongArrayListTest.kt
```diff
@@ -0,0 +1,30 @@
+package com.squareup.wire.internal
```
swankjesse, line 1: Square or JetBrains copyright header on these?

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Field.kt
```diff
@@ -140,6 +144,12 @@ data class Field(
     if (isPacked && !isPackable(linker, type!!)) {
       linker.errors += "packed=true not permitted on $type"
     }
+    if (useArray && !isPacked) {
+      linker.errors += "wire.use_array=true only permitted on packed fields"
```
swankjesse, line 148: Perfect, this is what I was asking for above. Yay!

wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/SchemaTest.kt
```diff
@@ -296,6 +296,44 @@ class SchemaTest {
...
+               |  repeated float c = 3 [wire.use_array=true];
+               |}
+               """.trimMargin()
+        )
+      }
+      fail()
+    } catch (expected: SchemaException) {
+      assertThat(expected).hasMessage(
+        """
+            |wire.use_array=true only permitted on packed fields
```
swankjesse, line 320: SWEEET test!

wire-tests/src/commonTest/proto-kotlin/com/squareup/wire/protos/kotlin/alltypes/AllTypes.kt
```diff
@@ -2491,6 +2671,126 @@ public class AllTypes(
...
+              }
+              array_int64!!.add(com.squareup.wire.ProtoAdapter.INT64.decodePrimitive(reader))
+            }
+            607 -> {
+              if (array_uint64 == null) {
+                val minimumByteSize = 1
+                val initialCapacity = (reader.nextFieldMinLengthInBytes() / minimumByteSize)
+                  .coerceAtMost(Int.MAX_VALUE.toLong())
+                  .toInt()
+                array_uint64 = LongArrayList(initialCapacity)
```
swankjesse, line 2740: Should `LongArrayList` have a constructor that takes a length in bytes?

swankjesse: (that’d reduce the total amount of generated code)

Maybe even a factory function:

```
LongArrayList.forDecoding(minLengthInBytes: Long)
```

swankjesse (approved):
This is wonderful

swankjesse (approved):

PR #2437, opened 2023-04-13
Adds schemaIndex to WireField

### LINKED ISSUE #2432: Wrong constructor parameter order on Kotlin classes
Currently it is not possible to use Wire to generate Kotlin classes and use them with Moshi to decode JSON data.

The error:

> java.lang.IllegalArgumentException: method com.example.proto.Subscriptions.<init> argument 1 has type java.lang.String, got java.lang.Integer
  at java.lang.reflect.Constructor.newInstance0(Native Method)
  at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
  at com.squareup.wire.KotlinConstructorBuilder.build(KotlinConstructorBuilder.kt:112)

If you check this line
https://github.com/square/wire/blob/master/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/KotlinConstructorBuilder.kt#L116

It is clear that it is wrongly assumed that the order of the fields returned by `Class.getDeclaredFields` matches the order of the constructor parameters. But that's actually not true as the order is undefined (and in my case is alphabetical):
Screenshot 2023-04-10 at 23 32 49

While the constructor is as this:
Screenshot 2023-04-10 at 23 33 49

Here is a link to the documentation:

https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getDeclaredFields--
> The elements in the returned array are not sorted and are not in any particular order.

oldergod:
First commit generates everything so check the second commit please!

Wire was relying on `Class.getDeclaredFields()` to guarantee the fields order which it does NOT. The PR adds a new value to `WireField` which lets one sort the declared fields in the order they were set in the constructor.

Fixes #2432

### DIFF
--- a/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
+++ b/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
@@ -1073,20 +1073,25 @@ class KotlinGenerator private constructor(
   ): List<Pair<ParameterSpec, PropertySpec>> {
     val result = mutableListOf<Pair<ParameterSpec, PropertySpec>>()
 
+    var index = 0
+    val fieldsAndOneOfFieldTags = message.fieldsAndOneOfFields.map { it.tag }
+    val printSchemaIndex = fieldsAndOneOfFieldTags != fieldsAndOneOfFieldTags.sorted()
+
     for (fieldOrOneOf in message.fieldsAndFlatOneOfFieldsAndBoxedOneOfs()) {
       when (fieldOrOneOf) {
         is Field -> result.add(
           constructorParameterAndProperty(
-            message,
-            fieldOrOneOf,
-            nameAllocator
+            message = message,
+            field = fieldOrOneOf,
+            nameAllocator = nameAllocator,
+            schemaIndex = if (printSchemaIndex) index++ else null,
           )
         )
         is OneOf -> result.add(
           constructorParameterAndProperty(
-            message,
-            fieldOrOneOf,
-            nameAllocator
+            message = message,
+            oneOf = fieldOrOneOf,
+            nameAllocator = nameAllocator,
           )
         )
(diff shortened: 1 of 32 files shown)


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -1022,20 +1022,22 @@ class KotlinGenerator private constructor(
...
       when (fieldOrOneOf) {
         is Field -> result.add(
           constructorParameterAndProperty(
-            message,
-            fieldOrOneOf,
-            nameAllocator
+            message = message,
+            field = fieldOrOneOf,
+            nameAllocator = nameAllocator,
+            constructorParameterIndex = index++,
```
Egorand, line 1033: Does `message.fieldsAndFlatOneOfFieldsAndBoxedOneOfs()` return the fields in the order they're declared in the constructor? Or maybe I'm misunderstanding the approach?

wire-runtime/src/commonMain/kotlin/com/squareup/wire/WireField.kt
```diff
@@ -62,6 +62,11 @@ annotation class WireField(
    * field is part of a oneof.
    */
   val oneofName: String = "",
+  /**
+   * Index this field in the primary constructor if available. If set to `-1`, this type was
```
Egorand, line 66: nit: Index "of" this field.

Egorand (approved):
Is there a failing test that this change fixes?

swankjesse:
I don’t like how much extra generated code is necessary to fix this bug. Is there another way? Would like to pair on this or something.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
oldergod: No, your understanding is correct.

PR #2439, opened 2023-04-26
Update descriptor.proto

sergei-lapin:
In our project we wanted to use `debug_redact` option but stumbled into its absence in Wire's `descriptor.proto`.  

This is to update `descriptor.proto` to [7dbe742](https://github.com/protocolbuffers/protobuf/blob/7dbe742e6c4bea916c54713e31811412d2076351/src/google/protobuf/descriptor.proto) version.


oldergod:
Sweet, thank you

PR #2443, opened 2023-04-28
Change how we declare functions to encode and decode primitive arrays 

swankjesse:


swankjesse:
Note this is not backwards-compatible for anyone using the optimized arrays that were introduced in 4.5.6.

wire-runtime/api/wire-runtime.api
```diff
@@ -32,22 +32,6 @@ public final class com/squareup/wire/AnyMessage$Companion {
 	public final fun pack (Lcom/squareup/wire/Message;)Lcom/squareup/wire/AnyMessage;
 }
 
-public final class com/squareup/wire/DoubleProtoAdapter : com/squareup/wire/ProtoAdapter {
```
oldergod, line 35: Is this like picasso where CI would fail if we don't manually update ourselves those files?

oldergod (approved):

wire-runtime/api/wire-runtime.api
oldergod: pretty sweet https://github.com/Kotlin/binary-compatibility-validator#tasks

swankjesse: Exactly!

PR #2455, opened 2023-05-09
Swift: Heap types should use DynamicMemberLookup for storage

lickel:
This removes a ton of symbols while not reducing type safety at all.
Xcode will also correctly handle things like code completion for you.


wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -272,12 +280,15 @@ class SwiftGenerator private constructor(
                   storageParams += CodeBlock.of("%1N: %1N", oneOf.name)
                 }
                 addStatement(
-                  "self.storage = %T(%L)", storageType,
```
lickel, line 275: This was a latent bug

wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -399,6 +399,7 @@ data class SwiftTarget(
         val modulePath = context.outDirectory
         val typeName = generator.generatedTypeName(type)
         val swiftFile = SwiftFileSpec.builder(typeName.moduleName, typeName.simpleName)
+          .addImport("Foundation")
```
lickel, line 402: This was needed because of the similar SwiftPoet 1.5 bug around extensions.

Because `init` doesn't explicitly use `Foundation.Data`, it was not getting _implicitly_ imported.
The extension then needs `Foundation.Data` in order to define `unknownFields` which was never imported.

lickel: If we move to a world where the initializer is minimal, we would run into the same importing problem.

PR #2456, opened 2023-05-10
fix: Produce Descriptors for messages without fields after pruning

jvmakine:
Empty messages are converted to EnclosingType at pruning. These were previously ignored at SchemaEncoder. This adds encoding for nested types in EnclosingTypes.


chris-ryan-square (approved):

swankjesse (approved):

swankjesse:
Nice fix@

PR #2457, opened 2023-05-11
Explicitely adds jvm variants to the BOM

oldergod:


Egorand (approved):
Do they not get added automatically?

Issue #2440, opened 2023-04-26
Gradle build is broken since 4.5.3 when using wire.protoLibrary = true


sergei-lapin:
Since 4.5.3 Gradle modules configured to be proto libs via `wire.protoLibrary = true` are destined to raise warning (older Gradle versions) or fail build (newer Gradle versions) due to the undeclared task dependencies:

```
Reason: Task ':processResources' uses this output of task ':generateMainProtos' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
```

[Here](https://github.com/sergei-lapin/issue-reproducers/tree/wire-gradle-issues) you can find a MRE (commit f97d92c)

This is reproducible with Gradle at least 7.6.1 (tested with 7.6.1 and 8.1.1)

oldergod:
What's the step to see the reason showing somewhere?

sergei-lapin:
closing as it was resolved with the release of 4.7.0

PR #2462, opened 2023-05-17
Explicitly add dependency on processResources

oldergod:
I'm not able to repro in test so that sucks.


Egorand (approved):

sergei-lapin:
should this fix #2440 ?

PR #2464, opened 2023-05-19
If a valid grpc-status header is present, raise a GrpcException rathe…

jpulgarin:
…r than an IOException, even when we see a transport error

Right now, when Wire encounters a response with an empty message, but with a valid, non-0 `grpc-status` header, it raises an `IOException`. This doesn't seem right to me: I'd like my Wire client to let me know that the server reported an error even if it didn't include a message in the response.

This change makes it so that in `grpcResponseToException` we check for a valid, non-0 `grpc-status` header, and raise a `GrpcException` in response to it, before checking for a `transportException`.


swankjesse (approved):

swankjesse:
Nice usability improvement!

PR #2486, opened 2023-06-19
Add support for custom logger

oldergod:


Egorand (approved):

swankjesse (approved):
Having to create these via reflection feels bad. It’s kind of an awkward API for a CLI to have? I don‘t have much for better ideas.

oldergod:
The API is symmetric to what we built for custom SchemaHandler so I'm good with it.

PR #2489, opened 2023-06-22
Dump internal APIs as well so we know when we break them

oldergod:
Although internal they are public. It'd be good for us to know when they break so we have confidence in breaking them!


Egorand (approved):
Good call, good to have visibility!

PR #2490, opened 2023-06-23
Introduce Wire's Visitor

oldergod:
There was discussions around getting WireLogger to become somewhat a listener but it's an interface which would break the world every time we'd wanna make a change, and we can only set one.

I would like something like an EventListener but which would also be able to log, do whatever it wants, even throwing if something doesn't please it. One use case we have would be to force pruning rules exhaustion and throw if there are unused roots for instance.

I'm open to suggestions if something seems off on the software design level.

### DIFF
--- a/wire-compiler/api/wire-compiler.api
+++ b/wire-compiler/api/wire-compiler.api
@@ -194,11 +194,12 @@ public final class com/squareup/wire/schema/TargetKt {
 }
 
 public final class com/squareup/wire/schema/WireRun {
-	public fun <init> (Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/util/List;Ljava/util/Map;Z)V
-	public synthetic fun <init> (Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/util/List;Ljava/util/Map;ZILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/util/List;Ljava/util/Map;ZLjava/util/List;)V
+	public synthetic fun <init> (Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/util/List;Ljava/util/Map;ZLjava/util/List;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun component1 ()Ljava/util/List;
 	public final fun component10 ()Ljava/util/Map;
 	public final fun component11 ()Z
+	public final fun component12 ()Ljava/util/List;
 	public final fun component2 ()Ljava/util/List;
 	public final fun component3 ()Ljava/util/List;
 	public final fun component4 ()Ljava/util/List;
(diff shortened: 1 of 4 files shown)


wire-compiler/api/wire-compiler.api
```diff
@@ -207,11 +208,12 @@ public final class com/squareup/wire/schema/WireRun {
 	public final fun component7 ()Ljava/lang/String;
 	public final fun component8 ()Ljava/lang/String;
 	public final fun component9 ()Ljava/util/List;
-	public final fun copy (Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/util/List;Ljava/util/Map;Z)Lcom/squareup/wire/schema/WireRun;
-	public static synthetic fun copy$default (Lcom/squareup/wire/schema/WireRun;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/util/List;Ljava/util/Map;ZILjava/lang/Object;)Lcom/squareup/wire/schema/WireRun;
+	public final fun copy (Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/util/List;Ljava/util/Map;ZLjava/util/List;)Lcom/squareup/wire/schema/WireRun;
```
oldergod, line 211: Why is `WireRun` a data class anyway???

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Auditor.kt
```diff
@@ -0,0 +1,100 @@
...
+ *  - moveTypesStart
+ *  - moveTypesEnd
+ *  - schemaHandlersStart // Looping over all handlers.
+ *    - schemaHandlerStart
+ *    - schemaHandlerEnd
+ *  - schemaHandlersEnd
+ *  - validate
+ *  - end
+ */
+abstract class Auditor {
```
Egorand, line 24: Should we maybe call it "interceptor"? It's a common pattern, particularly in Square OS libraries. Don't hate "auditor", just feel like using a common name for a common pattern will aid comprehension.

Egorand, line 24: Any advantages/downsides over using an interface with default method implementations?

Egorand (approved):

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Auditor.kt
JakeWharton: "EventListener" is also used in three or four libraries. I don't think this is an "interceptor" because you're not given an opportunity to actually intercept anything, just observe.

wire-compiler/api/wire-compiler.api
JakeWharton: Get rid of it! No `data class`es in public APIs!

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Auditor.kt
oldergod: I didn't wanna pick `EventListener` because I wanna convey it doesn't only listen, they can make the work throw with the `validate()` method.

oldergod: I truly don't know

JakeWharton: Hmmm I see. There's "visitor" or "callback" as other common nouns for things which are similar in nature. Not sure what else. I think "auditor" only stands out because it's not a term that you come across frequently and have intuition about.

oldergod: I didn't know about Visitor but I like it.

oldergod:
updated to Visitor

swankjesse:
Discussed offline. My recommendation is to make the validation features built-in, and to rename this to EventListener.

PR #2491, opened 2023-06-23
[swift] Always fully-qualify Foundation.Data to prevent name collisions

amorde:
If a group of protos has a message named `Data`, swiftpoet will only automatically qualify references to `Foundation.Data` from within the implementation of the `Data` message. Any other files in the group will have an unqualified reference to `Data` in their implementations which will default to the custom message type instead of the Foundation type.


wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -82,6 +82,9 @@ class SwiftGenerator private constructor(
   private val encoder = DeclaredTypeName.typeName("Swift.Encoder")
   private val writableKeyPath = DeclaredTypeName.typeName("Swift.WritableKeyPath")
 
+  // Always fully-qualified in case a message type is named `Data`
+  private val data = DeclaredTypeName.typeName("Foundation.Data", true)
```
amorde, line 86: I initially tried searching through the `schema` to see if one of the defined messages was named `Data`, but I couldn't quite figure that out

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -82,6 +82,9 @@ class SwiftGenerator private constructor(
   private val encoder = DeclaredTypeName.typeName("Swift.Encoder")
   private val writableKeyPath = DeclaredTypeName.typeName("Swift.WritableKeyPath")
 
+  // Always fully-qualified in case a message type is named `Data`
```
oldergod, line 85: https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

```suggestion
  // Always fully-qualified in case a message type is named `Data`.
```

oldergod (commented):
Is there a test you could write or code you could generate which would compile fail if this fix wasn't in?

amorde:
> Is there a test you could write or code you could generate which would compile fail if this fix wasn't in?

Yes - all you need is a message type named `Data` generated in the same module as another message type. I've tried to create a test locally but I'm having some trouble getting the tests to work (Kotlin / Java noob here 😅 )

wire-runtime-swift/src/main/swift/Codable/StringEncoded+Data.swift
```diff
@@ -16,26 +16,26 @@
 
 import Foundation
 
-extension Data : StringCodable {
+extension Foundation.Data : StringCodable {
```
lickel, line 19: We shouldn't need to modify anything in `wire-runtime-swift/src/main/swift`

wire-runtime-swift/src/test/swift/CodableTests.swift
```diff
@@ -52,7 +52,7 @@ extension CodableTests {
             opt_uint64: 4,
             opt_float: 5,
             opt_double: 6,
-            opt_bytes: Data(hexEncoded: "0123456"),
+            opt_bytes: Foundation.Data(hexEncoded: "0123456"),
```
lickel, line 55: Would it make sense to instead make a few type aliases?

```swift
typealias Bytes = Data // Wire.Bytes
typealias UnknownFields = Data // Wire.UnknownFields
```

Basically to encourage semantic naming

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -930,9 +935,21 @@ class SwiftGenerator private constructor(
     includeDefaults: Boolean = true
   ) = apply {
     type.fields.forEach { field ->
-      val fieldType = field.typeName
+      var typeName = when (field.type) {
+        ProtoType.BYTES -> if (field.isMap) {
+          dataMap
```
lickel, line 940: This seems wrong?
I would definitely want test cases for things like `[String: Data]`, `[Data: String]` (if that's possible), etc

dnkoutso: makes sense

wire-runtime-swift/src/test/swift/CodableTests.swift
dnkoutso: will update

efirestone: Doesn't that introduce the same problem, but with a message named `Bytes` (and/or `UnknownFields`)?

lickel: It's for sure a similar problem. The UnknownFields PR is explicitly `Wire.UnknownFields`.

I would guess that `message Bytes` wouldn't happen due to there being a `bytes` type, but we could also be explicitly Wire.Bytes.

My thinking was that Data was doing double-duty for two very different use cases.
I'm also 100% OK with using `Foundation.Data` for `bytes`.

oldergod:
What's missing?

lickel:
@oldergod My two unresolved comments are still valid.

dnkoutso:
Most definitely this needs to be resolved https://github.com/square/wire/pull/2491#discussion_r1244447286

wire-runtime-swift/src/main/swift/Codable/StringEncoded+Data.swift
dnkoutso: @lickel not sure why I recall seeing an error. Either way this is reverted and the build still passes.

wire-runtime-swift/src/test/proto/person.proto
```diff
@@ -37,4 +41,9 @@ message Person {
 
   repeated string aliases = 5;
   repeated int32 ids = 6;
+
+  required Data data = 7;
+
+  map<string, Data> person_to_data = 8;
+  map<string, bytes> person_to_byte_data = 9;
```
dnkoutso, line 48: @alickel added this as a test to showcase generated code being correct.

dnkoutso: generated code for tests (not committed so you cant see the diff here)

```swift
public struct Person {

    public var name: String
    public var id: Int32
    public var email: String?
    public var phone: [Person.PhoneNumber]
    public var aliases: [String]
    public var ids: [Int32]
    public var data: Data
    public var person_to_data: [String : Data]
    public var person_to_byte_data: [String : Foundation.Data]
    public var unknownFields: Foundation.Data = .init()
...
```

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
dnkoutso: done should be fixed now.

dnkoutso:
This is now ready.

dnkoutso:
@oldergod not sure whats wrong with the multiplatform CI job.

oldergod:
Many unfortunate flakiness. If one of either (push or pull_request) passes, good to go.

dnkoutso:
All passed!

oldergod (approved):

wire-runtime-swift/src/test/swift/ProtoReaderTests.swift
```diff
@@ -105,48 +105,54 @@ final class ProtoReaderTests: XCTestCase {
...
     func testDecodeMessage() throws {
-        let data = Data(hexEncoded: """
+        let data = Foundation.Data(hexEncoded: """
             0A       // (tag 1 | Length Delimited)
             04       // Length of name
             4C756B65 // Name value "Luke"
             10       // (tag 2 | Varint)
             05       // ID value 5
+            3A       // (tag 7 | Data)
+              04       // Length 4 for Data
```
lickel, line 135: - [ ] Nit: weird alignment

wire-runtime-swift/src/test/swift/ProtoWriterTests.swift
```diff
@@ -227,17 +227,23 @@ final class ProtoWriterTests: XCTestCase {
...
-            08       // Length 8
+            0E       // Length 10 for Person
               0A       // (Tag 1 | Length Delimited)
               04       // Length 4 for name
-              4C756B65 // Value "Luke"
+              4C756B65 // UTF-8 Value "Luke"
               10       // (Tag 2 | Varint)
               05       // Value 5
+              3A       // (Tag 7 | Data)
+                04       // Length 4 for Data
```
lickel, line 243: - [ ] Nit: alignment

wire-runtime-swift/src/test/swift/ProtoWriterTests.swift
```diff
@@ -567,6 +574,11 @@ final class ProtoWriterTests: XCTestCase {
             02           // repeated id 2
             30           // (Tag 6 | Varint)
             03           // repeated id 3
+            3A           // (tag 7 | Data)
+              04           // Length 4 for Data
+              0A           // (Tag 1 | Length Delimited)
```
lickel, line 579: - [ ] nit: alignment

lickel (approved):

wire-runtime-swift/src/test/swift/ProtoReaderTests.swift
dnkoutso: I do not think it is due to it being a new message type nested?

PR #2493, opened 2023-06-28
[swift] Create `UnknownFields` typealias.

dnkoutso:
This is the initial fix that @amorde had done here https://github.com/square/wire/pull/2491.

If we add `message Data` in our `.proto` file it exposes a second issue where a proto defines a message `Data` and a field of type `bytes` exists within the same module.

For now this is an incremental fix that we can take and then proceed further with landing https://github.com/square/wire/pull/2491

### DIFF
--- a/wire-runtime-swift/src/main/swift/UnknownFields.swift
+++ b/wire-runtime-swift/src/main/swift/UnknownFields.swift
@@ -1,0 +1,20 @@
+/*
+ * Copyright 2023 Square Inc.
+ *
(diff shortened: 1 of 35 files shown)


dnkoutso:
Confirmed internally this fixes our issue.

lickel (approved):

dnkoutso:
Upon more discussion we prefer with the overall approach to just emit `Foundation.Data` to fix this issue.

Closing this PR for now and will focus to land https://github.com/square/wire/pull/2491 instead.

PR #2494, opened 2023-06-29
WireRun not to be a data class anymore

oldergod:


wire-compiler/api/wire-compiler.api
```diff
@@ -196,21 +196,6 @@ public final class com/squareup/wire/schema/TargetKt {
 public final class com/squareup/wire/schema/WireRun {
 	public fun <init> (Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/util/List;Ljava/util/Map;ZLjava/util/List;)V
 	public synthetic fun <init> (Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/util/List;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/util/List;Ljava/util/Map;ZLjava/util/List;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
-	public final fun component1 ()Ljava/util/List;
```
Egorand, line 199: Are we fine with breaking compatibility?

Egorand (approved):

wire-compiler/api/wire-compiler.api
oldergod: yes! Anybody using WireRun as a data class doesn't deserve attention!

JakeWharton (approved):
You can use Poko if you want to retain the generated equals/hashCode/toString without the other bad functions.

PR #2496, opened 2023-06-29
Pass target payload to custom schemahandler

oldergod:
I'm not sure why I didn't do it from the get-go.... This would break Java implementers (we have one in `squareup`) but I think we have to do it.
Right now something like
```
custom {
  includes = xxxx
  exclusive = xxx
}
```
would all be lost.

I wonder if I could make a `Any` like argument (maybe a String) for the schema handler to parse it as they wish and be able to set payloads. Our java/kotlin targets needed some so we can assume new targets will want something. => https://github.com/square/wire/pull/2497

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
+++ b/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
@@ -372,7 +372,12 @@ data class CustomTarget(
   }
 
   override fun newHandler(): SchemaHandler {
-    return schemaHandlerFactory.create()
+    return schemaHandlerFactory.create(
+      includes = includes,
+      excludes = excludes,
+      exclusive = exclusive,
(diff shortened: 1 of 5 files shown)


wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -372,7 +372,12 @@ data class CustomTarget(
   }
 
   override fun newHandler(): SchemaHandler {
-    return schemaHandlerFactory.create()
+    return schemaHandlerFactory.create(
+      includes = includes,
+      excludes = excludes,
+      exclusive = exclusive,
+      outDirectory = outDirectory,
```
oldergod, line 379: I wonder if I need to use reflexion here, check if the method exist before executing it.

swankjesse (approved):

PR #2500, opened 2023-06-30
Promoting WireRun to wire-schema

oldergod:
So it'll be passable to the EventListener.

Basically moving WireRun and the interface Target (which is not a sealed class anymore) to wire-schema, and some collaterals.


wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SchemaLoader.kt
```diff
@@ -15,13 +15,14 @@
...
 
+import okio.FileSystem
 import okio.IOException
 
 /**
  * Load proto files and their transitive dependencies and parse them. Keep track of which files were
  * loaded from where so that we can use that information later when deciding what to generate.
  */
-expect class SchemaLoader : Loader, ProfileLoader {
+expect class SchemaLoader(fileSystem: FileSystem) : Loader, ProfileLoader {
```
oldergod, line 25: otherwise I cannot call it from `wire-schema/commonMain`

swankjesse (approved):

PR #2505, opened 2023-07-03
Open GrpcClient

### LINKED ISSUE #2484: Ability to use custom gRPC clients on Multiplatform
Hello!

I am currently developing a client for a service which uses Protobuf-over-WebSocket implementation instead of traditional HTTP requests, but to avoid boilerplate, I decided to use Wire's gRPC bindings.

The protobufs provide RPC information, so I can generate them, however:
- `CLIENT` role has the needed ADAPTER's in the method bindings, however it heavily depends on OkHttp/Wire gRPC runtime:
```kotlin
import com.squareup.wire.GrpcCall
import com.squareup.wire.GrpcClient
import com.squareup.wire.GrpcMethod

public class GrpcCommunity(
  private val client: GrpcClient,
) : Community {
  public override fun ClearSinglePartnerEventsAppPriority():
      GrpcCall<CCommunity_ClearSinglePartnerEventsAppPriority_Request, CCommunity_ClearSinglePartnerEventsAppPriority_Response>
      = client.newCall(GrpcMethod(
      path = "/Community/ClearSinglePartnerEventsAppPriority",
      requestAdapter = CCommunity_ClearSinglePartnerEventsAppPriority_Request.ADAPTER,
      responseAdapter = CCommunity_ClearSinglePartnerEventsAppPriority_Response.ADAPTER
  ))
```
Here, `CCommunity_ClearSinglePartnerEventsAppPriority_Request.ADAPTER` and `CCommunity_ClearSinglePartnerEventsAppPriority_Response.ADAPTER` are binded, removing the boilerplate in client implementation. It is a preferred approach, because everything is settled, but it has a direct dependency to `GrpcClient` which is not `open`, making custom implementations impossible.

- `SERVER` role does not provide ADAPTER's except for package name in `WireRpc`:
```kotlin
@WireRpc(
  path = "/CloudConfigStore/Download",
  requestAdapter = "steam.webui.cloudconfigstore.CCloudConfigStore_Download_Request#ADAPTER",
  responseAdapter = "steam.webui.cloudconfigstore.CCloudConfigStore_Download_Response#ADAPTER",
  sourceFile = "webui/service_cloudconfigstore.proto",
)
public suspend fun Download(request: CCloudConfigStore_Download_Request): CCloudConfigStore_Download_Response
```
Here, `requestAdapter` and `responseAdapter` are defined as package name references - which requires reflection on JVM and is not available at all in K/Native.

I've tried developing a custom workaround implementation for `GrpcClient`, but this will lead to conflicts and other errors if a library's user decides to import Wire's gRPC client for it's own needs.

Ideally, this could be an ability to provide custom imports compatible with `GrpcClient`, maybe something like an shared interface, using a configuration variable:
```kotlin
wire {
  kotlin {
    rpcRole = "client"
    rpcClientName = "com.example.MyGrpcClient"
  }
}
```

This issue is somewhat similar to #1962, but request/response adapter reflection was not mentioned in that issue.

Thanks!

oldergod:
So that people can inject their own implementation. Fixes #2484
Would that be a bad idea to do so?

For info, Wire currently generates something like this for an RPC:

```kotlin
public class GrpcWhiteboardClient(
  private val client: GrpcClient,
) : WhiteboardClient {
  override fun Whiteboard(): GrpcStreamingCall<WhiteboardCommand, WhiteboardUpdate> =
      client.newStreamingCall(GrpcMethod(
      path = "/com.squareup.wire.whiteboard.Whiteboard/Whiteboard",
      requestAdapter = WhiteboardCommand.ADAPTER,
      responseAdapter = WhiteboardUpdate.ADAPTER
  ))
}
```

### DIFF
--- a/wire-grpc-client/api/wire-grpc-client.api
+++ b/wire-grpc-client/api/wire-grpc-client.api
@@ -23,12 +23,12 @@ public final class com/squareup/wire/GrpcCalls {
 	public static final fun grpcStreamingCall (Lkotlin/jvm/functions/Function3;)Lcom/squareup/wire/GrpcStreamingCall;
 }
 
-public final class com/squareup/wire/GrpcClient {
+public abstract class com/squareup/wire/GrpcClient {
 	public synthetic fun <init> (Lokhttp3/Call$Factory;Lokhttp3/HttpUrl;JLkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun create (Lkotlin/reflect/KClass;)Lcom/squareup/wire/Service;
 	public final fun newBuilder ()Lcom/squareup/wire/GrpcClient$Builder;
-	public final fun newCall (Lcom/squareup/wire/GrpcMethod;)Lcom/squareup/wire/GrpcCall;
-	public final fun newStreamingCall (Lcom/squareup/wire/GrpcMethod;)Lcom/squareup/wire/GrpcStreamingCall;
+	public abstract fun newCall (Lcom/squareup/wire/GrpcMethod;)Lcom/squareup/wire/GrpcCall;
+	public abstract fun newStreamingCall (Lcom/squareup/wire/GrpcMethod;)Lcom/squareup/wire/GrpcStreamingCall;
 }
 
 public final class com/squareup/wire/GrpcClient$Builder {
--- a/wire-grpc-client/src/commonMain/kotlin/com/squareup/wire/GrpcClient.kt
+++ b/wire-grpc-client/src/commonMain/kotlin/com/squareup/wire/GrpcClient.kt
@@ -15,7 +15,7 @@
  */
 package com.squareup.wire
 
-expect class GrpcClient {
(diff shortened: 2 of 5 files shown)


JakeWharton:
Abstract class seems like a better idea for this with a private implementation on the JVM. The `newBuilder()` API on the JVM is unfortunate, though, as it couples you to OkHttp with the `Builder`. But I guess an alternate implementation could throw unsupported if one really wanted to exist.

oldergod:
Jake, I updated the code. Is it what you had in mind?

JakeWharton:
For the class, yes. Exactly. And then I think you can make `newCall` and `newStreamingCall` be `abstract` as well and then you don't need the throwing implementations. That way it behaves basically like an interface for the other platforms.

oldergod:
Got it, that's great to get rid of the throwing implementation indeed. Updated.

JakeWharton (approved):
Perfecto!

oldergod:
Jake, did we actually prevent people from adding their own JVM implementation? I don't see how they can do it now that there is an `actual GrpcClient private constructor(`. Did I miss something?

PR #2506, opened 2023-07-03
fix: support Unit values in grpc-server-generator

jvmakine:
Adds a special case for returning `Empty` from a grpc call in the grpc service compatibility generator. 

We map `Empty` to `kotlin.Unit` which does not have `ADAPTER` and previously resulted in invalid generated code.
Here we return a null stream as the InputStream instead, and when parsing a stream, always return `Unit` immediately. 

Also fixes similar issue with returning timestamps, durations, structs, and some primitive value messages.

### DIFF
--- a/wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/ImplBaseGenerator.kt
+++ b/wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/ImplBaseGenerator.kt
@@ -26,6 +26,7 @@ import com.squareup.kotlinpoet.PropertySpec
 import com.squareup.kotlinpoet.TypeSpec
 import com.squareup.kotlinpoet.UNIT
 import com.squareup.kotlinpoet.asClassName
+import com.squareup.wire.schema.ProtoType
 import com.squareup.wire.schema.Rpc
 import com.squareup.wire.schema.Service
 import java.io.InputStream
@@ -150,7 +151,7 @@ object ImplBaseGenerator {
                 .addModifiers(KModifier.OVERRIDE)
                 .addParameter(ParameterSpec(name = "value", type = className))
                 .returns(InputStream::class)
-                .addCode(CodeBlock.of("return %T.ADAPTER.encode(value).inputStream()", className))
+                .addCode(streamCodeFor(it!!, className))
                 .build()
             )
             .addFunction(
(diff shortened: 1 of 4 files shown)


oldergod:
We actually have an adapter for this one https://github.com/square/wire/blob/master/wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoAdapter.kt#L1271-L1288
Could you reuse it?

oldergod:
Defined for the JVM here: https://github.com/square/wire/blob/master/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/ProtoAdapter.kt#L318

jvmakine:
@oldergod Updated, and also added adapters for other possible message types. Would appreciate another look 🙇

wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/ImplBaseGenerator.kt
```diff
@@ -26,6 +26,8 @@ import com.squareup.kotlinpoet.PropertySpec
 import com.squareup.kotlinpoet.TypeSpec
 import com.squareup.kotlinpoet.UNIT
 import com.squareup.kotlinpoet.asClassName
+import com.squareup.wire.*
```
oldergod, line 29: no wildcard please

wire-grpc-server-generator/src/test/golden/unitService.kt
```diff
@@ -0,0 +1,146 @@
+import com.google.protobuf.DescriptorProtos
```
oldergod, line 1: copyright

oldergod (approved):

wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/ImplBaseGenerator.kt
jvmakine: Ah, a new laptop, I need to update my language settings 🤦

wire-grpc-server-generator/src/test/golden/unitService.kt
jvmakine: This is a directory of "golden" generated files to compare test runs against, so can't really add a copyright here without adding extra logic to strip it out in the tests. However, I assume it is OK, as this is generated code, and non of the other files here have the copyright either?

oldergod: I see.
Maybe in the future, we could add the headers we add elsewhere to make it clear.

An example:
```
// Source: squareup.wire.benchmarks.proto2.AllTypes in squareup/wire/benchmarks/all_types_proto2.proto
```

jvmakine: Yeah, that would be good. I might see if I can open a second PR for that

PR #2508, opened 2023-07-03
chore: add generated code comment to the golden test files under the grpc-server adapter tests

jvmakine:
The golden test files are generated before the standard generated code comment is added to the file. (It is included correctly in the final generated code)

This adds a comment saying that the code is generated code for tests, and cleans up the tests a bit as well.


wire-grpc-server-generator/src/test/java/com/squareup/wire/kotlin/grpcserver/GoldenTestUtils.kt
```diff
@@ -0,0 +1,19 @@
+package com.squareup.wire.kotlin.grpcserver
```
oldergod, line 1: Copyright here as well

oldergod (approved):

PR #2509, opened 2023-07-03
Add options to OneOf

### LINKED ISSUE #2503: OneOf field is missing option value
Protobuf3 now support option in oneOf field. However, in Wire it doesn't have option in the OneOf field

Example:
```

extend google.protobuf.OneofOptions {
  optional int32 my_option = 12345;
}

message Object{
  oneof oneofField {
    option (my_option) = 54321;
    string oneOfOne = 102;
    int32 oneOfTwo = 103;
  }
}

```

In Wire [OneOf](https://github.com/square/wire/blob/master/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/OneOf.kt) class doesn't have Options.

oldergod:
And thus allow the reading of its values.

There are few breaking APIs but it's kinda low-level. I wonder if many people instantiate `OneOfElement` or `OneOf` manually.

Fixes #2503

### DIFF
--- a/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/RootTest.kt
+++ b/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/RootTest.kt
@@ -34,7 +34,7 @@ class RootTest {
 
     val location = Location.get("sample/src/main/proto", "squareup/dinosaurs/dinosaur.proto")
     val roots = location.roots(fs)
-    assertThat(roots.size == 1)
+    assertThat(roots.size == 1).isTrue()
 
     // Standalone files resolve because we have a base directory.
     assertThat(roots[0].resolve("squareup/dinosaurs/dinosaur.proto")).isEqualTo(roots[0])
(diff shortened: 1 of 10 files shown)


wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/RootTest.kt
```diff
@@ -34,7 +34,7 @@ class RootTest {
 
     val location = Location.get("sample/src/main/proto", "squareup/dinosaurs/dinosaur.proto")
     val roots = location.roots(fs)
-    assertThat(roots.size == 1)
+    assertThat(roots.size == 1).isTrue()
```
Egorand, line 37: Can we do

```suggestion
    assertThat(roots.size).isEqualTo(1)
```

instead? The failure message will be much more descriptive (e.g. "expected <1> but was <5>" instead of "expected true but was false").

Egorand (approved):

wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/RootTest.kt
oldergod: Probably yes. This was an old test that I fixed because it was not testing *anything*

PR #2515, opened 2023-07-05
Detect and fail on type name collisions

### LINKED ISSUE #2124: Detect and fail on type name collisions
In #2122 we discovered that Wire was loading the same `.proto` file twice via two different paths.

I consider this Schema to be broken because some files and types were repeated. But this error was very clumsy because it was failing on duplicated enum constants, not duplicated paths or types.

We should detect duplicated type names and fail the schema load.

oldergod:
Fix #2124


Egorand (approved):
Nice!

PR #2516, opened 2023-07-05
Add serialVersionUID to AndroidMessage

### LINKED ISSUE #2247: AndroidMessage do not generate a serialVersionUID
AndroidMessage do not generate a serialVersionUID

oldergod:
Message defines it but it looks like even if the parent defines it, it will not be used by the serialization process. So we're adding one. Fixes #2247

### DIFF
--- a/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/AndroidMessage.kt
+++ b/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/AndroidMessage.kt
@@ -42,6 +42,8 @@ abstract class AndroidMessage<M : Message<M, B>, B : Message.Builder<M, B>> prot
   }
 
   companion object {
+    private const val serialVersionUID = 0L
+
     /** Creates a new [Parcelable.Creator] using `adapter` for serialization. */
     @JvmStatic fun <E> newCreator(adapter: ProtoAdapter<E>): Parcelable.Creator<E> {
       return ProtoAdapterCreator(adapter)


wire-runtime/src/jvmMain/kotlin/com/squareup/wire/AndroidMessage.kt
```diff
@@ -42,6 +42,8 @@ abstract class AndroidMessage<M : Message<M, B>, B : Message.Builder<M, B>> prot
   }
 
   companion object {
+    private const val serialVersionUID = 0L
```
Egorand, line 45: We should probably generate this field for every type instead of declaring it in the superclass? According to the [docs](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/Serializable.html), "It is also strongly advised that explicit serialVersionUID declarations use the private modifier where possible, since such declarations apply only to the immediately declaring class--serialVersionUID fields are not useful as inherited members".

Also, we probably need to explicitly mark it as `@JvmStatic`?

oldergod: When abstract, I think the implementing type is good, no?

Egorand: Can't see anything mentioning "abstract" in the docs, so not sure!

oldergod: Interesting.....

https://pl.kotl.in/fO_za757Y

It looks like it doesn't matter that it is set in Message or AndroidMessage. It has to be there in the final class.
Maybe we could add it and set its value to Wire's version.

Do you have experience with this @JakeWharton @swankjesse ?

oldergod: ok we actually do generate them for both java and kotlin
https://github.com/square/wire/blob/master/wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java#L697
https://github.com/square/wire/blob/master/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt#L595

oldergod: We could eventually change the version and make it match Wire's but don't know if that's a valid concern.

PR #2521, opened 2023-07-06
Bump to Gradle 7.6.2

oldergod:


Egorand (approved):
Why not 8.2?

oldergod:
> Why not 8.2?

I tried and I'm getting an avalanche of weird errors I don't know how to fix. So, I'm trying to update to latest 7.x, fix all warnings and then try 8.2

PR #2525, opened 2023-07-07
WireSchemaTests as MPP module

oldergod:


Egorand (approved):

wire-schema-tests/build.gradle.kts
```diff
@@ -15,20 +13,112 @@ if (project.rootProject.name == "wire") {
...
-  implementation(libs.okio.core)
-  implementation(libs.okio.fakefilesystem)
-  testImplementation(libs.assertj)
-  testImplementation(libs.kotlin.test.junit)
-  testImplementation(projects.wireTestUtils)
-}
+kotlin {
+  jvm {
+    withJava()
+  }
```
JakeWharton, line 19:
```suggestion
  jvm()
```
I don't see any Java sources in here, right?

JakeWharton (approved):

wire-schema-tests/build.gradle.kts
oldergod: Looks like `MavenPublishBaseExtension` fails without it with `SoftwareComponentInternal with name 'java' not found`

PR #2528, opened 2023-07-07
Introduce Spotless for the whole repo Kotlin & Java

oldergod:


build.gradle.kts
```diff
@@ -56,15 +56,38 @@ allprojects {
...
+        ktlint(libs.versions.ktlint.get()).editorConfigOverride(
+          mapOf("ktlint_standard_filename" to "disabled"),
+        )
+        trimTrailingWhitespace()
+        endWithNewline()
+        toggleOffOn()
+        // We cannot use it because the licensing job happens before. The header will be replaced and
+        // spotless will throw. We have to manually exclude stuff for now.
+        // See https://github.com/diffplug/spotless/discussions/1738
+        toggleOffOnRegex("""^(// Code generated by Wire protocol buffer compiler[\s\S]*)$""")
```
JakeWharton, line 74: Clever

build.gradle.kts
```diff
@@ -56,15 +56,38 @@ allprojects {
...
+      java {
+        target("**/*.java")
+        googleJavaFormat(libs.googleJavaFormat.get().version)
+        trimTrailingWhitespace()
+        endWithNewline()
+        toggleOffOn()
+        // We cannot use it because the licensing job happens before. The header will be replaced and
+        // spotless will throw. We have to manually exclude stuff for now.
+        // See https://github.com/diffplug/spotless/discussions/1738
+        // toggleOffOnRegex("""^(// Code generated by Wire protocol buffer compiler[\s\S]*)$""")
```
JakeWharton, line 87: Why keep if commented out?

wire-bom/build.gradle.kts
```diff
@@ -22,6 +22,7 @@ dependencies {
     api(projects.wireSchema)
     api(projects.wireSchema.group + ":wire-schema-jvm:" + projects.wireSchema.version)
     api(projects.wireSchemaTests)
+    api(projects.wireSchemaTests.group + ":wire-schema-jvm:" + projects.wireSchemaTests.version)
```
JakeWharton, line 25: Isn't this from another PR?

JakeWharton (approved):
Good enough

Screen Shot 2023-07-07 at 3 15 17 PM

wire-bom/build.gradle.kts
oldergod: fixed

build.gradle.kts
oldergod: oops

build.gradle.kts
oldergod: Would have been if it worked!

oldergod: https://github.com/diffplug/spotless/pull/1749

JakeWharton: Did you consider using excludes by path? Surely the generated code are all under individual folders and not mixed in with normal sources, right?

oldergod: Exclude by path already exist, right? And I'm actually using it because the thing I tried above didn't work. I'd want something to set in one place and never touch again. I don't know how Spotless will take my PR but that seems a better solution to me that manually adding paths

PR #2532, opened 2023-07-12
Don't use default configuration for project dependencies

oldergod:
There's another place I'm doing this and this needs addressing because Gradle 8 will throw on them.


Egorand (approved):
Looks even simpler!

PR #2536, opened 2023-07-12
feat: adds  CoroutineContext parameter into BindableAdapter constructor

jclyne:
The BindableAdapter currently does not visibility into the base <service>Impl class in order to set a custom CoroutineContext. This exposes a context parameter with, with an empty default, that passes it through to the underlying pass impl class. This is only generated if suspendingCalls is enabled

This should be backward compatible, since it has a default value and generates `@JvmOverloads`. It also keeps the lambda as the last parameter.

### DIFF
--- a/wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/BindableAdapterGenerator.kt
+++ b/wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/BindableAdapterGenerator.kt
@@ -62,9 +62,10 @@ object BindableAdapterGenerator {
                   // via a context parameter.
                   this.addParameter(
                     ParameterSpec.builder("context", CoroutineContext::class)
-                      .defaultValue("kotlin.coroutines.EmptyCoroutineContext")
+                      .defaultValue("%T", ClassName("kotlin.coroutines", "EmptyCoroutineContext"))
                       .build(),
                   )
+                  this.addAnnotation(JvmOverloads::class)
                 }
               }
               .apply { addRpcConstructorParameters(generator, this, service, options) }
--- a/wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/BindableAdapterGenerator.kt
+++ b/wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/BindableAdapterGenerator.kt
@@ -58,7 +58,8 @@ object BindableAdapterGenerator {
                     type = ExecutorService::class,
                   )
                 } else {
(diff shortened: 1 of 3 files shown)


wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/BindableAdapterGenerator.kt
```diff
@@ -55,6 +57,13 @@ object BindableAdapterGenerator {
                     name = "streamExecutor",
                     type = ExecutorService::class,
                   )
+                } else {
+                  // suspending calls optionally allow for adding to the CoroutineContext
```
oldergod, line 61: https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/BindableAdapterGenerator.kt
```diff
@@ -55,6 +57,14 @@ object BindableAdapterGenerator {
                     name = "streamExecutor",
                     type = ExecutorService::class,
                   )
+                } else {
+                  // For suspending calls, optionally allow for adding to the CoroutineContext
+                  // via a context parameter.
+                  this.addParameter(
+                    ParameterSpec.builder("context", CoroutineContext::class)
+                      .defaultValue("kotlin.coroutines.EmptyCoroutineContext")
```
JakeWharton, line 65:
```suggestion
                      .defaultValue("%T", ClassName("kotlin.coroutines", "EmptyCoroutineContext"))
```

wire-grpc-server-generator/src/test/golden/unitService.kt
```diff
@@ -128,8 +128,9 @@ public object MyServiceWireGrpc {
   }
 
   public class BindableAdapter(
+    context: CoroutineContext = kotlin.coroutines.EmptyCoroutineContext,
     private val service: () -> MyServiceServer,
```
JakeWharton, line 132: This is a binary-breaking signature change for the JVM. You will need to generate `@JvmOverloads` on the constructor in order to retain compatibility.

JakeWharton (approved):

oldergod:
Probably need to `./gradlew apiDump` btw

jvmakine (approved):
Nice! Thank you for doing this

jclyne:
> ./gradlew apiDump

Should this update files that need to be committed?

oldergod:
Got it, thank you.
We can ignore the windows failure if it's memory related. I'll merge on green

PR #2538, opened 2023-07-13
Removes inputFiles from WireTask

oldergod:
Doesn't seem worth it. It removes an input from WireTask which I like.


oldergod:
I'm not sure. Breaking the task API doesn't seem such a big deal to me. I wonder if people ever uses it

PR #2539, opened 2023-07-13
Swift: remove unused variables

jszumski:
Fixes two warnings for unused variables:

> wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt:1020:13 Variable 'typeName' is never used
> 
> wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt:1021:13 Variable 'fieldName' is never used


oldergod (approved):

dnkoutso (commented):
👍👍

PR #2548, opened 2023-07-22
Wrap non-default nio filesystems with new Okio APIs

oldergod:


wire-compiler/src/main/java/com/squareup/wire/schema/fileSystems.kt
```diff
@@ -17,10 +17,11 @@ package com.squareup.wire.schema
 
 import java.nio.file.FileSystem
 import java.nio.file.FileSystems
+import okio.FileSystem.Companion.asOkioFileSystem
 
 internal fun FileSystem.toOkioFileSystem(): okio.FileSystem {
   return when {
     this == FileSystems.getDefault() -> okio.FileSystem.SYSTEM
-    else -> error("Wire doesn't support non-default file system: $this")
+    else -> this.asOkioFileSystem()
```
oldergod, line 25: or maybe just replace wire's `toOkioFileSystem()` and use okio's `asOkioFileSystem()` all the time?

oldergod: updated

Egorand (approved):

PR #2549, opened 2023-07-26
Only add ProtoTarget output to resources, not sources

oldergod:
And vice versa. Should fix a bunch a bugs where one of the folder were treated in the other form and the IDE would go red and not resolve anything.

image


Egorand (approved):
nit: fix up the commit message? It currently says "...resources, not resources".

oldergod:
Thank you, I fixed it

PR #2550, opened 2023-07-27
Update google/protobuf protos

oldergod:
From https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/

### DIFF
--- a/wire-schema/src/jvmMain/resources/google/protobuf/descriptor.proto
+++ b/wire-schema/src/jvmMain/resources/google/protobuf/descriptor.proto
@@ -146,9 +146,6 @@ message ExtensionRangeOptions {
     // and enums.
     optional string type = 3;
 
-    // Deprecated. Please use "repeated".
-    optional bool is_repeated = 4 [deprecated = true];
-
     // If true, indicates that the number is reserved in the extension range,
     // and any extension field with the number will fail to compile. Set this
     // when a declared extension field is deleted.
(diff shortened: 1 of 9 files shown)


Egorand (approved):

wire-schema/src/jvmMain/resources/google/protobuf/descriptor.proto
```diff
@@ -768,6 +789,14 @@ message EnumValueOptions {
   // this is a formalization for deprecating enum values.
   optional bool deprecated = 1 [default = false];
 
+  // Any features defined in the specific edition.
+  optional FeatureSet features = 2;
+
+  // Indicate that fields annotated with this enum value should not be printed
+  // out when using debug formats, e.g. when the field contains sensitive
+  // credentials.
+  optional bool debug_redact = 3 [default = false];
```
oldergod, line 798: https://github.com/square/wire/issues/2551

Maybe we keep ignoring it or we ignore the comment saying "when using debug formats" (what is a debug format?)

wire-schema/src/jvmMain/resources/google/protobuf/descriptor.proto
```diff
@@ -854,6 +889,99 @@ message UninterpretedOption {
...
+// ===================================================================
+// Features
+
+// TODO(b/274655146) Enums in C++ gencode (and potentially other languages) are
+// not well scoped.  This means that each of the feature enums below can clash
+// with each other.  The short names we've chosen maximize call-site
+// readability, but leave us very open to this scenario.  A future feature will
+// be designed and implemented to handle this, hopefully before we ever hit a
+// conflict here.
+message FeatureSet {
```
oldergod, line 901: If we ever wanna support editions when generating code, that's the logic we need to check

PR #2555, opened 2023-07-31
Provide proto output to sourceSet's resources

oldergod:
The test doesn't actually fail without the fix but I confirmed within plasma that it now works. I don't know how to repro within Wire itself :/

cc @mmollaverdi

### DIFF
--- a/wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
+++ b/wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
@@ -787,6 +787,25 @@ class WirePluginTest {
   }
 
   @Test
+  fun protoLibrary() {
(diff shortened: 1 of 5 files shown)


Egorand (approved):

wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
```diff
@@ -786,6 +786,25 @@ class WirePluginTest {
...
 
+  @Test
+  fun protoLibrary() {
+    val fixtureRoot = File("src/test/projects/proto-library")
+
+    val result = gradleRunner.runFixture(fixtureRoot) {
+      withArguments("jar", "--stacktrace", "--info").build()
+    }
+
+    println(result.output)
```
mmollaverdi, line 797: 🙈

mmollaverdi (approved):

PR #2559, opened 2023-08-13
feat: Adjust Gradle plugin `outputs` to be computed lazily

Nava2:
At Faire, we have Gradle convention plugins in use to share common configuration of our many (~50) protobuf projects. To do this, we leverage a [Gradle Convention plugin](https://docs.gradle.org/current/samples/sample_convention_plugins.html) to wrap our common settings. One issue we have is that we want to make the `out` directory configurable by scripts, but due to how Wire requires computation at wire configuration time, it is not possible.

This PR moves the computation of targets to be wrapped in a `ListProperty<>` and pushes the execution of `Action<>` to be when tasks are configured. The `plugin-playground` project shows a simplified case.

All of the non-test changes are to lazily compute values via providers.

### DIFF
--- a/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireOutput.kt
+++ b/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireOutput.kt
@@ -31,7 +31,7 @@ import javax.inject.Inject
  * as destination directories and configuration options). This includes registering output
  * directories with the project so they can be compiled after they are generated.
  */
-abstract class WireOutput {
+sealed class WireOutput {
   /** Set this to override the default output directory for this [WireOutput]. */
   var out: String? = null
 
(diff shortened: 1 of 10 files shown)


wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireOutput.kt
```diff
@@ -31,7 +31,7 @@ import javax.inject.Inject
  * as destination directories and configuration options). This includes registering output
  * directories with the project so they can be compiled after they are generated.
  */
-abstract class WireOutput {
+sealed class WireOutput {
```
oldergod, line 34: Is this required for your PR to pass? If not, could you remove all collateral changes to make the review easier please?

Nava2:
To do this right is a lot more effort than I have ability to tackle! I started pulling on the thread, but there's a lot more paths than I expected that require switching to use `Property<>` values rather than straight kotlin properties. 😢 

I don't have capacity to work on this, but I do think it'd be a great project to tackle incrementally!

PR #2564, opened 2023-08-21
Support wildcard in member pruning rules

oldergod:


Egorand (approved):
Interesting! What's an example of a proto with a member that has a qualified name?

oldergod:
> What's an example of a proto with a member that has a qualified name?

For instance,

```protobuf
package a;
message A {
  string one = 1;
}
```

```protobuf
package b;
extend A {
  string two = 2;
}
```

To prune `two` without pruning `one`, one would do `prune(a.A#b.two)`

Egorand:
Wouldn't `prune(b.A#two)` achieve the same result? Or can you do `prune(a.A#two)`?

oldergod:
`a.A#two` doesn't work, no. And I think it's thus possible to have a "real" field named `one` and an extension field named `one` as well.

dnkoutso (approved):

PR #2594, opened 2023-08-29
Explicitly force project dependencies to JVM

oldergod:
Best effort as of now.


oldergod:
We can use https://github.com/square/wire/pull/2593 later to test fixes.

martinbonnin (approved):

PR #2595, opened 2023-08-29
Adds warning when one defines a protoLibrary without source sets

oldergod:
If we don't find a `resources` sets to add a dependency to, the `:jar` task or whatever it is will not emit the `.proto` files automatically.


wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -266,10 +266,15 @@ class WirePlugin : Plugin<Project> {
       if (extension.protoLibrary) {
         val sourceSets = project.extensions.getByType(SourceSetContainer::class.java)
         // Note that there are no source sets for some platforms such as native.
-        if (sourceSets.isNotEmpty()) {
+        // TODO(Benoit) Probably should be checking for other names than `main`. As well, source
+        //  sets might be created 'afterEvaluate'. Does that mean we should do this work in
+        //  `afterEvaluate` as well? See: https://kotlinlang.org/docs/multiplatform-dsl-reference.html#source-sets
+        if (sourceSets.isNotEmpty() && sourceSets.findByName("main") != null) {
```
martinbonnin, line 272: Do you need to check `isNotEmpty()` here?

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -266,10 +266,15 @@ class WirePlugin : Plugin<Project> {
...
           sourceSets.getByName("main") { main: SourceSet ->
             main.resources.srcDir(protoOutputDirectory)
           }
+        } else {
+          project.logger.warn("${project.displayName} doesn't have a 'main' source sets. The .proto files will not automatically be added to the artifact. You have to manually add them.")
```
martinbonnin, line 277: No need to tell users to add manually if the plugin is not going to be able to consume there afterwards?

```suggestion
          project.logger.warn("${project.displayName} doesn't have a 'main' source sets. The .proto files will not automatically be added to the artifact.")
```

PR #2615, opened 2023-09-06
Update grpc to v1.57.2

oldergod:


wire-grpc-server/build.gradle.kts
```diff
@@ -21,7 +22,13 @@ dependencies {
   implementation(libs.grpc.stub) {
     exclude(group = "com.google.guava", module = "guava")
   }
-  implementation("com.google.guava:guava:21.0")
+  implementation(libs.checker.qual)
+  implementation(libs.guava){
+    attributes {
+      // We help Gradle pick between the jre and android of Guava.
```
Egorand, line 28: Is it needed? Does Gradle resolve it to the Android version by default?

Egorand (approved):

wire-grpc-server/build.gradle.kts
oldergod: It doesn't resolve it at all actually so I had to tell him

image

And others are seeing it: https://github.com/google/guava/issues/6612

oldergod: I talked with @martinbonnin and we cannot repro without the attribute so I tried a commit without any of them.

martinbonnin: 🤞

PR #2626, opened 2023-09-08
Organize Dokka and fix navigation

oldergod:


wire-gradle-plugin/build.gradle.kts
```diff
@@ -74,21 +68,6 @@ val test by tasks.getting(Test::class) {
   systemProperty("knative", System.getProperty("knative"))
 }
 
-if (project.rootProject.name == "wire") {
-  configure<MavenPublishBaseExtension> {
-    configure(
-      GradlePlugin(javadocJar = Javadoc(), sourcesJar = true)
```
oldergod, line 80: This doesn't seem to be necessary anymore and it actually prevent the plugin to be published so removing.

PR #2630, opened 2023-09-14
Never prune descriptor.proto fileOptions if internal types are marked

oldergod:
Some repo were pruning `descriptor.proto` and add it to a Jar (whatever) but Wire would produce corrupted files where all types were not resolvable. Now, we don't let this happen.


swankjesse (commented):
I think not emitting `google/protobuf/descriptor.proto` is better.

What’s weird about emitting `google/protobuf/descriptor.proto` is that somebody has that file on their source path.  I wouldn’t have expected that.

PR #2632, opened 2023-09-14
Never emit google/protobuf/descriptor.proto

oldergod:
I renamed `containsRelativePaths` to `containsExactlyInAnyOrderAsRelativePaths` because it is what is does.

### DIFF
--- a/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
+++ b/wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
@@ -254,7 +254,13 @@ data class ProtoTarget(
         val outDirectory = context.outDirectory
 
         for (protoFile in schema.protoFiles) {
-          if (!context.inSourcePath(protoFile) || protoFile.isEmpty()) continue
+          if (!context.inSourcePath(protoFile) ||
+            protoFile.isEmpty() ||
+            // We never emit the `.proto` files we are embedding within Wire.
+            isWireRuntimeProto(protoFile.location.path)
+          ) {
+            continue
+          }
 
           val relativePath = protoFile.location.path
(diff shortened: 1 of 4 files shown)


wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
```diff
@@ -254,7 +254,13 @@ data class ProtoTarget(
         val outDirectory = context.outDirectory
 
         for (protoFile in schema.protoFiles) {
-          if (!context.inSourcePath(protoFile) || protoFile.isEmpty()) continue
+          if (!context.inSourcePath(protoFile) ||
+            protoFile.isEmpty() ||
+            // We never emit it.
+            protoFile.location.toString() == "google/protobuf/descriptor.proto"
```
swankjesse, line 260: Do likewise for our other magic files? Like Wire extensions and stuff?

wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt
```diff
@@ -217,6 +217,25 @@ class WireRunTest {
       .contains("message Red {")
   }
 
+  @Test
+  fun protoTargetNeverEmitsGoogleProtobufDescriptor() {
```
swankjesse, line 221: 👍🏻  Nice test

swankjesse (approved):

wire-compiler/src/main/java/com/squareup/wire/schema/Target.kt
oldergod: I modified the code to never emit any runtime protos

PR #2633, opened 2023-09-15
Allow setting up a custom logger in WireCompiler

oldergod:
Brought back some code of #2486 which got reverted later on.
Dimitris needs it for the Bazel integration.


oldergod:
I should follow up and add it to Gradle extensions as well

dnkoutso (approved):

PR #2636, opened 2023-09-15
Add explicit overloads for getters/setters

lickel:
There's two things here:

* Reordering the structures in a consistent order
    * Public object
    * Public object conformances
    * Object Sub-types
    * Object Sub-type conformances
    * Storage object
    * Storage object conformances
* Adding explicit overloads for getters/setters which _should_ improve performance
    * If needed I can unwind it further and call storage directly

I could not remove `dynamicMemberLookup` without breaking `@Defaulteded`

### DIFF
--- a/wire-runtime-swift/src/main/swift/propertyWrappers/Heap.swift
+++ b/wire-runtime-swift/src/main/swift/propertyWrappers/Heap.swift
@@ -22,6 +22,7 @@ import Foundation
  Use of this wrapper is required on large protobuf messages because large (and especially
  recursive) structs can overflow the Swift runtime stack.
  */
+@available(*, deprecated, message: "Replace with CopyOnWrite. Note that property wrapper has different semantics")
 @propertyWrapper
 public final class Heap<T> {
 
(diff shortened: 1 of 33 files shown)


wire-runtime-swift/src/main/swift/propertyWrappers/Heap.swift
```diff
@@ -22,6 +22,7 @@ import Foundation
  Use of this wrapper is required on large protobuf messages because large (and especially
```
efirestone, line 22: Can we make sure this comment gets copied onto CopyOnWrite before this is fully removed (or now)?

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -1257,24 +1302,25 @@ class SwiftGenerator private constructor(
...
     storageName: String,
     storageType: DeclaredTypeName,
+    oneOfEnumNames: Map<OneOf, DeclaredTypeName>,
   ) {
     if (!type.isHeapAllocated) {
       println("Generating storage property delegates for a non-heap allocated type?!")
     }
 
     val propertyVariable = TypeVariableName.typeVariable("Property")
+    val signature = FunctionSignatureSpec.builder()
```
efirestone, line 1312: Do we want/need this dynamic member lookup method anymore?

efirestone (approved):

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
lickel: Yup. We need it for `$foo`, even if `foo` explicitly exists.

PR #2650, opened 2023-09-23
[swift] Provide a default value for sub fields and common types as per proto spec

dnkoutso:
This implements the proto spec:

> If the default value is not specified for an optional element, a type-specific default value is used instead: for strings, the default value is the empty string. For bytes, the default value is the empty byte string. For bools, the default value is false. For numeric types, the default value is zero. For enums, the default value is the first value listed in the enum’s type definition. This means care must be taken when adding a value to the beginning of an enum value list. See the Updating A Message Type section for guidelines on how to safely change definitions.

### DIFF
--- a/wire-runtime-swift/src/main/swift/Redactable.swift
+++ b/wire-runtime-swift/src/main/swift/Redactable.swift
@@ -44,16 +44,20 @@ extension Redactable {
             guard let label = label else {
                 return "\(value)"
             }
-            if RedactedKeys(rawValue: label) != nil {
+            var strippedLabel = label
+            if (label.hasPrefix("_")) {
+                strippedLabel = String(label.dropFirst(1))
+            }
+            if RedactedKeys(rawValue: strippedLabel) != nil {
                 // This is a redacted field, but if it's nil then that's ok to print
                 if "\(value)" != "nil" {
-                    return "\(label): <redacted>"
+                    return "\(strippedLabel): <redacted>"
                 }
             }
             if value is String {
-                return "\(label): \"\(value)\""
+                return "\(strippedLabel): \"\(value)\""
             }
-            return "\(label): \(value)"
+            return "\(strippedLabel): \(value)"
         }
 
         if fields.count > 0 {
--- a/wire-runtime-swift/src/test/swift/sample/Dinosaur.swift
+++ b/wire-runtime-swift/src/test/swift/sample/Dinosaur.swift
@@ -8,12 +8,15 @@ public struct Dinosaur {
     /**
      * Common name of this dinosaur, like "Stegosaurus".
      */
+    @ProtoDefaulted
     public var name: String?
     /**
      * URLs with images of this dinosaur.
      */
     public var picture_urls: [String] = []
+    @ProtoDefaulted
     public var length_meters: Double?
+    @ProtoDefaulted
     public var mass_kilograms: Double?
     public var period: Period?
     public var unknownFields: Foundation.Data = .init()
@@ -36,10 +39,10 @@ extension Dinosaur {
         mass_kilograms: Swift.Double? = nil,
         period: Period? = nil
     ) {
-        self.name = name
+        self._name.wrappedValue = name
         self.picture_urls = picture_urls
-        self.length_meters = length_meters
-        self.mass_kilograms = mass_kilograms
+        self._length_meters.wrappedValue = length_meters
+        self._mass_kilograms.wrappedValue = mass_kilograms
         self.period = period
     }
 
@@ -60,6 +63,13 @@ extension Dinosaur : Hashable {
 extension Dinosaur : Sendable {
 }
 #endif
+
+extension Dinosaur : ProtoDefaultedValue {
+
+    public static var defaultedValue: Dinosaur {
+        Dinosaur()
+    }
+}
 
 extension Dinosaur : ProtoMessage {
 
@@ -91,10 +101,10 @@ extension Dinosaur : Proto2Codable {
         }
         self.unknownFields = try protoReader.endMessage(token: token)
 
-        self.name = name
+        self._name.wrappedValue = name
         self.picture_urls = picture_urls
-        self.length_meters = length_meters
-        self.mass_kilograms = mass_kilograms
+        self._length_meters.wrappedValue = length_meters
+        self._mass_kilograms.wrappedValue = mass_kilograms
         self.period = period
     }
 
@@ -114,10 +124,10 @@ extension Dinosaur : Codable {
 
     public init(from decoder: Swift.Decoder) throws {
         let container = try decoder.container(keyedBy: Wire.StringLiteralCodingKeys.self)
-        self.name = try container.decodeIfPresent(Swift.String.self, forKey: "name")
+        self._name.wrappedValue = try container.decodeIfPresent(Swift.String.self, forKey: "name")
         self.picture_urls = try container.decodeProtoArray(Swift.String.self, firstOfKeys: "pictureUrls", "picture_urls")
-        self.length_meters = try container.decodeIfPresent(Swift.Double.self, firstOfKeys: "lengthMeters", "length_meters")
-        self.mass_kilograms = try container.decodeIfPresent(Swift.Double.self, firstOfKeys: "massKilograms", "mass_kilograms")
+        self._length_meters.wrappedValue = try container.decodeIfPresent(Swift.Double.self, firstOfKeys: "lengthMeters", "length_meters")
+        self._mass_kilograms.wrappedValue = try container.decodeIfPresent(Swift.Double.self, firstOfKeys: "massKilograms", "mass_kilograms")
         self.period = try container.decodeIfPresent(Period.self, forKey: "period")
     }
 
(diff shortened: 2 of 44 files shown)


wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -158,6 +158,14 @@ class SwiftGenerator private constructor(
       else -> null
     }
 
+  private val Field.defaultedValue: CodeBlock?
+    get() = default?.let {
+      return defaultFieldInitializer(type!!, it)
+    } ?: if (isMessage && !isRequiredParameter && !isCollection) {
+        val subType = schema.getType(type!!) as MessageType
+        if (subType!!.fields.all { !it.isRequiredParameter }) CodeBlock.of("%T()", subType.typeName) else null
+    } else null
```
oldergod, line 167: Would that read better?
```suggestion
  private val Field.defaultedValue: CodeBlock?
    get() {
      when (val default = default) {
        null -> {
          if (!isMessage || isRequiredParameter || isCollection) return null
          val subType = schema.getType(type!!) as MessageType
          if (subType.fields.any { it.isRequiredParameter }) return null
          return CodeBlock.of("%T()", subType.typeName)
        }

        else -> return defaultFieldInitializer(type!!, default)
      }
    }

```

dnkoutso: nice yes!

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -158,6 +158,37 @@ class SwiftGenerator private constructor(
       else -> null
     }
 
+  private val Field.defaultedValue: CodeBlock?
+    get() = default?.let {
+      return defaultFieldInitializer(type!!, it)
+    } ?: when {
+      !isOptional -> null
```
dnkoutso, line 165: what about non optional fields? I still think we should emit those? I think I got a Swift compiler error...

wire-runtime-swift/src/test/swift/sample/Dinosaur.swift
```diff
@@ -15,6 +15,7 @@ public struct Dinosaur {
     public var picture_urls: [String] = []
     public var length_meters: Double?
     public var mass_kilograms: Double?
+    @Defaulted(defaultValue: Period.CRETACEOUS)
```
dnkoutso, line 18: this is weird why isnt the rest of the fields `length_meters` defaulted...? looking.

dnkoutso: figured it out has to do with optionality on the typeName...fixing

wire-runtime-swift/src/test/swift/sample/Dinosaur.swift
```diff
@@ -8,13 +8,17 @@ public struct Dinosaur {
     /**
      * Common name of this dinosaur, like "Stegosaurus".
      */
+    @Defaulted(defaultValue: "")
```
lickel, line 11: - [ ] Should _probably_ update the README to include info about @Defaulted and this updated model

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -158,6 +158,44 @@ class SwiftGenerator private constructor(
...
+        null -> {
+          if (!isOptional) return null
+          if (type == ProtoType.ANY) return null
+          if (isMessage && !isRequiredParameter && !isCollection) {
+            val messageType = schema.getType(type!!) as MessageType
+            return if (messageType.fields.any { it.isRequiredParameter }) null else CodeBlock.of("%T()", messageType.typeName.makeNonOptional())
+          }
+          if (isEnum) {
+            val enumType = schema.getType(type!!) as EnumType
+            return if (enumType.constants.getOrNull(0) != null) CodeBlock.of("%T.%L", typeName.makeNonOptional(), enumType.constants[0].name) else null
```
lickel, line 174: Is `enumType.constants.getOrNull(0)` correct? That just looks like the first (mapped) constant, _not_ the 0-th value.

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -158,6 +158,44 @@ class SwiftGenerator private constructor(
...
+            INT -> return CodeBlock.of("%L", 0)
+            INT32 -> return CodeBlock.of("%L", 0)
+            INT64 -> return CodeBlock.of("%L", 0)
+            UINT32 -> return CodeBlock.of("%L", 0)
+            UINT64 -> return CodeBlock.of("%L", 0)
+            FLOAT -> return CodeBlock.of("%L", 0)
+            DOUBLE -> return CodeBlock.of("%L", 0)
+            STRING -> return CodeBlock.of("%S", "")
+            DATA -> return CodeBlock.of(
+              "%T(base64Encoded: %S)!",
```
lickel, line 187: Can you not just do `Data()` ?

lickel (commented):
This is _effectively_ doubling the storage requirements for everything... 
It's now making me a bit nervous...

Maybe we can do something more like [BetterCodable](https://github.com/marksands/BetterCodable/blob/master/Sources/BetterCodable/DefaultCodable.swift) and use statics for String, Number, etc...

The empty structs could also be solved by conforming to a protocol like `EmptyInitializable` 

```swift
@propertyWrapper
public struct DefaultZero<Value: Numeric> {
    public var wrappedValue: Value?
     
    public var projectedValue: Value {
        wrappedValue ?? .zero
    }
}

@propertyWrapper
public struct DefaultedEmpty<Value: EmptyInitializable> {
    public var wrappedValue: Value?

    public var projectedValue: Value {
        wrappedValue ?? Value.init()
    }
}
```

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
lickel: Based on the proto3 code, I think we want `enum.constants.filter { it.tag == 0 }.firstOrNull()`

lickel:
Discussed offline: we need Defaulted to have conditional conformance to `Redactable`

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
dnkoutso: Yes this is probably incorrect. There is also `enum.constant(0)` but I think this didnt work, will be updating and looking again.

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
dnkoutso: yeah it was an initial copy paste from below will be updating

wire-runtime-swift/src/test/swift/sample/Dinosaur.swift
dnkoutso: yeap, thanks for reminding

dnkoutso: done

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
dnkoutso: fixed

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
dnkoutso: fixed

wire-runtime-swift/src/main/swift/Redactable.swift
```diff
@@ -44,16 +44,20 @@ extension Redactable {
             guard let label = label else {
                 return "\(value)"
             }
-            if RedactedKeys(rawValue: label) != nil {
+            var strippedLabel = label
```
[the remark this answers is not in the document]
dnkoutso: Test caught this now because we emit `@Defaulted` identity types for many properties which now as an empty string exacerbated this issue which is now fixed and the test passes.

wire-tests-proto3-swift/src/main/swift/ContainsDuration.swift
```diff
@@ -5,6 +5,7 @@ import Wire
 
 public struct ContainsDuration {
 
+    @Defaulted(defaultValue: Duration())
```
dnkoutso, line 8: this seems to be a bug actually. Duration cannot be initialized as it has required fields....

dnkoutso: investigating...

dnkoutso: uh `messageType.fields` is empty for this proto...this is weird? I am updating the logic to handle this case...

dnkoutso: fixed by checking if `fields.isEmpty` but I do not love it, wonder why Wire wouldnt load those? `contains_duration.proto` has `import "google/protobuf/duration.proto";` ...

lickel (approved):

jamieQ:
> Discussed offline: we need Defaulted to have conditional conformance to `Redactable`

before this falls out of my memory: the conditional conformance did not appear necessary to address the test failures occurring with the nested, partially redacted types. the reason being that the `Redactable` conformance is present on the types the `Defaulted` property wrapper applies to, and that appears to be what the `Mirror` actually sees when constructing the recursive description (vs seeing the property wrapper type). e.g. the Mirror iteration logic sees `_defaultedProp: SomeRedactableType` vs `defaultedProp: Defaulted<SomeRedactableType>`, or at least that was my impression of what was happening.

PR #2652, opened 2023-09-25
Introduce internal plugin

oldergod:
Centralizing configurations. Not able to get the extension to work with included builds, no idea why.


build-support/src/main/kotlin/com/squareup/wire/buildsupport/WireBuildPlugin.kt
```diff
@@ -0,0 +1,382 @@
...
+    val spotless = extensions.getByName("spotless") as SpotlessExtension
+    val licenseHeaderFile = rootProject.file("gradle/license-header.txt")
+    spotless.apply {
+      // The nested build-support Gradle project contains Java sources. Use our root project to
+      // target its sources rather than duplicating the Spotless setup in multiple places.
+      java {
+        if (path == ":") {
+          target("build-support/settings/src/**/*.java")
+        } else {
+          target("src/**/*.java")
```
JakeWharton, line 123: Looks like you need an exclude here for test fixtures that have nested gradle builds

JakeWharton (approved):
Looks like a pretty good start

build-support/src/main/kotlin/com/squareup/wire/buildsupport/WireBuildPlugin.kt
oldergod: Is it what's creating the task dependency problems with spotless?

oldergod: Got it, I saw the CI failures

PR #2658, opened 2023-10-03
[swift] Ensure imports of parameterized type names are emitted

dnkoutso:
We are not emitting import statements when the type is parameterized e.g. for repeated fields.

This is actually an underlying bug in SwiftPoet which is probably fixed by latest version of KotlinPoet. SwiftPoet for the most part is a copy paste of KotlinPoet so updating SwiftPoet could help us potentially remove our custom import management layer within SwiftGenerator.kt.

For now we will proceed with this fix until SwiftPoet is updated which is a much bigger task.


wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -1206,6 +1206,12 @@ class SwiftGenerator private constructor(
       if (typeName is DeclaredTypeName) {
         imports.add(typeName.moduleName)
       }
+      if (field.isRepeated && typeName is ParameterizedTypeName) {
+        val declaredTypeName = typeName.typeArguments[0]
+        if (declaredTypeName is DeclaredTypeName) {
```
dnkoutso, line 1211: is there a better way to avoid this check again? Is `typeArguments[0]` always this type?

Also I seem not to handle maps but only arrays?

oldergod: Tests could help you confirm all of those

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -1206,6 +1206,12 @@ class SwiftGenerator private constructor(
       if (typeName is DeclaredTypeName) {
         imports.add(typeName.moduleName)
       }
+      if (field.isRepeated && typeName is ParameterizedTypeName) {
+        val declaredTypeName = typeName.typeArguments[0]
```
lickel, line 1210: Maybe do a for loop on typeArguments instead of just [0] ?

wire-tests-swift/src/main/swift/module_two/SwiftModuleTwoMessage.swift
```diff
@@ -0,0 +1,104 @@
...
+        return "type.googleapis.com/squareup.protos.kotlin.swift_module_two.SwiftModuleTwoMessage"
+    }
+
+}
+
+extension SwiftModuleTwoMessage : Proto2Codable {
+
+    public init(from protoReader: Wire.ProtoReader) throws {
+        var name: Swift.String? = nil
+        var module_one_names: [module_one.SwiftModuleOneMessage] = []
```
lickel, line 59: You should add a second .proto which has a map of cross module types.

wire-tests-swift/src/main/swift/module_two/SwiftModuleTwoMessage.swift
```diff
@@ -0,0 +1,104 @@
+// Code generated by Wire protocol buffer compiler, do not edit.
+// Source: squareup.protos.kotlin.swift_module_two.SwiftModuleTwoMessage in swift_module_two.proto
+import Foundation
+import Wire
+import module_one
+
+public struct SwiftModuleTwoMessage {
+
+    public var name: String?
+    public var module_one_names: [SwiftModuleOneMessage] = []
```
lickel, line 10: This needs to be a Heap type, so you need ~16 parameters in the type

dnkoutso: The error reproduces without it I believe, still working on it.

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
dnkoutso: I see this pattern been used everywhere else regarding `ParametrizedTypeName`

wire-tests-swift/src/main/swift/module_two/SwiftModuleTwoMessage.swift
dnkoutso: yeap agree

wire-tests-swift/src/main/swift/no-manifest/AllTypes.swift
dnkoutso, line 1: With the new `generateSwiftProto2ManifestTests` Gradle task a test setup issue arises.

When this task runs with the existing `generateSwiftProto2Tests` tasks both them run `outFile.deleteRecursively()`.

I split the tests into `no-manifest` and `manifest` for future tests related to manifest files and Swift.

wire-tests-swift/src/main/swift/manifest/module_two/SwiftModuleTwoMessage.swift
```diff
@@ -0,0 +1,189 @@
+// Code generated by Wire protocol buffer compiler, do not edit.
+// Source: squareup.protos.kotlin.swift_modules.SwiftModuleTwoMessage in swift_module_two.proto
+import Foundation
+import Wire
+import module_one
```
dnkoutso, line 5: This import is not emitted without my fix which was the bug.

build.gradle.kts
```diff
@@ -143,7 +146,8 @@ subprojects {
          project.name.endsWith("-benchmarks") ||
          project.name.contains("golden") ||
          project.name.contains("protoc") ||
-         project.displayName.contains("sample")
+         project.displayName.contains("sample") ||
+         project.displayName.contains("wire-tests-swift")
```
dnkoutso, line 150: we do not want to apply checkStyle under this folder for anything, this was causing Gradle configuration to fail.

wire-tests-swift/src/main/swift/module_two/SwiftModuleTwoMessage.swift
dnkoutso: done and a test case.

lickel (commented):
- [x] add map test in a new .proto file

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
dnkoutso: OK yes this works for maps too so updated to loop through. Test case was added and verified it was broken and fixed with my change.

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
dnkoutso: Confirmed my theory. Fixed now with a test.

.github/workflows/build.yml
```diff
@@ -150,7 +150,7 @@ jobs:
         run: |
           ./gradlew generateSwiftTests --stacktrace --warning-mode all
           if [ ! -z "$(git status --porcelain)" ]; then git diff; echo -e "\nGenerated files changed. Did you run ./gradlew generateSwiftProtos and generateSwiftTests?"; exit 1; fi
-          ./gradlew :wire-tests-swift:build --stacktrace --warning-mode all
+          ./gradlew -p wire-tests-swift build --stacktrace --warning-mode all
```
dnkoutso, line 153: This runs all targets underneath that project.

wire-tests-swift/manifest/module_three/SwiftModuleThreeMessage.swift
```diff
@@ -0,0 +1,190 @@
+// Code generated by Wire protocol buffer compiler, do not edit.
+// Source: squareup.protos.kotlin.swift_modules.SwiftModuleThreeMessage in swift_module_three.proto
+import Foundation
+import Wire
+import module_one
```
dnkoutso, line 5: For maps too, this import was missing and verified its fixed by this PR.

build.gradle.kts
oldergod: That's on me to followup here

oldergod (approved):

PR #2665, opened 2023-10-06
Introduce OpaqueTypes

oldergod:
I've thought about doing like TypeMover but it is actually pretty hard to modify the types of fields after linking happened.
Putting it in the linking step when resolving types felt good so I went this way. It doesn't print unused types, the `Linker` doesn't know when it's finished so it might require some API modeling.

I'll follow up with entries for the Gradle plugin and WireCompiler.

### DIFF
--- a/wire-compiler/src/test/java/com/squareup/wire/schema/LinkerTest.kt
+++ b/wire-compiler/src/test/java/com/squareup/wire/schema/LinkerTest.kt
@@ -18,7 +18,9 @@
 package com.squareup.wire.schema
 
 import com.squareup.wire.testing.add
+import kotlin.test.assertFailsWith
 import okio.Path
+import okio.Path.Companion.toPath
 import okio.fakefilesystem.FakeFileSystem
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Test
@@ -26,6 +28,7 @@ import org.junit.Test
 class LinkerTest {
   private val fs = FakeFileSystem().apply {
     if (Path.DIRECTORY_SEPARATOR == "\\") emulateWindows() else emulateUnix()
+    createDirectories("proto-path".toPath())
   }
 
   @Test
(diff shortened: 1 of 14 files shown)


wire-compiler/src/test/java/com/squareup/wire/schema/LinkerTest.kt
```diff
@@ -56,6 +59,272 @@ class LinkerTest {
...
+          |  optional bool decaf = 2;
+          |}
+          |
+          |enum Roast {
+          |  MEDIUM = 1;
+          |  DARK = 2;
+          |}
+      """.trimMargin(),
+    )
+    val schema = loadAndLinkSchema(opaqueTypes = listOf(ProtoType.get("cafe.EspressoShot")))
```
swankjesse, line 86: Nice

wire-compiler/src/test/java/com/squareup/wire/schema/LinkerTest.kt
```diff
@@ -56,6 +59,272 @@ class LinkerTest {
...
+        |package cafe;
+        |
+        |message CafeDrink {
+        |  optional int32 size_ounces = 1;
+        |
+        |  repeated EspressoShot shots = 2;
+        |}
+        |
+        |message EspressoShot {
+        |  optional bytes roast = 1;
```
swankjesse, line 162: I don’t think this should work this way. `bytes` are encoded as length-delimited, but enums are encoded using varints, and we can’t substitute one for the other while retaining the encoded form. We’d want enums to map to `int32`, or simply don’t support them.

wire-compiler/src/test/java/com/squareup/wire/schema/LinkerTest.kt
```diff
@@ -56,6 +59,272 @@ class LinkerTest {
...
+        |  optional bool decaf = 2;
+        |}
+        |
+        |enum Roast {
+        |  MEDIUM = 1;
+        |  DARK = 2;
+        |}
+        |
+        |extend CafeDrink {
+        |  repeated bytes shots = 2;
```
swankjesse, line 231: Nice

wire-compiler/src/test/java/com/squareup/wire/schema/LinkerTest.kt
```diff
@@ -56,6 +59,272 @@ class LinkerTest {
...
+        |package cafe;
+        |
+        |message CafeDrink {
+        |  optional int32 size_ounces = 1;
+        |
+        |  repeated bytes shots = 2;
+        |}
+        |
+        |message EspressoShot {
+        |  optional bytes roast = 1;
```
swankjesse, line 288: this enum is cursed!

wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt
```diff
@@ -270,6 +270,32 @@ class WireRunTest {
...
+      ),
+    )
+    wireRun.execute(fs, logger)
+
+    assertThat(fs.findFiles("generated")).containsExactlyInAnyOrderAsRelativePaths(
+      "generated/kt/squareup/colors/Blue.kt",
+    )
+    assertThat(fs.readUtf8("generated/kt/squareup/colors/Blue.kt"))
+      .contains("class Blue")
+      // The type `Triangle` has been opaqued.
```
swankjesse, line 295: I love the word opaqued

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Field.kt
```diff
@@ -45,7 +45,7 @@ data class Field(
 
   val default: String?,
 
-  private val elementType: String,
+  private var elementType: String,
```
swankjesse, line 48: Making this mutable feels bad. But not worse than the rest of linking, which is also very mutable.

wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/PrunerTest.kt
```diff
@@ -56,6 +56,51 @@ class PrunerTest {
     assertThat(pruned.getType("MessageB")).isNull()
   }
 
+  /**
+   * We test that all references of an opaque type are getting pruned correctly.
+   */
+  @Test
+  fun opaqueTypePrunesItsReferences() {
```
swankjesse, line 63: Oooooh very nice

oldergod:
Updated the doc so that we throw if one is trying to opaque an enum

PR #2669, opened 2023-10-09
[swift] WIP

dnkoutso:


wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/ManifestPartitionTest.kt
```diff
@@ -177,6 +177,62 @@ class ManifestPartitionTest {
...
+          .build(),
+      ),
+    )
+
+    val partitionedSchema = schema.partition(modules)
+
+    val firstPartition = partitionedSchema.partitions.getValue("first")
+
+    assertThat(firstPartition.types).contains(ProtoType.get("com.squareup.wire.first.Card"))
+    assertThat(firstPartition.types).contains(ProtoType.get("com.squareup.wire.first.Card.Brand"))
```
dnkoutso, line 234: This is not present as part of the partition despite that `second` module depends on `first` and includes a reference to a field `optional com.squareup.wire.first.Card.Brand brand = 1;`

PR #2671, opened 2023-10-10
Don't arbitrarily prune oneOf options

### LINKED ISSUE #2670: com.squareup.wire.schema.Pruner ignore OneOf options
Hi, I got an unexpected behaviour after using Pruner to optimise a proto file. Unexpected in my case = removing options (`my_custom_oneOf_option`)inside the OneOf message (but it should be reachable if understood correctly). 

Most likely this behaviour is because the OneOf fields are iterated inside Pruner, but the OneOf options are ignored.
https://github.com/square/wire/blob/126c5f4e2fe5470ae55f5218c9d9deb0657bf293/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Pruner.kt#L220

Example of falling test: 

```
@Test
public void should_mark_option_reachable() {
    // given
    var loader = new SchemaLoader(FileSystems.getDefault());
    var linker = new Linker(loader, new ErrorCollector(), true, true);
    var testFileLocation = Location.get(<TEST_FOLDER_WITH_TEST_PROTOFILE>);
    loader.initRoots(List.of(testFileLocation), List.of(testFileLocation));
    var loadedTestProtoFile = loader.load("test_event.proto");
    var schema = linker.link(List.of(loadedTestProtoFile));

    // when
    var optimizedSchema = schema.prune(new PruningRules.Builder().addRoot("test.oneOf.options.test.TestMessage").build());
    var optimizedProtoFile = optimizedSchema.protoFile("test_event.proto");
    
    // then
    assertThat(optimizedProtoFile.toSchema())
            .isEqualTo("""
                    // Proto schema formatted by Wire, do not edit.
                    // Source: test_event.proto
                    
                    syntax = "proto3";
                    
                    package test.oneOf.options.test;
                    
                    import "google/protobuf/descriptor.proto";
                    
                    message TestMessage {
                      oneof element {
                        option (my_custom_oneOf_option) = true;
                        string one = 1;
                        string two = 2;
                      }
                    }
                    """);
}
```

**test_event.proto** 

```
syntax = "proto3";

import "google/protobuf/descriptor.proto";
import "test_event_custom_option.proto";

package test.oneOf.options.test;

message TestMessage {
  oneof element {
    option (my_custom_oneOf_option) = true;
    string one = 1;
    string two = 2;
  }
}
```

**test_event_custom_option.proto** 

```
syntax = "proto3";

import "google/protobuf/descriptor.proto";

package test.oneOf.options;

extend google.protobuf.OneofOptions {
  bool my_custom_oneOf_option = 101400;
}
```

oldergod:
Fixes #2670

### DIFF
--- a/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Pruner.kt
+++ b/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Pruner.kt
@@ -185,8 +185,7 @@ class Pruner(
             result.add(field.type)
             options = field.options
           } else {
-            val oneOf = type.oneOf(member)
-            checkNotNull(oneOf) { "unexpected member: $member" }
+            val oneOf = checkNotNull(type.oneOf(member)) { "unexpected member: $member" }
             options = oneOf.options
           }
         } else if (type is EnumType) {
--- a/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Pruner.kt
+++ b/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Pruner.kt
@@ -105,8 +105,12 @@ class Pruner(
     val member = protoMember.member
     return when (val type = schema.getType(protoMember.type)) {
       is MessageType -> {
-        val field = type.field(member) ?: type.extensionField(member)!!
-        pruningRules.isFieldRetainedVersion(field.options)
+        val field = type.field(member) ?: type.extensionField(member)
+        if (field != null) {
+          pruningRules.isFieldRetainedVersion(field.options)
+        } else {
+          pruningRules.isFieldRetainedVersion(type.oneOf(member)!!.options)
+        }
       }
       is EnumType -> {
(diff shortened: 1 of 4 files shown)


wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Pruner.kt
```diff
@@ -177,9 +181,14 @@ class Pruner(
...
           val field = type.field(member) ?: type.extensionField(member)
-          checkNotNull(field) { "unexpected member: $member" }
-          result.add(field.type)
-          options = field.options
+          if (field != null) {
+            result.add(field.type)
+            options = field.options
+          } else {
+            val oneOf = type.oneOf(member)
+            checkNotNull(oneOf) { "unexpected member: $member" }
```
Egorand, line 189:
```suggestion
            val oneOf = checkNotNull(type.oneOf(member)) { "unexpected member: $member" }
```

Egorand (approved):

Issue #2589, opened 2023-08-23
Kotlin typesafe builder to help avoid the need for constructor use


cgruber:
For Kotlin gencode, we were using constructors a lot, before realizing the diamon-dependency-skew/ABI-incompatibility issues (obvious in retrospect).  But we like the terseness of the syntax, and the java builder syntax is bulky and feels... not so kotlin-y.

I built a quick workaround in our services - one form that let us use kotlin typesafe builders even for java-generated protos, but it's even cleaner for kotlin generated protos (because the Companion object is available to make extension methods) (see below for example).  After working with it a bit, it occurred to me that this could just be generated on the Companion object from the start, to allow for this syntax to be used if there is a Builder generated at all.

The generalized pattern is:
```
fun <M : Message<M, B>, B : Message.Builder<M, B>> buildProto(constructor: () -> B, initializer: B.() -> Unit): M {
  val builder = constructor.invoke()
  initializer.invoke(builder)
  return builder.build()
}
```

With this, one could comfortably do one-liners per builder, such as (for java, where there's no companion object to hang an extension function):
```
fun myProto(i: MyProto.Builder.() -> Unit) =  buildProto({ MyProto.Builder() }, i)

val myProto = myProto {
  prop1 = something
  prop2 = somethingElse
}
```

or for kotlin with a Companion object, the slightly more consistent and discoverable:
```
fun MyProto.Companion.build(i: MyProto.Builder.() -> Unit) =  buildProto({ MyProto.Builder() }, i)

val myProto = MyProto.build {
  prop1 = something
  prop2 = somethingElse
}
```

It occurred to me after this use, Wire could just generate this directly on the Companion object, and without the generics indirection, so it would look like so:

```
class MyProto(
  // fields
): Message<MyProto, MyProto.Builder>(ADAPTER, unknownFields) {
  // ...
  companion object {
    //...
    fun build(initializer: MyProto.Builder.() -> Unit) : MyProto {
      val builder = MyProto.Builder()
      initializer.invoke(builder)
      return builder.build()
    }
  }
}
```

I'm working up a PR for this, but I wanted to get some comment prior to proposing it in code.

oldergod:
That's an idea we've seen in the past and to be honest  I don't think this should be in Wire.

[...]

PR #2674, opened 2023-10-10
Generate Kotlin-esque build function for using closures in addition to Builders

rynonl:
https://github.com/square/wire/issues/2589

This adds a static `build` method on Messages to allow for a more Kotlin-esque style of building up proto objects, rather than forcing everyone into Builders now that constructors are generally deemed unsafe.

Usage:
```proto
// .proto
message MyMessage {
  optional int32 property = 1;
}
```
```kotlin
// Wire generated
companion object {
  ...

  public fun build(body: Builder.() -> Unit): MyMessage = Builder().apply(body).build()
}
```

```kotlin
// .kt
val message = MyMessage.build {
  property = 5
}
```
```

### CODE REVIEW

File: wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt [hunk 1]
```diff
@@ -548,6 +549,7 @@ class KotlinGenerator private constructor(
 
     addDefaultFields(type, companionBuilder, nameAllocator)
     addAdapter(type, companionBuilder)
+    if (buildersOnly || javaInterOp) addBuildClosure(type, companionBuilder)
```
rynonl, line 552: I added support for `javaInterop` since we get Builders there too - but I don't know exactly what the implications would be if Java code tried to use this (not a Java expert). Looking for feedback there especially.

adrw: I probably wouldn't include it for `buildersOnly` since this technically isn't a classic Java builder pattern, but would for `javaInterop` if this build closure is going to be default generated for all Kotlin targets.

It probably shouldn't be generated if the target is Java given I don't think Java supports Kotlin-style closures like this.

adrw:
Is it possible to not have to call `it` for all `it.property = value` assignments? That would align this implementation with Google's Kotlin support in protobuf and also be a bit more ergonomic. https://protobuf.dev/getting-started/kotlintutorial/


## Before
```
val person = com.protos.Person {
  it.id = "123"
  it.name = "Alice Bob"
}
```

## After
```
val person = com.protos.Person {
  id = "123"
  name = "Alice Bob"
}
```

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -1968,6 +1970,26 @@ class KotlinGenerator private constructor(
...
   }
 
+  private fun addBuildClosure(type: MessageType, companionBuilder: TypeSpec.Builder) {
+    val body = buildCodeBlock {
+      addStatement("val builder = Builder()")
+      addStatement("buildFn(builder)")
+      addStatement("return builder.build()")
+    }
+    val buildFn = FunSpec.builder("build")
+      .addParameter("buildFn",
```
JakeWharton, line 1980: Generally we avoid abbreviations in names. I think "builder" or "body" would work here.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -1968,6 +1970,26 @@ class KotlinGenerator private constructor(
...
+        LambdaTypeName.get(
+          parameters = arrayOf(TypeVariableName("Builder")),
+          returnType = Unit::class.asClassName()
+        )
```
JakeWharton, line 1984: Probably more ergonomic to change the builder type to be the receiver rather than an argument.

Also please pass in the `builderClassName` from the calling function and use that to reference the type here...

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -1968,6 +1970,26 @@ class KotlinGenerator private constructor(
     }
   }
 
+  private fun addBuildClosure(type: MessageType, companionBuilder: TypeSpec.Builder) {
+    val body = buildCodeBlock {
+      addStatement("val builder = Builder()")
```
JakeWharton, line 1975: ...and also here
```suggestion
      addStatement("val builder = %T()", builderClassName)
```

JakeWharton (approved):
Please also add a test somewhere that uses this function and compares it to a regular built value to confirm it works.

Like
```kotlin
@Test fun works() {
  val a = Foo.build {
    name = "hey"
    age = 21
  }
  val b = Foo.Builder()
    .name("hey")
    .age(21)
    .build()
  assertThat(a).isEqualTo(b)
}
```

rynonl:
`Is it possible to not have to call it for all it.property = value assignments?`

It is, but it introduces a variable shadowing issue:

```
val thing = "foo"

Message.build {
  thing = thing // this is a self referential assignment, the outer `thing` is completely inaccessible inside this block
}
```

The workaround is to rename the outer `thing` something else that doesn't conflict with a property name on the proto. It's not the end of the world, and I would be happy with that approach if others are.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
rynonl: See comment below on using builder as the receiver rather than the argument.

👍  on builderClassName

rynonl:
`Using this.thing also does not work, because thing implicitly refers to this.thing anyway...that's why I took the it approach so that you're explicitly referencing the argument and not the receiver.`

Actually, hold that thought. I had tested this previously but looks like maybe I had done something wrong. If we pass the builder as the receiver, looks like maybe we *can* use `this.` to differentiate when the outer val has a name conflict.

```
val x = 5

Message.build {
  x = x             // error
  this.x = x     // expected behavior
}
```

Are folks more into that? Seems cleaner for the happy case with a fairly straightforward workaround when name conflicts arise (which shouldn't actually be all that often since Wire generates snake_case properties and folks typically use camelCase variable names)

rynonl:
Changed to use Builder as receiver in [this commit](https://github.com/square/wire/pull/2674/commits/907e54ed1d53e8b2607a836674f84fcceb1d0349), I think that's probably the right path here.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -1968,6 +1971,23 @@ class KotlinGenerator private constructor(
     }
   }
 
+  private fun addBuildClosure(type: MessageType, companionBuilder: TypeSpec.Builder, builderClassName: ClassName) {
+    val buildFn = FunSpec.builder("build")
```
oldergod, line 1975: Can you rename to `buildFunction` or whatever explicit name this could be

wire-kotlin-generator/src/test/java/com/squareup/wire/kotlin/KotlinGeneratorTest.kt
```diff
@@ -1960,6 +1960,32 @@ class KotlinGeneratorTest {
...
+        |}
+        |
+        """.trimMargin(),
+      )
+    }
+    val code = KotlinWithProfilesGenerator(schema)
+      .generateKotlin("SomeMessage", buildersOnly = false, javaInterop = false)
+    assertThat(code).doesNotContain("inline fun build(body: Builder.() -> Unit): SomeMessage")
+    val buildersOnlyCode = KotlinWithProfilesGenerator(schema)
+      .generateKotlin("SomeMessage", buildersOnly = true)
```
oldergod, line 1982:
```suggestion
      .generateKotlin("SomeMessage", buildersOnly = true, javaInterop = false)
```

wire-kotlin-generator/src/test/java/com/squareup/wire/kotlin/KotlinGeneratorTest.kt
```diff
@@ -1960,6 +1960,32 @@ class KotlinGeneratorTest {
...
+    assertThat(buildersOnlyCode).contains("inline fun build(body: Builder.() -> Unit): SomeMessage")
+    val javaInteropCode = KotlinWithProfilesGenerator(schema)
+      .generateKotlin("SomeMessage", javaInterop = true)
```
oldergod, line 1985:
```suggestion
      .generateKotlin("SomeMessage", buildersOnly = false, javaInterop = true)
```

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -1968,6 +1971,23 @@ class KotlinGenerator private constructor(
     }
   }
 
+  private fun addBuildClosure(type: MessageType, companionBuilder: TypeSpec.Builder, builderClassName: ClassName) {
```
oldergod, line 1974:
```suggestion
  /**
   * Adds a closure into the [companionBuilder] allowing the creation of an instance via the Kotlin
   * DSL.
   *
   * Example
   * ```
   * companion object {
   *   public inline fun build(body: Builder.() -> Unit): AllTypes = Builder().apply(body).build()
   * }
   * ```
   */
  private fun addBuildClosure(type: MessageType, companionBuilder: TypeSpec.Builder, builderClassName: ClassName) {
```

oldergod (approved):
👍

oldergod:
I think it'd be good to hide it from Java callers using `@JvmSynthetic`. I'll do that in followup I think

rynonl:
@oldergod Sorry missed your comments! Followup incoming.

PR #2677, opened 2023-10-12
Allow GrpcClient to be implemented on all platforms

### LINKED ISSUE #2484: Ability to use custom gRPC clients on Multiplatform

oldergod:
Including JVM or common. Following up on https://github.com/square/wire/pull/2505 and we didn't quite do that.
Obviously breaking some APIs. Most common breaking change for client users is around having now to import `newCall` or `create` for they became extension functions.
I am not really sure it's the right way.

Goals are to let anybody implement their own GrpcClient on common, jvm, js, native. Right now, that's only possible for js and native. With this PR, I think all would be cool.

Fixes #2484
cc @iTaysonLab @RaphaelJenni


oldergod:
Jesse is telling me we could create the option at the OkHttp level. We'll pair next week and see how it goes.

wire-grpc-client/api/wire-grpc-client.api
```diff
@@ -24,9 +24,7 @@ public final class com/squareup/wire/GrpcCalls {
 }
 
 public abstract class com/squareup/wire/GrpcClient {
-	public synthetic fun <init> (Lokhttp3/Call$Factory;Lokhttp3/HttpUrl;JLkotlin/jvm/internal/DefaultConstructorMarker;)V
-	public final fun create (Lkotlin/reflect/KClass;)Lcom/squareup/wire/Service;
-	public final fun newBuilder ()Lcom/squareup/wire/GrpcClient$Builder;
+	public fun <init> ()V
```
swankjesse, line 27:
```suggestion
	public fun <init> ()V
	public final fun create (Lkotlin/reflect/KClass;)Lcom/squareup/wire/Service;
	public final fun newBuilder ()Lcom/squareup/wire/GrpcClient$Builder;```

File: wire-grpc-client/api/wire-grpc-client.api [hunk 2]
```diff
@@ -41,6 +39,11 @@ public final class com/squareup/wire/GrpcClient$Builder {
 	public final fun minMessageToCompress (J)Lcom/squareup/wire/GrpcClient$Builder;
 }
 
+public final class com/squareup/wire/GrpcClientKt {
+	public static final fun create (Lcom/squareup/wire/GrpcClient;Lkotlin/reflect/KClass;)Lcom/squareup/wire/Service;
+	public static final fun newBuilder (Lcom/squareup/wire/GrpcClient;)Lcom/squareup/wire/GrpcClient$Builder;
+}
+
```
swankjesse, line 46: This shouldn’t be a public API.
```suggestion
```

swankjesse (approved):
This is great!

oldergod:
Updated the code!

PR #2681, opened 2023-10-16
Don't crash in the Gradle plugin assigning isCanBeConsumed

swankjesse:
This PR is attempting to prevent this crash:

    org.gradle.api.GradleException: Cannot change the allowed usage of configuration ':protos:protoSourceCopy', as it was locked upon creation to the role: 'Custom Role'.
    This role permits the following usage:
            Consumable - this configuration can be selected by another project as a dependency (but this behavior is marked deprecated)
            Resolvable - this configuration can be resolved by this project to a set of files
            Declarable Against - this configuration can have dependencies added to it
    Ideally, each configuration should be used for a single purpose.
            at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration.assertUsageIsMutable(DefaultConfiguration.java:1787)
            at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration.preventIllegalMutation(DefaultConfiguration.java:1556)
            at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration.validateMutation(DefaultConfiguration.java:1510)
            at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration.setCanBeConsumed(DefaultConfiguration.java:1866)
            at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration_Decorated.setCanBeConsumed(Unknown Source)
            at com.squareup.wire.gradle.WirePlugin.setupWireTasks(WirePlugin.kt:147)

I believe the root cause is a regression fixed in Gradle 8.4 https://github.com/gradle/gradle/commit/edc8ab0aae1578ea1849111b52f663e14cd26308

### DIFF
--- a/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
+++ b/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
@@ -17,6 +17,7 @@
 
 package com.squareup.wire.gradle
 
+import java.lang.reflect.Array as JavaArray
 import com.squareup.wire.VERSION
 import com.squareup.wire.gradle.internal.libraryProtoOutputPath
 import com.squareup.wire.gradle.internal.targetDefaultOutputPath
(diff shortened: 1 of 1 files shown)


swankjesse:
This is unnecessary. Upgrading to Gradle 8.4 is sufficient.

oldergod:
Related to #2676

PR #2684, opened 2023-10-16
[swift] Use In32 type for enums versus UInt32

dnkoutso:


wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -1628,7 +1628,7 @@ class SwiftGenerator private constructor(
     val enumName = type.typeName
     return TypeSpec.enumBuilder(enumName)
       .addModifiers(PUBLIC)
-      .addSuperType(UINT32)
+      .addSuperType(INT32)
```
dnkoutso, line 1631: @JakeWharton This was introduced as UINT32 3 years ago with initial Swift support. Would it be possible to take a look at this change?

lickel (approved):
- [x] additional validation

JakeWharton (approved):

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
JakeWharton: No idea why. The docs mention it's signed (two's compliment) in multiple places.

dnkoutso: thank you!

oldergod (approved):
maybe you can add an random enum somewhere in the generated swift proto with a tag of -1 and confirm it compiles

lickel:
I asked for an enum that _every_ supported language compiles.

dnkoutso:
Tests added.

MikeSilvis (approved):

PR #2686, opened 2023-10-17
Restore ProtoDefaulted in Proto2

lickel:


dnkoutso (approved):

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/EnumType.kt
```diff
@@ -29,6 +29,7 @@ data class EnumType(
   override val documentation: String,
   override val name: String,
   val constants: List<EnumConstant>,
+  val defaultConstantValue: Int,
```
oldergod, line 32: What is a default constant value?

oldergod: this is also breaking APIs

oldergod:
Could we have a test to see what it's actually addressing?

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/EnumType.kt
lickel: It's the `tag` value of an enum. The "default" value for a proto is defined to always be the first case of that enum.
Since constants can (and are) pruned, I am persisting this at initialization time.

In proto3 land, it should always be `0` as the spec requires the first constant to have a value of `0`.
In proto2 land, there are no guarantees.

lickel:
> Could we have a test to see what it's actually addressing?

There's now two
* The explicit `Period.defaultValue == CRETACEOUS` test (which has a default value of `1`)
* Manifest compiles verify pruned enums perform as expected

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/EnumType.kt
lickel: Sorry, let me add to the PR description

PR #2689, opened 2023-10-18
[swift] Better handling for edge case field names

dnkoutso:


lickel (approved):

dnkoutso:
Swift is green and this only touches Swift generation merging.

PR #2693, opened 2023-10-20
Use platform slashes in order to please spotless on windows

oldergod:
The excludes seems not to be working on windows.


build-support/src/main/kotlin/com/squareup/wire/buildsupport/WireBuildPlugin.kt
```diff
@@ -110,19 +111,22 @@ class WireBuildPlugin : Plugin<Project> {
   }
 
   private fun Project.configureCommonSpotless() {
+    // CI runs on Windows as well, so we make sure Spotless is using the correct paths.
+    val slash = Path.DIRECTORY_SEPARATOR
```
JakeWharton, line 115: This is also available as `File.pathSeparator`, no dependency needed

JakeWharton (approved):

oldergod:
The problem wasn't about the slash but my path being wrong. I updated it seems good now.

PR #2706, opened 2023-11-13
Swift: Fix for @ProtoDefaulted

lickel:
Across module boundaries, `Message.fields` is often (always?) empty. When that happens, we can't correctly know that we want `@ProtoDefaulted` or not. Given it's unknowable in the current context, it chooses not to emit them.


wire-runtime-swift/src/test/swift/ProtoEnumTests.swift
```diff
@@ -0,0 +1,72 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import Foundation
+
+import Foundation
+import XCTest
+@testable import Wire
+
+final class ProtoEnumTests: XCTestCase {
```
lickel, line 22: These are new tests for other work that happened in v4.9.2.

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -186,7 +186,8 @@ class SwiftGenerator private constructor(
       if (type == ProtoType.ANY) return false
       if (isMessage) {
         val messageType = schema.getType(type!!) as MessageType
-        return messageType.supportsEmptyInitialization
+
+        return messageType.supportsEmptyInitialization && messageType.fields.isNotEmpty()
```
dnkoutso, line 190: This is the actual fix/change. I believe it is the correct fix overall if for any reason we cannot load the `fields` of a message we do not know whether we can provide a default value for it.

oldergod (approved):

dnkoutso (approved):

PR #2710, opened 2023-11-15
Bump to Kotlin 1.9.21

efirestone:


build-support/build.gradle.kts
```diff
@@ -79,8 +79,6 @@ allprojects {
   tasks.withType<KotlinCompile>().configureEach {
     kotlinOptions {
       jvmTarget = "11"
-      // Disable optimized callable references. See https://youtrack.jetbrains.com/issue/KT-37435
-      freeCompilerArgs += "-Xno-optimized-callable-references"
```
efirestone, line 83: According to console messages, this option is no longer supported.

wire-grpc-client/api/wire-grpc-client.api
```diff
@@ -60,6 +60,14 @@ public final class com/squareup/wire/GrpcMethod {
 	public final fun getResponseAdapter ()Lcom/squareup/wire/ProtoAdapter;
 }
 
+public final class com/squareup/wire/GrpcResponse {
```
efirestone, line 63: The biggest change in this PR is to create this wrapper adapter around `okhttp3.Response`. The reasoning is that with Kotlin 1.9.20, an `actual` implementation can't provide default argument values within an `actual` method implementation when using typealiases to fulfill the implementation. The definition for `Response.header(…)` includes the parameter `defaultValue: String? = null`, and so we can no longer do `actual type alias GrpcResponse = okhttp3.Response`.

Instead we create a concrete `GrpcResponse` class within the JVM platform implementation which wraps `okhttp3.Response`.

wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/BlockingMessageSource.kt
```diff
@@ -66,7 +66,7 @@ internal class BlockingMessageSource<R : Any>(
         queue.put(Failure(e))
       }
 
-      override fun onResponse(call: Call, response: GrpcResponse) {
+      override fun onResponse(call: Call, response: Response) {
```
efirestone, line 69: Previously `okhttp3.Response` and `GrpcResponse` were interchangeable within the JVM implementation. Now they're not a typealias of each other, so we need to explicitly reference the `okhttp3` version.

wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/platform.kt
```diff
@@ -15,12 +15,27 @@
...
 package com.squareup.wire.internal
 
+import com.squareup.wire.GrpcResponse
 import java.lang.reflect.Method
+import okio.IOException
 import okio.Sink
 import okio.Source
 import okio.gzip
 
-actual typealias Call = okhttp3.Call
```
efirestone, line 23: This typealias no longer works either, since `okhttp3.Call` references `okhttp3.Response`, which is no longer our implementation of `GrpcResponse`.

wire-runtime/build.gradle.kts
```diff
@@ -12,6 +13,7 @@ kotlin {
   jvm {
     withJava()
   }
+  jvmToolchain(11)
```
efirestone, line 16: 1.9.20 is pickier about the JVM compiler and the kotlin compiler being the same, so some of these have been made explicit so that they match.

JakeWharton:
There should be a .21 this week that we should wait for. Native libraries are badly broken in .20.

oldergod (approved):
I'm afraid we're gonna break all cash-app services............................... but not sure we can avoid it. I should try internally on a few maybe

wire-grpc-client/build.gradle.kts
```diff
@@ -1,11 +1,11 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
```
JakeWharton, line 1: Let's revert all changes to this file. It's a style tweak and an unused import so shouldn't be needed.

wire-runtime/build.gradle.kts
```diff
@@ -6,9 +6,7 @@ plugins {
 }
 
 kotlin {
-  jvm {
-    withJava()
-  }
+  jvm().withJava()
```
JakeWharton, line 9: Revert this too

JakeWharton (approved):

wire-grpc-client/src/commonMain/kotlin/com/squareup/wire/GrpcResponse.kt
```diff
@@ -19,14 +19,13 @@ package com.squareup.wire
...
 import com.squareup.wire.internal.addSuppressed
 import kotlin.jvm.JvmName
-import kotlin.jvm.JvmOverloads
 import okio.IOException
 
 expect class GrpcResponse {
   @get:JvmName("body")
   val body: GrpcResponseBody?
 
-  @JvmOverloads fun header(name: String, defaultValue: String? = null): String?
```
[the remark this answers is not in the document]
JakeWharton: Wouldn't the binary dump tool have highlighted this change if it was needed? Or is it not applied to this module?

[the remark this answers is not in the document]
oldergod: I think the binary tool only reflect JVM APIs. Since we were typealiasing `okHttp3.response`, it didn't print `GrpcResponse` until now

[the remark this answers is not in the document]
oldergod: We're not implementing anything on js/native so let's forget about it now

build-support/build.gradle.kts
```diff
@@ -23,7 +23,7 @@ buildscript {
 plugins {
   `kotlin-dsl`
   `java-gradle-plugin`
-  kotlin("jvm") version "1.9.10"
+  kotlin("jvm") version "1.9.21"
```
JakeWharton, line 26:
```suggestion
  kotlin("jvm") version "1.9.21"
```

JakeWharton (approved):

oldergod:
I'll make a release before merging this one

oldergod:
I wanna see if there is a less breaking way to upgrade

chris-hatton:
I'm keen on this as the basis for attempting to implement support for a WASM/JS target.
Would it be trivial enough to bump this PR to `1.9.22` to ensure access to the most recent WASM/JS developments?

oldergod:
Done in #2820

PR #2715, opened 2023-11-16
[swift] Fix decoding and encoding of negative numbers.

dnkoutso:


wire-runtime-swift/src/main/swift/ProtoCodable/ProtoIntCodable+Implementations.swift
```diff
@@ -48,7 +48,12 @@ extension Int32: ProtoIntCodable, ProtoDefaultedValue {
             writer.writeVarint(zigZagEncoded())
         case .variable:
             // int32 fields
-            writer.writeVarint(UInt32(bitPattern: self))
+            if (self >= 0) {
+                writer.writeVarint(UInt32(bitPattern: self))
+            } else {
+                // Must sign-extend.
+                writer.writeVarint(UInt64(bitPattern: Int64(self)))
```
dnkoutso, line 55: This is the breaking change in encoding in which we now correctly encode negative values as unsigned int 64 as per protobuf spec.

wire-runtime-swift/src/main/swift/ProtoCodable/ReadBuffer.swift
```diff
@@ -116,44 +116,8 @@ extension ReadBuffer {
         return UInt64(littleEndian: value)
     }
 
-    /**
-     * Reads a raw varint from the stream. If larger than 32 bits, discard the upper bits.
-     */
-    func readVarint32() throws -> UInt32 {
```
JakeWharton, line 122: It still may be worth keeping this function around but changing its implementation to `return UInt32(truncatingIfNeeded: readVarint64())`. This does two things:
- Ensures that callsites are a bit more self-documenting as to what they expect the bit-width of the varint to be.
- Provides an optimization point should we eventually discover this is a hotspot. Wire's Kotlin has an optimization for this variant of varint (heh). I'm not saying we need to do the optimization eagerly for Swift, but if we inline this into all the callsites then doing the optimization later means going and having to find all those callsites.

We could even mark it as inline or whatever. Is that a thing? I've forgotten all my Swift...

JakeWharton (approved):

lickel (approved):

wire-runtime-swift/src/main/swift/ProtoCodable/ReadBuffer.swift
lickel: You can mark as inline, but my reason to avoid doing so was that it might be a problem when you _want_ Int32, but go through the UInt32 version. 

We would need to write tests to verify that the truncation of eg -5 and -2^31 behave sanely

dnkoutso: > It still may be worth keeping this function around but changing its implementation to `return UInt32(truncatingIfNeeded: readVarint64())`

Yes I actually considered the same thing.

efirestone: In cases where we know the value is unsigned, such as for encoded message lengths (probably the most common case by a long shot), this is existing implementation is still likely to be slightly faster since there's no need to allocate a UInt64, just to then truncate it.

It would be nice to profile this quickly to verify (even by just running one of the tests 10,000X with both implementations and comparing). If the difference is negligible, then I'm all for just having the one parsing implementation.

If there's a difference between the two implementations, then I think I'm inclined to keep this but maybe rename it to something like `readUnsignedVarint64()` and make sure the documentation is clear on what it can be used for. Then we'd use that method in places like parsing `.lengthEncoded` messages and `readFieldKey()`.

efirestone (approved):
Great find and thanks for fixing. One thought, but nothing blocking at all.

oldergod:
Do the modified tests cover everything or is there now a new case we support and if yes, let's add a test

dnkoutso:
@oldergod yes and no. We were able to reproduce the actual poblem with the existing tests and some of the existing tests catch the encoding issue.

Whats not present here is the roundtrip from `protoc` <> `Wire` for encode/decode dance. I think we are OK as it is with existing test coverage as I am not sure how much effort it is to add the round trip tests.

oldergod:
Sounds good!

wire-runtime-swift/src/main/swift/ProtoCodable/ReadBuffer.swift
dnkoutso: +1, will measure it!

wire-runtime-swift/src/main/swift/ProtoCodable/ProtoIntCodable+Implementations.swift
```diff
@@ -104,10 +109,10 @@ extension Int64: ProtoIntCodable, ProtoDefaultedValue {
             self = try Int64(bitPattern: reader.readFixed64())
         case .signed:
             // sint64 fields
-            self = try reader.readVarint64().zigZagDecoded()
+            self = try reader.readVarint().zigZagDecoded()
         case .variable:
             // int64 fields
-            self = try Int64(bitPattern: reader.readVarint64())
+            self = try Int64(truncatingIfNeeded: reader.readVarint())
```
dnkoutso, line 115: TODO: I think this needs to be `bitPattern` when converting from UInt64 to Int64.

wire-runtime-swift/src/main/swift/ProtoCodable/ReadBuffer.swift
dnkoutso: @efirestone I measured it like this:

```swift
    func testDecodeVarintInt32() throws {
        let data = Foundation.Data(hexEncoded: "08_FBFFFFFFFFFFFFFFFF01")!
        measureMetrics([.wallClockTime], automaticallyStartMeasuring: false) {
            do {
                try test(data: data) { reader in
                    let value = try reader.decode(tag: 1) {
                        self.startMeasuring()
                        let value = try reader.decode(Int32.self, encoding: ProtoIntEncoding.variable)
                        self.stopMeasuring()
                        return value
                    }
                    XCTAssertEqual(value, -5)
                }
            } catch {
                XCTFail("Exception thrown during performance test.")
            }
        }
    }
```

I got a before and after measurements after 1000 repetitions. I did not see any difference between the two at all.

The total runtime of 1000s after was 260s vs the before was 261s, even faster but honestly could be attributed to various random factors.

Feel free to inspect this tiny html page I generated for it.
[wire_encode_measure.zip](https://github.com/square/wire/files/13425441/wire_encode_measure.zip)

Does that cover your concern around performance metrics with this change? Anything you'd like me to look into further?

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ Version 4.9.3
 
 _Unreleased_
 
+* Breaking: Encoding and decoding negative ints was broken in Swift and did not match protobuf spec.
```
lickel, line 9: - [ ] Change "ints" to "`int32`s" (and back ticked appropriately)

dnkoutso: fixing and will update more on this.

dnkoutso: Updated and included more info

wire-runtime-swift/src/main/swift/ProtoCodable/ReadBuffer.swift
dnkoutso: ping @efirestone if you have time to take a look at this!

dnkoutso: Going to go ahead and merge this and proceed so we can have a release to fix the issue. It seems performance is not affected by this change.

PR #2716, opened 2023-11-17
Enable the use of transitive dependencies in protoPath

frojasg:
* Enable the use of transtive dependencies by default in protoPath. This will allow users to transtively add needed proto files to Wire.
* Add new project to validate the behavior


frojasg:
Oooops. I started my changes in a old branch. I will create a new PR

wire-library/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -56,7 +56,7 @@ class WirePlugin : Plugin<Project> {
     project.configurations.create("protoPath")
       .also {
         it.isCanBeConsumed = false
-        it.isTransitive = false
+        it.isTransitive = true
```
oldergod, line 59: Won't this pull-in all dependencies? like if we depend on `my.service:client:version`, we could get the `.proto` files but as well, all `.class`, etc?

Issue #2695, opened 2023-10-25
502 status code results in receiver hanging as channel is not closed


chrisatweave:
I am using the latest stable version `4.9.1`.

When a server returns a [502](https://github.com/square/wire/blob/f0245a2d59fada53dfa5251d7697e6b2e41c86a9/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/grpc.kt#L182) the [checkGrpcResponse](https://github.com/square/wire/blob/f0245a2d59fada53dfa5251d7697e6b2e41c86a9/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/grpc.kt#L173) throws an exception that is not caught. Resulting in the my `receiver.consumeEach` hanging.

I found issue [2038](https://github.com/square/wire/pull/2038) that has a very similar fix, it seems that [response.messageSource(responseAdapter)](https://github.com/square/wire/blob/f0245a2d59fada53dfa5251d7697e6b2e41c86a9/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/grpc.kt#L101) is not within the try catch that calls `close(exeption)` close the channel, resulting in the receiver hanging.

My code is pretty simple

```
val (sender, receiver) = serviceCall.executeIn(this)

// Dispatch data to sync
data.forEach { sender.send(it) }
sender.close()

// Process sync results
receiver.consumeEach { results.add(it) }
```

`receiver.consumeEach` in this case hangs forever as the exception never closes the receiver.

It seems the Worker thread is killed without notifying the caller and the channel is left open.

```
Callback failure for call to [ommited]...
java.io.IOException: expected gRPC but was HTTP status=520, content-type=null
  	at com.squareup.wire.internal.GrpcKt.checkGrpcResponse(grpc.kt:187)
  	at com.squareup.wire.internal.GrpcKt.messageSource(grpc.kt:173)
  	at com.squareup.wire.internal.GrpcKt$readFromResponseBodyCallback$1$onResponse$1.invokeSuspend(grpc.kt:101)
  	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
  	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:108)
  	at kotlinx.coroutines.EventLoopImplBase.processNextEvent(EventLoop.common.kt:280)
[...]

PR #2718, opened 2023-11-22
Add test for streaming hanging when unexpected response is received

sergei-lapin:
The reproducer for https://github.com/square/wire/issues/2695


oldergod:
I've push a fix, let me know what you think

oldergod (approved):

PR #2744, opened 2023-12-11
[swift] Further guard against incomplete schema for default value calculation

dnkoutso:
When invoking Wire against a single proto, if that proto depends on another proto the Wire schema wont link the second proto. This means that when we try to calculate the default value to use for a potential type ProtoA refers in ProtoB we crash because `encodeMode` is not loaded.

```kotlin
  // Null until this field is linked.
  var encodeMode: EncodeMode? = null
    private set
```

A general fix is for Wire to perhaps "soft"-link proto dependencies when invoking it against a single proto. I do not know much about the linker and how it works, for now this prevents a crasher.

### DIFF
--- a/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
+++ b/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
@@ -165,7 +165,7 @@ class SwiftGenerator private constructor(
       // These types' fields aren't properly loaded
       false
     } else {
-      fields.isEmpty() || fields.all { !it.isRequiredParameter }
+      fields.isEmpty() || fields.all { it.encodeMode != null && !it.isRequiredParameter }
     }
 
   private val EnumType.supportsEmptyInitialization: Boolean


JakeWharton:
The linker ensures you have a correct view of your proto. Wire was never designed to parse protos without a complete view of the graph of messages. The linker validates that, and constructs references so when you ask for things like the type of a field or its default a trustworthy value is always available. Relaxing this, at least viewed in isolation, does not seem like a good idea.

lickel (approved):

dnkoutso:
Would it be possible to generally execute Wire and only produce a set of files for a given proto while having the schema loaded? Is that what `includes` does? It seems SwiftGenerator does not have `includes` and `roots` is not the same as if I pass `roots` I also get output for that module. 

We are looking to have significant granularity in our protos produced by Wire to avoid bottlenecks in our dependency graph and it is something that Bazel generally recommends to go after...

JakeWharton:
The manifest support that was built for Swift initially somewhat does this, but it doesn't operate by file. It operates on types. I suppose you could find the types by enumerating those within the desired files once the schema was linked. Beyond that it allows partitioning the code generation in a way that forms a dependency tree. This was initially meant to be done manually using the manifest, but there's certainly a way that it could be done programmatically using the existing bazel modules as a 1:1 mapping.

dnkoutso:
Based on Jake's comments I am exploring further how we invoke Wire in our monorepo to see if this issue can overall be circumvented.

PR #2747, opened 2023-12-12
Add `ProtoReader` API for length-delimited stream reads

### LINKED ISSUE #633: How to read length-delimited messages?
This may be a dumb question, but I can't figure out the answer by looking at the runtime or generated source code. I have a Bluetooth connection on Android over which I'm sending length-delimited messages (`parseDelimitedFrom()` and `writeDelimitedTo()` in Google's protobuf API). With Wire, writing looks like this:

``` java
writer.writeVarint32(Msg.ADAPTER.encodedSize(msg));
Msg.ADAPTER.encode(writer, msg);
sink.flush();
```

How do I read the response? If I call `Msg.ADAPTER.decode(reader)`, it immediately tries to read the next tag without figuring out how long the message is, and ProtoReader doesn't expose any way for me to limit the message length manually (and it's all `final`, so I can't add this functionality myself).

I don't want to read the whole message into a byte[] and decode from there because my messages may be 4-8 KB in size and it is more efficient to decode them as the data is being received instead of waiting until the entire message is buffered in memory.

It looks like my best option right now is to implement a wrapper around BufferedSource, read the delimiter manually, and have exhausted() return true when the limit is reached. That's a lot of work for basic protobuf functionality. Am I missing some existing way of doing this?

JakeWharton:
Closes #633

### DIFF
--- a/wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoReader.kt
+++ b/wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoReader.kt
@@ -131,6 +131,28 @@ class ProtoReader(private val source: BufferedSource) {
     endMessageAndGetUnknownFields(token)
   }
 
+  /** Reads and returns the length of the next message in a length-delimited stream. */
+  @Throws(IOException::class)
+  fun nextLengthDelimited(): Int {
+    check(state == STATE_TAG || state == STATE_LENGTH_DELIMITED) {
+      "Unexpected call to nextDelimited()"
+    }
+    return internalNextLengthDelimited()
+  }
+
+  private fun internalNextLengthDelimited(): Int {
+    nextFieldEncoding = FieldEncoding.LENGTH_DELIMITED
+    state = STATE_LENGTH_DELIMITED
+    val length = internalReadVarint32()
+    if (length < 0) throw ProtocolException("Negative length: $length")
+    if (pushedLimit != -1L) throw IllegalStateException()
+    // Push the current limit, and set a new limit to the length of this value.
+    pushedLimit = limit
+    limit = pos + length
+    if (limit > pushedLimit) throw EOFException()
+    return length
+  }
+
   /**
    * Reads and returns the next tag of the message, or -1 if there are no further tags. Use
    * [peekFieldEncoding] after calling this method to query its encoding. This silently skips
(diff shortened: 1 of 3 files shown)


wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoReader.kt
```diff
@@ -131,6 +131,28 @@ class ProtoReader(private val source: BufferedSource) {
     endMessageAndGetUnknownFields(token)
   }
 
+  /** Reads and returns the length of the next message in a length-delimited stream. */
+  @Throws(IOException::class)
+  fun nextLengthDelimited(): Int {
+    check(state == STATE_TAG || state == STATE_LENGTH_DELIMITED) {
+      "Unexpected call to nextDelimited()"
+    }
```
JakeWharton, line 139: It's not clear this is useful. Could eliminate and only have the one function.

wire-runtime/src/commonTest/kotlin/com/squareup/wire/ProtoReaderTest.kt
```diff
@@ -32,4 +32,25 @@ class ProtoReaderTest {
     assertEquals(-1, reader.nextTag())
     reader.endMessageAndGetUnknownFields(token)
   }
+
+  @Test fun lengthDelimited() {
+    val encoded = "0208020608ffffffff07".decodeHex()
```
oldergod, line 37:
```suggestion
    // Bytes representing:
    // ── 0 ┐
    //      ╰- 0 ┐
    //           ╰- 1: 2147483647
    val encoded = "0208020608ffffffff07".decodeHex()
```

wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoReader.kt
oldergod: Could we make it clearer by telling which state we're in? I don't mind whether we keep it or not

oldergod (approved):

wire-runtime/src/commonTest/kotlin/com/squareup/wire/ProtoReaderTest.kt
JakeWharton: This is not a single proto. It's two protos prefixed with their varint32-encoded lengths. I'll add a comment.

wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoReader.kt
JakeWharton: The message is the same as other places in this file. I can update them all separately to include the current state.

wire-runtime/src/commonTest/kotlin/com/squareup/wire/ProtoReaderTest.kt
oldergod: sweet

PR #2754, opened 2023-12-15
Ensure ConsoleWireLogger does not print anything if in quiet mode

dnkoutso:


dnkoutso:
Actually we found out `--quiet` is gone and in lieu of it we can just pass an empty logger.

PR #2758, opened 2023-12-18
[swift] Update to SwiftPoet 1.6.1

dnkoutso:
SwiftPoet 1.6.1 includes this one fix https://github.com/outfoxx/swiftpoet/pull/97. Because SwiftPoet handles imports much better now and has fixed quite a few, if not all, of edge cases we no longer need Wire's custom import support system that we have added in order to circumvent various issues.

- [x] To be super safe we will first do an internal release 4.9.4-square02 that includes SwiftPoet 1.6.0 and then re-do another one with 1.6.1 and this PR.


wire-runtime-swift/src/test/swift/sample/Period.swift
```diff
@@ -1,6 +1,5 @@
 // Code generated by Wire protocol buffer compiler, do not edit.
 // Source: squareup.geology.Period in squareup/geology/period.proto
-import Foundation
```
dnkoutso, line 3: This is OK as our custom import system within Wire was always assuming Foundation is needed...

oldergod (approved):
Nice to see all that removed code

wire-tests-swift/no-manifest/src/main/swift/AllTypes.swift
```diff
@@ -7,11 +7,11 @@ import Wire
 public struct AllTypes {
 
     @CopyOnWrite
-    private var storage: AllTypes.Storage
+    private var storage: Storage
```
dnkoutso, line 10: This is also OK since SwiftPoet takes care of this now.

lickel (approved):

PR #2764, opened 2023-12-19
[swift] Always qualify names for types that conflict with common Swift types

dnkoutso:
In rare occasions a proto file could declare a type name that collides with one of Swift's standard types from Foundation or stdlib. 

There is a chance this can be fixed in SwiftPoet by building in this mechanism into it instead of Wire but for now this is a cheap lookup Wire can do.

Added a test case that fails without this change:

```
/Users/dimitris/Development/wire/wire-tests-swift/manifest/module_two/SwiftModuleTwoMessage.swift:11:16: warning: stored property 'order' of 'Sendable'-conforming struct 'SwiftModuleTwoMessage' has non-sendable type '<<error type>>'
    public var order: SortOrder?
               ^
```


wire-tests-swift/manifest/module_two/SwiftModuleTwoMessage.swift
```diff
@@ -8,6 +8,7 @@ public struct SwiftModuleTwoMessage {
 
     @ProtoDefaulted
     public var name: String?
+    public var order: module_one.SortOrder?
```
dnkoutso, line 11: Without this fix this would output `public var order: SortOrder?` which fails to compile.

lickel (approved):

ccclayton (approved):

.github/workflows/build.yml
```diff
@@ -134,7 +134,7 @@ jobs:
         # This builds the wire-swift-runtime, verifies that generated protos match what's currently
         # checked into the repo both for the runtime itself and for its tests, and runs the tests.
         run: |
-          ./gradlew :wire-runtime-swift:build --stacktrace --warning-mode all
```
oldergod, line 137: @dnkoutso please put them back next time. They help when things go wrong, and I don't think we gain much by taking them out?

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -1633,6 +1633,40 @@ class SwiftGenerator private constructor(
 
     fun builtInType(protoType: ProtoType): Boolean = protoType in BUILT_IN_TYPES.keys
 
+    private val SWIFT_COMMON_TYPES = setOf(
```
oldergod, line 1636: Not something we want swift poet to do?

oldergod: That's what the name allocator is for

.github/workflows/build.yml
dnkoutso: its extremely hard to scroll through to find a build failure like I did include them. 

Just take a look at a sample here:

Screenshot 2023-12-20 at 12 39 04 PM

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
dnkoutso: In the PR description I wrote:

> There is a chance this can be fixed in SwiftPoet by building in this mechanism into it instead of Wire but for now this is a cheap lookup Wire can do.

If NameAllocator is the correct path to go then yes SwiftGenerator still does not use it so we could delete this in the near future when NameAllocator is used.

dnkoutso: I believe once this TODO is resolved https://github.com/outfoxx/swiftpoet/blob/main/src/main/java/io/outfoxx/swiftpoet/NameAllocator.kt#L121-L123 we can adopt the SwiftPoet NameAllocator here.

.github/workflows/build.yml
oldergod: `ctrl+f` mate

PR #2776, opened 2024-01-07
Provide a built in public EmptyWireLoggerFactory class

dnkoutso:
It seems there is no `--quiet` option in Wire because it is probably obviated by the usage of `--logger_factory_class` argument.

This PR makes it easy for scripts that use Wire not to have to provide their own JAR for a built in "quiet" mode by exposing an empty logger factory and implementation. This now simply works by adding `"--logger_factory_class=com.squareup.wire.schema.EmptyWireLoggerFactory"` when invoking Wire.

### DIFF
--- a/wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt
+++ b/wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt
@@ -937,7 +937,7 @@ class WireRunTest {
       context = SchemaHandler.Context(
         fileSystem = fs,
         outDirectory = "out".toPath(),
-        logger = NULL_LOGGER,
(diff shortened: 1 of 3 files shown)


wire-compiler/src/test/java/com/squareup/wire/schema/WireRunTest.kt
```diff
@@ -1717,24 +1717,4 @@ class WireRunTest {
       logs.add("unusedExcludesInTarget: $unusedExcludes")
     }
   }
-
-  class CustomLoggerFactory : WireLogger.Factory {
```
dnkoutso, line 1721: This was unused so it was deleted.

oldergod:
sweet, thank you

PR #2780, opened 2024-01-11
FileDescriptor generation - descriptor maps chunks

chris-ryan-square:
With large protos, it is possible to have the generated DescriptorMap be too large and break the compiler with a message similar to:

>  org.jetbrains.org.objectweb.asm.MethodTooLargeException: Method too large: .../XXXXServiceWireGrpc.<clinit>

This PR splits the descriptor map into smaller chunks, with each created in their own function to avoid the Method too large.

### DIFF
--- a/wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/FileDescriptorGenerator.kt
+++ b/wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/FileDescriptorGenerator.kt
@@ -86,7 +86,7 @@ object FileDescriptorGenerator {
     encodedListChunked.forEachIndexed { index: Int, subEncodedList: List<Pair<String, String>> ->
       subDescriptorMapCodeBlock(builder, subEncodedList, index)
       if (index > 0) {
-        initLines.add(" +\n")
+        initLines[index - 1] += " +\n"
       }
       initLines.add("$DESCRIPTOR_MAP_FUNCTION_PREFIX$index()")
     }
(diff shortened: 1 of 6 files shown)


wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/FileDescriptorGenerator.kt
```diff
@@ -67,19 +75,53 @@ object FileDescriptorGenerator {
...
-          encoded.toList().fold(CodeBlock.builder().addStatement("mapOf(")) { b, (name, data) ->
+          descriptorMapInitializer.build(),
+        ).build(),
+    )
+  }
+
+  private fun subDescriptorMapCodeBlock(
+    builder: TypeSpec.Builder,
+    encodedList: List<Pair<String, String>>,
+    index: Int = 0,
```
wsutina, line 116: Just a minor comment

Is the default value needed? Since we always know the index, can we drop it `0`?

chris-ryan-square: Yeah, I'll do that

wsutina:
## 🧪 Local Testing 
We tested this locally by publishing it to maven local `./gradlew publishToMavenLocal` and used it in our project.

### Test results
- XXXXServiceWireGrpc is generated as expected ✅ 
```kotlin
public object XXXServiceWireGrpc {
  private val descriptorMap: Map<String, DescriptorProtos.FileDescriptorProto> =
        createDescriptorMap0() +
    createDescriptorMap1() +
    createDescriptorMap2() +
    ...
    createDescriptorMap16()
```
- No longer see `Method too large` error and our project built successfully ✅

swankjesse (approved):

PR #2781, opened 2024-01-16
Implement equals/hash for Duration and Instant on native and JS

### LINKED ISSUE #2724: com.squareup.wire.Instant should implement equals()/hashCode()
com.squareup.wire.Instant has a overridden `equals()/hashCode()` implementation on the JVM and a default implementation on non-JVM.

This breaks `equals()` of the generated classes. For example, this test passes on the JVM and fails on native (linuxX64):

```protobuf
// pizza.proto:
syntax = "proto3";
import "google/protobuf/timestamp.proto";

message Pizza {
  google.protobuf.Timestamp created_at = 1;
}
```

```kotlin
@Test
fun testPizza() {
    val pizzaPayload = Pizza(created_at = ofEpochSecond(1701191500, 0)).encode()
    val pizza1 = Pizza.ADAPTER.decode(pizzaPayload)
    val pizza2 = Pizza.ADAPTER.decode(pizzaPayload)
    assertEquals(pizza1, pizza2)
}
```

oldergod:
fixes #2724

### DIFF
--- a/wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/Internal.kt
+++ b/wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/Internal.kt
@@ -19,7 +19,9 @@
 
 package com.squareup.wire.internal
 
+import com.squareup.wire.Duration
 import com.squareup.wire.FieldEncoding
+import com.squareup.wire.Instant
 import com.squareup.wire.ProtoAdapter
 import com.squareup.wire.ProtoReader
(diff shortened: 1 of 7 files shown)


Egorand (approved):
Is it possible to share code by e.g. declaring extension functions in the `common` module and calling them from `js` and `native`? There's not a ton of code duplication though, so not a big deal.

oldergod:
Shared the code!

Egorand:
:sweet-kotlin:

PR #2785, opened 2024-01-17
Add escapeKotlinKeywords parameter to Kotlin target

Egorand:


oldergod:
👍 probably need to `apiDump`

oldergod (approved):
🚀

PR #2788, opened 2024-01-19
[swift] clean up unneeded storage visibility

dnkoutso:


wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -105,8 +104,6 @@ class SwiftGenerator private constructor(
 
   private val deprecated = AttributeSpec.builder("available").addArguments("*", "deprecated").build()
 
-  private val storageVisibility: Modifier = PUBLIC
```
dnkoutso, line 108: No need to do:
```kotlin
    val visibility: Modifier = if (forStorageType) {
      storageVisibility
    } else {
      PUBLIC
    }
```

as this is always `PUBLIC`.

lickel (approved):
Do you want to also do that global formatting you had done in our pairing?

dnkoutso:
I thought about it but it seems that I might be triggering spotless if I do this. I think there is an inconsistency between my IDE settings and overall formatting for the project, for now skipping.

PR #2792, opened 2024-01-22
Extract MessageType.isExtensible as extension function in SwiftGen

oldergod:
I would like to reduce the API changes within wire-schema as much as possible A bit like  Wire Swift was evolving into its own repo; same applies for Wire Java and Wire Kotlin.
Extracting the method created here https://github.com/square/wire/pull/2790

### DIFF
--- a/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/MessageType.kt
+++ b/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/MessageType.kt
@@ -50,9 +50,6 @@ data class MessageType(
 ) : Type() {
   private var deprecated: Any? = null
 
-  val isExtensible: Boolean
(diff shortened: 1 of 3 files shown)


lickel (approved):

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/MessageType.kt
```diff
@@ -50,9 +50,6 @@ data class MessageType(
 ) : Type() {
   private var deprecated: Any? = null
 
-  val isExtensible: Boolean
```
lickel, line 53: Fwiw longer term, Kotlin could do the same extension stuff as swift is planning.

dnkoutso (approved):

Issue #807, opened 2019-01-20
Add the @Generated annotation


nielsz:
If we add this annotation to a class, we give a signal that we didn't write this code ourselves. 
If we then (for example) push this code to Sonarqube, they won't give any errors on generated code that we can't change.

https://docs.oracle.com/javase/8/docs/api/javax/annotation/Generated.html

PR #2794, opened 2024-01-24
New API: extraTypeAnnotations

swankjesse:
The goal of this new API is to allow users to specify extra annotations that'll apply to the Wire-generated classes. This is specifically useful for the @Immutable annotation offered by Jetpack Compose, and could be useful for other annotations.

### DIFF
--- a/docs/wire_compiler.md
+++ b/docs/wire_compiler.md
@@ -492,6 +492,9 @@ wire {
 
     // True for emitted services to implement one interface per RPC.
     singleMethodServices = false
+
+    // Fully-qualified type names of annotations to be added to each generated type.
+    extraTypeAnnotations = listOf<String>()
   }
 }
 ```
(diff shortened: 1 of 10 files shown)


JakeWharton:
The Compose compiler allows you to provide a list of external types that are stable. So if one provided Wire's `Message` I think you wouldn't need this feature (Compose will traverse the hierarchy of a type). Unless there are actually other use cases in mind.

oldergod:
Coil did it a few weeks ago https://github.com/coil-kt/coil/pull/1962/files

oldergod:
Other usecase would be https://github.com/square/wire/issues/807

swankjesse:
I’m going to shelve this for now because I believe the Compose configuration file can solve my immediate problem.

I think there’s potentially something useful for Wire to do to hint to Compose about what’s immutable.

autonomousapps:
> Other usecase would be #807

JaCoCo will automatically excluded generated files from code coverage reports.

PR #2797, opened 2024-01-25
Prune imports no longer used by retained entities in the schema

mpeyper:
This is a WIP but wanted to get it up to get some feedback on the approach. This started with a discussion with @swankjesse after I discovered that imports are not necessarily being pruned from files when they are no longer needed by retained entities in the schema.

The existing logic appear to retain an imports when the file is being retained in the schema at all, regarless of whether the import is still being used by the file (and not by another file).

This change attempts to resolve the referenced types in the file and resolve them back to their import locations so that the list of actually required imports can be determined.

It's currently passing all existing test cases as well as the new test we added to replicate the issue, but I'm not sure if I'm capturing all ways a type can be referenced at the moment. Any insights into wierd edge cases that I should consider are very welcome.

I'll add some comments in the files where I'm not sure about things.

### DIFF
--- a/wire-grpc-server-generator/src/test/java/com/squareup/wire/kotlin/grpcserver/FileDescriptorGeneratorTest.kt
+++ b/wire-grpc-server-generator/src/test/java/com/squareup/wire/kotlin/grpcserver/FileDescriptorGeneratorTest.kt
@@ -46,7 +46,9 @@ internal class FileDescriptorGeneratorTest {
       |package test;
       |import "imported.proto";
       |
-      |message Test {}
+      |message Test {
+      |  optional Imported imported = 1;
+      |}
       |
         """.trimMargin(),
       ),
--- a/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
+++ b/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
@@ -129,29 +129,51 @@ data class ProtoFile(
     return result
   }
 
-  private fun referencedTypes(schema: Schema): List<Type> {
+  /**
+   * Returns all types references by types and subtypes defined in the proto file.
+   */
+  private fun referencedTypes(): List<ProtoType> {
     val types = typesAndNestedTypes()
     val messages = types.filterIsInstance<MessageType>()
+    val enums = types.filterIsInstance<EnumType>()
 
-    val messageTypes = messages
+    val messageFieldTypes = messages
       .flatMap { it.fieldsAndOneOfFields }
       .mapNotNull { it.type }
-
-    val messageOptions = messages.map { it.options }
-    val fieldOptions = messages.flatMap { it.fieldsAndOneOfFields.map { it.options } }
-    val oneOfOptions = messages.flatMap { it.oneOfs }.map { it.options }
-    val optionTypes = (messageOptions + oneOfOptions + fieldOptions).flatMap { it.fields().asMap().keys }
 
     val rpcTypes = services
       .flatMap { service -> service.rpcs }
       .flatMap { rpc -> listOfNotNull(rpc.requestType, rpc.responseType) }
 
-    return (messageTypes + optionTypes + rpcTypes).distinct().mapNotNull { schema.getType(it) }.toList()
+    val extendTypes = mutableListOf<ProtoType>().apply {
+      addAll(extendList.flatMap { it.fields }.mapNotNull { it.type })
+      addAll(types.flatMap { it.nestedExtendList }.flatMap { it.fields }.mapNotNull { it.type })
+    }
+
+    val optionTypes = mutableListOf<Options>().apply {
+      add(options) // file options
+      addAll(messages.map { it.options }) // message options
+      addAll(messages.flatMap { it.fields }.map { it.options }) // field options
+      addAll(messages.flatMap { it.oneOfs }.map { it.options }) // one-of options
+      addAll(enums.map { it.options }) // enum options
+      addAll(enums.flatMap { it.constants }.map { it.options }) // enum value options
+      addAll(services.map { it.options }) // service options
+      addAll(services.flatMap { it.rpcs }.map { it.options }) // method options
+    }.flatMap { it.fields().asMap().keys }
+
+    return (messageFieldTypes + rpcTypes + extendTypes + optionTypes).distinct().toList()
+  }
+
+  /** Returns a new proto file that omits unnecessary imports. */
+  @Suppress("unused") // left for backwards compatibility
+  fun retainImports(retained: List<ProtoFile>): ProtoFile {
+    val schema = Schema(retained)
+    return retainImports(schema)
   }
 
   /** Returns a new proto file that omits unnecessary imports. */
   fun retainImports(schema: Schema): ProtoFile {
-    val referencedTypes = referencedTypes(schema)
+    val referencedTypes = referencedTypes().mapNotNull { schema.getType(it) }
 
     val typeLocations = referencedTypes.map { type -> type.location }
 
@@ -164,9 +186,13 @@ data class ProtoFile(
 
     val retainedImports = imports.filter { referencedImports.contains(it) }
 
-    return if (imports.size != retainedImports.size) {
+    val protoFilesInSchema = schema.protoFiles.map { it.location.path }.toSet()
+
+    val retainedPublicImports = publicImports.filter { protoFilesInSchema.contains(it) }
+
+    return if (imports.size != retainedImports.size || publicImports.size != retainedPublicImports.size) {
       val result = ProtoFile(
-        location, retainedImports, publicImports, packageName, types, services,
+        location, retainedImports, retainedPublicImports, packageName, types, services,
         extendList, options, syntax,
       )
       result.javaPackage = javaPackage
--- a/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
+++ b/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
@@ -186,9 +186,12 @@ data class ProtoFile(
 
     val retainedImports = imports.filter { referencedImports.contains(it) }
 
-    val protoFilesInSchema = schema.protoFiles.map { it.location.path }.toSet()
+    val nonEmptyProtoFilesInSchema = schema.protoFiles
+      .filter { it.types.isNotEmpty() || it.services.isNotEmpty() || it.extendList.isNotEmpty() }
+      .map { protoFile -> protoFile.location.path }
+      .toSet()
 
-    val retainedPublicImports = publicImports.filter { protoFilesInSchema.contains(it) }
+    val retainedPublicImports = publicImports.filter { nonEmptyProtoFilesInSchema.contains(it) }
 
     return if (imports.size != retainedImports.size || publicImports.size != retainedPublicImports.size) {
       val result = ProtoFile(
@@ -242,10 +245,6 @@ data class ProtoFile(
         protoFileElement.publicImports, packageName, types, services, wireExtends, options,
         protoFileElement.syntax,
       )
-    }
-
-    private fun findProtoFile(protoFiles: List<ProtoFile>, path: String): ProtoFile? {
-      return protoFiles.find { it.location.path == path }
     }
   }
 }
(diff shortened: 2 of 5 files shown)


wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
```diff
@@ -129,24 +129,37 @@ data class ProtoFile(
     return result
   }
 
+  private fun referencedTypes(schema: Schema): List<Type> {
+    val types = typesAndNestedTypes()
+    val messages = types.filterIsInstance<MessageType>()
```
mpeyper, line 134: Is there a better way to get the message types defined in the proto file?

wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/PrunerTest.kt
```diff
@@ -1462,6 +1462,54 @@ class PrunerTest {
...
+   * even if those imports were unnecessary.
+   */
+
+  @Test
+  fun importPruningIsPrecise() {
+    val schema = buildSchema {
+      add(
+        "message.proto".toPath(),
+        """
+             |import 'footer.proto';
```
mpeyper, line 1476: Previously this import was being retained event the `AnotherMessage` was being pruned, along with the need to keep importing it.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
```diff
@@ -129,24 +129,37 @@ data class ProtoFile(
...
+
+    val messageTypes = messages
+      .flatMap { it.fieldsAndOneOfFields }
+      .mapNotNull { it.type }
+
+    val optionTypes = (messages.map { it.options } + messages.flatMap { it.oneOfs }.map { it.options }).flatMap { it.fields().asMap().keys }
```
mpeyper, line 140: Note to self, I need to clean this line up and make it more readable.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
```diff
@@ -129,24 +129,40 @@ data class ProtoFile(
...
+    val messageOptions = messages.map { it.options }
+    val fieldOptions = messages.flatMap { it.fieldsAndOneOfFields.map { it.options } }
+    val oneOfOptions = messages.flatMap { it.oneOfs }.map { it.options }
+    val optionTypes = (messageOptions + oneOfOptions + fieldOptions).flatMap { it.fields().asMap().keys }
+
+    val rpcTypes = services
+      .flatMap { service -> service.rpcs }
+      .flatMap { rpc -> listOfNotNull(rpc.requestType, rpc.responseType) }
+
+    return (messageTypes + optionTypes + rpcTypes).distinct().mapNotNull { schema.getType(it) }.toList()
```
mpeyper, line 149: * `messageTypes` are the field types defined in messages
* `optionTypes` are the option types for the message and fields, and one-ofs
  * The import paths of any option extensions are resolved below in `retainImports`
* `rpcTypes` are the request and response types defined in services

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
```diff
@@ -129,24 +129,40 @@ data class ProtoFile(
...
-    }
+  fun retainImports(schema: Schema): ProtoFile {
+    val referencedTypes = referencedTypes(schema)
+
+    val typeLocations = referencedTypes.map { type -> type.location }
+
+    val extensionLocations = referencedTypes
+      .filterIsInstance<MessageType>()
+      .flatMap { type -> type.extensionFields }
+      .map { field -> field.location }
```
mpeyper, line 161: With the existing test suite, The locations that have been important to retain are the location of a referenced type itself (`typeLocations`), and the location of any option extensions (`extensionLocations`).

Am I missing any other places I should be looking for locations?

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
```diff
@@ -129,24 +129,40 @@ data class ProtoFile(
...
+
+    val referencedImports = (typeLocations + extensionLocations).map { it.path }.toSet()
+
+    val retainedImports = imports.filter { referencedImports.contains(it) }
```
mpeyper, line 165: The logic here is that I want to ensure that I only remove existing imports and never add new ones.  I had a case where `google/protobuf/descriptor.proto` was being inserted into the imports if I simply used `referencedImports` as te new `retainedImports`.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
```diff
@@ -129,24 +129,40 @@ data class ProtoFile(
...
 
     return if (imports.size != retainedImports.size) {
       val result = ProtoFile(
```
mpeyper, line 168: It's actually on the next line, but I wonder if there is anything that _should_ happen with `publicImports`. Like, if the `public import` is no retained in the schema, perhaps we should prune it out of the file as well?

wire-grpc-server-generator/src/test/java/com/squareup/wire/kotlin/grpcserver/FileDescriptorGeneratorTest.kt
```diff
@@ -46,7 +46,9 @@ internal class FileDescriptorGeneratorTest {
       |package test;
       |import "imported.proto";
       |
-      |message Test {}
+      |message Test {
+      |  optional Imported imported = 1;
```
mpeyper, line 50: A side effect of this change is that `descriptor.dependencies.size` was 0 without this change becuase the unused import is getting pruned away. I'm not sure how much of an issue that is in reality though?

oldergod: no problem I think

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Pruner.kt
```diff
@@ -53,8 +53,9 @@ class Pruner(
   }
 
   private fun retainImports(protoFiles: List<ProtoFile>): List<ProtoFile> {
+    val schema = Schema(protoFiles)
     return protoFiles.map { protoFile ->
-      protoFile.retainImports(protoFiles)
+      protoFile.retainImports(schema)
```
oldergod, line 58: Could we do the transformation of `protoFiles` to `schema` in the `retainImports` method so the public API doesn't change?

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
oldergod: I think that's cool

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
oldergod: Not sure if we need to retain enum's and constants' options?

oldergod (approved):
🚀 LGTM

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
mpeyper: I remembered later that I found a list of a bunch of different option types somewhere. I’ll take a look at that and see if there’s others to add as well.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Pruner.kt
mpeyper: We definitely could. The main concern is we’d be doing the same transformation multiple times for each proto file, but I’m not sure how expensive that is and if it would matter.

Alternatively I could add an overload that takes the schema and have the existing api delegate to it to so the change is only additive instead of breaking?

oldergod: Yes, ok

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
```diff
@@ -129,24 +129,40 @@ data class ProtoFile(
     return result
   }
 
+  private fun referencedTypes(schema: Schema): List<Type> {
```
swankjesse, line 132: I think this should return `ProtoType` objects, not `Type` objects. Ie. the call to `mapNotNull()` should be the caller’s responsibility.

(Just to make this more general purpose)

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
swankjesse: Good call

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
swankjesse: Yeah, good idea

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Pruner.kt
swankjesse: Thumbs up to adding an overload

swankjesse (approved):

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
mpeyper: I went with `Type` for consistency with `typesAndNestedTypes`, but happy to change to `ProtoType` if that’s preferred?

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
mpeyper: I've update this now to look for all of [these option types](https://github.com/square/wire/blob/master/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Options.kt#L405-L419) (at least where I believe they would be in the model). Please confirm the new logic makes sense.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
mpeyper: I've added this, but will need to write some test cases around it.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Pruner.kt
mpeyper: Added and removed the breaking change.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
mpeyper: Done.

swankjesse: Thanks.

typesAndNestedTypes() doesn’t need to look things up in a schema because the types are defined within the file.

This thing doesn’t have that information so doing the schema lookup is extra work.

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
mpeyper: Test added.

mpeyper:
I've added a bunch of tests to test the various ways a type can be referenced and also to test the new `import public` pruning logic. I think I have all cases covered now and I think this is ok to merge now.

oldergod (approved):
I pushed a commit to satisfy spotless. Very good tests, thank you

swankjesse (approved):

oldergod:
🚀

PR #2800, opened 2024-01-27
Fix over pruning imports when only option types are being extended in a file

mpeyper:
I introduced a bug in #2797 where if the only thing in a file is an extension to an option type, the import for `google/protobuf/descriptor.proto` was being pruned as we were not adding the type being extended to the list of referenced types.

This change does exactly that, plus a test to ensure we don't miss it in the future.

### DIFF
--- a/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/PrunerTest.kt
+++ b/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/PrunerTest.kt
@@ -1750,7 +1750,33 @@ class PrunerTest {
     )
 
     val extension = pruned.protoFile("extension.proto")!!
-    assertThat(extension.imports).containsExactly("title.proto")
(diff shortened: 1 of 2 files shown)


adrw (approved):

wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/PrunerTest.kt
```diff
@@ -1750,7 +1750,33 @@ class PrunerTest {
     )
 
     val extension = pruned.protoFile("extension.proto")!!
-    assertThat(extension.imports).containsExactly("title.proto")
+    assertThat(extension.imports).containsExactly("message.proto", "title.proto")
```
mpeyper, line 1753: This was actually not producing a valid proto file after the pruning either as it had removed the required import for the message being extended.

swankjesse (approved):

PR #2801, opened 2024-01-29
Add public dependencies when encoding the schema

mpeyper:
I'm not sure if the is the correct way to encode the public imports, but I'm basing it off the wording in the JavaDoc [here](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/DescriptorProtos.FileDescriptorProto.html#getPublicDependencyList--).

There is an issue where if a file is `import public` it was not being encoded in the `descriptorMap` of generated servers (generated [here](https://github.com/square/wire/blob/master/wire-grpc-server-generator/src/main/java/com/squareup/wire/kotlin/grpcserver/FileDescriptorGenerator.kt#L72)), causing startup failures in services.


lyonlai (approved):

wsutina (approved):
✨

Issue #2769, opened 2023-12-27
Global request interceptor


sergeys-opera:
I couldn't find a way in Wire to intercept all the requests and handle their results. Without that implementing things like authentication is much harder.

The way it works in `io.grpc` is that you implement and add a `io.grpc.ClientInterceptor` when creating a `io.grpc.Channel`. For example, the code authenticating requests might look like this (note that the existing auth token is invalidated on receiving `Status.UNAUTHENTICATED` from the server):
```kotlin
class GrpcAuthInterceptor (private val authTokenHolder: AuthTokenHolder) : ClientInterceptor {
    override fun <ReqT : Any, RespT : Any> interceptCall(
        method: MethodDescriptor<ReqT, RespT>,
        callOptions: CallOptions,
        next: Channel
    ): ClientCall<ReqT, RespT> = object : ForwardingClientCall.SimpleForwardingClientCall<ReqT, RespT>(
        next.newCall<ReqT, RespT>(method, callOptions)
    ) {
        override fun start(responseListener: Listener<RespT>, headers: Metadata) {
            val needsAuth = headers.needsAuth()
            if (!needsAuth) return super.start(responseListener, headers)

            val authToken = runBlocking { authTokenHolder.getAuthToken() }
            if (authToken != null) {
                headers.authenticate(authToken)
            }
            super.start(
                if (authToken != null) responseListener(authToken, responseListener) else responseListener,
                headers
            )
        }
    }

    private fun <RespT : Any> responseListener(
        authToken: String,
        original: ClientCall.Listener<RespT>
    ) = object : ForwardingClientCallListener.SimpleForwardingClientCallListener<RespT>(original) {
        override fun onClose(status: Status, trailers: Metadata) {
            super.onClose(status, trailers)
            if (status == Status.UNAUTHENTICATED) {
                runBlocking {
                    authTokenHolder.invalidate(authToken)
                }
            }
        }
    }
}
```

However, it's hard to do the same in Wire as there is no single point of request handling.

[...]

PR #2814, opened 2024-01-31
Exposing GrpcResponse.grpcResponseToException

oldergod:
To help with #2769. cc @sergeys-opera

### DIFF
--- a/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/grpc.kt
+++ b/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/grpc.kt
@@ -1,0 +1,61 @@
+/*
+ * Copyright (C) 2024 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire
+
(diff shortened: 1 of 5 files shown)


wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/grpc.kt
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Base64
+import okhttp3.Headers.Companion.headersOf
+import okio.IOException
+
+/** Returns an exception if the gRPC call didn't have a grpc-status of 0. */
+fun GrpcResponse.grpcResponseToException(suppressed: IOException? = null): IOException? {
+  var trailers = headersOf()
+  var transportException = suppressed
+  try {
+    trailers = trailers()
```
sergeys-opera, line 27: This method is guaranteed to throw if the response hasn't been consumed yet. It would be good to document this behavior or ensure it doesn't happen by buffering the response body here via `response.peekBody()`

swankjesse: Good catch.

wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/grpc.kt
```diff
@@ -0,0 +1,61 @@
...
+ * limitations under the License.
+ */
+package com.squareup.wire
+
+import java.util.Base64
+import okhttp3.Headers.Companion.headersOf
+import okio.IOException
+
+/** Returns an exception if the gRPC call didn't have a grpc-status of 0. */
+fun GrpcResponse.grpcResponseToException(suppressed: IOException? = null): IOException? {
```
swankjesse, line 23: I don’t think I like the signature of this function as a public API.

It’s job is to produce an exception if the response itself is malformed or bad.

Maybe this?

```
fun GrpcResponse.toExceptionIfFailed(suppressed: IOException? = null): IOException? {
}
```

swankjesse (commented):
Who would call this?

oldergod:
We'll try working on interceptors instead soon.

PR #2816, opened 2024-02-01
Fix RealGrpcCall timeout

JGulbronson:
`LateInitTimeout` doesn't actually work properly, and overwrites the new with the old, instead of the old with the new. I'm not convinced it's needed, and since it's internal to Wire, I'm hoping I can just delete it and use `ForwardingTimeout` instead.

Added a test to verify that the timeout is preserved after `initCall` is called.


wire-grpc-tests/src/test/java/com/squareup/wire/GrpcClientTest.kt
```diff
@@ -85,6 +86,9 @@ class GrpcClientTest {
   private lateinit var incompatibleRouteGuideService: IncompatibleRouteGuideClient
   private var callReference = AtomicReference<Call>()
 
+  // Just set to a large enough value we won't timeout in tests
+  private val okHttpClientTimeout = Duration.ofSeconds(117)
```
swankjesse, line 90: This value is SO BANANAS

swankjesse (approved):

wire-grpc-tests/src/test/java/com/squareup/wire/GrpcClientTest.kt
oldergod: 🍌

oldergod (approved):
Thank you

PR #2793, opened 2024-01-22
[swift] new Swift extensions API

dnkoutso:
In this PR we are proposing a new way to emit extension fields.

The problem with Wire today is that it is flattening (inlining) extension fields making it very hard (or impossible) to deterministically expect where a file for a given proto will be output. The end result is that the module manifest is not honored and while the roots of a specific module as explicitly specified, the generated files might end up being emitted under a different module.

Consider this example:

`swift_module_one.proto`
```proto
syntax = "proto2";

package squareup.protos.kotlin.swift_modules;

message SwiftModuleOneMessage {
    required string name = 1;
}
```

`swift_module_two.proto`
```proto
syntax = "proto2";

package squareup.protos.kotlin.swift_modules;

import "swift_module_one.proto";

message SwiftModuleTwoMessage {
  optional string name = 1;

  message NestedMessage {
    optional SwiftModuleOneMessage module_type = 1;

    extend SwiftModuleOneMessage {
      optional ExtensionMessage extension_message = 1000;
    }
  }
}

message ExtensionMessage {
  optional string value = 1;
}
```

And this manifest:

```yaml
---
module_one:
  roots:
  - squareup.protos.kotlin.swift_modules.SwiftModuleOneMessage
module_two:
  roots:
  - squareup.protos.kotlin.swift_modules.SwiftModuleTwoMessage
  - squareup.protos.kotlin.swift_modules.SwiftModuleTwoMessage.NestedMessage
  - squareup.protos.kotlin.swift_modules.ExtensionMessage
  dependencies:
  - module_one
```

If Wire is invoked you will see that `ExtensionMessage.swift` is actually emitted in `module_one` folder rather than `module_two` where it is defined in the manifest. That is because the extension is causing the schema analyzer to choose a different (but as things are designed now correct) path.

Here's the output:

```
Writing .SwiftModuleOneMessage declared in swift_module_one.proto to wire-tests-swift/manifest/module_one (target=Swift)
Writing .SwiftModuleTwoMessage declared in swift_module_two.proto to wire-tests-swift/manifest/module_one (target=Swift)
Writing .ExtensionMessage declared in swift_module_two.proto to wire-tests-swift/manifest/module_one (target=Swift)
```

As you can see both `SwiftModuleTwoMessage ` and `ExtensionMessage` were output inside `module_one` folder despite being an explicit root of `module_two`.

[...]

PR #2823, opened 2024-02-09
[swift] Declare extensions in Swift test proto

dnkoutso:


dnkoutso:
Actually nevermind will do it as part of this https://github.com/square/wire/pull/2793

PR #2829, opened 2024-02-13
Don't warn about an unused emitting rule, when that rule is '*'

swankjesse:


wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/WireRun.kt
```diff
@@ -327,12 +327,14 @@ class WireRun(
     }
 
     for (emittingRules in targetToEmittingRules.values) {
-      if (emittingRules.unusedIncludes().isNotEmpty()) {
-        logger.unusedIncludesInTarget(emittingRules.unusedIncludes())
+      val unusedIncludes = emittingRules.unusedIncludes()
+      if (unusedIncludes.isNotEmpty() && unusedIncludes != setOf("*")) {
```
jhump, line 331: Maybe worth some context here, similar to the comment on the test, to explain how it could be `["*"]` and why to ignore it?

jhump (approved):

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/WireRun.kt
swankjesse: Yeah, great idea

swankjesse, line 331:
```suggestion
      // The '*' here is the default includes rule. It's okay if this is unused.
      if (unusedIncludes.isNotEmpty() && unusedIncludes != setOf("*")) {
```

PR #2838, opened 2024-02-27
Get Gradle to do more of the work in our Gradle plugin

swankjesse:
In particular leave more of the dependency management up to Gradle.

### DIFF
--- a/wire-gradle-plugin/api/wire-gradle-plugin.api
+++ b/wire-gradle-plugin/api/wire-gradle-plugin.api
@@ -16,13 +16,6 @@ public class com/squareup/wire/gradle/CustomOutput : com/squareup/wire/gradle/Wi
 	public synthetic fun toTarget (Ljava/lang/String;)Lcom/squareup/wire/schema/Target;
 }
 
-public final class com/squareup/wire/gradle/InputLocation : java/io/Serializable {
-	public fun <init> (Ljava/lang/String;Ljava/util/List;Ljava/util/List;)V
-	public final fun getExcludes ()Ljava/util/List;
-	public final fun getIncludes ()Ljava/util/List;
-	public final fun getPath ()Ljava/lang/String;
-}
-
 public class com/squareup/wire/gradle/JavaOutput : com/squareup/wire/gradle/WireOutput {
 	public fun <init> ()V
 	public final fun getAndroid ()Z
@@ -149,10 +142,6 @@ public final class com/squareup/wire/gradle/WireExtension$ProtoRootSet {
 	public final fun srcJar (Lorg/gradle/api/provider/ProviderConvertible;)V
 	public final fun srcProject (Ljava/lang/String;)V
 	public final fun srcProject (Lorg/gradle/api/internal/catalog/DelegatingProjectDependency;)V
-}
-
-public final class com/squareup/wire/gradle/WireInputKt {
-	public static final fun toLocations (Lcom/squareup/wire/gradle/InputLocation;Lorg/gradle/api/internal/file/FileOperations;Ljava/io/File;)Ljava/util/List;
 }
 
 public abstract class com/squareup/wire/gradle/WireOutput {
--- a/wire-gradle-plugin/api/wire-gradle-plugin.api
+++ b/wire-gradle-plugin/api/wire-gradle-plugin.api
@@ -17,17 +17,10 @@ public class com/squareup/wire/gradle/CustomOutput : com/squareup/wire/gradle/Wi
 }
 
 public final class com/squareup/wire/gradle/InputLocation : java/io/Serializable {
-	public static final field Companion Lcom/squareup/wire/gradle/InputLocation$Companion;
-	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/String;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
-	public static final fun get (Lorg/gradle/api/Project;Ljava/lang/String;)Lcom/squareup/wire/gradle/InputLocation;
-	public static final fun get (Lorg/gradle/api/Project;Ljava/lang/String;Ljava/lang/String;)Lcom/squareup/wire/gradle/InputLocation;
-	public final fun getBase ()Ljava/lang/String;
+	public fun <init> (Ljava/lang/String;Ljava/util/List;Ljava/util/List;)V
+	public final fun getExcludes ()Ljava/util/List;
+	public final fun getIncludes ()Ljava/util/List;
 	public final fun getPath ()Ljava/lang/String;
-}
-
-public final class com/squareup/wire/gradle/InputLocation$Companion {
-	public final fun get (Lorg/gradle/api/Project;Ljava/lang/String;)Lcom/squareup/wire/gradle/InputLocation;
-	public final fun get (Lorg/gradle/api/Project;Ljava/lang/String;Ljava/lang/String;)Lcom/squareup/wire/gradle/InputLocation;
 }
 
 public class com/squareup/wire/gradle/JavaOutput : com/squareup/wire/gradle/WireOutput {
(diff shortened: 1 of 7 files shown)


wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -218,7 +197,16 @@ class WirePlugin : Plugin<Project> {
       val task = project.tasks.register(taskName, WireTask::class.java) { task: WireTask ->
         task.group = GROUP
         task.description = "Generate protobuf implementation for ${source.name}"
-        task.source(protoSourceInput.configuration)
+
+        // Flatten all the input files here. Changes to any of them will cause the task to re-run.
```
oldergod, line 201: Yes!

wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
```diff
@@ -86,21 +86,10 @@ class WirePluginTest {
...
-            |wire {
-            |  sourcePath {
-            |    srcDir 'dirPath'
-            |    include 'relativePath'
-            |  }
-            |}
-      """.trimMargin(),
-    )
+    assertThat(result.task(":generateMainProtos")?.getOutcome())
+      .isEqualTo(TaskOutcome.NO_SOURCE)
```
oldergod, line 92: I'm cool with it

oldergod (approved):
Very exciting.

oldergod:
I pushed a commit with spotless

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -136,32 +134,13 @@ class WirePlugin : Plugin<Project> {
...
-        protoSourceInput.addPaths(project, defaultSourceFolders(source))
-      }
-
-      val inputFiles = project.layout.files(protoSourceInput.inputFiles, protoPathInput.inputFiles)
-
-      val projectDependencies =
-        (protoSourceInput.dependencies + protoPathInput.dependencies).filterIsInstance<ProjectDependency>()
-      for (projectDependency in projectDependencies) {
-        projectDependenciesJvmConfiguration.dependencies.add(projectDependency)
+      if (extension.protoSourceProtoRootSets.all { it.roots.isEmpty() }) {
```
martinbonnin, line 140: Note: this probably resolves configurations eagerly. I think this is the case already so not a huge deal in the grand scheme of things but would be a good candidate for future improvements.

wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
```diff
@@ -186,17 +175,20 @@ class WirePluginTest {
...
-            |  sourcePath {
-            |    srcDir 'dirPath'
-            |    include 'relativePath'
-            |  }
-            |}
+        |For individual files, use the following syntax:
+        |wire {
+        |  sourcePath {
+        |    srcDir 'dirPath'
+        |    include 'relativePath'
```
martinbonnin, line 189: For those cases, I like using a syntax that works in both Groovy and Kotlin:

```suggestion
        |    srcDir("dirPath")
        |    include("relativePath")
```

It's less "idiomatic" in Groovy though so either work with me

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
martinbonnin: ~Is it even required? Given there is `sourceInput` and `protoInput` below, might be possible for `WireTask` to extend `DefaultTask` instead of `DefaultTask`?~

Edit I read [below](https://github.com/square/wire/pull/2838/files#r1503972524) we want `NO-SOURCE` if no source is found. In which case, I think `protoSource` should be in source and `protoPath` passed out of the band?

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt
```diff
@@ -16,183 +16,86 @@
...
-      configuration.dependencies.add(dependency)
-    }
+private fun ProtoRootSet.inputLocation(file: File): InputLocation {
+  val includes = when {
+    includes.isEmpty() && excludes.isEmpty() -> listOf("**/*.proto")
+    else -> includes
   }
+  // We store [file] relative to the [project] in order to not invalidate the cache when we don't
+  // have to.
+  return InputLocation(project.relativePath(file.path), includes, excludes)
```
martinbonnin, line 39: Note: It's technically ok to move the directory around as long as its contents stay the same so there might still be spurious cache invalidations here.

martinbonnin (approved):
I left a few non-blocking comments, mainly for possible future improvements. 

This change is a nice simplification 👍

wire-gradle-plugin/api/wire-gradle-plugin.api
```diff
@@ -114,15 +107,9 @@ public class com/squareup/wire/gradle/WireExtension {
...
 	public final fun getOutputs ()Ljava/util/List;
-	public final fun getProtoJars ()Ljava/util/Set;
 	public final fun getProtoLibrary ()Z
-	public final fun getProtoPaths ()Ljava/util/Set;
-	public final fun getProtoTrees ()Ljava/util/Set;
 	public final fun getRejectUnusedRootsOrPrunes ()Z
 	public final fun getRules ()Ljava/lang/String;
-	public final fun getSourceJars ()Ljava/util/Set;
-	public final fun getSourcePaths ()Ljava/util/Set;
-	public final fun getSourceTrees ()Ljava/util/Set;
```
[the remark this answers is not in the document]
swankjesse: Yep, exactly. These should have always been implementation details.

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
swankjesse: Great idea. We can move this behaviour to execution time.

swankjesse: In follow-up!

wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
swankjesse: Great idea.

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt
swankjesse: Remote caching is the big thing I wanna preserve. I think we get that as long as the paths are relative to the project.

swankjesse:
Build failure in samples means I’ve made an unintentional behaviour change and our tests didn’t catch it! Yikes!

```
> Task :samples:android-lib-java-sample:javaPreCompileDebug
e: file:///home/runner/work/wire/wire/samples/android-app-variants-sample/src/testRelease/java/com/squareup/wire/android/app/variants/ReleaseUnitTest.kt:50:23 Unresolved reference: ReleaseType
```

oldergod:
Samples are also used as tests. Way simpler to work with that tests against the plugin directly.

swankjesse:
Ugh, the failing Windows build shows me I’ve made a design error in my implementation.

The string I’m passing to `InputLocation()` is the relative path on the file system from my projects’ directory to the .jar or directory. For example, given these inputs:
  projectDirectory=/Users/jwilson/Development/greeter/hello
  path=/Users/jwilson/Development/greeter/hello/src/main/proto
It correctly returns a relative path output:
  result=src/main/proto

Unfortunately if my path isn’t in the project directory, the result I’m producing is bananas and won’t work with Gradle’s cache:
  projectDirectory=/Users/jwilson/Development/greeter/hello
  path=/Users/jwilson/.gradle/caches/com/example/library/protos.jar
This produces a nonsense relative path:
  result=../../../.gradle/caches/com/example/library/1.2.3/protos-1.2.3.jar
In particular, the relative path is now conditional on where my project and library directories are on the file system, and I don’t want that.

I think instead I need to change `InputLocation` to be either a `File` relative to the project, or a symbolic identifier for whatever library it’s pulling in, like say
  library=com.example.library:protos:1.2.3

Ugh.

swankjesse:
Here’s a tracking issue to fix that path problem: https://github.com/square/wire/issues/2842

wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
swankjesse: https://github.com/square/wire/pull/2844

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
swankjesse: I think you’re strictly correct, and also I don’t think it’ll matter much in practice?
Both skipping the task and running it when there’s no sources will both be super fast and rare.

martinbonnin: Agreed 👍

PR #2840, opened 2024-02-27
Propagates Request timeout as grpc-timeout.

frojasg:
This is a proof of concept of how it would look like to propagate the request timeout as part of the request metadata following the grp-timeout in the grpc spec [1]

Open questions:
* What is the best way to serialize the timeout into the header depending on the units
* Should we propagate deadlines as well?
* Why TimeoutMarshaller is compleining about the space between the timeout number and the unit [2]

[1] https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md
[2] https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/GrpcUtil.java#L699-L700

### DIFF
--- a/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/GrpcClient.kt
+++ b/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/GrpcClient.kt
@@ -17,6 +17,7 @@ package com.squareup.wire
 
 import com.squareup.wire.internal.RealGrpcCall
 import com.squareup.wire.internal.RealGrpcStreamingCall
+import java.util.concurrent.TimeUnit
 import kotlin.reflect.KClass
 import okhttp3.Call
 import okhttp3.OkHttpClient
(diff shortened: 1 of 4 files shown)


oldergod:
Here's the spec around that
https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

> - Timeout → "grpc-timeout" TimeoutValue TimeoutUnit
> - TimeoutValue → {positive integer as ASCII string of at most 8 digits}
> - TimeoutUnit → Hour / Minute / Second / Millisecond / Microsecond / Nanosecond
>   - Hour → "H"
>   - Minute → "M"
>   - Second → "S"
>   - Millisecond → "m"
>   - Microsecond → "u"
>   - Nanosecond → "n"

Exemple: `grpc-timeout = 1S`

wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/GrpcClient.kt
```diff
@@ -81,13 +84,45 @@ actual abstract class GrpcClient actual constructor() {
           for ((key, value) in requestMetadata) {
             addHeader(key, value)
           }
+
+          if (timeout.hasDeadline()) {
+            addHeader("grpc-timeout", serializeTimeout(timeout.deadlineNanoTime()))
```
frojasg, line 89: Does the order of the headers matter? The spec says list this one to be present after `te` header. Should I reorder this?

oldergod: I don't think so? The spec only lists the headers in a specific order but doesn't say they have to be sent into this specific one, right?

oldergod (approved):

wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/GrpcClient.kt
frojasg: yup, right.

PR #2841, opened 2024-02-27
[wire-grpc-tests] Reads assertions from backend in tests

frojasg:
* Validates the assertions on request shape.
* Question: Should we add a beforeEach rule for this?


wire-grpc-tests/src/test/java/com/squareup/wire/GrpcClientTest.kt
```diff
@@ -1494,6 +1494,8 @@ class GrpcClientTest {
     assertThat(responseChannel.read()).isEqualTo(Feature(name = "house"))
     assertThat(responseChannel.read()).isNull()
     assertThat(grpcCall.responseMetadata!!["response-lucky-animal"]).isEqualTo("horse")
+
+    mockService.awaitSuccessBlocking()
```
frojasg, line 1498: Should we move this to a before each rule? it seems is not being evaluated on all test and I'm not sure if that is intentional or not.

PR #2846, opened 2024-02-28
Reject unused prunes and roots by default

swankjesse:
This flips the default from lenient to strict. The option to manually configure this is now necessary for anyone who wants the lenient behavior.


oldergod (approved):
🚀

PR #2848, opened 2024-02-29
Resolve configurations more lazily

swankjesse:
We're still resolving it when the input locations properties are required. I'd prefer to not resolve until the task runs, but that's a bit more difficult.

### DIFF
--- a/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
+++ b/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
@@ -17,6 +17,7 @@
 
 package com.squareup.wire.gradle
 
+import java.lang.reflect.Array as JavaArray
 import com.squareup.wire.VERSION
 import com.squareup.wire.gradle.internal.libraryProtoOutputPath
 import com.squareup.wire.gradle.internal.targetDefaultOutputPath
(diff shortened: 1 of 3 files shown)


wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -207,16 +207,14 @@ class WirePlugin : Plugin<Project> {
...
-          buildList {
-            for (rootSet in protoSourceProtoRootSets) {
-              addAll(rootSet.roots)
-            }
-            for (rootSet in protoPathProtoRootSets) {
-              addAll(rootSet.roots)
-            }
-          },
-        )
+        for (rootSet in protoSourceProtoRootSets) {
```
swankjesse, line 210: Calling `.roots` resolves the configuration; but passing a configuration is lazy!

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
```diff
@@ -235,8 +233,8 @@ class WirePlugin : Plugin<Project> {
         if (extension.protoLibrary) {
           task.protoLibraryOutput.set(File(project.libraryProtoOutputPath()))
         }
-        task.sourceInput.set(protoSourceProtoRootSets.inputLocations)
-        task.protoInput.set(protoPathProtoRootSets.inputLocations)
+        task.sourceInput.set(project.provider { protoSourceProtoRootSets.inputLocations })
+        task.protoInput.set(project.provider { protoPathProtoRootSets.inputLocations })
```
swankjesse, line 237: To make this lazier we need to refer to the configurations symbolically. I don’t know how to do that.

oldergod: 👍

oldergod (approved):

PR #2851, opened 2024-03-01
Remove deprecated SchemaHandler#create method

oldergod:


wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SchemaHandler.kt
```diff
@@ -192,20 +192,23 @@ abstract class SchemaHandler {
   /** Implementations of this interface must have a no-arguments public constructor. */
   @JvmDefaultWithCompatibility
   interface Factory : Serializable {
-    @Deprecated("Wire does not call this method anymore. Implement the other 'create' method to receive the payload associated with the schema handler. This method will be removed in Wire 5.")
```
Egorand, line 195: Will Wire 5 be the next release?

Egorand (approved):

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SchemaHandler.kt
oldergod: Soon™
We already merged breaking changes.

We might release some `4.9.x` but 5 is coming.

PR #2854, opened 2024-03-06
Add use-site target field to custom options annotation spec

alvinsee:
## Description
This change adds a use-site target for fields in the kotlin generated classes. This enables the annotations targeted for a field to be available through the java field via reflection.

### DIFF
--- a/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
+++ b/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
@@ -1190,7 +1190,10 @@ class KotlinGenerator private constructor(
           )
         }
         for (annotation in optionAnnotations(field.options)) {
+          // Adds default use-site target for kotlin properties.
           addAnnotation(annotation)
+          // Adds use-site target for java fields.
+          addAnnotation(annotation.toBuilder().useSiteTarget(FIELD).build())
         }
         addAnnotation(wireFieldAnnotation(message, field, schemaIndex))
         if (javaInterOp) {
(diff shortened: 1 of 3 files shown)


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -1190,7 +1191,10 @@ class KotlinGenerator private constructor(
           )
         }
         for (annotation in optionAnnotations(field.options)) {
-          addAnnotation(annotation)
+          // adds use-site target for kotlin properties
+          addAnnotation(annotation.toBuilder().useSiteTarget(PROPERTY).build())
+          // adds use-site target for java fields
```
oldergod, line 1196: https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

```suggestion
          // Adds use-site target for kotlin properties.
          addAnnotation(annotation.toBuilder().useSiteTarget(PROPERTY).build())
          // Adds use-site target for java fields.
```

wire-tests/src/commonTest/proto-kotlin/com/squareup/wire/protos/custom_options/FooBar.kt
```diff
@@ -40,37 +40,53 @@ import kotlin.collections.List
 import okio.ByteString
 
 public class FooBar(
-  @MyFieldOptionOneOption(17)
```
oldergod, line 43: This is breaking then I think?

alvinsee: I applied targets to property and field. I think the default is property, but I can remove the explicitness

swankjesse:
I don’t like doubling up the amount of generated code here. 

Is it already available via reflection on the getter?

JakeWharton:
Can we make these all `@JvmField` as a hack?

PR #2880, opened 2024-03-26
Fix a bug where outputs weren't deterministic

swankjesse:
This was only problematic when inputs weren't deterministic. We started seeing this with the changes to our Gradle setup.


mpeyper (approved):

oldergod (approved):
Dope, thank you

PR #2881, opened 2024-03-26
fix KotlinConstructorBuilder on Android SDK below 26

### LINKED ISSUE #2876: NoSuchMethodError for java.lang.Constructor#getParameterCount
This method was not added until Java 8, and thus is not available on Android API levels lower than 26:

```
Fatal Exception: java.lang.NoSuchMethodError: No virtual method getParameterCount()I in class Ljava/lang/reflect/Constructor; or its super classes (declaration of 'java.lang.reflect.Constructor' appears in /system/framework/core-oj.jar)
       at com.squareup.wire.KotlinConstructorBuilder.build(KotlinConstructorBuilder.kt:91)
       at com.squareup.wire.MessageJsonAdapter.fromJson(MessageJsonAdapter.kt:85)
       at com.squareup.wire.MessageJsonAdapter.fromJson(MessageJsonAdapter.kt:24)
       at com.squareup.moshi.internal.NullSafeJsonAdapter.fromJson(NullSafeJsonAdapter.java:41)
       ...        
```

I'm happy to put up a PR with a fix, but am curious to know what an appropriate test would look like given that there aren't currently any JDK-specific or Android-specific tests already in the codebase.

damianw:
Fixes #2876. I also added a sample app that can run on API 24+ per @oldergod's request, and verified that it no longer crashes.

Screenshot_1711467123


samples/android-app-kotlin-minsdk-sample/src/main/java/com/squareup/wire/android/app/kotlin/minsdk/MainActivity.kt
```diff
@@ -0,0 +1,39 @@
...
+  private val jsonAdapter = Moshi.Builder()
+    .add(WireJsonAdapterFactory())
+    .build()
+    .adapter<SomeText>()
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+    setContentView(R.layout.activity_main)
+    val someText = checkNotNull(jsonAdapter.fromJson("""{"value": "Hi"}"""))
+    findViewById<TextView>(R.id.text_view).text = someText.value_
```
oldergod, line 37: Since we're fixing a runtime problem, we need this code to run on CI to be useful. We thus need a test that execute this path.

oldergod (approved):
Thanks for the PR

Could we get rid of all the mipmap folders but one?

samples/android-app-kotlin-minsdk-sample/src/main/java/com/squareup/wire/android/app/kotlin/minsdk/MainActivity.kt
oldergod: hmmmm does that mean we need androidTests with an emulator...? hmmm

damianw: Yup - I would love to write a test but I am both unfamiliar with the CI setup, and I don't think the current build configuration supports any of the ways I can think of to add a regression test for this, so your input would be helpful on what you think would be best. I don't want to introduce something unmaintainable just for this one test.

Right now I can think of a few ways this might be done, all of which are pretty icky in the context of what exists in the project already:
- Run unit tests on an older JDK: don't think the project has any support for this
- Use a mocking framework (e.g. MockK or Mockito): one is not already used in the project
- Use Robolectric: I don't even know if it would pick this up, but it is not part of this project
- Run instrumentation tests: Could use Gradle managed devices, or Firebase Test Lab. For the former, would need bare metal machines (which are probably already being used in the form of Mac agents?)

Let me know what you think.

damianw: Just what I need, thanks! Will do.

By the way, it looks like `com.squareup.wire.GrpcClientTest > streamingRequestFailureWithUnexpectedHttpCode` might be [flaky](https://github.com/square/wire/actions/runs/8467516909/job/23198506065?pr=2881#step:6:1392)

damianw: Added - fwiw I wasn't able to use assertj because it depends on other JDK APIs that can't be desugared

wire-runtime/src/jvmMain/kotlin/com/squareup/wire/KotlinConstructorBuilder.kt
```diff
@@ -108,7 +108,6 @@ internal class KotlinConstructorBuilder<M : Message<M, B>, B : Message.Builder<M
     .sortedBy { it.wireField.schemaIndex }
 
   private class ProtoField(
-    // TODO(Benoit) Delete if unused?
```
oldergod, line 111: This was waiting for you it seems

oldergod (approved):

oldergod:
Thanks a lot Damian, chef-kiss PR.

damianw:
Thanks for merging! Are there any plans to release another stable version (`4.9.9`?) in the near future?

oldergod:
`4.9.9` has been released. Let me know how it went

damianw:
Amazing, thank you! Will do 🙇

damianw:
Just wanted to close this out since I forgot to follow up - we updated to the new version and have not had any more problems 🙌 . Thanks for your help!

PR #2887, opened 2024-03-29
[swift] deprecate extension fields with new API

dnkoutso:
tiny regression from https://github.com/square/wire/pull/2793

### DIFF
--- a/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
+++ b/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
@@ -1201,8 +1201,11 @@ class SwiftGenerator private constructor(
                   addDoc("\n%L\n", field.documentation.sanitizeDoc())
                 }
                 addDoc("\nSource: %L\n", field.location.withPathOnly())
-              }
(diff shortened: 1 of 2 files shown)


wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -1201,8 +1201,11 @@ class SwiftGenerator private constructor(
                   addDoc("\n%L\n", field.documentation.sanitizeDoc())
                 }
                 addDoc("\nSource: %L\n", field.location.withPathOnly())
-              }
-              .apply {
+
+                if (field.isDeprecated) {
+                  addAttribute(deprecated)
```
dnkoutso, line 1206: verified generated code includes this:

```swift
    /**
     *
     * Source: extensible.proto
     */
    @available(*, deprecated)
    public var ext_string_deprecated: String? {
        get {
            self.parseUnknownField(fieldNumber: 1016, type: String.self)
        }
        set {
            self.setUnknownField(fieldNumber: 1016, newValue: newValue)
        }
    }
```

lickel (approved):

Issue #2837, opened 2024-02-26
Wire doesn't appear to be compatible with the configuration cache in gradle 8+


staktrace:
Standalone minimal reproducer project is at https://github.com/staktrace/wire-bug-20240226-cc. Basically, when wire is run with the configuration cache enabled and gradle 8+, pretty basic operations fail as wire is unable to locate all the necessary input files.

This appears to be similar to https://github.com/square/wire/issues/2815

staktrace:
AFAICT wire 5.0.0-alpha.square.3 fixes this problem. But we should probably still add this as a testcase in wire so it doesn't regress

staktrace:
Adding a test in https://github.com/square/wire/pull/2890

staktrace:
So although the minimal reproducer project in https://github.com/staktrace/wire-bug-20240226-cc works now, I'm still seeing a problem in real-world usage that looks the same (configuration cache enabled, gradle 8+, wire 5.0+) so I'm going to reopen this issue while I investigate.

staktrace:
5.0.0-alpha.square.9 at least seems to not have the problem any more. I guess I'll close this

oldergod:
@staktrace I don't think this version exists though?

staktrace:
Lol we're using so it definitely does exist! https://artifactory.global.square/ui/native/square-public/com/squareup/wire/wire-bom/5.0.0-alpha.square.9/

PR #2890, opened 2024-04-04
Add test for configuration cache bug

staktrace:
This is a test for the scenario in https://github.com/square/wire/issues/2837 which has since been fixed already.

### DIFF
--- a/wire-gradle-plugin/src/test/projects/configuration-cache-failure/gradle.properties
+++ b/wire-gradle-plugin/src/test/projects/configuration-cache-failure/gradle.properties
@@ -1,0 +1,4 @@
+org.gradle.caching=false
+org.gradle.unsafe.configuration-cache=true
+org.gradle.configuration-cache=true
+VERSION_NAME=5.0.0-SNAPSHOT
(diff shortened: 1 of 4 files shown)


wire-gradle-plugin/src/test/projects/configuration-cache-failure/gradle.properties
```diff
@@ -0,0 +1,4 @@
+org.gradle.caching=false
+org.gradle.unsafe.configuration-cache=true
+org.gradle.configuration-cache=true
+VERSION_NAME=5.0.0-SNAPSHOT
```
swankjesse, line 4: omit this one? I don’t wanna have to keep it up-to-date

adrw (approved):

swankjesse (approved):

wire-gradle-plugin/src/test/projects/configuration-cache-failure/gradle.properties
staktrace: It's required by the runner: https://github.com/square/wire/blob/b5bf223468acc788941fe088ddd098906a56aae2/wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt#L1492

oldergod:
Thank you!

wire-gradle-plugin/src/test/projects/configuration-cache-failure/gradle.properties
oldergod: We have it somewhere else as well I think, our releasing steps update it so should be good

PR #2893, opened 2024-04-05
Don't override RealGrpcCall.timeout when it's manually set

JGulbronson:
Currently we always update the timeout on `RealGrpcCall` with the timeout from the underlying `OkHttpClient`. This means that if the user has manually updated the timeout, it gets overwritten.

### DIFF
--- a/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/RealGrpcCall.kt
+++ b/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/RealGrpcCall.kt
@@ -139,7 +139,11 @@ internal class RealGrpcCall<S : Any, R : Any>(
     val result = grpcClient.newCall(method, requestMetadata, requestBody, timeout)
     this.call = result
     if (canceled) result.cancel()
-    (timeout as ForwardingTimeout).setDelegate(result.timeout())
+    // If the timeout doesn't have a deadline or timeout, then the user
+    // didn't set the timeout on this Call manually.
+    if (!timeout.hasDeadline() && (timeout.timeoutNanos() == 0L)) {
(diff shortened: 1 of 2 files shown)


wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/RealGrpcCall.kt
```diff
@@ -139,7 +139,11 @@ internal class RealGrpcCall<S : Any, R : Any>(
     val result = grpcClient.newCall(method, requestMetadata, requestBody, timeout)
     this.call = result
     if (canceled) result.cancel()
-    (timeout as ForwardingTimeout).setDelegate(result.timeout())
+    // If the timeout doesn't have a deadline or timeout, then the user
+    // didn't set the timeout on this Call manually.
```
oldergod, line 143: > then the user didn't set the timeout on this Call manually

Does this imply there are other ways to set it?

oldergod (approved):

wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/RealGrpcCall.kt
JGulbronson: Well it gets set by default to `ForwardingTimeout(Timeout())` when it's initialized.

MariusVolkhart:
@JGulbronson thanks for implementing this! Was looking for it in a 4.x version and saw it in the changeling of 5.x.

However, the timeout doesn't get pushed down to the okhttp `Call` - should it? Right now, it seems setting a timeout or a deadline manually doesn't result in okhttp cancelling the call if the deadline is exceeded. Setting a timeout via the `OkHttpClient.Builder` does result in immediate cancellation.

PR #2897, opened 2024-04-08
Add doc to our Wire extensions/output

oldergod:
So that we show something when DSLing within gradle.

image

### DIFF
--- a/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
+++ b/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
@@ -207,24 +207,36 @@ open class WireExtension(
     return result
   }
 
+  /**
+   * Defines a Java target. See [com.squareup.wire.schema.JavaTarget]
+   */
   fun java(action: Action<JavaOutput>) {
     val javaOutput = objectFactory.newInstance(JavaOutput::class.java)
     action.execute(javaOutput)
     outputs += javaOutput
   }
 
+  /**
+   * Defines a Kotlin target. See [com.squareup.wire.schema.KotlinTarget]
+   */
   fun kotlin(action: Action<KotlinOutput>) {
     val kotlinOutput = objectFactory.newInstance(KotlinOutput::class.java)
     action.execute(kotlinOutput)
     outputs += kotlinOutput
   }
 
+  /**
+   * Defines a Proto target. See [com.squareup.wire.schema.ProtoTarget]
+   */
   fun proto(action: Action<ProtoOutput>) {
     val protoOutput = objectFactory.newInstance(ProtoOutput::class.java)
     action.execute(protoOutput)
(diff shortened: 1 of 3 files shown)


wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
```diff
@@ -207,24 +207,36 @@ open class WireExtension(
     return result
   }
 
+  /**
+   * Defines a Java target. See [com.squareup.wire.schema.JavaTarget]
```
Egorand, line 211: Comments should be sentences!

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
```diff
@@ -277,16 +295,19 @@ open class WireExtension(
     internal val includes = mutableListOf<String>()
     internal val excludes = mutableListOf<String>()
 
+    /** Sets a directory. Example: "src/main/proto". */
     fun srcDir(dir: String) {
       isEmpty = false
       sourceDirectoriesAndLocalJars += project.file(dir)
     }
 
+    /** Sets one or more directory. */
```
Egorand, line 304: Shouldn't it be plural?

```suggestion
    /** Sets one or more directories. */
```

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
```diff
@@ -323,10 +348,18 @@ open class WireExtension(
       project.dependencies.add("protoProjectDependenciesJvm", dependencyNotation)
     }
 
+    /**
+     * If set, only the files being defined as included will be treated.
```
Egorand, line 352: nit:

```suggestion
     * If set, only the files defined as included will be processed.
```

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
```diff
@@ -323,10 +348,18 @@ open class WireExtension(
...
+     * Example: "com/example/important.proto".
+     */
     fun include(vararg includePaths: String) {
       includes += includePaths
     }
 
+    /**
+     * If set, all the files being defined as excluded will be ignored.
```
Egorand, line 360: nit:

```suggestion
     * If set, all the files defined as excluded will be ignored.
```

Egorand (approved):

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
oldergod: déjà vu !!!

Thank you, all addressed.

PR #2899, opened 2024-04-08
[swift] Always conform parent struct type to ProtoExtensible

dnkoutso:


wire-tests-swift/no-manifest/src/main/swift/AllTypes.swift
```diff
@@ -848,6 +848,9 @@ public struct AllTypes {
 
 }
 
+extension AllTypes : ProtoExtensible {
```
lickel, line 851: This also should be generating:
```
public var ext_opt_int32: Int32? {]
    get { storage.ext_opt_int32 }
    set { storage.ext_opt_int32 = newValue }
}
```

lickel: Otherwise you'd be paying `dynamicMemberLookup` costs

dnkoutso: Will be updating.

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -1177,10 +1178,71 @@ class SwiftGenerator private constructor(
     )
   }
 
+  private fun TypeSpec.Builder.generateMessageExtensionStorageDelegates(
+    type: MessageType,
+    storageName: String,
+    structType: DeclaredTypeName,
+    fileMembers: MutableList<FileMemberSpec>,
+  ) {
+    if (!type.isHeapAllocated) {
```
dnkoutso, line 1187: copy pasted warning pattern from other methods that do similar stuff.

wire-tests-swift/no-manifest/src/main/swift/AllTypes.swift
dnkoutso: done

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -1192,17 +1254,21 @@ class SwiftGenerator private constructor(
...
+                if (!forStorageType && field.documentation.isNotBlank()) {
                   addDoc("\n%L\n", field.documentation.sanitizeDoc())
                 }
-                addDoc("\nSource: %L\n", field.location.withPathOnly())
+                if (!forStorageType) {
+                  addDoc("\nSource: %L\n", field.location.withPathOnly())
+                }
 
-                if (field.isDeprecated) {
+                if (!forStorageType && field.isDeprecated) {
```
lickel, line 1271: I _think_ we want to mark as deprecated regardless?

dnkoutso: we are in the other method for generating storage delegates?

lickel (approved):

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
dnkoutso: nm spoke offline we want in both and updated.

PR #2909, opened 2024-04-16
Allow custom options to be passed in WireCompiler

oldergod:


wire-compiler/src/main/java/com/squareup/wire/WireCompiler.kt
```diff
@@ -372,6 +381,11 @@ class WireCompiler internal constructor(
             schemaHandlerFactoryClass = arg.substring(SCHEMA_HANDLER_FACTORY_CLASS_FLAG.length)
           }
 
+          arg.startsWith(CUSTOM_OPTION_FLAG) -> {
+            val (key, value) = arg.substring(SCHEMA_HANDLER_FACTORY_CLASS_FLAG.length).split(',')
+            customOptions[key] = value
```
oldergod, line 386: @JakeWharton made it simpler

JakeWharton: This is different that what you asked, though

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.kt
```diff
@@ -372,6 +381,11 @@ class WireCompiler internal constructor(
             schemaHandlerFactoryClass = arg.substring(SCHEMA_HANDLER_FACTORY_CLASS_FLAG.length)
           }
 
+          arg.startsWith(CUSTOM_OPTION_FLAG) -> {
+            val (key, value) = arg.substring(SCHEMA_HANDLER_FACTORY_CLASS_FLAG.length).split(',')
```
JakeWharton, line 385:
```suggestion
            val (key, value) = arg.substring(SCHEMA_HANDLER_FACTORY_CLASS_FLAG.length).split(',', limit = 2)
```

JakeWharton (approved):

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.kt
oldergod: Yep yep I made the API simpler. I was going with `options=key,value[,key,value]`

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.kt
oldergod: Nice

dnkoutso (approved):

PR #2920, opened 2024-04-25
expose the grpc url when converting Response to GrpcException

traviscj:


swankjesse (approved):

oldergod:
Thankfully this was tested. Once you updated the now failing tests, we should be good to go.

PR #2923, opened 2024-05-01
No eager source access on Task

oldergod:


martinbonnin (commented):
LGTM but TBH I'd rather remove `source()` and `SourceTask` altogether in the long run. 

How important is it that some tasks have `NO-SOURCE` outcome vs just `SUCCESS` and empty output dir? I've always felt `NO-SOURCE` was unnecessary there.

autonomousapps (approved):
I tested this locally and it resolves our issue. Thanks! <3

oldergod:
> How important is it that some tasks have NO-SOURCE outcome vs just SUCCESS and empty output dir? I've always felt NO-SOURCE was unnecessary there.

We have conflicting logic. If Gradle thinks there's no source, we go NO-SOURCE but if it thinks there's source, Wire loads and find nothing, we throw. This would be a breaking change to modify any of that.

Issue #2725, opened 2023-11-28
Adding special value UNRECOGNIZED to the enum type


JurajBegovac:
Are there any plans for Wire to add an **UNRECOGNIZED** value to enums, similar to the functionality provided by the proto3 compiler for Java? 

https://protobuf.dev/reference/java/java-generated/#enum
```
The protocol buffer compiler will generate a Java enum type called Foo with the same set of values. If you are using proto3, it also adds the special value UNRECOGNIZED to the enum type. 
```

The context for my query is that I'm working on a project where we're considering switching to Wire for generating Kotlin code from our proto3 schema. However, I've noticed that not all enums in our existing proto schema include a special type in the first position (value 0). This feature of handling unknown enum values automatically is important for our project's compatibility and robustness, and I'm interested in knowing if Wire plans to support something similar for Kotlin. If this feature is not planned, is there a possibility for me to implement it myself using some custom handler or workaround?

Thank you in advance

oldergod:
I'm surprised protoc didn't throw on that. Is there a lenient flag you can turn on to get that behaviour?

oldergod:
The doc says it adds it all the time? https://protobuf.dev/reference/java/java-generated/#enum Even if the example had a constant with a `0` value which is the default...

What is happening...

oldergod:
`UNRECOGNIZED(-1),`

oldergod:
Got it. Wire since the beginning has been returning null. Maybe we could put it behind an option for Java and Kotlin generation... 🤔 

Java:
```java
    public static CurrencyCodeType fromValue(int value) {
      switch (value) {
        case 0: return CURRENCYCODETYPE_UNKNOWN;
        case 1: return CURRENCYCODETYPE_EUR;
        case 2: return CURRENCYCODETYPE_USD;
        default: return null;
      }
    }
```

Kotlin:
```kotlin
      public fun fromValue(`value`: Int): CurrencyCodeType? = when (value) {
        0 -> CURRENCYCODETYPE_UNKNOWN
        1 -> CURRENCYCODETYPE_EUR
        2 -> CURRENCYCODETYPE_USD
        else -> null
      }
```

JurajBegovac:
I don't believe the issue lies with returning null (my example also returns null). The potential problem, as I see it, could arise with the decode function in messages that utilize enum. 

Take a look at the following example, which is a snippet from the Price source code generated using Wire:

```
override fun decode(reader: ProtoReader): Price {
[...]

PR #2934, opened 2024-05-09
Read unknown enum tag

oldergod:
Continuing the work for #2725


psymoon (approved):

wire-runtime/src/jvmMain/kotlin/com/squareup/wire/Messages.kt
```diff
@@ -0,0 +1,65 @@
...
+import okio.IOException
+
+/**
+ * Reads through the [message's unknown fields][Message.unknownFields] and return the first value
+ * matching the [property]'s tag as set in the [WireField] annotation attached to it. If the tag is
+ * not found, it will return null. If the tag is found but the value attached to it isn't a varint,
+ * the method will throw an [IOException].
+ */
+@Throws(IOException::class)
+inline fun <reified M : Message<M, B>, B : Message.Builder<M, B>, reified P : Enum<P>> Message<M, B>.unknownEnumTag(property: KProperty1<M, P?>): Int? {
```
swankjesse, line 29: I think we need to do the reflection machines thing here. Create something like this:

```
interface UnknownEnumTagReader<M, B, P> {
  fun get(message: M): P?
}
```

That way we don’t have to do the KProperty reflection stuff each time we want to read a field

wire-runtime/src/jvmMain/kotlin/com/squareup/wire/Messages.kt
```diff
@@ -0,0 +1,65 @@
...
+  return null
+}
+
+/**
+ * Reads through the [message's unknown fields][Message.unknownFields] and return all values
+ * matching the [property]'s tag as set in the [WireField] annotation attached to it. If the tag is
+ * found but the value attached to it isn't a varint, the method will throw an [IOException].
+ */
+@Throws(IOException::class)
+inline fun <reified M : Message<M, B>, B : Message.Builder<M, B>, reified P : Enum<P>> Message<M, B>.unknownEnumTags(property: KProperty1<M, List<P>>): List<Int> {
```
swankjesse, line 50: This is for repeated properties?

swankjesse (approved):

wire-runtime/src/jvmMain/kotlin/com/squareup/wire/Messages.kt
oldergod: Yes, `repeated MyEnum my_enums = 1`

oldergod: I wanna rename it to `unknownEnumValue` and `unknownEnumValues` by the way.

oldergod:
@psymoon @swankjesse I did a big revamp following a pairing Jesse and I did. Please, have a look

PR #2938, opened 2024-05-13
[swift] handle edge for defaulted value when it collides a sub-type name

dnkoutso:
We hit an edge case in which a  sub-type in a message is the same name as the message itself. The Swift compiler trips thinking the sub-field is the one to provide as the default value which does not seem to work.

This makes sure we avoid this collision. 

A test is added that failed to compile before this change.


wire-tests/src/commonTest/proto/kotlin/swift_edge_cases.proto
```diff
@@ -20,3 +20,10 @@ package squareup.protos.kotlin.swift_modules;
 message SwiftEdgeCases {
   optional string return = 1;
 }
+
+message Screen {
+  oneof screen {
```
dnkoutso, line 25: This is the case in which a field name is the same as the enclosing type name.

jhatin94 (approved):

lickel (approved):

PR #2946, opened 2024-05-16
Support dependency on taskProvider (among other things)

### LINKED ISSUE #2933: `srcDir` should accept `TaskProvider`
If someone is sourcing protos from an external source, you currently have to wire (heh) directories like this:

```kotlin
schemas {
  get {
    saveTo("./src/main/proto")
    file("squareup/cash/whatever/api.proto")
  }
}

tasks.matching { it.name == "generateCommonMainProtos" }.configureEach {
  dependsOn("getSchemas")
}

wire {
  sourcePath {
    srcDir("src/main/proto")
  }
}
```

This is error prone and counter to Gradle best practices.

In the same way that Kotlin-producing tasks can be added directly as a `srcDir` with which Gradle resolves the `@OutputDirectory`, I should be able to add proto-producing tasks (or rather, `TaskProvider`s) to Wire's `srcDir`.

oldergod:
fixes #2933

### DIFF
--- a/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
+++ b/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
@@ -17,9 +17,11 @@ package com.squareup.wire.gradle
 
 import com.squareup.wire.schema.EventListener
 import java.io.File
+import kotlin.LazyThreadSafetyMode.NONE
 import org.gradle.api.Action
 import org.gradle.api.Project
 import org.gradle.api.artifacts.MinimalExternalModuleDependency
+import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.internal.catalog.DelegatingProjectDependency
 import org.gradle.api.internal.file.FileOrUriNotationConverter
 import org.gradle.api.provider.Provider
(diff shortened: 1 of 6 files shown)


staktrace (commented):
Seems reasonable enough. I'm concerned about the change in ABI, though. There could be gradle plugins that are compiled against the old version of wire (with `srcDir(String)`) which will break if used in a build with the new version of wire on the buildscript classpath. It might be better to lave `srcDir` alone and add a new function that takes `Any` so that there's no ABI breakage

oldergod:
I can keep srcDir(string) alongside it

staktrace (approved):
Thanks, LGTM

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
```diff
@@ -291,11 +293,25 @@ open class WireExtension(
...
+    /**
+     * Adds a set of source. The given paths are evaluated as per [Project.files][org.gradle.api.Project.files].
+     */
+    fun srcDir(fileCollection: Any) {
+      isEmpty = false
+      files.from(fileCollection)
+    }
+
     /** Sets a directory. Example: "src/main/proto". */
+    @Deprecated("Deprecated in favor of 'srcDir(fileCollection)'")
```
JakeWharton, line 314: If you are deprecating this function, seems like the one below it should go as well.

JakeWharton (approved):

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
```diff
@@ -291,11 +293,25 @@ open class WireExtension(
     /** Calling this will resolve the configuration. */
     internal val roots: Set<File>
       get() = configuration.files + sourceDirectoriesAndLocalJars
+    private val files: ConfigurableFileCollection by lazy(NONE) {
+      val files = project.files()
+      project.dependencies.add(configuration.name, files)
```
martinbonnin, line 298: TIL you could pass a `FileCollection` as a dependency

martinbonnin (approved):
🚀

PR #2957, opened 2024-05-21
Trim leading dots when parsing message/enum/service

oldergod:
Related to https://github.com/square/wire/issues/2927

Not so happy with the... ad-hoc-ness of it all to be honest.

### DIFF
--- a/wire-schema/src/commonTest/kotlin/com/squareup/wire/schema/internal/parser/ProtoParserTest.kt
+++ b/wire-schema/src/commonTest/kotlin/com/squareup/wire/schema/internal/parser/ProtoParserTest.kt
@@ -322,6 +322,19 @@ class ProtoParserTest {
   }
 
   @Test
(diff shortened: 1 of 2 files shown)


wire-schema/src/commonTest/kotlin/com/squareup/wire/schema/internal/parser/ProtoParserTest.kt
```diff
@@ -321,6 +321,19 @@ class ProtoParserTest {
     assertThat(type.documentation).isEqualTo("Test all the things!")
   }
 
+  @Test
+  fun leadingDotsAreNormalized() {
+    val proto = """
+        |message ..package.Message {}
```
swankjesse, line 327: reallly??? Does protoc do this?

swankjesse (approved):

oldergod:
This fix sucks.

PR #2959, opened 2024-05-22
Support very long field names with builders

oldergod:
cc @jpulgarin

### DIFF
--- a/wire-tests/src/jvmKotlinInteropTest/proto-kotlin/com/squareup/wire/protos/kotlin/Percents.kt
+++ b/wire-tests/src/jvmKotlinInteropTest/proto-kotlin/com/squareup/wire/protos/kotlin/Percents.kt
@@ -35,11 +35,25 @@ public class Percents(
   )
   @JvmField
   public val text: String? = null,
+  /**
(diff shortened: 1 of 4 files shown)


wire-tests/src/jvmKotlinInteropTest/proto-kotlin/com/squareup/wire/protos/kotlin/Percents.kt
```diff
@@ -35,11 +35,25 @@ public class Percents(
...
+   * Testing very long field name in builders.
+   */
+  @field:WireField(
+    tag = 2,
+    adapter = "com.squareup.wire.ProtoAdapter#INT32",
+    schemaIndex = 1,
+  )
+  @JvmField
+  public
+      val very_very_long_field_name_placeholder_1_with_very_long_same_placeholder_2_and_different_placeholder_3:
```
Egorand, line 48: Interesting, looks like `public` and `val` on separate lines compiles, but we can probably improve this in KotlinPoet.

Egorand (approved):

PR #2961, opened 2024-05-23
Encoding and JSON support for single unknown enum fields

oldergod:
When an enum has the specific proto3 generated `UNRECOGNIZED` constant and that the value is set to it, the message needs not to send anything on encoding + needs some magic on JSON serialization to replace it by the actual constant tag.

Look at the first commit to see the actual change without the generated code noise.

#2725

### DIFF
--- a/wire-moshi-adapter/src/main/java/com/squareup/wire/MessageJsonAdapter.kt
+++ b/wire-moshi-adapter/src/main/java/com/squareup/wire/MessageJsonAdapter.kt
@@ -73,7 +73,8 @@ internal class MessageJsonAdapter<M : Message<M, B>, B : Message.Builder<M, B>>(
       val index = option / 2
       val fieldBinding = messageAdapter.fieldBindingsArray[index]
 
-      // We peekJson in order to be able to read the value twice if need be.
+      // We peekJson in order to be able to read the value twice if need to access for building
+      // unknown values.
       val peekInput = input.peekJson()
       val value = jsonAdapters[index].fromJson(peekInput)
       val protoAdapter = fieldBinding.adapter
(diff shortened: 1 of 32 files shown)


wire-moshi-adapter/src/main/java/com/squareup/wire/MessageJsonAdapter.kt
```diff
@@ -71,14 +71,28 @@ internal class MessageJsonAdapter<M : Message<M, B>, B : Message.Builder<M, B>>(
...
+        protoAdapter.syntax == Syntax.PROTO_3 &&
+        protoAdapter is EnumAdapter<*>
+      ) {
+        // The tag for this enum's constant is unknown to our runtime. It has been assigned  to
+        // `UNRECOGNIZED(-1)` and we put its actual value in the unknown fields.
+        builder.addUnknownField(fieldBinding.tag, FieldEncoding.VARINT, input.nextInt())
+      } else {
+        // We need to skip it since we passed our peeked Json to the json adapter.
+        input.skipValue()
+      }
```
oldergod, line 90: I have currently no idea how to do that on the Gson side

psymoon (approved):

wire-moshi-adapter/src/main/java/com/squareup/wire/MessageJsonAdapter.kt
```diff
@@ -71,14 +71,28 @@ internal class MessageJsonAdapter<M : Message<M, B>, B : Message.Builder<M, B>>(
         continue
       }
       val index = option / 2
+      val fieldBinding = messageAdapter.fieldBindingsArray[index]
 
-      val value = jsonAdapters[index].fromJson(input)
+      // We peekJson in order to be able to read the value twice if need be.
```
psymoon, line 76:
```suggestion
      // We peekJson in order to be able to read the value twice if need to access for building unknown values.
```

PR #2966, opened 2024-05-24
Add support for parsing extensionRangeOptions

oldergod:
https://protobuf.com/docs/language-spec#extension-ranges

### DIFF
--- a/wire-schema/api/wire-schema.api
+++ b/wire-schema/api/wire-schema.api
@@ -1203,16 +1203,18 @@ public final class com/squareup/wire/schema/internal/parser/ExtendElement {
 }
 
 public final class com/squareup/wire/schema/internal/parser/ExtensionsElement {
-	public fun <init> (Lcom/squareup/wire/schema/Location;Ljava/lang/String;Ljava/util/List;)V
-	public synthetic fun <init> (Lcom/squareup/wire/schema/Location;Ljava/lang/String;Ljava/util/List;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Lcom/squareup/wire/schema/Location;Ljava/lang/String;Ljava/util/List;Ljava/util/List;)V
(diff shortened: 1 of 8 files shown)


wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/SchemaTest.kt
```diff
@@ -1010,7 +1010,7 @@ class SchemaTest {
       }
       fail()
     } catch (expected: IllegalStateException) {
-      assertThat(expected).hasMessage("Syntax error in /sourcePath/message.proto:2:17: expected '{'")
+      assertThat(expected).hasMessage("Syntax error in /sourcePath/message.proto:2:17: expected '{' but was 's'")
```
Egorand, line 1013: Nice!

wire-schema/api/wire-schema.api
```diff
@@ -1203,16 +1203,18 @@ public final class com/squareup/wire/schema/internal/parser/ExtendElement {
 }
 
 public final class com/squareup/wire/schema/internal/parser/ExtensionsElement {
-	public fun <init> (Lcom/squareup/wire/schema/Location;Ljava/lang/String;Ljava/util/List;)V
-	public synthetic fun <init> (Lcom/squareup/wire/schema/Location;Ljava/lang/String;Ljava/util/List;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Lcom/squareup/wire/schema/Location;Ljava/lang/String;Ljava/util/List;Ljava/util/List;)V
```
Egorand, line 1206: Should we be concerned about breaking binary compat?

Egorand (approved):

wire-schema/api/wire-schema.api
oldergod: I'll put that in Wire 5 so I would not say so.

PR #2975, opened 2024-06-05
Support assignment of Java array-types annotations from Kotlin classes

oldergod:


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -1465,7 +1465,9 @@ class KotlinGenerator private constructor(
   }
 
   private fun List<*>.toArrayFieldInitializer(protoType: ProtoType): CodeBlock = buildCodeBlock {
-    add("[")
+    // We use named parameters so that the call-site compiles when the annotation has been
+    // generated as a Java interface, and not Kotlin.
```
swankjesse, line 1469: 👍🏻

swankjesse (approved):

PR #2991, opened 2024-06-14
Filter extend fields based on the emitting rules

oldergod:
Fixes https://github.com/cashapp/misk/pull/3286

This PR filters extend fields (like options) based on the file they are generated in, looking at their package name. Before that, even if one has multiple targets, the first target would generate all the extend fields, all the time.

### DIFF
--- a/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SchemaHandler.kt
+++ b/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SchemaHandler.kt
@@ -174,11 +174,15 @@ abstract class SchemaHandler {
       }
     }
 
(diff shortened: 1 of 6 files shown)


wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/SchemaHandler.kt
```diff
@@ -174,12 +174,16 @@ abstract class SchemaHandler {
       }
     }
 
-    // TODO(jwilson): extend emitting rules to support include/exclude of extension fields.
```
adrw, line 177: Nice!

adrw (approved):

swankjesse (approved):

PR #2993, opened 2024-06-17
EnumMode enum option

oldergod:
This takes precedence over the newly created `KotlinTarget#enumMode` to allow granular migration.

Extension registered here https://github.com/protocolbuffers/protobuf/pull/17158

### DIFF
--- a/wire-gradle-plugin/src/test/projects/kotlin-enum-mode/src/main/proto/enum/geology/period.proto
+++ b/wire-gradle-plugin/src/test/projects/kotlin-enum-mode/src/main/proto/enum/geology/period.proto
@@ -1,0 +1,33 @@
+syntax = "proto2";
+
+package enum.geology;
+
+import "wire/extensions.proto";
+
+option java_package = "com.squareup.enum.geology";
+
+// Target is set to "enum_class" so the option should not have any effect.
+enum Period {
+  option (wire.enum_mode) = "enum_class";
+  CRETACEOUS = 1;
+  JURASSIC = 2;
+  TRIASSIC = 3;
+}
+
+// Target is set to "enum_class" so the option should takes precedence.
+enum Continent {
+  option (wire.enum_mode) = "sealed_class";
+  AFRICA = 0;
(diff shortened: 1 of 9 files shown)


wire-schema/src/jvmMain/resources/wire/extensions.proto
```diff
@@ -43,6 +43,18 @@ extend google.protobuf.FieldOptions {
...
+extend google.protobuf.EnumOptions {
+  /**
+   * Defines how an enum type is to be generated. This is only supported by Kotlin generation.
+   * - 'enum_class': the enum type will be generated as a Kotlin enum class.
+   * - 'sealed_class': the enum type will be generated as a Kotlin sealed class.
+   *
+   * When set, the value of this option takes precedence over the global enum mode that may be set
+   * at the Kotlin target level in the Wire Gradle plugin.
+   */
+  optional string enum_mode = 1190;
```
oldergod, line 55: even though `enum_class` and `sealed_class` are Kotlin words, I think we'd wanna use this option on Java generation as well later, even if they are less explicit.
Hence the option name to `enum_mode` instead of something like `kotlin_enum`

wire-gradle-plugin/src/test/projects/kotlin-enum-mode/src/main/proto/enum/geology/period.proto
```diff
@@ -0,0 +1,33 @@
...
+
+package enum.geology;
+
+import "wire/extensions.proto";
+
+option java_package = "com.squareup.enum.geology";
+
+// Target is set to "enum_class" so the option should not have any effect.
+enum Period {
+  option (wire.enum_mode) = "enum_class";
```
swankjesse, line 11: Love this

wire-schema/src/jvmMain/resources/wire/extensions.proto
swankjesse: I like it

swankjesse (approved):

PR #2994, opened 2024-06-18
Revert code we don't need anymore

oldergod:
cc @deanpapastrat


deanpapastrat (approved):
Looks good to me since this effectively a rollback to the version pre `5.0.0-alpha02-square04`.

deanpapastrat:
Although maybe the commit should be titled `inefficient`?

swankjesse (approved):

PR #2998, opened 2024-06-20
Allow edition parsing

oldergod:
defaulting to proto2 behavior, which is kinda wrong.


wire-runtime/src/commonMain/kotlin/com/squareup/wire/Syntax.kt
```diff
@@ -15,20 +15,51 @@
  */
 package com.squareup.wire
 
-/** Syntax version. */
-enum class Syntax(private val string: String) {
-  PROTO_2("proto2"),
-  PROTO_3("proto3"),
-  ;
+/** Syntax and edition version. */
+sealed class Syntax(internal val string: String) {
```
swankjesse, line 19: I’m tempted to say we create a new thing, `Syntax2` or something, that offers these things, and keep `Syntax` as-is? Making a big binary-incompatible change to `wire-runtime` has a potentially large user impact?

(Why is `Syntax` in the runtime API?)

wire-schema/src/commonTest/kotlin/com/squareup/wire/schema/internal/parser/ProtoParserTest.kt
```diff
@@ -627,6 +628,48 @@ class ProtoParserTest {
...
+    try {
+      ProtoParser.parse(location, proto)
+      fail()
+    } catch (e: IllegalStateException) {
+      assertThat(e).hasMessage("Syntax error in file.proto:2:1: too many syntax or edition definitions")
+    }
+  }
+
+  @Test
+  fun rejectUnknownEditions() {
```
swankjesse, line 657: Love these tests

wire-runtime/src/commonMain/kotlin/com/squareup/wire/Syntax.kt
oldergod: Because it's passed as a parameter to adapters

```kotlin
expect abstract class ProtoAdapter<E>(
  fieldEncoding: FieldEncoding,
  type: KClass<*>?,
  typeUrl: String?,
  syntax: Syntax,
  identity: E? = null,
  sourceFile: String? = null,
) {
```

PR #3005, opened 2024-06-25
Prevent conflicts from same name options of different types

oldergod:


wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -423,7 +423,22 @@ private static void putAllExtensions(
         if (!eligibleAsAnnotationMember(schema, field)) continue;
 
         ProtoMember protoMember = extend.member(field);
-        memberToJavaName.put(protoMember, annotationName(protoFile, field, new ClassNameFactory()));
+        ClassName annotationName =
+            annotationName(protoFile, field, new ClassNameFactory(), "Option");
+        if (memberToJavaName.containsValue(annotationName)) {
+          // To avoid conflicts for same named options on different types, we generate a more
```
Egorand, line 429: Is it possible for us to always pick suffix based on the option type, so that all message options will be named `XxxMessageOption` and all field names - `XxxFieldOption`?

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
```diff
@@ -423,7 +423,22 @@ private static void putAllExtensions(
...
+          // precise name. i.e. 'ObjectiveOption' will become 'ObjectiveFieldOption'.
+          String extendSimpleName = extend.getType().getSimpleName();
+          memberToJavaName.put(
+              protoMember,
+              annotationName(
+                  protoFile,
+                  field,
+                  new ClassNameFactory(),
+                  extendSimpleName.substring(0, extendSimpleName.length() - 1)));
```
Egorand, line 438: Will `extendSimpleName` here always have the value `FieldOptions`, or it can also have the value `MessageOptions`?

Egorand (approved):

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
oldergod: That'd be the ideal solution but the feature has already shipped and we'll be breaking all clients with this change 😢 
So we decided to do that only when conflicts arise.

wire-java-generator/src/main/java/com/squareup/wire/java/JavaGenerator.java
oldergod: Can be any option type, `FieldOption`, `MessageOption`, `EnumValueOption`, etc. This is how we avoid the conflict

PR #3077, opened 2024-08-19
ProtoReader32, a specialization for Kotlin/JS

swankjesse:


wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -356,12 +367,12 @@ internal inline fun <E> ProtoAdapter<E>.commonEncodeByteString(value: E): ByteSt
 
 @Suppress("NOTHING_TO_INLINE")
 internal inline fun <E> ProtoAdapter<E>.commonDecode(bytes: ByteArray): E {
-  return decode(Buffer().write(bytes))
+  return decode(ByteArrayProtoReader(bytes))
```
swankjesse, line 370: I gotta revert this and the one below before merging. Otherwise we call a generated function that might not exist in already-compiled libraries.

wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -356,12 +367,12 @@ internal inline fun <E> ProtoAdapter<E>.commonEncodeByteString(value: E): ByteSt
...
 }
 
 @Suppress("NOTHING_TO_INLINE")
 internal inline fun <E> ProtoAdapter<E>.commonDecode(bytes: ByteString): E {
-  return decode(Buffer().write(bytes))
+  return decode(ByteArrayProtoReader(bytes))
```
swankjesse, line 375: I gotta revert this and the one below before merging. Otherwise we call a generated function that might not exist in already-compiled libraries.

swankjesse:
Next steps:

 * Rename to ProtoReader32
 * Make it an interface with factory functions like `fun ProtoReader32(ByteArray, Int, Int)`
 * Create a function `ProtoReader32.asProtoReader()`
 * Make the base implementation of `ProtoAdapter.decode(ProtoReader32)` delegate by calling `asProtoReader()`
 * Make it opt-in to generate `decode()` functions that accept a `ProtoReader32`
 * Test all combinations of read functions

wire-runtime/api/wire-runtime.api
```diff
@@ -266,31 +272,59 @@ public final class com/squareup/wire/ProtoAdapter$EnumConstantNotFoundException
...
+	public abstract fun readFixed32 ()I
+	public abstract fun readFixed64 ()J
+	public abstract fun readString ()Ljava/lang/String;
+	public abstract fun readUnknownField (I)V
+	public abstract fun readVarint32 ()I
+	public abstract fun readVarint64 ()J
+	public abstract fun skip ()V
+}
+
+public final class com/squareup/wire/ProtoReader32Kt {
```
swankjesse, line 320: I don’t love how this API shows up to Java users

wire-runtime/src/commonMain/kotlin/com/squareup/wire/ByteArrayProtoReader32.kt
```diff
@@ -0,0 +1,469 @@
+// This class is derived from the CodedInputByteBuffer class in Google's "Nano" Protocol Buffer
+// implementation. The original copyright notice, list of conditions, and disclaimer for those
+// classes is as follows:
```
swankjesse, line 3: Keeping this 'cause it’s copied from a thing that had it

wire-runtime/src/commonMain/kotlin/com/squareup/wire/ByteArrayProtoReader32.kt
```diff
@@ -0,0 +1,469 @@
...
+import com.squareup.wire.internal.ProtocolException
+import com.squareup.wire.internal.and
+import com.squareup.wire.internal.shl
+import okio.Buffer
+import okio.ByteString
+import okio.ByteString.Companion.toByteString
+import okio.EOFException
+import okio.IOException
+
+internal class ByteArrayProtoReader32(
```
swankjesse, line 56: This is mostly a copy of ProtoReader.kt.

wire-runtime/src/commonMain/kotlin/com/squareup/wire/ByteArrayProtoReader32.kt
```diff
@@ -0,0 +1,469 @@
...
+  private var pushedLimit: Int = -1
+
+  /** The encoding of the next value to be read. */
+  private var nextFieldEncoding: FieldEncoding? = null
+
+  /** Pooled buffers for unknown fields, indexed by [recursionDepth]. */
+  private val bufferStack = mutableListOf<Buffer>()
+
+  /** Lazily-initialized. */
+  private var protoReader: ProtoReader32AsProtoReader? = null
```
swankjesse, line 84: This variable and the function below are not in ProtoReader. (Should I just use `by lazy` ? I don’t have an intuition on the runtime cost of that)

wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoReader32.kt
```diff
@@ -0,0 +1,183 @@
...
+   * Returns the min length of the next field in bytes. Some encodings have a fixed length, while
+   * others have a variable length. LENGTH_DELIMITED fields have a known variable length, while
+   * VARINT fields could be as small as a single byte.
+   */
+  fun nextFieldMinLengthInBytes(): Int
+}
+
+/** Reads each tag, handles it, and returns a byte string with the unknown fields. */
+@JvmName("-forEachTag") // hide from Java
+inline fun ProtoReader32.forEachTag(tagHandler: (Int) -> Any): ByteString {
```
swankjesse, line 155: This is not a member function in ProtoReader32, but it is a member function in ProtoReader

wire-tests/src/jvmKotlinProtoReader32Test/proto-kotlin/com/squareup/wire/protos/kotlin/alltypes/AllTypes.kt
```diff
@@ -0,0 +1,4122 @@
+// Code generated by Wire protocol buffer compiler, do not edit.
```
swankjesse, line 1: 100% generated code

wire-runtime/src/commonMain/kotlin/com/squareup/wire/ByteArrayProtoReader32.kt
JakeWharton: Seems fine as-is

wire-runtime/src/jsMain/kotlin/com/squareup/wire/ProtoAdapter.kt
```diff
@@ -92,6 +92,11 @@ actual abstract class ProtoAdapter<E> actual constructor(
   /** Read a non-null value from `reader`. */
   actual abstract fun decode(reader: ProtoReader): E
 
+  /** Read a non-null value from `reader`. */
+  actual open fun decode(reader: ProtoReader32): E {
+    return decode(reader.asProtoReader())
+  }
```
JakeWharton, line 98: Is this codepath tested? I saw the new reader being tested and I saw the old reader being tested on a type with the new function. Didn't see the new reader being tested on a type which does not contain the new function, though.

JakeWharton (approved):

wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoReader.kt
```diff
@@ -47,7 +47,17 @@ import okio.IOException
 /**
  * Reads and decodes protocol message fields.
  */
-class ProtoReader(private val source: BufferedSource) {
+open class ProtoReader(private val source: BufferedSource) {
```
Egorand, line 50: Would it simplify the design if we introduced a new interface, e.g. `IProtoReader<Tag>`, and have both `ProtoReader` and `ProtoReader32` implement it? Or that would break backwards compat?

Egorand (approved):

wire-runtime/src/commonTest/kotlin/com/squareup/wire/ProtoReader32Test.kt
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
```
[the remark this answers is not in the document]
swankjesse: Fixed!

wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoReader.kt
swankjesse: I would love it, but the generic `Tag` would be boxed on every use, potentially at significant runtime cost.

wire-runtime/src/commonMain/kotlin/com/squareup/wire/ProtoReader32.kt
swankjesse: Because the new target type is an interface, I think it fits best to implement as an extension function. (I know that we can put implementations in interface functions now, but we’d only be doing that for symmetry).

swankjesse: Also I don’t like the idea of having an inline function that’s polymorphic.

wire-runtime/src/jsMain/kotlin/com/squareup/wire/ProtoAdapter.kt
swankjesse: Ooooh I think it might be accidentally tested, but I think explicitly testing it is much better. Done.

PR #3078, opened 2024-08-28
Fix Gradle project isolation issue when reading a property

liutikas:
Project.getProperties is not Gradle project isolation safe as it attempts to also include properties from parent projects.

Instead use providers.gradleProperty that only checks gradle properties in a given project.

Project isolation feature description
https://docs.gradle.org/current/userguide/isolated_projects.html

### DIFF
--- a/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
+++ b/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WirePlugin.kt
@@ -297,7 +297,7 @@ class WirePlugin : Plugin<Project> {
     if (!hasJavaOutput && !hasKotlinOutput) return
 
     // Indicates when the plugin is applied inside the Wire repo to Wire's own modules.
-    val isInternalBuild = project.properties["com.squareup.wire.internal"].toString() == "true"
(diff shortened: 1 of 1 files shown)


JakeWharton:
This feature relies on being able to read from the parent. Can we force the read on the root project directly?

JakeWharton:
I guess, alternatively, we could apply the property to every project on which the plugin is applied. But reading directly from the root would be preferred, if possible.

liutikas:
When you say "being able to read from parent" I assume you mean being able to read the value set in root gradle.properties https://github.com/square/wire/blob/1aaeb3ca00fc46026bbfe84529692a76dd03d5ca/gradle.properties#L19

is so, then you don't need to read it from the root project, as the property is automatically applied to every project.

Or is there a different place where this is set?

JakeWharton:
Nope, that is the only place.

liutikas:
> Nope, that is the only place.

I think we are all set then. To validate, I created https://github.com/liutikas/propertiesallthewaydown and it does do that propagation.

liutikas:
Cleaned up the CL, i correctly had left toString in there. Can we please rerun the checks? It passes locally now

oldergod:
@liutikas I just released `4.9.11` which does contain your fix.

PR #3093, opened 2024-09-05
Kotlin 2.0.20

oldergod:


.kotlin/metadata/kotlinTransformedMetadataLibraries/com.squareup.okio-okio-3.9.0-appleMain-BlIr1w.klib
JakeWharton, line 1: You need to `.gitignore` this directory.

JakeWharton (approved):

liutikas:
Is there a strong reason for wire-runtime to move to targeting kotlin 2.x and depending on kotlin-stdlib 2.x?

Kotlin 2.0 has been a pain for many to upgrade to, making wire-runtime Kotlin 2.x forces all the user libraries to also transitively force their consumers to upgrade KGP. I would like to pick up https://github.com/square/wire/pull/3078 but to do that that would force androidx.benchmark to also start forcing its users to move to KGP 2.x and we'd like to avoid that.

oldergod:
@liutikas could you not enforce the Kotlin version in your project?

liutikas:
I don't quite follow. We use wire-runtime, if it starts targeting kotlin
2.0 or it adds kotlin-stdlib 2.0 that targets kotlin 2.0, for add these to
the class path of our users. That means it is no longer safe for any of our
users to consume our library using kotlinc 1.9

oldergod:
Sorry I'm not following. Why would it not be safe?

liutikas:
Similar to Java language version that allows you to use new language features in a newer version, kotlin does that same thing with Kotlin language version. If you create a binary using a newer Kotlin language version target it is no longer safe for an old Kotlinc compiler to use APIs from this binary as language version bump is an explicit breaking of backwards compatibility support. An analogous move would be a jar that was targeting java 21 language version should not be use (and it will in fact fail hard) by javac 17

PR #3100, opened 2024-09-11
CompileOnly for the KGP in wire-gradle-plugin

oldergod:


martinbonnin (approved):

JakeWharton:
Is this actually used in a way that works when KGP is not on the classpath?

oldergod:
It's more so that the version isn't transitively passed downstream. The cash-server monorepo has collateral damage with Wire being on 2.x.
I'm not 100% sure this will fix it but that's the try

martinbonnin:
> Is this actually used in a way that works when KGP is not on the classpath?

The wire plugin could fail with a nice message if no KGP is on the classpath (either using reflection or `pluginManager.withId("org.jetbrains.kotlin...") {}`).
`compileOnly` gives more freedom to consumers to chose their KGP version.

JakeWharton:
1. Wire should not require KGP on the classpath since it can generate only Java
2. The Cash server not being able to upgrade Kotlin is not Wire's problem. They can stay on old versions of Wire, force a lower KGP version, or use a KGP exclude. Yes the dependency should be migrated to `compileOnly`, but in a way that's tested for point 1 and not to band aid someone's project being stuck on old Kotlin.

martinbonnin:
> Wire should not require KGP on the classpath since it can generate only Java

Very good point 👍

JakeWharton:
3. If we are concerned about people needing to select their own versions then consider depending on minimum-supported versions rather than the latest. This will also draw issues from people complaining about dependencies with CVEs because they rely on your transitive version, so it's a trade-off.

oldergod:
Thanks for the explanation. The PR as is sounds good to me.

martinbonnin:
> in a way that's tested for point 1

That's also a very good point. There's a very good chance that changing to `compileOnly` breaks the java-only consumers because linking the `WirePlugin` class now fails due to missing KGP symbols. Ideally the plugin is refactored so that "kotlin-only" portions are not accessed unless KGP is present but that's a much bigger lift.

PR #3123, opened 2024-09-26
New @WireEnclosingType annotation, so R8 doesn't prune too much

swankjesse:
We've observed a surprising behavior where Kotlin reflection assumes the nested classes of non-Kotlin classes are themselves non-Kotlin classes.

This assumption is reasonable, but breaks if R8 prunes the `@Metadata` annotation on an enclosing class. The horrible result is a nested class that has a `@Metadata` annotation but isn't considered to be a Kotlin class. Yuck.

This marker annotation should help us to configure R8 to avoid this unfortunate situation, and it might prove useful for other people as well.


JakeWharton (approved):
Not a marker interface? Probably doesn't matter much.

swankjesse:
I started with a marker interface, then changed my mind! I want to leave the opportunity to later expose metadata from the originating message, and an annotation feels like the right fit for that.

oldergod (approved):

mattprecious (approved):

wire-runtime/src/commonMain/kotlin/com/squareup/wire/WireEnclosingType.kt
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * When the enclosing message is pruned, Wire still generates a class for that enclosing message to
+ * preserve the class names and structure.
+ *
+ * This marks such messages and is intended for use by tools like R8 to preserve them.
+ *
+ * https://developer.android.com/build/shrink-code
+ */
+annotation class WireEnclosingType
```
mattprecious, line 26: Late feedback, but does this need a target or a retention specified?

swankjesse: Lets do the same retention that we use for `WireField`

Issue #3127, opened 2024-09-30
Android Client Base64 in Android 7


Xiao1209:
App in Android7 or Android7.1.1 happened this Exception

```
java.lang.NoClassDefFoundError: Failed resolution of: Ljava/util/Base64; (Ask Gemini)
	at com.squareup.wire.internal.GrpcKt.grpcResponseToException(grpc.kt:219)
	at com.squareup.wire.internal.RealGrpcCall.readExactlyOneAndClose(RealGrpcCall.kt:112)
	at com.squareup.wire.internal.RealGrpcCall.access$readExactlyOneAndClose(RealGrpcCall.kt:32)
	at com.squareup.wire.internal.RealGrpcCall$execute$2$2.onResponse(RealGrpcCall.kt:70)
	at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:519)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
	at java.lang.Thread.run(Thread.java:761)
```

Then I fond this code in grpc.kt

```
internal fun GrpcResponse.grpcResponseToException(suppressed: IOException? = null): IOException? {
  grpcStatus?.toIntOrNull()?.takeIf { it != 0 }?.let { grpcStatusInt ->
    (trailers["grpc-status-details-bin"] ?: header("grpc-status-details-bin"))?.let {
      try {
        grpcStatusDetailsBin = Base64.getDecoder().decode(it)
      } catch (e: IllegalArgumentException) {
        throw IOException(
          "gRPC transport failure, invalid grpc-status-details-bin" +
            " (HTTP status=$code, grpc-status=$grpcStatus, grpc-message=$grpcMessage)",
          e,
        )
      }
    }

    return GrpcException(GrpcStatus.get(grpcStatusInt), grpcMessage, grpcStatusDetailsBin)
  }
  return null // Success.
}
```

How Can I update this code?Please

Xiao1209:
Sorry,The code layout is not easy to identify.I update at here
```kotlin
/** Returns an exception if the gRPC call didn't have a grpc-status of 0. */
internal fun GrpcResponse.grpcResponseToException(suppressed: IOException? = null): IOException? {
  var trailers = headersOf()
  var transportException = suppressed
  try {
    trailers = trailers()
  } catch (e: IOException) {
[...]

PR #3139, opened 2024-10-09
Use Okio for base64 decoding

oldergod:
Fixes https://github.com/square/wire/issues/3127

### DIFF
--- a/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/grpc.kt
+++ b/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/grpc.kt
@@ -19,7 +19,6 @@ import com.squareup.wire.GrpcException
 import com.squareup.wire.GrpcStatus
 import com.squareup.wire.ProtoAdapter
 import java.io.Closeable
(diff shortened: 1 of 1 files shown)


wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/grpc.kt
```diff
@@ -228,7 +229,7 @@ internal fun Response.grpcResponseToException(suppressed: IOException? = null):
       }
     }
 
-    return GrpcException(GrpcStatus.get(grpcStatusInt), grpcMessage, grpcStatusDetailsBin, url)
+    return GrpcException(GrpcStatus.get(grpcStatusInt), grpcMessage, grpcStatusDetailsBin?.toByteArray(), url)
```
oldergod, line 232: do I need to `.use { it.toByteArray() }` ?

JakeWharton: Nope

JakeWharton (approved):

PR #3142, opened 2024-10-10
Switch to Palantir's JavaPoet

swankjesse:


swankjesse:
https://github.com/square/javapoet/pull/1014

PR #3147, opened 2024-10-15
Introduce opaque_types for the WireCompiler

oldergod:


gen-tests.gradle.kts
```diff
@@ -213,6 +213,26 @@ val generateAndroidCompactTests by tasks.creating(JavaExec::class) {
   )
 }
 
+// OPAQUE TYPES
+
+val generateOpaqueTypeTests by tasks.creating(JavaExec::class) {
+  group = "Generate Tests"
+  description = "Generates compact Java classes for Android from the test protos"
```
dnkoutso, line 220: nit: is this description copy pasta'd?

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.kt
```diff
@@ -98,6 +98,11 @@ import okio.openZip
  *
  * The `--android-annotations` flag will add the `Nullable` annotation to optional fields.
  *
+ * The `opaque_types` takes a list of qualified named Protobuf types separated by a ','. All types
```
dnkoutso, line 101: nit maybe add `\`--opaque_types\`` to match the rest?

dnkoutso (approved):

gen-tests.gradle.kts
oldergod: fixed!

wire-compiler/src/main/java/com/squareup/wire/WireCompiler.kt
oldergod: fixed!

PR #3151, opened 2024-10-17
[swift] Fix not marking oneof field as `@Indirect` when detecting a cycle

dnkoutso:


soorinpark (approved):

wire-runtime-swift/src/main/swift/propertyWrappers/Indirect.swift
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 @propertyWrapper
-public enum Indirect<Value : ProtoCodable> {
```
dnkoutso, line 17: There is no use of ProtoCodable.

lickel (approved):

luispadron (approved):

PR #3152, opened 2024-10-20
Add wasmJs support

JavierSegoviaCordoba:
The code inside the Wasm source sets is a copy of the JS source sets. I don't know if it would be better to create an intermediate target (something like the web) or keep the duplication for the future.


.github/workflows/build.yml
```diff
@@ -41,11 +41,11 @@ jobs:
 
       - name: Test Native and JS
         run: |
-          ./gradlew -Dkjs=true -Dknative=true -Pswift=false samples:native:build samples:js:build samples:multi-target:build --stacktrace --warning-mode all
+          ./gradlew -Dkjs=true -Dkwasm=true -Dknative=true -Pswift=false samples:native:build samples:js:build samples:multi-target:build --stacktrace --warning-mode all
```
oldergod, line 44:
```suggestion
          ./gradlew -Dkjs=true -Dkwasm=true -Dknative=true -Pswift=false samples:native:build samples:js:build samples:wasm:build samples:multi-target:build --stacktrace --warning-mode all
```

oldergod (commented):
Looks good to me. I need Jake's eyes on it as well.

PR #3185, opened 2024-11-25
grpc: Attempt to read trailers always when parsing grpc exception [DO NOT MERGE]

DoumanAsh:
This should prevent missing grpc-status in case of inability to read message correctly

Unless this is changed, any failure to read body will result in generic IO exception when grpc-status is in trailers

### DIFF
--- a/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/grpc.kt
+++ b/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/grpc.kt
@@ -203,10 +203,10 @@ private fun Response.checkGrpcResponse() {
 internal fun Response.grpcResponseToException(suppressed: IOException? = null): IOException? {
   var trailers = headersOf()
   var transportException = suppressed
-  if (suppressed == null) {
-    try {
-      trailers = trailers()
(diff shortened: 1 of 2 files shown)


DoumanAsh:
I've been notified of https://github.com/square/wire/pull/3087

Actually that makes sense if body read fails, but it doesn't make sense if body is just empty so I'm not sure if this PR is worth it

oldergod (commented):
Thanks for the PR. Do you think you could write a test in `GrpcClientTest` ?

DoumanAsh:
I tried to simulate situation that we encountered in wild but I think it doesn't go correctly through this part
https://github.com/square/wire/blob/master/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/RealGrpcCall.kt#L107-L109

For some reason even when I send empty body, it successfully passes despite exception supposed to be NULL
I'm not very familiar with internals but I guess this test suite doesn't simulate real client behavior

DoumanAsh:
This is fundamentally not correct approach as trailers() might not be available due to how buggy okhttp library is
Correct approach would be to re-write grpc call to correctly handle empty response as distinct case instead of failing to read body

PR #3187, opened 2024-11-26
Promote google/protobuf/api.proto as a runtime proto

oldergod:
this includes its dependencies. This is useful for people who want to define a path on their RPCs without having to load protobuf on their proto path manually.


JakeWharton (approved):
I don't know what this means but if you want a green checkmark I am your guy

oldergod:
Protobuf exposes option so that someone can embed the path of an RPC in the proto schema like so
```
  rpc MyRpc (MyRequest) returns (MyResponse) {
    option (google.api.http) = {
      post: "/cash/my-path"
    };
  }
```

Before this PR, people wanting to use this option would have to manually add `google/protobuf/api.proto` to their path. After the PR, Wire does it by default.

Issue #3202, opened 2024-12-05
[iOS] Heap buffer overflow in ProtoWriter.swift when writing messages with more than 5 layers of nesting


amorde:
While running with the Address Sanitizer enabled I found a buffer overflow in `ProtoWriter.swift`.

![Image](https://github.com/user-attachments/assets/fb2c8da2-cceb-4711-a0a9-cf8a42a69931)

![Image](https://github.com/user-attachments/assets/4413310c-b428-4361-be78-1d3fa377b90c)

The issue is the `+ 1` in the failing line combined with the fact that the caller increments the index immediately before: https://github.com/square/wire/blob/67b73b50ef79c063a2dacf675a15a9c65828371c/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoWriter.swift#L711-L714

It attempts to copy and deinitialize memory that has not been initialized in the source buffer, specifically out of bounds by 1.

I've written a test that can reproduce this with the address sanitizer enabled and will open a PR with it.

PR #3203, opened 2024-12-06
Fix buffer overflow and data corruption when a type has more than 5 layers of nesting

amorde:
Fixes https://github.com/square/wire/issues/3202

### DIFF
--- a/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoWriter.swift
+++ b/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoWriter.swift
@@ -708,13 +708,13 @@ public final class ProtoWriter {
         }
 
         if let isProto3 = isProto3 {
-            messageStackIndex += 1
-            if messageStackIndex >= messageStackCapacity {
+            if messageStackIndex + 1 >= messageStackCapacity {
                 expandMessageStack()
             }
             let frame = MessageFrame(
                 isProto3: isProto3
             )
+            messageStackIndex += 1
             messageStack.advanced(by: messageStackIndex).initialize(to: frame)
 
             // Cache this value as an ivar for quick access.
--- a/wire-runtime-swift/src/test/proto/nested5.proto
+++ b/wire-runtime-swift/src/test/proto/nested5.proto
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Square Inc.
+ * Copyright 2024 Square Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,11 +14,12 @@
  * limitations under the License.
  */
 syntax = "proto2";
-import "nested5.proto";
 
-message Nested4 {
+import "nested6.proto";
+
+message Nested5 {
   required string name = 1;
 
-  optional Nested5 nested = 2;
+  optional Nested6 nested = 2;
 }
 
--- a/wire-runtime-swift/src/test/swift/ProtoWriterTests.swift
+++ b/wire-runtime-swift/src/test/swift/ProtoWriterTests.swift
@@ -255,6 +255,56 @@ final class ProtoWriterTests: XCTestCase {
         """)
     }
 
+    func testEncodeMessageWithStackExpansion() throws {
+        let writer = ProtoWriter()
+        let message = Nested1(name: "name1") { v1 in
+          v1.nested = Nested2(name: "name2") { v2 in
+            v2.nested = Nested3(name: "name3") { v3 in
+              v3.nested = Nested4(name: "name4") { v4 in
+                v4.nested = Nested5(name: "name5") { v5 in
+                  v5.nested = Nested6(name: "name6")
+                }
+              }
+            }
+          }
+        }
+
+        try writer.encode(tag: 1, value: message)
+
+        assertBufferEqual(writer, """
+            0A       // (Tag 1 | Length Delimited)
+            34       // Length 52 for Nested1
+              0A         // (Tag 1 | Length Delimited)
+              05         // Length 5 for name1
+              6E616D6531 // UTF-8 Value "name1"
+              12         // (Tag 2 | Length Delimited)
+              2B         // Length 43
+                0A         // (Tag 1 | Length Delimited)
+                05         // Length 5 for name2
+                6E616D6532 // UTF-8 Value "name2"
+                12         // (Tag 2 | Length Delimited)
+                22         // Length 34
+                  0A         // (Tag 1 | Length Delimited)
+                  05         // Length 5 for name3
+                  6E616D6533 // UTF-8 Value "name3"
+                  12         // (Tag 2 | Length Delimited)
+                  19         // Length 25
+                    0A         // (Tag 1 | Length Delimited)
+                    05         // Length 5 for name4
+                    6E616D6534 // UTF-8 Value "name4"
+                    12         // (Tag 2 | Length Delimited)
+                    10         // Length 16
+                      0A         // (Tag 1 | Length Delimited)
+                      05         // Length 5 for name5
+                      6E616D6535 // UTF-8 Value "name5" 
+                      12         // (Tag 2 | Length Delimited)
+                      07         // Length 7
+                        0A         // (Tag 1 | Length Delimited)
+                        05         // Length 5 for name6
+                        6E616D6536 // UTF-8 Value "name6"
+        """)
+    }
+
     func testEncodeString() throws {
         let writer = ProtoWriter()
         try writer.encode(tag: 1, value: "foo")
--- a/wire-runtime-swift/src/test/proto/nested6.proto
+++ b/wire-runtime-swift/src/test/proto/nested6.proto
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Square Inc.
+ * Copyright 2024 Square Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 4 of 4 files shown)


wire-runtime-swift/src/test/swift/ProtoWriterTests.swift
```diff
@@ -255,6 +255,56 @@ final class ProtoWriterTests: XCTestCase {
...
+                  v5.nested = Nested6(name: "name6")
+                }
+              }
+            }
+          }
+        }
+
+        try writer.encode(tag: 1, value: message)
+
+        assertBufferEqual(writer, """
```
amorde, line 274: this test fails without the change.

here's the bytes after the fix, followed by the bytes before the fix. before the fix one of the bytes is actually missing - the `0A (Tag 1 | Length Delimited)` byte that preceeds the `name2` string value

      // 0A 34 0A 05 6E616D6531 12 2B 0A 05 6E616D6532 12 22 0A 05 6E616D6533 12 19 0A 05 6E616D65 3412 10 0A 05 6E616D6535 1207 0A 05 6E616D6536
      // 0A 34 0A 05 6E616D6531 12 2B XX 05 6E616D6532 12 22 0A 05 6E616D6533 12 19 0A 05 6E616D65 3412 10 0A 05 6E616D6535 1207 0A 05 6E616D6536
      //                               \
      //                                \_ missing byte

dnkoutso (approved):
💯 great work!

cc @lickel for one more additional review.

wire-runtime-swift/src/main/swift/ProtoCodable/ProtoWriter.swift
```diff
@@ -778,7 +778,7 @@ public final class ProtoWriter {
         let newCapacity = messageStackCapacity * 2
 
         let newMessageStack = UnsafeMutablePointer<MessageFrame>.allocate(capacity: newCapacity)
-        newMessageStack.moveInitialize(from: messageStack, count: messageStackIndex + 1)
+        newMessageStack.moveInitialize(from: messageStack, count: messageStackCapacity)
```
lickel, line 781: Should this just be the stack index? (Eg just remove the +1)

Expanding the stack doesn't necessarily mean you're at the end of the stack, does it?

I think I'd want to see a few more test cases like:

Size is 4, write 2 bytes, then try to append a 4 byte message. I would guess that index is 2, and your change would shift the index to 4.

wire-runtime-swift/src/test/proto/nested5.proto
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright 2020 Square Inc.
```
dnkoutso, line 2: nit for these to use 2024?

wire-runtime-swift/src/test/swift/ProtoWriterTests.swift
jamieQ: this doesn't seem to repro for me... did you have to do anything special? what toolchain were you using?

wire-runtime-swift/src/main/swift/ProtoCodable/ProtoWriter.swift
jamieQ: i think we don't want to use `messageStackCapacity` since in theory it could still have uninitialized memory that we presumably should not read when copying it over into the new storage location.

IMO whatever the ultimate change, the code should probably be symmetric with the analogous logic in `ProtoReader` ([here](https://github.com/square/wire/blob/67b73b50ef79c063a2dacf675a15a9c65828371c/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoReader.swift#L1001)), and that codepath should also probably get a comparable test case.

if the invariant we want is: '`messageStackIndex + 1` is always the number of initialized elements in the message stack buffer', then we should make sure that always holds. in this case maybe changing where the index increment & test happens would make the overall logic more straightforward & consistent. e.g. we'd change this: https://github.com/square/wire/blob/67b73b50ef79c063a2dacf675a15a9c65828371c/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoWriter.swift#L711-L714

to something like what is done in `ProtoReader`: https://github.com/square/wire/blob/67b73b50ef79c063a2dacf675a15a9c65828371c/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoReader.swift#L136-L140

amorde: I had originally done it that way but was worried about some other issue cropping up. Since `ProtoReader` has similar code I'll go back to that version

wire-runtime-swift/src/test/swift/ProtoWriterTests.swift
amorde: I was using Xcode 16.0

amorde: what part doesn't repro for you?

wire-runtime-swift/src/main/swift/ProtoCodable/ProtoWriter.swift
naftaly: Typically if working if an unsafe language we wouldn't really allow this type of random access with pointer and size tracking a bit scattered. It would be great if we could do the same here. Maybe we could encapsulate the message stack into it's own class which closes up access to its internals, it might help a bunch with a few of the problems enumerated by @jamieQ.

wire-runtime-swift/src/test/swift/ProtoWriterTests.swift
amorde: Hmm, I can't seem to repro the corrupted data either. Perhaps I encountered this while testing some other fix? I'm not sure. I was copying those hex strings from the assertions that were failing 🤔

wire-runtime-swift/src/main/swift/ProtoCodable/ProtoWriter.swift
amorde: I agree, that was a much bigger refactor than I was prepared to do at the moment. I'd like to fix this issue without requiring a big refactor

amorde: > I think I'd want to see a few more test cases like:
> 
> Size is 4, write 2 bytes, then try to append a 4 byte message. I would guess that index is 2, and your change would shift the index to 4.

This stack appears to be specific to length delimited fields, not bytes, and tracks the number of lengths that have been written or something along those lines. So in the test case it goes like this:

1. Start writing the entire message `Nested 1` with `try writer.encode(tag: 1, value: message)`
2. in `beginLengthDelimitedEncode`, increment `messageStackIndex` from -1 to 0.
3. Start writing the `nested` property of `Nested1` via `try protoWriter.encode(tag: 2, value: self.nested)` in `encode(to:)`
4. in `beginLengthDelimitedEncode`, increment `messageStackIndex` from 0 to 1.
5. Start writing the `nested` property of `Nested2` via `try protoWriter.encode(tag: 2, value: self.nested)` in `encode(to:)`
6. in `beginLengthDelimitedEncode`, increment `messageStackIndex` from 1 to 2.
7. Start writing the `nested` property of `Nested3` via `try protoWriter.encode(tag: 2, value: self.nested)` in `encode(to:)`
8. in `beginLengthDelimitedEncode`, increment `messageStackIndex` from 2 to 3.
9. Start writing the `nested` property of `Nested4` via `try protoWriter.encode(tag: 2, value: self.nested)` in `encode(to:)`
10. in `beginLengthDelimitedEncode`, increment `messageStackIndex` from 3 to 4.
11. Start writing the `nested` property of `Nested5` via `try protoWriter.encode(tag: 2, value: self.nested)` in `encode(to:)`
12. in `beginLengthDelimitedEncode`, increment `messageStackIndex` from 4 to 5.

At this point we are at index 5 which is equal to the initial capacity set here: 

https://github.com/square/wire/blob/67b73b50ef79c063a2dacf675a15a9c65828371c/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoWriter.swift#L52

but we have not yet initialized that _index_. The number of bytes we have initialized is 5 (indices 0-4).

Now that `messageStackIndex == 5`, we hit the call to `expandMessageStack()` 
https://github.com/square/wire/blob/67b73b50ef79c063a2dacf675a15a9c65828371c/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoWriter.swift#L712-L713

In `expandMessageStack()` it attempts to copy and deinitialize 6 bytes from the source buffer because `messageStackIndex + 1 == 6` 

https://github.com/square/wire/blob/67b73b50ef79c063a2dacf675a15a9c65828371c/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoWriter.swift#L781

which copies garbage data into index 5. We then call `initialize(to:)` at index 5. This is potentially an invalid use of `initialize(to:)`, depending on whether the `MessageFrame` type is considered a "trivial type" as mentioned in the [docs](https://developer.apple.com/documentation/swift/unsafemutablepointer/initialize(to:)):

> The destination memory must be uninitialized or the pointer’s Pointee must be a trivial type

wire-runtime-swift/src/test/swift/ProtoWriterTests.swift
amorde: Assuming this was a false signal I encountered then its a relief that the actual data that was encoded was not impacted

jamieQ: > what part doesn't repro for you?

the test still passes even if i run with the implementation change as it was vs the fixed version.

> Assuming this was a false signal I encountered then its a relief that the actual data that was encoded was not impacted

agreed. but does the test actually fail then without the change? obviously good to have it regardless so that something is exercising the 'grow the buffer' path, but also makes me feel it might provide less safety than it might aspire to. ideally these tests should probably run in a mode with ASAN on as part of CI.

jamieQ (approved):

Issue #1448, opened 2020-03-04
Large proto definitions may generate methods which are too large to be compiled


damianw:
Wire version: 3.1.0

For example, [Perfetto's FtraceEvent](https://android.googlesource.com/platform/external/perfetto/+/master/protos/perfetto/trace/ftrace/ftrace_event.proto) generates [this Kotlin file](https://gist.github.com/damianw/661b963c8c9eb4337088a3dc9d182c0d). Attempting to compile it (with Kotlin 1.3.70) results in:

```
e: java.lang.IllegalStateException: Backend Internal error: Exception during file facade code generation
File being compiled: file:///.../build/generated/source/wire/perfetto/protos/FtraceEvent.kt
The root cause java.lang.RuntimeException was thrown at: org.jetbrains.kotlin.codegen.ClassFileFactory$OutputClassFile.asByteArray(ClassFileFactory.java:315)
        at org.jetbrains.kotlin.backend.common.CodegenUtil.reportBackendException(CodegenUtil.kt:247)
        at org.jetbrains.kotlin.codegen.PackageCodegenImpl.generate(PackageCodegenImpl.java:78)
        at org.jetbrains.kotlin.codegen.DefaultCodegenFactory.generatePackage(CodegenFactory.kt:88)
        at org.jetbrains.kotlin.codegen.DefaultCodegenFactory.generateModule(CodegenFactory.kt:67)
        ...
Caused by: java.lang.RuntimeException: Error generating class file perfetto/protos/FtraceEvent$Companion$ADAPTER$1.class (compiled from [/.../build/generated/source/wire/perfetto/protos/FtraceEvent.kt]): Method too large: perfetto/protos/FtraceEvent$Companion$ADAPTER$1.redact (Lperfetto/protos/FtraceEvent;)Lperfetto/protos/FtraceEvent;
        at org.jetbrains.kotlin.codegen.ClassFileFactory$OutputClassFile.asByteArray(ClassFileFactory.java:315)
        at org.jetbrains.kotlin.cli.common.output.OutputUtilsKt.writeAll(outputUtils.kt:32)
        at org.jetbrains.kotlin.cli.common.output.OutputUtilsKt.writeAll(outputUtils.kt:42)
        at org.jetbrains.kotlin.cli.jvm.compiler.KotlinToJVMBytecodeCompiler.writeOutput(KotlinToJVMBytecodeCompiler.kt:101)
        ...
[...]

PR #3214, opened 2024-12-12
Kotlin, non-builder: chunk redact function every 100 fields

damianw:
This adds a workaround to an issue discussed in #1448, where the generated `redact` function is too large to be compiled.

It's much more likely to hit this issue with `redact` than the other functions because of [KT-73789](https://youtrack.jetbrains.com/issue/KT-73789/), which results in methods which are significantly longer than necessary specifically due to the repeated `?.let` construction.


oldergod (approved):
thank you for the fix!
I've replaced the test with a golden file which will also confirm compilation is good.
And did a quick refactor so that we clearly see the sequential mutation.

```kotlin
override fun redact(`value`: HundredsFields): HundredsFields {
  var result = redact0(`value`)
  result = redact1(result)
  result = redact2(result)
  return result
}
```

PR #3215, opened 2024-12-15
Fix path normalization for Windows OS

DimaVilda:
**Issue**:
The issue goes from kafka-ui thread https://github.com/kafbat/kafka-ui/pull/261 which uses 'Location.get(...)' function  from your library.
- Location paths don't work correctly on Windows due to path separator differences
- Windows uses backslashes (\) while proto imports consistently use forward slashes (/)
- This causes multiple NPEs on Linker.link(...) method in kafka-ui tests on Windows as the path from  'Location.get(...)' gets incorrect result.

**Example**:
If in Location.get(...) we pass:
1st param base - C:\Users\username\IdeaProjects\kafka-ui\api\target\test-classes\protobuf-serde
2nd param path - language\language.proto
the result is - C:\Users\username\IdeaProjects\kafka-ui\api\target\test-classes\**protobuf-serde/language\language.proto** 

And then Linker.link(...) throws NPE:
Screenshot 2024-12-15 130433

**Solution**:
- Normalized paths in Location.get(...) method to consistently use forward slashes
- Added unit test coverage for Location.get(...) method

### DIFF
--- a/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Location.kt
+++ b/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/Location.kt
@@ -70,7 +70,9 @@ data class Location(
       base: String,
       path: String,
     ): Location {
-      return Location(base.trimEnd('/'), path, -1, -1)
+      val normalizedBase = base.replace('\\', '/')
+      val normalizedPath = path.replace('\\', '/')
+      return Location(normalizedBase.trimEnd('/'), normalizedPath, -1, -1)
     }
   }
 }
--- a/wire-schema/src/commonTest/kotlin/com/squareup/wire/schema/LocationTest.kt
+++ b/wire-schema/src/commonTest/kotlin/com/squareup/wire/schema/LocationTest.kt
@@ -1,8 +1,23 @@
+/*
+ * Copyright (C) 2024 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.squareup.wire.schema
 
-import kotlin.test.Test
 import assertk.assertThat
 import assertk.assertions.isEqualTo
+import kotlin.test.Test
 
 class LocationTest {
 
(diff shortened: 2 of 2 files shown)


oldergod:
@DimaVilda Thanks for the PR. If existing tests are not affected, I'm happy to merge your fix. It looks like some tests on failing on Windows right now.

DmytroVilda:
> @DimaVilda Thanks for the PR. If existing tests are not affected, I'm happy to merge your fix. It looks like some tests on failing on Windows right now.

hi @oldergod ! Np, I will check it asap and in case I was wrong will fix it

DimaVilda:
> @DimaVilda Thanks for the PR. If existing tests are not affected, I'm happy to merge your fix. It looks like some tests on failing on Windows right now.

hi again there @oldergod 
I figured out what is the issue of a failed test. Ehh, it's a bit long topic but I try to put it shortly and then we can decide the next steps:
The problem of a failed SchemaLoaderTest test is here: 

  private val fs = FakeFileSystem().apply {
    if (Path.DIRECTORY_SEPARATOR == "\\") emulateWindows() else emulateUnix()
  }

the in emptyPackagedProtoMessage() test we get NPE:
    loader.initRoots(
      sourcePath = listOf(Location.get(fs.workingDirectory.toString())),  --> NPE
      protoPath = listOf(Location.get(fs.workingDirectory.toString())),
    )

Cause:
When we pass reverted 'F:/' it fails because of how the current implementation of okio jvm Path toPath(). This function only recognizes Windows volume paths (e.g., "F:") when they use backslashes, despite Windows accepting **both** forward and backward slashes.
Example:
Call "F:/".toPath()
Expected result: A path representing "F:/"
Actual result: A path representing just "F:" (without the slash)

This issue comes from:
private fun Buffer.startsWithVolumeLetterAndColon(slash: ByteString): Boolean {
    if (slash != BACKSLASH) return false  // This causes forward slashes to fail
    ...
}

While "F:\\".toPath() works correctly, "F:/".toPath() doesn't, even though both formats are valid on Windows. Since Windows accepts both slash types interchangeably, the path parsing should ideally handle both formats consistently.
We can either fix test or put this question to okio Path team like why they don't accept both SLASH and BACKSLASH in the volume letter check.
Hope my point is clear

oldergod:
It doesn't look like it's part of the spec though that Windows supports or not forward slashes: https://retrocomputing.stackexchange.com/a/28348

@swankjesse did you look into it when making Okio's FileSystem?

DimaVilda:
hey hi @oldergod and @swankjesse !
You got any updates about it ?

DimaVilda:
hi there!
Do any updates regarding this ?
 @oldergod and @swankjesse

oldergod:
Not really. It looks like something that should be addressed in Okio if needs be though. I think an issue with a quick failing test over there would help a lot https://github.com/square/okio/

oldergod:
Addressing it in https://github.com/square/okio/pull/1802 and https://github.com/square/wire/pull/3576

PR #3217, opened 2024-12-17
Add support for mutable messages in Wire's Kotlin Generator.

tikurahul:
* Sometimes its very useful to avoid needing to create new object graph instances when all we need to do is to serialize the object graph into a `ProtoSink`.
* Mutable messages solves for this, for code that is particularly performance sensitive where a large number of objects would have otherwise been created which adds unnecesssary GC pressure.
* Mutable messages should be **used very carefully**. There are **no guarantees of safety whatsoever**.
* Mutable messages don't use `Builder`s and so `newBuilder()` throws an `UnsupportedOperationException()`. 
* Mutable messages don't support `copy(...)`.
* When using mutable messages, `hashCode()` is no longer cached.
* Mutable messages don't support `redact()`.

Test: Added a unit test in `KotlinGeneratorTest`.


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -700,7 +706,10 @@ class KotlinGenerator private constructor(
       .returns(BOOLEAN)
 
     val body = buildCodeBlock {
-      addStatement("if (%N === this) return·true", otherName)
+      if (!generateMutableMessages) {
+        // This is true iff the message is not mutable
```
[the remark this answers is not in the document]
tikurahul: Updated the comment.

wire-golden-files/src/main/kotlin/squareup/wire/mutable/MutableHeader.kt
```diff
@@ -0,0 +1,112 @@
...
+import okio.ByteString
+
+public class MutableHeader(
+  @field:WireField(
+    tag = 1,
+    adapter = "com.squareup.wire.ProtoAdapter#UINT64",
+    schemaIndex = 0,
+  )
+  public var id: Long? = null,
+  unknownFields: ByteString = ByteString.EMPTY,
```
JakeWharton, line 39: I think we need to make `unknownFields` be `open` on the base type so that it can be overridden here as mutable. Otherwise there's no way to clear or copy in unknown fields.

wire-golden-files/src/main/kotlin/squareup/wire/mutable/MutableHeader.kt
```diff
@@ -0,0 +1,112 @@
...
+) : Message<MutableHeader, Nothing>(ADAPTER, unknownFields) {
+  @Deprecated(
+    message = "Shouldn't be used in Kotlin",
+    level = DeprecationLevel.HIDDEN,
+  )
+  override fun newBuilder(): Nothing = throw AssertionError("newBuilder() is unsupported for Mutable message types")
```
JakeWharton, line 45: This should probably be https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-unsupported-operation-exception/

wire-golden-files/src/main/kotlin/squareup/wire/mutable/MutableHeader.kt
```diff
@@ -0,0 +1,112 @@
...
+            else -> reader.readUnknownField(tag)
+          }
+        }
+        return MutableHeader(
+          id = id,
+          unknownFields = unknownFields
+        )
+      }
+
+      override fun redact(`value`: MutableHeader): MutableHeader = throw UnsupportedOperationException("redact() is unsupported for Mutable message types")
```
JakeWharton, line 107: I wouldn't capitalize "mutable" here.

JakeWharton (approved):

wire-golden-files/src/main/kotlin/squareup/wire/mutable/MutableHeader.kt
tikurahul: That makes sense. Given that this requires changes to `wire-runtime`, should I make the a change in the `wire-runtime` first (a separate PR), and then land the corresponding change in `wire-kotlin-generator` ? (after having landed this).

wire-golden-files/src/main/kotlin/squareup/wire/mutable/MutableHeader.kt
tikurahul: Done.

wire-golden-files/src/main/kotlin/squareup/wire/mutable/MutableHeader.kt
tikurahul: Done.

wire-golden-files/src/main/kotlin/squareup/wire/mutable/MutableHeader.kt
JakeWharton: I think it's okay to do here. Then we correctly "blame" the change on needing to override for mutable support in case anyone goes spelunking in the future.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -700,7 +717,10 @@ class KotlinGenerator private constructor(
       .returns(BOOLEAN)
 
     val body = buildCodeBlock {
-      addStatement("if (%N === this) return·true", otherName)
+      if (!mutableTypes) {
+        // We cannot rely on referential equality if the message is mutable.
```
JakeWharton, line 721: Is this true? I don't think this is true.

oldergod: False. True, it's false. True, that's my fault.

tikurahul: Clarified my comment. Please tell me if this reads better.

wire-golden-files/src/main/kotlin/squareup/wire/mutable/MutableHeader.kt
tikurahul: Okay, made the change to `wire-runtime` and updated the generator.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
tikurahul: Reworded this.

JakeWharton: I meant the whole conditional can go I think. Mutable or not, `===` should always be a valid shortcut for `equals` for any object.

tikurahul: You are right 😆. Will fix that and send another update.

wire-golden-files/src/main/kotlin/squareup/wire/mutable/MutableHeader.kt
```diff
@@ -0,0 +1,112 @@
...
+  public var id: Long? = null,
+  override var unknownFields: ByteString = ByteString.EMPTY,
+) : Message<MutableHeader, Nothing>(ADAPTER, unknownFields) {
+  @Deprecated(
+    message = "Shouldn't be used in Kotlin",
+    level = DeprecationLevel.HIDDEN,
+  )
+  override fun newBuilder(): Nothing = throw UnsupportedOperationException("newBuilder() is unsupported for mutable message types")
+
+  override fun equals(other: Any?): Boolean {
```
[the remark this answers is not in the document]
tikurahul: Did you mean that we should only support referential equality ? 

I think it's also okay for us to also compare the values. Otherwise it might just end up being confusing.

wire-golden-files/src/main/kotlin/squareup/wire/mutable/MutableHeader.kt
```diff
@@ -0,0 +1,112 @@
...
+    tag = 1,
+    adapter = "com.squareup.wire.ProtoAdapter#UINT64",
+    schemaIndex = 0,
+  )
+  public var id: Long? = null,
+  override var unknownFields: ByteString = ByteString.EMPTY,
+) : Message<MutableHeader, Nothing>(ADAPTER, unknownFields) {
+  @Deprecated(
+    message = "Shouldn't be used in Kotlin",
+    level = DeprecationLevel.HIDDEN,
```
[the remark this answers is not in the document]
tikurahul: This existed previously so I can't really take credit 😆

Issue #3226, opened 2024-12-31
Support for parsing weak imports


paul35621:
I use `ProtoParser` to detect if a file with a proto extension is really a Protocol Buffers file. Sometimes someone uses a weak import and parsing fails. It would be nice if weak imports are supported.

[This is relevant parsing code is here.](https://github.com/square/wire/blob/54e032b1dd520fefdfce4799fda8470c4a97f1c3/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/internal/parser/ProtoParser.kt#L115)

I made [pull request 3227](https://github.com/square/wire/pull/3227) that implements the parsing. (not the linking)

## [Specification](https://protobuf.com/docs/language-spec#imports)

```
import = "import" [ "weak" | "public" ] strLit ";"
```

A "weak" import means that the import is not required as a runtime dependency. Fields that depend on types defined in these imports can also be marked with a field option named `weak`.

Use of weak imports and weak field options is strongly discouraged and is supported by few target runtimes.

PR #3227, opened 2024-12-31
Added support for weak imports in ProtoParser, ProtoFile and ProtoFileElement

paul35621:
I changed `ProtoParser`, `ProtoFile` and `ProtoFileElement` to support weak imports.

I didn't try to compile this, but if it does compile then the new behaviour would be that weak imports are ignored, since I added a field `weakImports` to `ProtoFile` and `ProtoFileElement` that is not read outside of these classes. The current situation is that parsing fails when the file contains a weak import, so I consider that an improvement.

See [issue 3226](https://github.com/square/wire/issues/3226).

### DIFF
--- a/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
+++ b/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/ProtoFile.kt
@@ -25,6 +25,7 @@ data class ProtoFile(
   val location: Location,
   val imports: List<String>,
   val publicImports: List<String>,
+  val weakImports: List<String>,
   val packageName: String?,
   val types: List<Type>,
   val services: List<Service>,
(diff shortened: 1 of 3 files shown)


oldergod:
I can't push to this branch but I've modified an existing test here: https://github.com/square/wire/compare/bquenaudon.2025-01-06.weakimports?expand=1
also dumped API and spotless

paul35621:
Alright, great! Should I integrate that in this pull request, or what would be the next step?

JakeWharton:
Yes, please add a test to this PR so we can validate that it and the others pass with the new changes.

paul35621:
@JakeWharton Can you edit this pull request or make a new one? The code with the test is already on the **bquenaudon.2025-01-06.weakimports** branch. I've tried to figure out how to make a pull request, but it's above my head.

PR #3233, opened 2025-01-07
Fix code generation for mutable types.

tikurahul:
* Handle `enum`s gracefully. They will never need to be `Mutable`.
* For `optional` `string` types, use the `!!` operator when dispatching a call to `sanitize(...)`. There are no guarantees of safety here anyway, so this should be okay.

Test: Add golden file tests.


wire-golden-files/src/main/kotlin/squareup/wire/mutable/MutablePacket.kt
```diff
@@ -36,14 +38,17 @@ public class MutablePacket(
     schemaIndex = 0,
   )
   public var header_: MutableHeader? = null,
+  payload: List<MutablePayload> = emptyList(),
```
oldergod, line 41: Maybe food for follow up but do you want mutable lists here?

oldergod (approved):

PR #3245, opened 2025-01-10
Be more graceful when handing null values

tikurahul:
Test: Updated golden tests.

### DIFF
--- a/wire-golden-files/src/main/kotlin/squareup/wire/mutable/MutablePayload.kt
+++ b/wire-golden-files/src/main/kotlin/squareup/wire/mutable/MutablePayload.kt
@@ -93,7 +93,7 @@ public class MutablePayload(
 
   override fun toString(): String {
     val result = mutableListOf<String>()
-    if (preamble != null) result += """preamble=${sanitize(preamble!!)}"""
+    if (preamble != null) result += """preamble=${sanitize(preamble ?: "")}"""
     if (content != null) result += """content=$content"""
     if (type != null) result += """type=$type"""
     if (footers.isNotEmpty()) result += """footers=${sanitize(footers)}"""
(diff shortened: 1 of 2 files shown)


wire-golden-files/src/main/kotlin/squareup/wire/mutable/MutablePayload.kt
```diff
@@ -93,7 +93,7 @@ public class MutablePayload(
 
   override fun toString(): String {
     val result = mutableListOf<String>()
-    if (preamble != null) result += """preamble=${sanitize(preamble!!)}"""
```
oldergod, line 96: If `preamble` is not null when would `preamble ?: ""` make sense?

oldergod (approved):

wire-golden-files/src/main/kotlin/squareup/wire/mutable/MutablePayload.kt
JakeWharton: I think doing `?.let` would be best, but there's that compiler bug with them...

tikurahul: Was the thought that it's confusing to read ? 

I considered one other thing which is to make `sanitize(...)` accept `String?` but I would have to rename the list variant to `sanitizeList` to disambiguate. Thought this was better than that.

tikurahul:
Sorry, there is not that much different. I was trying to avoid the use of the `!!` operator but that might be clearer perhaps.

oldergod:
@tikurahul let's use `?.let {}` on the whole line? Does that satisfy your case?

tikurahul:
I feel like what's there right now might be okay; and better than then `?.let { ...}`. The `!!` is easier to read than the former.

JakeWharton:
`?.let` produces the same conditional but omits the exception code from `!!`. It's what we use in other functions like `redact` already.

Issue #2748, opened 2023-12-12
Options on external enum constants are not linked


damianw:
This might be a more general issue than just enum constants, and I also might be "holding it wrong", but I can't figure out if it's intentional that options on external enum constants are not linked.

The use case I have in mind is to encode metadata about particular enum constants directly in protobuf, and use this in a custom `SchemaHandler`. This doesn't seem to be possible unless I add some kind of stub to force Wire to link those types.

A test for `OptionsLinkingTest.kt`:

```Kotlin
  @Test
  fun extensionTypesInExternalFileWithEnumOptions() {
    fs.add(
      "source-path/a.proto",
      """
        |syntax = "proto3";
        |
        |package src;
        |
        |import "extensions.proto";
        |import "range.proto";
        |
        |message B {
        |  option (lib.my_message_options) = {
        |    a: A_ONE
        |  };
        |}
      """.trimMargin()
    )
    fs.add(
      "proto-path/extensions.proto",
      """
        |syntax = "proto3";
        |
        |package lib;
        |
        |import "google/protobuf/descriptor.proto";
        |import "range.proto";
        |
        |message MyEnumValueOptions {
        |  Range length = 1;
        |}
        |
        |message MyMessageOptions {
        |  A a = 1;
        |}
        |
        |extend google.protobuf.EnumValueOptions {
        |  MyEnumValueOptions my_enum_value_options = 22002;
        |}
        |
        |extend google.protobuf.MessageOptions {
        |  MyMessageOptions my_message_options = 22002;
        |}
        |
        |enum A {
        |  A_UNSPECIFIED = 0;
        |  A_ONE = 1 [
        |    (my_enum_value_options) = {
        |      length: {
        |        min: 20
        |        max: 80
        |      }
        |    }
        |  ];
        |}
      """.trimMargin()
    )
    fs.add(
      "proto-path/range.proto",
      """
        |syntax = "proto3";
        |
        |package lib;
        |
        |message Range {
        |  double min = 1;
        |  double max = 2;
        |}
      """.trimMargin()
    )
    val schema = loadAndLinkSchema()

    // Assert "EnumValueOptions" type is present
    val enumValueOptionsType = schema.getType("google.protobuf.EnumValueOptions") as MessageType
    assertThat(enumValueOptionsType.extensionField("lib.my_enum_value_options")).isNotNull()

    // Assert option on enum constant is present
    val enumType = schema.getType("lib.A") as EnumType
[...]

PR #3254, opened 2025-01-21
add Gradle, CLI options to load exhaustively

damianw:
This PR adds a flag to the CLI and setting to the Gradle plugin to enable the `loadExhaustively` option.

It's currently not possible to correctly implement certain custom schema handlers without this option. In our case, we are using [the Google API HTTP options](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto), which introduce options that can refer to fields by nested path (e.g. `field1.nestedField.anotherField`). By default, Wire may not link fields referenced by such options.

An ideal solution might be some kind of API to tell Wire about additional types/fields/etc that need to be linked, but I don't know how possible that is. In the meantime, exposing this option unblocks implementing schema handlers that use options in this way.

Another use-case can be seen in #2748.


wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
```diff
@@ -1398,6 +1398,16 @@ class WirePluginTest {
     assertThat(File(outputRoot, "Period.kt")).exists()
   }
 
+  @Test
+  @Ignore("Probable ClassLoader problem which makes the test fails")
```
oldergod, line 1402: What's happening here?

oldergod: You could otherwise skip that and only write a test in WireRunTest or somewhere around that

damianw: I tried to write this test, but there's a classloader issue that is preventing classes defined in the build file from working properly (something throws `NoClassDefFoundError` for `SchemaLoader.Factory`). Multiple other tests are ignored because of this issue:
- [`listener`](https://github.com/square/wire/blob/8e6b7f55adca7208a7928c7131f1980797e1090d/wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt#L252-L254)
- [`customOutput`](https://github.com/square/wire/blob/8e6b7f55adca7208a7928c7131f1980797e1090d/wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt#L925-L927)

This could probably be done with `buildSrc` or an included build, though it would be good to refactor those tests to fix them in the same way. I could try to fix all of them in this PR or could punt it to a later PR that fixes all of them.

> You could otherwise skip that and only write a test in WireRunTest or somewhere around that

Yup, this PR adds a test to `WireRunTest` 👍

oldergod: I missed it thank you.
Let's remove this ignored test then and that's good with me

oldergod (approved):

PR #3260, opened 2025-01-23
[swift] Fix conformance to ProtoMessage for heap allocated protos

dnkoutso:


wire-tests-swift/no-manifest/src/main/swift/AllTypes.swift
```diff
@@ -1911,6 +1911,14 @@ extension AllTypes : Proto2Codable {
 
 }
 
+extension AllTypes : ProtoMessage {
```
dnkoutso, line 1914: This is incorrectly missing.

lickel (approved):

PR #3302, opened 2025-03-04
Avoid eagerly resolving the configuration

staktrace:
Calling task.source.isEmpty does some sort of eager resolution of the configurations that were added to the task source. This in turn can trigger other actions within Gradle and cause weird behaviour (in our case, it caused other plugins to get applied and eventually tripped a mutation guard.

The wire rootset code has a isEmpty flag it keeps which, according to the comments, is there explicitly to avoid resolving the configuration, so we can use that instead.

### DIFF
--- a/wire-gradle-plugin/src/test/projects/lazy-configuration-resolution/gradle.properties
+++ b/wire-gradle-plugin/src/test/projects/lazy-configuration-resolution/gradle.properties
@@ -1,0 +1,2 @@
+org.gradle.configureondemand=true
+VERSION_NAME=5.4.0-SNAPSHOT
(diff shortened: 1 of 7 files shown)


wire-gradle-plugin/src/test/projects/lazy-configuration-resolution/gradle.properties
```diff
@@ -0,0 +1,2 @@
+org.gradle.configureondemand=true
+VERSION_NAME=5.4.0-SNAPSHOT
```
oldergod, line 2: this line might not be necessary?

oldergod (approved):

wire-gradle-plugin/src/test/projects/lazy-configuration-resolution/gradle.properties
staktrace: it's needed for https://github.com/square/wire/blob/cb4cfd74bd764de5f83fb06b374f65727920893d/wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt#L1431-L1432

PR #3306, opened 2025-03-31
Add Wire Binary Compatibility plugin

HLGreer:
This change adds a Kotlin compiler plugin that rewrites Wire-generated callsites to be more resilient to schema changes. This first PR only rewrites uses of the generated constructor.

### Why:
When a new field is added to a proto `Message`, existing usages of the Wire-generated class constructor can cause class conflicts. This is often encountered this when there are duplicate or transitive dependencies pulling in different versions of the same proto, resulting in errors shaped as `java.lang.NoSuchMethod` at runtime. 

When faced with this situation, a frequent recommendation is to refactor the codebase to use the class' `Builder`. Rather than requiring individuals to update their codebases, we can adapt their Kotlin code instead.

### Next:
I will follow up with a second PR that will rewrite the Wire-generated `copy()` method. The copy method currently uses the generated constructor.


wire-binary-compatibility-kotlin-plugin/src/main/resources/META-INF/services/org.jetbrains.kotlin.compiler.plugin.CommandLineProcessor
oldergod, line 1: Do we need this file and the next?

oldergod (approved):

settings.gradle.kts
```diff
@@ -60,7 +60,7 @@ include(":wire-schema-tests")
 include(":wire-swift-generator")
 include(":wire-test-utils")
 include(":wire-tests")
-if (startParameter.projectProperties.get("swift") != "false") {
+if (false && startParameter.projectProperties.get("swift") != "false") {
```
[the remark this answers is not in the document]
HLGreer: Oops indeed! Let me remove that

wire-binary-compatibility-gradle-plugin/src/main/kotlin/com/squareup/wire/binarycompatibility/gradle/WireBinaryCompatibilityPlugin.kt
```diff
@@ -0,0 +1,43 @@
...
+@Suppress("unused") // Created reflectively by Gradle.
+class WireBinaryCompatibilityPlugin : KotlinCompilerPluginSupportPlugin {
+  override fun isApplicable(kotlinCompilation: KotlinCompilation<*>): Boolean = true
+
+  override fun getCompilerPluginId(): String = "com.squareup.wire.binarycompatibility.kotlin"
+
+  override fun getPluginArtifact(): SubpluginArtifact = SubpluginArtifact(
+    groupId = "com.squareup.wire.binarycompatibility",
+    artifactId = "wire-binary-compatibility-kotlin-plugin",
+    version = "2.6.0-SNAPSHOT",
```
staktrace, line 33: hard-coding the version here seems wrong?

wire-binary-compatibility-gradle-plugin/build.gradle.kts
```diff
@@ -0,0 +1,36 @@
...
+  `kotlin-dsl`
+  kotlin("jvm")
+  id("org.jetbrains.dokka")
+  id("com.vanniktech.maven.publish.base")
+}
+
+dependencies {
+  implementation(kotlin("gradle-plugin-api"))
+  implementation(project(":wire-binary-compatibility-kotlin-plugin"))
+  implementation(libs.kotlin.gradlePlugin)
```
staktrace, line 16: I suspect this dependency on the kotlin gradle plugin should be `compileOnly` to avoid polluting downstream consumers classpath. @autonomousapps do you know? Perhaps the `gradle-plugin-api` dependency as well?

wire-binary-compatibility-kotlin-plugin/src/main/kotlin/com/squareup/wire/binarycompatibility/kotlin/BurstCompilerPluginRegistrar.kt
staktrace, line 1: Filename is copypasta, should probably be updated

wire-binary-compatibility-kotlin-plugin/src/main/kotlin/com/squareup/wire/binarycompatibility/kotlin/BurstCompilerPluginRegistrar.kt
```diff
@@ -0,0 +1,42 @@
+/*
```
HLGreer, line 1: One more name fixup needed here

wire-binary-compatibility-gradle-plugin/build.gradle.kts
oldergod: You're right

wire-binary-compatibility-kotlin-plugin/src/main/kotlin/com/squareup/wire/binarycompatibility/kotlin/BurstCompilerPluginRegistrar.kt
HLGreer: Fixed!

wire-binary-compatibility-gradle-plugin/src/main/kotlin/com/squareup/wire/binarycompatibility/gradle/WireBinaryCompatibilityPlugin.kt
HLGreer: I've moved these maven coordinates to a simple BuildConfig object, and fixed the version (another copy-paste miss) -- this is the only place these values are referenced, the intent is to just keep this lightweight!

wire-binary-compatibility-gradle-plugin/README.md
```diff
@@ -0,0 +1,43 @@
...
+val newDinosaur = Dinosaur(
+  name = "triceratops",
+  avg_length_meters = 9.0,
+)
+```
+
+When a new field is added to the Dinosaur schema, if there are competing versions of the compiled class and the new
+version is resolved at runtime, the usage above may encounter an error:
+```
+java.lang.NoSuchMethodError: 'void com.squareup.dinosaurs.Dinosaur<init>(java.lang.String, java.lang.Double,
```
swankjesse, line 32: YASS!

wire-binary-compatibility-gradle-plugin/src/main/kotlin/com/squareup/wire/binarycompatibility/gradle/BuildConfig.kt
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire.binarycompatibility.gradle
+
+object BuildConfig {
```
swankjesse, line 18: Usually this file can be generated from gradle properties.

swankjesse (approved):
So rad

wire-binary-compatibility-gradle-plugin/src/main/kotlin/com/squareup/wire/binarycompatibility/gradle/BuildConfig.kt
HLGreer: I was looking at some prior art! I figured that is a good follow up unless someone felt strongly to have it generated here now

JakeWharton: It must be generated, at least the version component, or it will not work properly in the destination project. It's only a few lines in the Gradle build and you can copy it from Burst or Zipline probably without any changes.

wire-binary-compatibility-gradle-plugin/build.gradle.kts
staktrace: I think you got the change slightly backwards - the `libs.kotlin.gradlePlugin` should be `compileOnly` and the `project("...")` one should be `implementation

staktrace: with that change made I have no other concerns, thanks for doing this!

autonomousapps: Yes, I agree with what Kats has said. While this plugin only makes sense in the context of a Kotlin project, we don't want to force a Kotlin version update on anyone if we can avoid it, so we should use `compileOnly(libs.kotlin.gradlePlugin)`.

wire-binary-compatibility-gradle-plugin/build.gradle.kts
```diff
@@ -0,0 +1,36 @@
...
+      description = "Rewrites Wire callsites to be resilient to API changes"
+      implementationClass = "com.squareup.wire.binarycompatibility.gradle.WireBinaryCompatibility"
+    }
+  }
+}
+
+configure<MavenPublishBaseExtension> {
+  configure(
+    GradlePlugin(
+      javadocJar = JavadocJar.Empty()
```
autonomousapps, line 33: Out of curiosity, why publish an empty javadoc jar?

wire-binary-compatibility-gradle-plugin/src/main/kotlin/com/squareup/wire/binarycompatibility/gradle/BuildConfig.kt
autonomousapps: I think maybe this PR also demonstrates this: https://github.com/square/gradle-dependencies-sorter/pull/123/files

wire-binary-compatibility-gradle-plugin/build.gradle.kts
HLGreer: Coming back to this -- thanks for the pointers @staktrace / @autonomousapps! I've swapped around the change.

The `"gradle-plugin-api"` dependency couldn't be `compileOnly`, that had to be left as `implementation` (tests would fail otherwise)

staktrace: I suspect that you could do it as `compileOnly` if you also add it to the `testImplementation` or `testRuntimeOnly` configuration. I'm still a bit concerned that having this dependency as `implementation` will infect downstream consumers. Looking at the pom file I think it transitively brings in the kotlin-gradle-plugins-bom which will cause downstream consumers to get upgraded to the version of kotlin being used by wire.

Issue #3312, opened 2025-05-14
WireTask cache uses non-relocatable inputs


staktrace:
On one machine, running `gradle -Dorg.gradle.caching.debug=true out:shared-protos:generateMainProtos` produced output that included this:

```
Appending input value fingerprint for 'protoInput$wire_gradle_plugin' to build cache key: f42ccfbdb1cbc2c2563dc5692c65b47b	
```

On a different machine with the exact same code/version/command, it produced this:
```
Appending input value fingerprint for 'protoInput$wire_gradle_plugin' to build cache key: da595be19d6f097db8e557b3f70d063a	
```

This indicates that the WireTask is misconfiguring the cache inputs, probably by treating a path as absolute instead of relative. This makes the gradle build cache entry for the task non-relocatable across machines and hurts build performance.

staktrace:
I have a reproducer for this. The main problem is that although the code at https://github.com/square/wire/blob/b319fc53cffb6b1fa4dae270cc818e2c857414f2/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt#L39-L42 is relativiz-ing paths inside the project, it's not doing the same for paths inside the gradle user home dir, which is where `libs.whatever` references resolve to. But really the better solution seems to be to not resolve the configuration to files and then have those files individually added as the task input. I'm exploring fixes.

PR #3313, opened 2025-05-16
Update InputLocation to be more cache-friendly

staktrace:
Instead of storing individual resolved paths, we should use gradle's
`@Classpath` annotation to indicate that we've got a set of input paths
that are a classpath. Importantly, this will normalize files across
varying gradle user home dirs, so jars that are resolved into the gradle
user home dir will not have their full absolute path baked into the
task input.

Fixes https://github.com/square/wire/issues/3312

### DIFF
--- a/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt
+++ b/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt
@@ -20,6 +20,8 @@ import com.squareup.wire.schema.Location
 import java.io.EOFException
 import java.io.File
 import java.io.RandomAccessFile
+import kotlin.io.relativeTo
+import kotlin.io.startsWith
 import org.gradle.api.internal.file.FileOperations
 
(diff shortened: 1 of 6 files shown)


wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireInput.kt
```diff
@@ -23,24 +23,14 @@ import java.io.RandomAccessFile
 import org.gradle.api.internal.file.FileOperations
 
 internal val List<ProtoRootSet>.inputLocations: List<InputLocation>
-  get() = flatMap { rootSet ->
-    rootSet.roots.map { file ->
-      rootSet.inputLocation(file)
-    }
-  }
+  get() = map { rootSet -> rootSet.inputLocation() }
```
staktrace, line 26: This avoids creating an InputLocation per-file, but intead creates an InputLocation per-ProtoRootSet, with all the files in the ProtoRootSet treated as a "classpath"

oldergod (approved):
Amazing, thank you

PR #3314, opened 2025-05-21
Fix crash when inlined option map values have multiple fields

tejasna:


oldergod (approved):

oldergod:
@tejasna Thanks for the fix!

tejasna:
@oldergod  Is there a release scheduled in the near future? We’re looking to integrate this change into our workflow 🙏

oldergod:
@tejasna `5.3.2` coming soon

PR #3342, opened 2025-06-26
Simplify some publishing code

JakeWharton:
This works around a lifecycle problem where the publishing code was attempting to modify a Gradle property after it was frozen. Since we were setting values which were the default already, we can just remove the call.


oldergod (approved):
🙏

PR #3350, opened 2025-07-23
[swift] Properly disambiguate OneOf enum if it has the same name as enclosing type.

dnkoutso:
If a `oneof` field is named the same as one of the chain of messages it belongs to and it also declares a message type as an one its options, then it can cause ambiguous Swift compile errors when it comes to scoping due to the way code is emitted.

This change fixes this edge case by scanning the chain of message types and names if they completely match the name of the `oneof` declaration and appends `_OneOf` to the type to help disambiguate.

A test was expanded that fails without this change and passes with it and also adds more edge case scenarios to verify it works.

### DIFF
--- a/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
+++ b/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
@@ -626,7 +626,7 @@ class SwiftGenerator private constructor(
             addStatement("var %N: %T = %L", field.safeName, localType, initializer)
           }
           type.oneOfs.forEach { oneOf ->
-            val enumName = oneOfEnumNames.getValue(oneOf)
+            val enumName = oneOfSafeDeclaredTypeName(oneOf, type, oneOfEnumNames)
             addStatement("var %N: %T = nil", oneOf.name, enumName.makeOptional())
           }
           if (type.declaredFieldsAndOneOfFields.isNotEmpty()) {
@@ -1063,7 +1063,7 @@ class SwiftGenerator private constructor(
     }
     if (includeOneOfs) {
       type.oneOfs.forEach { oneOf ->
-        val enumName = oneOfEnumNames.getValue(oneOf).makeOptional()
+        val enumName = oneOfSafeDeclaredTypeName(oneOf, type, oneOfEnumNames)
         addParameter(
           ParameterSpec.builder(oneOf.name, enumName)
             .defaultValue("nil")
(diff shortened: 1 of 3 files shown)


wire-tests/src/commonTest/proto/kotlin/swift_edge_cases.proto
```diff
@@ -37,6 +37,11 @@ message Screen {
   oneof screen {
     string oneof_string = 1;
     int32 oneof_int32 = 2;
+    SubMessage oneof_sub_message = 3;
```
dnkoutso, line 40: Compiling generated Swift code without the change in the SwiftGenerator above fails. 

Also verified in our internal repo this fixes this edge case.

dnkoutso: Also I should note there are other protos that do not have the same name as the enclosing message name that from this change do not use the `_OneOf` suffix. This suffix is only used in this edge case.

wire-tests/src/commonTest/proto/kotlin/swift_edge_cases.proto
```diff
@@ -35,8 +35,30 @@ message SwiftEdgeCases {
 
 message Screen {
   oneof screen {
-    string oneof_string = 1;
-    int32 oneof_int32 = 2;
+    string screen_oneof_string = 1;
+    int32 screen_oneof_int32 = 2;
+    SubMessage screen_oneof_sub_message = 3;
```
dnkoutso, line 40: adding this to this edge case proto starts failing compilation

wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
```diff
@@ -1649,6 +1649,27 @@ class SwiftGenerator private constructor(
...
+    var currentType: ProtoType? = type.type
+
+    while (currentType != null) {
+      val simpleName = currentType.simpleName
+      if (oneOf.name.equals(simpleName, ignoreCase = true)) {
+        val oneOfTypeName = oneOfEnumNames.getValue(oneOf)
+        return oneOfTypeName.peerType("${oneOfTypeName.simpleName}_OneOf")
+      }
+
+      val enclosingName = currentType.enclosingTypeOrPackage ?: break
```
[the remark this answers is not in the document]
dnkoutso: I think its the same, the `continue will run the loop once more but `enclosingName` will be `null` and therefore `currentType` will be `null` and fall to `return oneOfEnumNames.getValue(oneOf)` without the `_OneOf` suffix.

PR #3356, opened 2025-08-05
[swift] Add more well known types in Swift runtime

dnkoutso:
We just map those types to the built in Swift native types.


jszumski (approved):

wire-tests-swift/no-manifest/src/main/swift/AllSwiftWrappers.swift
```diff
@@ -0,0 +1,609 @@
...
+        var double_value: Double? = nil
+        var float_value: Float? = nil
+        var int64_value: Int64? = nil
+        var uint64_value: UInt64? = nil
+        var int32_value: Int32? = nil
+        var uint32_value: UInt32? = nil
+        var bool_value: Bool? = nil
+        var string_value: String? = nil
+        var bytes_value: Foundation.Data? = nil
+        var rep_double_value: [Double] = []
```
dnkoutso, line 388: hmmm @efirestone I suspect this needs to be `[Double?]` if that is what these Google wrapper types were made for right?

PR #3359, opened 2025-08-06
[swift] Improve performance of topological sort algorithm for partitions

dnkoutso:
On larger scale repos that involve many partitions and hundreds of transitive deps, the topological sort of the graph can be quite slow. 

In our case, an internal proto took ~2 minutes to run the Wire compiler (not including the Swift compilation time). This change updates to use the [Kahn topological sort algorithm](https://www.geeksforgeeks.org/dsa/topological-sorting-indegree-based-solution/) for faster speed. We saw the same proto take 10s-15s to run the compiler after this change.

While faster, the algorithm does not produce a stable order so afterwards we sort by depth as the rest of the code was processing dependencies in a bottom-up order.

Added new tests and existing tests should further validate this.

### DIFF
--- a/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/DirectedAcyclicGraph.kt
+++ b/wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/DirectedAcyclicGraph.kt
@@ -45,20 +45,40 @@ internal class DirectedAcyclicGraph<N>(
   }
 
   fun topologicalOrder(): List<N> {
-    val seen = LinkedHashSet<N>() // Insertion order is important to produce the final list!
-    val queue = ArrayDeque<N>().apply { addAll(seeds) }
-    while (queue.isNotEmpty()) {
-      val currentName = queue.removeFirst()
-      val currentDependencies = edges(currentName).toList()
-      if (seen.containsAll(currentDependencies)) {
-        seen += currentName
-        queue += incomingEdges(currentName)
-      } else {
-        // All dependencies have not been seen so move to the back of the line to try again later.
-        queue += currentName
+    val incomingEdges = mutableMapOf<N, Int>()
+    for (vertex in nodes) {
+      if (vertex !in incomingEdges) {
+        incomingEdges[vertex] = 0
+      }
+      for (edge in edges(vertex)) {
+        incomingEdges[edge] = (incomingEdges[edge] ?: 0) + 1
       }
     }
-    return seen.toList()
+
+    val queue = ArrayDeque<N>()
+    for ((vertex, edges) in incomingEdges) {
+      if (edges == 0) queue += vertex
+    }
+
+    val result = mutableListOf<N>()
+
+    while (queue.isNotEmpty()) {
+      val vertex = queue.removeFirst()
+      result += vertex
+
+      for (edge in edges(vertex)) {
+        incomingEdges[edge] = (incomingEdges[edge] ?: 0) - 1
+        if (incomingEdges[edge] == 0) {
+          queue += edge
+        }
+      }
+    }
+
+    check(result.size == incomingEdges.size) {
+      "Graph contains a cycle, topological sort not possible!"
+    }
+
+    return result
   }
 
   fun transitiveNodes(node: N): Set<N> {
--- a/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/DirectedAcyclicGraphTest.kt
+++ b/wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/DirectedAcyclicGraphTest.kt
@@ -1,0 +1,116 @@
+/*
+ * Copyright (C) 2025 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire.schema
+
+import assertk.assertThat
+import assertk.assertions.containsExactly
+import assertk.assertions.hasMessage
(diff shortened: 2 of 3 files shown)


wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/DirectedAcyclicGraph.kt
```diff
@@ -15,6 +15,8 @@
  */
 package com.squareup.wire.schema
 
+typealias Queue<T> = ArrayDeque<T>
```
JakeWharton, line 18: This puts the name into the public API which seems unnecessary. It's only used in one place, why not reference the type directly?

dnkoutso: will remove! It was just testing some stuff in the beginning and this was left over.

dnkoutso: fixed

wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/DirectedAcyclicGraphTest.kt
```diff
@@ -0,0 +1,121 @@
...
+import kotlin.test.fail
+
+class DirectedAcyclicGraphTest {
+
+  @Test
+  fun singleNode() {
+    val nodes = listOf("A")
+    val graph = DirectedAcyclicGraph(nodes) { emptyList() }
+    val order = graph.topologicalOrder()
+    assertThat(listOf("A")).isEqualTo(order)
```
JakeWharton, line 32:
```suggestion
    assertThat(order).isEqualTo(listOf("A")
```
You can also do
```suggestion
    assertThat(order).containsExactly(listOf("A")
```
but have to import it

wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/DirectedAcyclicGraphTest.kt
```diff
@@ -0,0 +1,121 @@
...
+      "C" to listOf("D"),
+      "D" to emptyList(),
+    )
+    val graph = DirectedAcyclicGraph(nodes) { edges[it].orEmpty() }
+    val order = graph.topologicalOrder()
+    assertThat(order).isEqualTo(listOf("A", "B", "C", "D"))
+    assertThat(order.indexOf("A")).isLessThan(order.indexOf("B"))
+    assertThat(order.indexOf("A")).isLessThan(order.indexOf("C"))
+    assertThat(order.indexOf("B")).isLessThan(order.indexOf("D"))
+    assertThat(order.indexOf("C")).isLessThan(order.indexOf("D"))
```
JakeWharton, line 83:
```suggestion
```
These are redundant checks given the prior assertion

wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/DirectedAcyclicGraphTest.kt
```diff
@@ -0,0 +1,121 @@
...
+    val edges = mapOf(
+      "A" to listOf("C"),
+      "B" to listOf("C"),
+      "C" to emptyList(),
+    )
+    val graph = DirectedAcyclicGraph(nodes) { edges[it].orEmpty() }
+    val order = graph.topologicalOrder()
+    assertThat(order).isEqualTo(listOf("A", "B", "C"))
+    assertThat(order.indexOf("A")).isLessThan(order.indexOf("C"))
+    assertThat(order.indexOf("B")).isLessThan(order.indexOf("C"))
```
JakeWharton, line 100:
```suggestion
```

wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/ManifestPartitionTest.kt
```diff
@@ -103,7 +103,7 @@ class ManifestPartitionTest {
           .build(),
       ),
       "feature" to WireRun.Module(
-        dependencies = setOf("common"),
+        // dependencies = setOf("common"),
```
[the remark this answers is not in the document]
dnkoutso: fixed

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/DirectedAcyclicGraph.kt
```diff
@@ -45,20 +45,40 @@ internal class DirectedAcyclicGraph<N>(
...
+        if (incomingEdges[edge] == 0) {
+          queue += edge
+        }
       }
     }
-    return seen.toList()
+
+    if (result.size != incomingEdges.size) {
+      error("Graph contains a cycle, topological sort not possible!")
+    }
```
JakeWharton, line 79: Just a style nit
```suggestion
    check(result.size == incomingEdges.size) {
      "Graph contains a cycle, topological sort not possible!"
    }
```

JakeWharton (approved):

wire-schema/src/jvmTest/kotlin/com/squareup/wire/schema/DirectedAcyclicGraphTest.kt
dnkoutso: done

wire-schema/src/commonMain/kotlin/com/squareup/wire/schema/DirectedAcyclicGraph.kt
```diff
@@ -45,20 +45,40 @@ internal class DirectedAcyclicGraph<N>(
   }
 
   fun topologicalOrder(): List<N> {
-    val seen = LinkedHashSet<N>() // Insertion order is important to produce the final list!
-    val queue = ArrayDeque<N>().apply { addAll(seeds) }
+    val incomingEdges = mutableMapOf<N, Int>()
```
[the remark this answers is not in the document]
romainguy: I wasn't thinking of just the boxing of the ints, but also the allocations on insertion/iteration/etc. No idea if it's worth it since I don't know anything about Wire :)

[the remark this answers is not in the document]
JakeWharton: This code usually runs on the JVM during codegen, so it's not in the critical path of serialization or anything.

PR #3360, opened 2025-08-07
[swift] Fix edge case for oneof field named `self`

dnkoutso:
We have a rather badly named `message Self` in one of our protos that is throwing off again the Swift compiler.

SwiftPoet does the right thing to escape it with backticks however other parts of the generated code start failing.

For Wire, we rename \`Self\` with Self_


wire-tests/src/commonTest/proto/kotlin/swift_edge_cases.proto
```diff
@@ -38,12 +38,15 @@ message Screen {
     string screen_oneof_string = 1;
     int32 screen_oneof_int32 = 2;
     SubMessage screen_oneof_sub_message = 3;
+    Self self = 4;
```
dnkoutso, line 41: generated code fails to compile without the above change.

jszumski (approved):

PR #3365, opened 2025-08-21
Parse trailers even when an error occurs

ahmedre:
It is possible for gRPC error responses to be empty and for servers to send the error information within trailers without also sending them in headers. In these cases, Wire will propagate an exception due to no response body and will not attempt to read trailers, relying only on headers instead.

This patch allows Wire to attempt to parse the trailers for non-streaming requests instead of only relying on headers.

### DIFF
--- a/wire-grpc-client/src/commonMain/kotlin/com/squareup/wire/internal/GrpcMessageSource.kt
+++ b/wire-grpc-client/src/commonMain/kotlin/com/squareup/wire/internal/GrpcMessageSource.kt
@@ -61,6 +61,8 @@ internal class GrpcMessageSource<T : Any>(
     }
   }
 
+  fun isEmptyBody(): Boolean = source.exhausted()
+
   fun readExactlyOneAndClose(): T {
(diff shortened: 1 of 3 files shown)


wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/grpc.kt
```diff
@@ -118,7 +118,7 @@ internal fun <R : Any> SendChannel<R>.readFromResponseBodyCallback(
               }
               exception = response.grpcResponseToException()
             } catch (e: IOException) {
-              exception = response.grpcResponseToException(e)
+              exception = response.grpcResponseToException(false, e)
```
ahmedre, line 121: intentionally only did this for unary calls because there's a specific test to disallow this for streaming (`duplexSuspend_responseBodyThrows`) - i suspect this is because when something goes wrong, the channel is explicitly closed, therefore making it not possible to read the trailers.

PR #3372, opened 2025-09-03
Clean build set up for wire projects

liutikas:
To help upgrade to Gradle 9.0.0 clean up a variety of issues in wire build setup.

- Migrate to com.gradleup.shadow and pick 8.3.9 that is compatible with Gradle 9.0.0
- Migrate away from using kotlinOptions DSL that has been deprecated
- Migrate away from using tasks.creating to tasks.registering
- Remove usage of metaInfo in Kotlin2JsCompile has the option has no effect since KGP 2.0 and has been removed in https://github.com/JetBrains/kotlin/commit/22436db


oldergod:
I rebased and pushed

oldergod (approved):

PR #3424, opened 2025-10-10
Gradle / Github Actions Optimizations

jgavris:
- Enable Gradle build cache for the local project (and Github Actions workflows).
- We should see some nice boosts on the actions runners on subsequent rebases.
- Gradle [configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html) in a future PR after I take a look at the classes that are incompatible with configuration cache.

For reference:
image

Gradle Build Scans / Benchmarks (not considering Github Actions caching, just local builds):

```sh
./gradlew clean check
```

- No Cache - ~10m

https://scans.gradle.com/s/ilkm7x7xozjgg

- Cache - ~3m

https://scans.gradle.com/s/3v5ydfa3xfxn4

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,10 @@ jobs:
           distribution: 'zulu'
           java-version-file: .github/workflows/.java-version
 
+      - uses: gradle/actions/setup-gradle@v5
+        with:
+          cache-read-only: false
+
       - name: Check test files
         run: |
           ./gradlew generateTests -Pswift=false --stacktrace --warning-mode all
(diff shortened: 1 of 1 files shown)


jgavris:
@JakeWharton @dnkoutso @martinbonnin I updated the description with some Gradle build scans, for `./gradlew clean check`, it drops from 10m to 3m. I think also need to add the `setup-gradle` step to all the other workflow steps to use the cache though.

jgavris:
@oldergod I just ran a 'full clean' as part of testing the build, the `./gradlew cleanGeneratedTests` task (has no callers as far as I can tell). Those files did not come back with `./gradlew generateTests`, so they seem to just be unused / lint.

jgavris:
Looks like the cache was populated for the first time on the rest of the jobs, if you want to re-run the workflow it should pull from them this time.

PR #3498, opened 2026-01-14
Small changes for okio consistency

cbrunsdon:
This picks off a few quick places where okio was not being used in favor of platform-specific calls which okio wraps which can be taken advantage of.

Wire uses/exposes okio pretty frequently so its probably safe to assume these were just oversights or legacy reasons nobody has cared to change themselves.


oldergod (approved):
Nice, thank you

PR #3502, opened 2026-01-21
Use kotlin.jvm annotations if jvmInterop

cbrunsdon:
There is a bit of background around the `internal.Jvm*` annotations here for background:

https://github.com/square/wire/pull/2504

They have two (minor?) issues:
- There are times `@Jvm*` are treated special, and certain setups of module dependencies and project layouts can cause IDE auto-completion, for instance, to not interpret the `internal.Jvm` correctly, redlining your perfectly-reasonable and runnable code
- They're in `internal` but they're public aliases which isn't amazing to ship to any projects including wire-runtime and other careless programmers (totally not me though...) might accidentally auto-complete those in instead of their expected `@Jvm` annotations. 

I don't actually quite follow the how/when/why the `internal.Jvm*` was required. It might be a historical issue that generating into /main wouldn't have those provided? That no longer seems to be the case as setting 
`jvmAnnotationPackage` in this commit to kotlin.jvm seems to run the suites just fine?

This change sets the annotations to reference the `kotlin.jvm` annotations if jvmInterop is enabled, which we should expect to be provided by anyone generating while explicitly asking for jvm compatibility. 

Either way, though this adds complexity to the generation process, I personally believe its worth it as a step to phase these out.

### DIFF
--- a/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
+++ b/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
@@ -141,9 +141,11 @@ class KotlinGenerator private constructor(
   private val emitProtoReader32: Boolean,
   private val mutableTypes: Boolean,
   private val explicitStreamingCalls: Boolean,
-  private val jvmAnnotationPackage: String,
 ) {
   private val nameAllocatorStore = mutableMapOf<Type, NameAllocator>()
+
+  private val jvmAnnotationPackage: String = if (javaInterOp) "kotlin.jvm" else "com.squareup.wire.internal"
+  private val useJavaInterop: Boolean = javaInterOp || buildersOnly
 
   @Suppress("RecursivePropertyAccessor")
   private val ProtoType.typeName: TypeName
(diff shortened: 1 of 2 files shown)


cbrunsdon:
Okay looking at that failure, sorry. I have `./gradlew check` up to wasm browser tests but have been selectively running things to verify.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -2638,7 +2629,7 @@ class KotlinGenerator private constructor(
       .build()
 
     return PropertySpec.builder(adapterName, adapterType)
-      .jvmField()
+      .jvmFieldIf(true, jvmAnnotationPackage)
```
oldergod: If we know it's true, let's inline the `addAnnotation` call, same everywhere we `true` it

oldergod (approved):

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
cbrunsdon: in lined them, thanks

Issue #3371, opened 2025-09-03
Wire uses applicationVariants and libraryVariants that is getting removed in AGP 9.0.0 stable


liutikas:
Repro project:

https://github.com/liutikas/wire-agp-9.0.0-legacy-variants-repro

Steps:

./gradlew assebleDebug

## Expected

Success

## Actual

```
* What went wrong:
A problem occurred configuring project ':app'.
> Failed to notify project evaluation listener.
   > API 'applicationVariants' is obsolete.
     It will be removed in version 10.0 of the Android Gradle plugin.
     The legacy variant API is disabled by default in AGP 9.0, but can be re-enabled by adding 
         android.enableLegacyVariantApi=true
     to this project's gradle.properties file.
     For more information, see https://developer.android.com/studio/releases/gradle-plugin-api-updates.
     
     REASON: It is currently called from the following trace:
     java.lang.Thread.getStackTrace(Thread.java:2450)
     com.android.build.gradle.internal.errors.DeprecationReporterImpl.reportDeprecatedApi(DeprecationReporterImpl.kt:71)
     com.android.build.gradle.BaseExtension.recordOldVariantApiUsage(BaseExtension.kt:157)
     com.android.build.gradle.AbstractAppExtension.getApplicationVariants(AbstractAppExtension.kt:63)
     com.android.build.gradle.internal.services.BuiltInKotlinServicesKt.toBaseVariant(BuiltInKotlinServices.kt:341)
     com.android.build.gradle.internal.TaskManager.maybeCreateKotlinTasks(TaskManager.kt:997)
     com.android.build.gradle.internal.VariantTaskManager.createPostApiTasks(VariantTaskManager.kt:149)
     com.android.build.gradle.internal.plugins.BasePlugin.createAndroidTasks(BasePlugin.kt:819)
     com.android.build.gradle.internal.plugins.BasePlugin$createTasks$2$1.call(BasePlugin.kt:587)
     com.android.build.gradle.internal.profile.NoOpAnalyticsConfiguratorService.recordBlock(NoOpAnalyticsConfiguratorService.kt:57)
     com.android.build.gradle.internal.plugins.BasePlugin$createTasks$2.accept(BasePlugin.kt:582)
     com.android.build.gradle.internal.plugins.BasePlugin$createTasks$2.accept(BasePlugin.kt:579)
     com.android.build.gradle.internal.crash.CrashReporting$afterEvaluate$1.execute(crash_reporting.kt:37)
[...]

PR #3503, opened 2026-01-22
Refactor to Factory pattern and use modern Android Variant API

omarismail94:
This refactor modernizes the Wire Gradle plugin by decoupling task creation from the core plugin logic and adopting the modern Android Gradle Plugin (AGP) APIs.

Key changes:
- Introduced WireTaskFactory and a sealed Source hierarchy (Android, Kotlin Multiplatform, JVM) to handle task registration across different project types, replacing the legacy WireSourceDirectorySet.
- Migrated Android task generation from the deprecated BaseExtension to the modern AndroidComponentsExtension using the onVariants and variant.sources APIs.

TESTED: ran ./gradlew clean check
FIXES: Issue #3371

### DIFF
--- a/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/kotlin/SourceRoots.kt
+++ b/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/kotlin/SourceRoots.kt
@@ -19,6 +19,10 @@ import com.android.build.api.variant.AndroidComponentsExtension
 import com.android.build.api.variant.Variant
 import com.squareup.wire.gradle.WireTask
 import com.squareup.wire.gradle.internal.targetDefaultOutputPath
+import com.squareup.wire.schema.CustomTarget
+import com.squareup.wire.schema.JavaTarget
+import com.squareup.wire.schema.KotlinTarget
+import com.squareup.wire.schema.ProtoTarget
 import com.squareup.wire.schema.Target
 import java.io.File
 import org.gradle.api.Project
@@ -48,7 +52,6 @@ internal fun forEachWireSource(
           name = variant.name,
           sourceSets = sourceSetNames.distinct(),
           variant = variant,
-          kotlin = hasKotlin,
         )
         sourceHandler(source)
       }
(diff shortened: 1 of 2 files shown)


wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/kotlin/SourceRoots.kt
```diff
@@ -15,207 +15,351 @@
...
+    generatedSourcesDirectories.toList().forEachIndexed { i, _ ->
+      variant.sources.java?.addGeneratedSourceDirectory(wireTask) { it.outputDirectories[i] }
+      if (kotlin) {
+        variant.sources.kotlin?.addGeneratedSourceDirectory(wireTask) { it.outputDirectories[i] }
+      }
     }
   }
 }
+
+private val SOURCE_FUNCTION by lazy(LazyThreadSafetyMode.NONE) {
```
liutikas: Comment was lost in the refactor. This is needed to support KGP 1.6 and 1.7, given that AGP upgrade now requires a much newer AGP that has newer KGP can we drop this and stop using reflection?

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/kotlin/SourceRoots.kt
```diff
@@ -15,207 +15,351 @@
...
+  }
+  val generatedSourcesDirectories: Set<File> =
+    targets
+      // Emitted `.proto` files have a special treatment. Their root should be a resource, not a
+      // source. We exclude the `ProtoTarget` and we'll add its output to the resources below.
+      .filterNot { it is ProtoTarget }
+      .map { target -> project.file(target.outDirectory) }
+      .toSet()
+
+  val protoTarget = targets.filterIsInstance<ProtoTarget>().firstOrNull()
```
liutikas: Should we fail if there is more than one ProtoTarget? We seem to silently ignore the rest.

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/kotlin/SourceRoots.kt
```diff
@@ -15,207 +15,351 @@
...
+
+  if (hasJavaOutput) {
+    project.tasks
+      .withType(JavaCompile::class.java)
```
liutikas: Does this run for Android projects? If so, we should not be wiring to JavaCompile tasks directly, instead we should use addGeneratedSources API.

If this doesn't run on Android projects lets name the method to make that clear.

PR #3505, opened 2026-01-26
Ensure compatibility with older Kotlin consumers

omarismail94:
Lowering the bundled stdlib version via coreLibrariesVersion. This prevents forcing a stdlib upgrade on our users. We:
* Set Kotlin API and language versions to 2.0.
* Explicitly set `coreLibrariesVersion` to 2.0.21 to maximize compatibility.
* Force JS and Wasm stdlib dependencies to match the toolchain version as work-around for KT-71032

TESTED: ./gradlew clean check . I also published to local repository and checked the .module files for the artifacts and the kotlin-stdlib was correctly set to 2.0.21

### DIFF
--- a/build-support/src/main/kotlin/com/squareup/wire/buildsupport/WireBuildPlugin.kt
+++ b/build-support/src/main/kotlin/com/squareup/wire/buildsupport/WireBuildPlugin.kt
@@ -24,9 +24,12 @@ import java.io.File
 import kotlinx.validation.ApiValidationExtension
 import kotlinx.validation.ExperimentalBCVApi
 import org.gradle.accessors.dm.LibrariesForLibs
+import org.gradle.api.GradleException
 import org.gradle.api.JavaVersion
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.artifacts.VersionCatalog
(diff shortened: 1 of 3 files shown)


oldergod (approved):

oldergod:
Coming back to this thread because the `implementation` scope on KGP ended up breaking downstream consumers in 7.0.0-alpha05.
The plugin POM now carries a *runtime* dependency on `kotlin-gradle-plugin` (2.3.21 as of alpha05).
On the reasons given above for moving away from `compileOnly` — I think all three are about `kotlin-stdlib`, not KGP, and none of them actually need KGP at `implementation` scope:

  1. *Consumers missing stdlib / too-old stdlib*: the plugin's own `kotlin-stdlib` dependency stays in the POM at compile scope (pinned by `coreLibrariesVersion` to 2.0.21) regardless of KGP's scope. `compileOnly` on KGP doesn't remove it.
  2. *Minimum stdlib bound*: same — that bound comes from the stdlib dependency, which is unaffected.
  3. *stdlib appearing in 6.0.0-alpha01 despite `compileOnly`*: expected, and correct — that's the plugin's own code dependency, not KGP's.

  I'm putting up a PR restoring `compileOnly` here https://github.com/square/wire/pull/3661

PR #3509, opened 2026-02-03
Fix large field number parsing by using unsigned right shift

xi0yu:
This fixes an issue where Wire fails to parse protobuf data when encountering very large field numbers (for example, 290,848,974).
The failure is caused by using a signed right shift (shr), which introduces sign extension for large values, leading to incorrect tag extraction.
Changed shr to ushr in nextTag() and skipGroup() methods to ensure proper handling of large field numbers.

Only two lines modified:
- Line 184: tag = tagAndFieldEncoding shr TAG_FIELD_ENCODING_BITS -> tag = (tagAndFieldEncoding ushr TAG_FIELD_ENCODING_BITS)
- Line 250: val tag = tagAndFieldEncoding shr TAG_FIELD_ENCODING_BITS -> val tag = (tagAndFieldEncoding ushr TAG_FIELD_ENCODING_BITS)


oldergod:
Thanks for the PR. Is writing a test for this gonna be difficult?

wire-tests/jvm-kotlin-interop/src/test/kotlin/com/squareup/wire/LargeFieldNumberTest.kt
```diff
@@ -0,0 +1,46 @@
...
+import kotlin.test.Test
+
+class LargeFieldNumberTest {
+
+  private val adapter = createRuntimeMessageAdapter(
+    LargeFieldMessage::class.java,
+    "square.github.io/wire/unknown",
+    Syntax.PROTO_2,
+    LargeFieldNumberTest::class.java.classLoader,
+  )
```
oldergod: You could access the generated adapter as well, I think?


```suggestion
  private val adapter = LargeFieldMessage.ADAPTER
```

PR #3514, opened 2026-02-10
[swift]: fix a few edge case memory bugs

jamieQ:
this change addresses two issues identified within the swift runtime:

1. an potential memory leak if unknown fields are added but message parsing does not terminate normally (i.e. an error is thrown before `endMessage()` completes)
2. a memory access violation if the nested message recursion limit is exceeded


dnkoutso (approved):

efirestone (approved):
Great finds!

jamieQ:
@codex review

Issue #2452, opened 2023-05-05
Kotlin - make jsonName casing configurable


DougSig:
Currently, the Kotlin generator sets `WireField.jsonName` to the camel case version of the field name. If I want to keep the field names as snake case, I have to do this for every field:
```
message ExampleMessage {
  ExampleField example_field = 1 [json_name = "example_field"];
}
```

oldergod:
This follows the spec so it's unlikely we change anything in the behavior here.
```
  // JSON name of this field. The value is set by protocol compiler. If the
  // user has set a "json_name" option on this field, that option's value
  // will be used. Otherwise, it's deduced from the field's name by converting
  // it to camelCase.
  optional string json_name = 10;
```

https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/descriptor.proto#L255-L259

jdbp-classpass:
I think `WireJsonTest` would be the place for the new tests? That's where the existing `writeIdentityValues` option is exercised.

PR #3521, opened 2026-02-24
Add preservingProtoFieldNames option

jdbp-classpass:
As discussed on https://github.com/square/wire/issues/2452, the upstream Google protobuf libraries (e.g. `JsonFormat`) can be configured to serialize with the unmodified field names from the proto files. Wire also supports this capability in Swift (https://github.com/square/wire/blob/ca1e09e297f96c52b4615f75e95780be0b10eeb7/wire-runtime-swift/src/main/swift/Codable/CodingUserInfoKey%2BWire.swift#L48-L53) but not in Java.

Adding a `preservingProtoFieldNames` option to `WireJsonAdapterFactory` (Moshi) and `WireTypeAdapterFactory` (Gson) allows Wire's JSON implementation to be configured for field name preservation as well.


oldergod (approved):
Let's dump the api `./gradlew apiDump` and push. I think it's fine since we have the jmvoverloads

PR #3538, opened 2026-03-11
Support repeated entries in byte arrays

oldergod:
Will fix #3537

Specs: https://protobuf.dev/programming-guides/encoding/#last-one-wins

image


JakeWharton (approved):
Rough performance hit, but should never happen

oldergod:
Can't really brag about performance if the spec is not honoured though :phylosophical:

PR #3566, opened 2026-04-14
Test confirming we don't duplicate sources in jar creation anymore

### LINKED ISSUE #3558: Wire v6 sourceReleaseJar task broken
After updating wire from v5.5.1 to v6.2.0, I'm getting the following error:

Execution failed for task ':proto-android:sourceReleaseJar'.
> Cannot copy file '/Users/alexandereggers/Developer/grpc-android/lib/build/generated/source/wire/release/myapp/proto/common/Blob.kt' to 'myapp/proto/common/Blob.kt' because file '/Users/alexandereggers/Developer/grpc-android/lib/build/generated/source/wire/release/myapp/proto/common/Blob.kt' has already been copied there.

oldergod:
Fixes #3558

Now that we're on AGP 9, the fix should be done without us having to do anything


oldergod:
Updated with test only now that we're on AGP9

JakeWharton (approved):

PR #3570, opened 2026-04-15
Use Gradle lazy properties for WireExtension API

oldergod:
Fixes #3543

Should not be breaking current callers

### DIFF
--- a/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
+++ b/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
@@ -26,6 +26,8 @@ import org.gradle.api.artifacts.MinimalExternalModuleDependency
 import org.gradle.api.artifacts.ProjectDependency
 import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.internal.catalog.DelegatingProjectDependency
+import org.gradle.api.provider.ListProperty
+import org.gradle.api.provider.Property
 import org.gradle.api.provider.Provider
 import org.gradle.api.provider.ProviderConvertible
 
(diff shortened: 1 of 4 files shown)


wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
```diff
@@ -36,88 +38,93 @@ open class WireExtension(
...
   /**
    * See [com.squareup.wire.schema.WireRun.onlyVersion].
    */
   fun onlyVersion(onlyVersion: String) {
-    this.onlyVersion = onlyVersion
+    this.onlyVersion.set(onlyVersion)
   }
 
-  fun permitPackageCycles() = permitPackageCycles
+  fun permitPackageCycles() = permitPackageCycles.get()
```
JakeWharton, line 98:
```suggestion
  fun permitPackageCycles() = permitPackageCycles.set(true)
```

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
```diff
@@ -36,88 +38,93 @@ open class WireExtension(
...
   /**
    * See [com.squareup.wire.schema.WireRun.permitPackageCycles].
    */
   fun permitPackageCycles(permitPackageCycles: Boolean) {
-    this.permitPackageCycles = permitPackageCycles
+    this.permitPackageCycles.set(permitPackageCycles)
   }
 
-  fun loadExhaustively() = loadExhaustively
+  fun loadExhaustively() = loadExhaustively.get()
```
JakeWharton, line 107:
```suggestion
  fun loadExhaustively() = loadExhaustively.set(true)
```

JakeWharton (approved):
Still can break people. Kotlin-based Gradle plugins will no longer be able to do `= whevs`.

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
oldergod: It was a getter before, why do we wanna change it to a setter?

oldergod:
Stuff like this are in to not break existing consumers. Are you referring to people setting up the Wire plugin in code?
```
  /** For Groovy DSL assignment: `protoLibrary = true`. */
  fun setProtoLibrary(value: Boolean) {
    protoLibrary.set(value)
  }
```

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireExtension.kt
JakeWharton: Oh I misread the binary diff. Previously it returned a boolean (which is misread as void) and now it returns a boxed boolean. Do you want to fix that?

JakeWharton:
Yes. Someone whose build plugin depends on the Wire extension will have to migrate from `dryRun = true` to `dryRun.set(true)`.

PR #3572, opened 2026-04-16
Avoid race in duplex pipe for streaming calls

### LINKED ISSUE #3370: GrpcStreamingCall client buffers packets and flushes only after timeout — not receiving events in real-time
I'm using com.squareup.wire.GrpcStreamingCall along with OkHttp in a Kotlin mobile app to stream gRPC events from a backend service. The issue I'm facing is that the client does not receive the streamed packets in real-time.

Instead, all packets are buffered internally and delivered together only when the server closes the connection or times out the session (e.g., after 5 minutes). This defeats the purpose of a streaming connection. 

There is no any problem from backend side, API's are working fine in Postman.

Expected Behavior
- Packets should be delivered immediately to the client as the server sends them (real-time).
- The GrpcStreamingCall.response flow should emit each message as it arrives on the wire.

Actual Behavior
- Packets are only emitted from the flow after several minutes — when the session is closed/timed out.
- All buffered packets arrive together at once, as if they were delayed or cached.

Please help me out in fixing this problem

oldergod:
Fixes #3370

From Claude

### Root Cause

`GrpcServerStreamingCall` was implemented by delegating to `GrpcStreamingCall`, which uses a `PipeDuplexRequestBody`, an asynchronous pipe where the request body is written by a background coroutine on `Dispatchers.IO`. This means the request message and the `HTTP/2 END_STREAM` flag are sent asynchronously, after the HTTP call is already enqueued.
Many gRPC server implementations wait for the client's half-close (`END_STREAM`) before starting to stream responses. With the duplex pipe approach, there's a race between when the server receives `END_STREAM` and when it decides to time out the idle stream, causing the reported ~5-minute delay.

### Fix

`RealGrpcServerStreamingCall` now has its own independent implementation using `newRequestBody()` (a non-duplex request body), identical to how `RealGrpcCall` works. The complete request message + `END_STREAM` is written synchronously in OkHttp's `writeTo()` call, ensuring the server receives the full request before any response reading begins.
The old delegation-based class is preserved as `GrpcStreamingCallServerStreamingAdapter`, still used by the test-double factory (`GrpcServerStreamingCall { ... }` in GrpcCalls.kt).

Supporting changes:

- `readFromResponseBodyCallback()` in `grpc.kt`: added an overload accepting `onResponseMetadata: (Map<String, String>) -> Unit` so both `RealGrpcStreamingCall` and `RealGrpcServerStreamingCall` can use it
- `BlockingMessageSource`: replaced `grpcCall: RealGrpcStreamingCall` with `onResponseMetadata: (Map<String, String>) -> Unit` for the same reason
- `GrpcClient.newServerStreamingCall()` — now creates `RealGrpcServerStreamingCall` directly instead of wrapping a streaming call.

### DIFF
--- a/wire-grpc-client/api/wire-grpc-client.api
+++ b/wire-grpc-client/api/wire-grpc-client.api
@@ -81,9 +81,14 @@ public final class com/squareup/wire/GrpcHttpUrlKt {
 
 public final class com/squareup/wire/GrpcMethod {
 	public fun <init> (Ljava/lang/String;Lcom/squareup/wire/ProtoAdapter;Lcom/squareup/wire/ProtoAdapter;)V
+	public fun <init> (Ljava/lang/String;Lcom/squareup/wire/ProtoAdapter;Lcom/squareup/wire/ProtoAdapter;Z)V
+	public fun <init> (Ljava/lang/String;Lcom/squareup/wire/ProtoAdapter;Lcom/squareup/wire/ProtoAdapter;ZZ)V
+	public synthetic fun <init> (Ljava/lang/String;Lcom/squareup/wire/ProtoAdapter;Lcom/squareup/wire/ProtoAdapter;ZZILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun getPath ()Ljava/lang/String;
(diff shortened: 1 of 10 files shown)


oldergod:
Does it look legit to you @swankjesse ?

Issue #2842, opened 2024-02-27
Pass dependencies symbolically to WireTask for better Gradle caching


swankjesse:
Right now we pass dependencies to WireTask as the absolute paths to the resolved artifact, like so:

```
/Users/jwilson/.m2/repository/com/squareup/wire/dinosaur/all-protos/1.0/all-protos-1.0.jar
```

This won’t work for shared Gradle caches. We should instead use a symbolic identifier like `com.squareup.wire.dinosaur:all-protos:1.0` that is independent of the current environment’s artifact directory.

martinbonnin:
At the end of the day, I think the task inputs should be `ConfigurableFileCollection` everywhere. `ConfigurableFileCollection`s can contain:
* a local zipTree
* a fileTree folder + include rules
* or a configuration with `project()` dependencies.

Which I think covers all the use cases? Then the task could be trimmed down to just:

```kotlin
abstract class WireTask @Inject constructor(
  objects: ObjectFactory,
  private val fileOperations: FileOperations,
) : DefaultTask() {

  @get:InputFiles
  @SkipWhenEmpty
  @get:PathSensitive(PathSensitivity.RELATIVE)
  abstract val protoSourceFiles: ConfigurableFileCollection

  @InputFiles
  @PathSensitive(PathSensitivity.RELATIVE)
  abstract val protoPathFiles: ConfigurableFileCollection

  // ...
}
```

(I thought `SourceTask` was needed for `@SkipWhenEmpty` but turns out it's not so it brings very little value compared to `DefaultTask`)

The nice bonus is that `ConfigurableFileCollection`s are "lazy". They won't try to resolve their configurations until the task starts executing.

oldergod:
Fixed by https://github.com/square/wire/pull/3313.

PR #3574, opened 2026-04-17
Test confirming Gradle caching improvements made in #3313

oldergod:
#2842 is already fixed in #3313. Adding a new test here.

A test where Maven dependencies are used as sourcePath `wire { sourcePath { srcJar("...") } }` was missing.


staktrace (approved):
sure

PR #3581, opened 2026-04-27
OneofMode to generate SealedClass for oneofs

### LINKED ISSUE #2749: Exhaustive checks in `oneof` fields
Is there a way to configure Wire to generate either an Enum / Sealed Class for `oneof` fields, to provide Kotlin developers with an exhaustive API for selecting the `oneof` field? It would be very helpful for detecting new cases at compile-time, rather then at run-time. 🙏 

I see that the sealed classes approach used to exist, but was removed, and several developers have asked for similar "exhaustive check" support in this issue: https://github.com/square/wire/issues/998. I have never contributed to this project before, but I would be willing to invest some time into this feature if it will be well-received.

oldergod:
Apologies for the long PR. I wanted to make everything sweet, and it's split into 4 commits.
Golden files + the interop test is the meat of the end result.

Fixes #2749


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -2899,6 +2987,68 @@ class KotlinGenerator private constructor(
       .build()
   }
 
+  /**
+   * Converts a snake_case field name to PascalCase for use as a sealed class subtype name.
+   * For example: `card_id` → `CardId`, `bank_account` → `BankAccount`.
+   */
+  private fun String.toSealedSubclassName(): String = split("_").joinToString("") { part -> part.replaceFirstChar { it.uppercaseChar() } }
```
scmorse, line 3057: Should sealed subtype names go through NameAllocator / existing Wire casing helpers instead of a custom split("_") conversion?

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -2899,6 +2987,68 @@ class KotlinGenerator private constructor(
...
+      val subclass = TypeSpec.classBuilder(subclassName)
+        .addModifiers(DATA)
+        .superclass(sealedClassName)
+        .primaryConstructor(
+          FunSpec.constructorBuilder()
+            .addParameter("value", valueType)
+            .build(),
+        )
+        .addProperty(
+          PropertySpec.builder("value", valueType)
```
scmorse, line 3096: Would it be worthwhile to include field options on the `value` property? e.g. deprecated/redacted/etc

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -2899,6 +2987,68 @@ class KotlinGenerator private constructor(
...
+        }
+      }
+
+    for (field in oneOf.fields) {
+      val subclassName = field.name.toSealedSubclassName()
+      val valueType = field.type!!.typeName
+      val subclass = TypeSpec.classBuilder(subclassName)
+        .addModifiers(DATA)
+        .superclass(sealedClassName)
+        .primaryConstructor(
```
scmorse, line 3090: Should the sealed class instances per-field use the builder pattern if `buildersOnly` is configured?

scmorse (commented):
I know you said this is still a WIP, but I noticed some other files that probably need some updates:

 - `wire-compiler/src/main/java/com/squareup/wire/WireCompiler.kt`
 - `wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/WireOutput.kt`
 - `wire-gradle-plugin/api/wire-gradle-plugin.api`
 - `wire-compiler/src/test/java/com/squareup/wire/WireCompilerTest.kt`
 - `docs/wire_compiler.md`

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
oldergod: I'm not sure I follow but I don't think so? You can see the golden file for referrence

wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/SealedOneOfBinding.kt
```diff
@@ -0,0 +1,81 @@
...
+internal class SealedOneOfBinding<M : Message<M, B>, B : Message.Builder<M, B>>(
+  private val messageField: Field,
+  builderType: Class<B>,
+  private val annotation: WireOneofField,
+  private val subclassType: Class<*>,
+  classLoader: ClassLoader? = messageField.declaringClass.classLoader,
+) : FieldOrOneOfBinding<M, B>() {
+
+  private val builderField: Field =
+    builderType.getDeclaredField(messageField.name).also { it.isAccessible = true }
```
scmorse, line 45: Should this use `by lazy {}`, like `valueField` does, so that it doesn't fail if builder types aren't enabled?

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -3170,6 +3361,7 @@ class KotlinGenerator private constructor(
       buildersOnly: Boolean = false,
       escapeKotlinKeywords: Boolean = false,
       enumMode: EnumMode = ENUM_CLASS,
+      oneofMode: OneofMode = OneofMode.FLAT,
```
scmorse, line 3422: I hit a binary compatibility issue testing this against cash-server. Adding `oneofMode` to `KotlinGenerator.Companion.get` changes the JVM signature of Kotlin’s generated `get$default` method, so existing compiled schema handlers/extensions that call `KotlinGenerator.get(...)` fail at runtime with `NoSuchMethodError`.

So I think that Cash's existing compiled extensions can’t run with the new Wire jars unless they are recompiled, is that okay? Maybe cash's extensions can just be recompiled and then updated simultaneously with the wire version upgrade.

Alternatively, maybe we could preserve the old overload and have it delegate to the new one with `oneofMode = OneofMode.FLAT`?

scmorse (commented):
- [x] Looks like `docs/wire_compiler.md` could use an update

wire-runtime/src/jvmMain/kotlin/com/squareup/wire/internal/SealedOneOfBinding.kt
oldergod: Code has been updated here

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
oldergod: Can do it!

oldergod: Addressed by adding a hidden constructor

oldergod (commented):
@scmorse WireCompiler.md is updated and fixed some old things in it

scmorse (approved):

staktrace:
Anything I should be aware of with respect to compile or runtime incompatibility with existing code?

staktrace (approved):
The general idea seems fine to me. The PR is surprisingly big for what it is and I don't have sufficient context on the wire codebase to properly review it. If there are any known backwards-compatibility issues please flag them explicitly.

oldergod:
> Anything I should be aware of with respect to compile or runtime incompatibility with existing code?

Nothing breaks unless someone uses the option and another value than `flat`

PR #3582, opened 2026-04-27
Use Gradle properties in nested Wire output DSL

steve-the-edwards:
I recently had this issue - https://github.com/square/wire/issues/3543 - but for the inner Custom Schema Handler DSL. 

So I thought we could extend this type of solution - https://github.com/square/wire/pull/3570 - to the DSL for this as well.

We need to keep them as Output's for as long as possible before converting to Target so that we can bring the Gradle Property/Provider plumbing through.

To adopt 'Gradle-idiomatic' naming we make the existing field names the Properties, and add setters if callers still want to send raw values.

### DIFF
--- a/docs/wire_compiler.md
+++ b/docs/wire_compiler.md
@@ -9,7 +9,7 @@ Getting Started
 ---------------
 
 The best way to configure and execute the Wire compiler is via our [Gradle][gradle] plugin. It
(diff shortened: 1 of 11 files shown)


oldergod (approved):
Added a small commit. Good to go

PR #3594, opened 2026-05-11
Add accessors for oneof's generated data classes

oldergod:
Created the extension on the sealed class, not the outer class, to make it clear it's going through a oneof

### DIFF
--- a/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
+++ b/wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
@@ -338,6 +338,16 @@ class KotlinGenerator private constructor(
     }
   }
 
+  fun generateSealedOneOfAccessors(type: Type): List<PropertySpec> {
+    if (oneofMode != OneofMode.SEALED_CLASS) return listOf()
+
+    return when (type) {
+      is MessageType -> type.sealedOneOfAccessors()
+      is EnclosingType -> type.nestedTypes.flatMap(::generateSealedOneOfAccessors)
+      is EnumType -> listOf()
+    }
+  }
+
   /**
    * Generates all [TypeSpec]s for the given [Service].
    *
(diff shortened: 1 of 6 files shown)


wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -3063,6 +3073,53 @@ class KotlinGenerator private constructor(
...
+      val accessorNameAllocator = NameAllocator(preallocateKeywords = !escapeKotlinKeywords)
+
+      for (field in oneOf.fields) {
+        val fieldName = if (field.name == field.type!!.simpleName || hasEponymousType(schema, field)) {
+          accessorNameAllocator.newName(legacyQualifiedFieldName(field), field)
+        } else {
+          accessorNameAllocator.newName(field.name, field)
+        }
+        val subclassName = sealedClassName.nestedClass(subclassNameAllocator[field])
+        accessors += PropertySpec.builder(fieldName, field.type!!.typeName.copy(nullable = true))
```
scmorse, line 3112: Do we need the `if (field.isDeprecated) {` and `for (annotation in optionAnnotations(field.options)) {` bits from `sealedOneOfClass` below, in order to propagate metadata / annotations on the extension fields too?

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -3314,18 +3371,18 @@ class KotlinGenerator private constructor(
 
   private fun MessageType.flatOneOfs(): List<OneOf> = when (oneofMode) {
     OneofMode.FLAT -> oneOfs.filter { it.fields.size < boxOneOfsMinSize }
-    else -> emptyList()
+    else -> listOf()
```
scmorse, line 3397: No problem, I'm just curious to understand why switching from `emptyList()` to `listOf()`, is that important?

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
```diff
@@ -338,6 +338,16 @@ class KotlinGenerator private constructor(
     }
   }
 
+  fun generateSealedOneOfAccessors(type: Type): List<PropertySpec> {
+    if (oneofMode != OneofMode.SEALED_CLASS) return listOf()
+
+    return when (type) {
+      is MessageType -> type.sealedOneOfAccessors()
+      is EnclosingType -> type.nestedTypes.flatMap(::generateSealedOneOfAccessors)
+      else -> listOf()
```
scmorse: Nit, would prefer to see this last case listed out for exhaustivity and future-proofing.

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
oldergod: `listOf` reads better! no reason for changing this in this PR

wire-kotlin-generator/src/main/java/com/squareup/wire/kotlin/KotlinGenerator.kt
oldergod: fixed!

JakeWharton (approved):
Should these be `inline`?

PR #3597, opened 2026-05-13
Add negative-length check in skipGroup

oldergod:
https://github.com/square/wire/pull/3595 but for the Wire 6 patch release


oldergod:
- [x] merge on green

PR #3599, opened 2026-05-14
Remove trailing spaces in generated Kotlin types

oldergod:
A KotlinPoet bug? no


JakeWharton (approved):
I saw this in the other diff but didn't feel like it was worth commenting. Do you mean KotlinPoet bug? If it's the one doing it, then yeah, it shouldn't emit them.

oldergod:
I don't think it's a KotlinPoet bug, the "problem" was that we were nudging new lines instead of letting KotlinPoet do the formatting as it sees fit

PR #3611, opened 2026-05-27
Correctly register all sources

### LINKED ISSUE #3607: Wire 7.0.0-alpha (1/2/3): Android variant Kotlin source set empty — KSP and compileKotlin see no user sources
## Summary

After upgrading to Wire 7.0.0-alpha (any of `alpha01`, `alpha02`, `alpha03`), KSP runs against an Android library variant but receives **zero** Kotlin source files. The Kotlin compile task for the variant likewise does not see the user's `.kt` files under `src/main/java`. Both `release` and `debug` variants reproduce identically.

For projects using KSP-driven libraries (e.g. `koin-ksp-compiler`), this surfaces as missing generated symbols at consumer sites — e.g. `Unresolved reference 'defaultModule'` / `Unresolved reference 'module'` when importing `org.koin.ksp.generated.*`.

## Versions

- Wire Gradle plugin: reproduced on `7.0.0-alpha01`, `alpha02`, and `alpha03`
- Android Gradle Plugin: `9.2.1`
- Kotlin: `2.2.21`
- KSP: `2.3.7`
- Gradle: `9.4.1`
- KSP processor in use: `io.insert-koin:koin-ksp-compiler:2.1.0`
- Build target: `com.android.library`

(Wire `6.4.0` works for KSP but breaks `sourceReleaseJar` per #3558 — the fix for which is what landed in 7.x and triggered this regression.)

## Module config

```kotlin
plugins {
    id("com.android.library")
    alias(libs.plugins.wire)
    alias(libs.plugins.ksp)
    alias(libs.plugins.kotlin.compose)
}

dependencies {
    implementation(libs.wire.runtime)
    implementation(libs.koin.annotations)
    ksp(libs.koin.ksp)
}

wire {
    kotlin { javaInterop = true }
    sourcePath { srcDir("src/main/proto") }
}
```

User Kotlin sources live under `src/main/java/...` (the conventional shared dir for Kotlin + Java co-located code in many Android codebases).

## Symptom — diagnostic output

`./gradlew :my-library:kspReleaseKotlin --rerun-tasks --info`:

```
> Task :my-library:kspReleaseKotlin
Task ':my-library:kspReleaseKotlin' is not up-to-date because:
  The input changes require a full rebuild for incremental task ...
i: [ksp] loaded provider(s): [org.koin.compiler.BuilderProcessorProvider]
BUILD SUCCESSFUL
```

Note: every other Android library module in the project emits a `Kotlin source files: <list>` line at this point during `--info`. The Wire-using module emits **nothing** — KSP sees an empty source set.

`my-library/build/generated/ksp/` is created but contains no `release/` or `debug/` subdirectory after the task runs (KSP only creates per-variant output dirs when it actually emits something).

The same is true for `kspDebugKotlin` — variant-independent.

## What rules in / out

- **Not specific to `release` variant** — debug variant reproduces identically.
- **Not specific to `singleVariant("release") { withSourcesJar() }`** — the KSP failure happens regardless of that publishing config.
- **Not a KSP / processor wiring issue** — the processor loads cleanly (`loaded provider(s): [...BuilderProcessorProvider]`).
- **Not a stale cache** — reproduced after `clean` and with `--rerun-tasks`, configuration cache disabled.
- **Specific to Wire 7.x** — same project on Wire `6.4.0` has a populated Kotlin source set and KSP generates output normally (it just hits #3558 on `sourceReleaseJar`).

## Hypothesis (best guess only)

The `KotlinTarget` source-root rework that closed #3558 / shipped via #3566 appears to leave the AGP variant's Kotlin source container in a state where the user's `src/main/java` `.kt` files are no longer visible to either `compileKotlin` or `ksp{Variant}Kotlin`. I suspect the variant API call that registers Wire's generated output is also displacing or overriding the variant's existing Kotlin source set rather than appending to it, but I haven't narrowed this down further — flagging in case the diagnosis is obvious to maintainers familiar with the AGP 9 variant API.

## Repro

Happy to put together a minimal repro on a public Gist if useful — the shape is just any `com.android.library` module with `wire` + `ksp` + Kotlin sources under `src/main/java`.

oldergod:
fixes https://github.com/square/wire/issues/3607

### DIFF
--- a/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/kotlin/SourceRoots.kt
+++ b/wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/kotlin/SourceRoots.kt
@@ -177,7 +177,9 @@ private class AndroidSource(
           variant.sources.java?.addGeneratedSourceDirectory(wireTask) { it.outputDirectoriesList[index] }
         }
         is KotlinOutput -> {
-          variant.sources.kotlin?.addGeneratedSourceDirectory(wireTask) { it.outputDirectoriesList[index] }
+          // Android Kotlin compilation and KSP also consume .kt files from Java source roots.
+          // Registering with variant.sources.kotlin can hide handwritten sources from KSP.
+          variant.sources.java?.addGeneratedSourceDirectory(wireTask) { it.outputDirectoriesList[index] }
         }
         is CustomOutput -> {
           // Custom targets are wildcards, so we add all output directories.
--- a/wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
+++ b/wire-gradle-plugin/src/test/kotlin/com/squareup/wire/gradle/WirePluginTest.kt
@@ -36,6 +36,7 @@ import com.squareup.wire.testing.withPlatformSlashes
 import com.squareup.wire.wireVersion
 import java.io.File
 import java.io.IOException
+import java.util.Properties
 import java.util.zip.ZipFile
 import kotlin.text.RegexOption.DOT_MATCHES_ALL
 import kotlin.text.RegexOption.MULTILINE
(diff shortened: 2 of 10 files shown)


wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/kotlin/SourceRoots.kt
```diff
@@ -177,7 +177,9 @@ private class AndroidSource(
           variant.sources.java?.addGeneratedSourceDirectory(wireTask) { it.outputDirectoriesList[index] }
         }
         is KotlinOutput -> {
-          variant.sources.kotlin?.addGeneratedSourceDirectory(wireTask) { it.outputDirectoriesList[index] }
```
oldergod, line 180: it was supposed to be fixed so........... not happy

oldergod: Maybe I need to find the AOSP ticket and share

JakeWharton (approved):

wire-gradle-plugin/src/main/kotlin/com/squareup/wire/gradle/kotlin/SourceRoots.kt
oldergod: confusing with another piece of code, all good

PR #3626, opened 2026-06-14
Skip null list elements in moshi fromJson

### LINKED ISSUE #3353: Ignore null lists in MessageJsonAdapter#fromJson
In `MessageJsonAdapter#fromJson`, we ignore null values with `if (value == null) continue`. But if the deserialized value is a `List<T?>` whose some element is null, we still call `fieldBinding.set(builder, value)`. 

Image
<br/><br/>

Later, when the `T` message is built, `immutableCopyOf("…", list)` checks `require(null !in result)` and throws an IllegalArgumentException because the list contains a null entry.

Image
<br/><br/>

~~Is it okay to open a PR that updates `fromJson` to skip lists where every element is null, or to only include non-null elements?~~

EDIT: I initially thought we could simply remove nulls, but there are cases—like `repeated google.protobuf.NullValue`—where null is officially permitted.

mvanhorn:
## Summary

`MessageJsonAdapter#fromJson` (and the gson `MessageTypeAdapter#read`) skipped a field only when the whole decoded value was null. For a repeated field, a literal `null` array element (e.g. `"items": [{...}, null]`) decodes into a non-null `List<T?>` whose element is null. That list passed the `value == null` guard and reached `fieldBinding.set(...)`, where building the message hit `Internal.immutableCopyOf`'s `require(null !in result) { "$name.contains(null)" }` and threw `IllegalArgumentException: <field>.contains(null)`.

This adds a shared `FieldOrOneOfBinding.withoutStrayNullElements(value)` helper that, for repeated non-struct fields, drops stray null list elements before they reach the builder. Struct fields (e.g. `repeated google.protobuf.NullValue`, `google.protobuf.ListValue`) are decoded by the struct adapter and may legitimately carry null entries, so they are left untouched. Both the moshi and gson adapters route through the helper.

## Why this matters

Decoding otherwise-valid JSON that happens to contain a `null` array element crashed instead of ignoring the unrepresentable entry, matching the report in https://github.com/square/wire/issues/3353 where @oldergod said a PR was welcome.

## Testing

`./gradlew :wire-tests:jvm-json-kotlin:test --tests 'com.squareup.wire.WireJsonTest'` — BUILD SUCCESSFUL (210 tests, parameterized across the moshi and gson libraries). Also ran `:wire-runtime:jvmTest`, `:wire-moshi-adapter-test:test`, and `:wire-gson-support:test` — all green. New cases cover a null element in a repeated message field, a repeated scalar field, and preservation of a null inside a struct list.

Fixes #3353

### DIFF
--- a/wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/FieldOrOneOfBinding.kt
+++ b/wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/FieldOrOneOfBinding.kt
@@ -89,4 +89,23 @@ abstract class FieldOrOneOfBinding<M, B> {
     if (isMap && syntax == Syntax.PROTO_3) return true
     return false
   }
+
+  /**
+   * Returns [value] with any null elements removed when it's the decoded JSON value for a repeated
+   * field. A literal `null` in a JSON array (e.g. `"items": [{...}, null]`) decodes to a null list
+   * element that isn't representable in a proto repeated field; left in place it would trip the
+   * `null !in result` check in [immutableCopyOf]. Struct fields (such as
+   * `repeated google.protobuf.NullValue`) are decoded by the struct adapter and may legitimately
+   * carry null entries, so they're returned unchanged.
+   */
+  fun withoutStrayNullElements(value: Any?): Any? {
+    if (!label.isRepeated || singleAdapter.isStructAdapter() || value !is List<*>) return value
+    if (null !in value) return value
+    return value.filterNotNull()
+  }
+
+  private fun ProtoAdapter<*>.isStructAdapter(): Boolean = this == ProtoAdapter.STRUCT_MAP ||
+    this == ProtoAdapter.STRUCT_LIST ||
+    this == ProtoAdapter.STRUCT_VALUE ||
+    this == ProtoAdapter.STRUCT_NULL
 }
--- a/wire-gson-support/src/main/java/com/squareup/wire/MessageTypeAdapter.kt
+++ b/wire-gson-support/src/main/java/com/squareup/wire/MessageTypeAdapter.kt
@@ -71,7 +71,7 @@ internal class MessageTypeAdapter<M : Message<M, B>, B : Message.Builder<M, B>>(
       // interpreted as the appropriate default value when parsed into a protocol buffer."
       if (value == null) continue
 
-      jsonField.fieldBinding.set(builder, value)
+      jsonField.fieldBinding.set(builder, jsonField.fieldBinding.withoutStrayNullElements(value))
     }
     input.endObject()
     return builder.build()
--- a/wire-moshi-adapter/src/main/java/com/squareup/wire/MessageJsonAdapter.kt
+++ b/wire-moshi-adapter/src/main/java/com/squareup/wire/MessageJsonAdapter.kt
@@ -79,7 +79,7 @@ internal class MessageJsonAdapter<M : Message<M, B>, B : Message.Builder<M, B>>(
       if (value == null) continue
 
       val fieldBinding = messageAdapter.fieldBindingsArray[index]
-      fieldBinding.set(builder, value)
+      fieldBinding.set(builder, fieldBinding.withoutStrayNullElements(value))
     }
     input.endObject()
     return builder.build()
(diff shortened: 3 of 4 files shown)


wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/FieldOrOneOfBinding.kt
```diff
@@ -89,4 +89,23 @@ abstract class FieldOrOneOfBinding<M, B> {
...
   }
+
+  /**
+   * Returns [value] with any null elements removed when it's the decoded JSON value for a repeated
+   * field. A literal `null` in a JSON array (e.g. `"items": [{...}, null]`) decodes to a null list
+   * element that isn't representable in a proto repeated field; left in place it would trip the
+   * `null !in result` check in [immutableCopyOf]. Struct fields (such as
+   * `repeated google.protobuf.NullValue`) are decoded by the struct adapter and may legitimately
+   * carry null entries, so they're returned unchanged.
+   */
```
oldergod, line 100: The comment sounds like an explanation that would live in this PR, but not in the code.
Simply explaining what it does is sufficient.

```suggestion
  /**
   * Normalizes a decoded JSON field value before assigning it to a builder.
   *
   * For repeated, non-struct fields, null list elements are discarded because proto repeated fields
   * cannot store null elements. Struct fields are left unchanged because `google.protobuf` struct
   * values use null as a valid JSON value.
   */
```

Also let's rename the method to `filterNullElements`

oldergod (approved):
Thank you!

Also please, could you sign our CLA so that I can merge it?
https://docs.google.com/forms/d/e/1FAIpQLSeRVQ35-gq2vdSxD1kdh7CJwRdjmUA0EZ9gRXaWYoUeKPZEQQ/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1

wire-runtime/src/commonMain/kotlin/com/squareup/wire/internal/FieldOrOneOfBinding.kt
oldergod: - [ ] waiting for a resolution before merging

oldergod: Do you wanna finish it @mvanhorn ?

mvanhorn:
Yes, sorry for the delay. Done in dda6a32.

Applied your suggestion for the KDoc verbatim and renamed `withoutStrayNullElements` to `filterNullElements`, updating both call sites (the moshi and gson adapters). You were right that the old comment was explaining why the change exists rather than what the method does, which belongs here in the PR rather than in the code.

The rename does not touch any `.api` dump: `FieldOrOneOfBinding` is not in one, and the binary-compatibility-validator dumps do not cover `com.squareup.wire.internal`, so there is nothing to regenerate.

mvanhorn:
@oldergod the resolution you were waiting on landed in `dda6a32`: your KDoc suggestion applied verbatim, and `withoutStrayNullElements` renamed to `filterNullElements` across both the moshi and gson call sites.

I also updated the description, which still used the old method name.

The only red check is the task-list bot tracking your unchecked box on `FieldOrOneOfBinding.kt` line 99, which I cannot tick. If you are happy with where it landed, ticking that box should clear the last blocker.

oldergod:
Let me look at it @mvanhorn

oldergod:
@mvanhorn please, [sign the CLA.](https://docs.google.com/forms/d/e/1FAIpQLSeRVQ35-gq2vdSxD1kdh7CJwRdjmUA0EZ9gRXaWYoUeKPZEQQ/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1)

oldergod:
@mvanhorn apologies, I checked the spec again today https://protobuf.dev/programming-guides/json/#null-values and it looks like we're doing the wrong thing here for _`null` values are not allowed within repeated fields_. The issue was probably misleading but we should not skip them if the specs says differently.

oldergod:
I think https://github.com/square/wire/pull/3727 would be best

PR #3639, opened 2026-06-28
Rename a nested type named `Builder` to avoid a collision

### LINKED ISSUE #593: A nested type named 'Builder' collides with our message builder.
This type should be mangled into `Builder_` via `NameAllocator`

eyupcanakman:
A message generates a nested `Builder`, so a proto type that is itself named `Builder` collided with it and the generated code did not compile. Both generators now reserve the name through a `NameAllocator`, renaming the nested type to `Builder_`. Existing output is unchanged.

Fixes #593

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ Change Log
 Unreleased
 ----------
 
+### Java
+
+* Fix a naming collision when a message has a nested type named `Builder` (#593)
+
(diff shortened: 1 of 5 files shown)


oldergod:
@eyupcanakman thank you! I've added a commit to be not rename anything when we don't need to.
Merging on green.

oldergod (approved):

PR #3644, opened 2026-07-02
FieldMask: runtime + schema awareness

oldergod:


wire-runtime/src/commonMain/kotlin/com/squareup/wire/FieldMask.kt
```diff
@@ -0,0 +1,37 @@
...
+ */
+package com.squareup.wire
+
+/**
+ * A set of symbolic field paths.
+ *
+ * Field masks are used to specify a subset of fields on a target message. Each path uses proto
+ * field names, separated by dots for nested fields.
+ */
+class FieldMask(paths: List<String> = emptyList()) {
```
adrw, line 24: Could this be a data class?

adrw (approved):

wire-runtime/src/commonMain/kotlin/com/squareup/wire/FieldMask.kt
oldergod: Any change are breaking for data classes, we have more control this way

PR #3676, opened 2026-08-14
Expose extension field numbers as generated Swift constants

loganblevins:
## What

The Swift code generator now emits `public static let` constants alongside every generated extension accessor, mirroring the existing `default_<name>` constants:

- `fieldNumber_<name>: UInt32` for every extension field
- `fieldEncoding_<name>: ProtoIntEncoding` for integer extension fields, whose accessors pass an explicit `encoding:`

```swift
extension FooBar {
    public var ext: FooBar.FooBarBazEnum? {
        get { ... }
        set { ... }
    }
    /**
     * Field number for the ext extension field.
     */
    public static let fieldNumber_ext: UInt32 = 101
}

extension AllTypes {
    public var ext_opt_sint32: Int32? {
        get { self.parseUnknownField(fieldNumber: 1003, type: Int32.self, encoding: .signed) }
        set { self.setUnknownField(fieldNumber: 1003, newValue: newValue, encoding: .signed) }
    }
    /**
     * Field number for the ext_opt_sint32 extension field.
     */
    public static let fieldNumber_ext_opt_sint32: UInt32 = 1003
    /**
     * Integer encoding for the ext_opt_sint32 extension field.
     */
    public static let fieldEncoding_ext_opt_sint32: ProtoIntEncoding = .signed
}
```

## Why

The generated extension accessors hardcode their decoding behavior: `try? ProtoDecoder().decode(...)`, i.e. the default `.throwError` enum strategy with the error swallowed. Consumers that need different semantics — a `.returnNil` decoder so a single unrecognized enum value doesn't nil the entire payload, or an `unknownFields[tag]` presence check to distinguish an absent extension from an undecodable one — must call `parseUnknownField`/`setUnknownField` directly, and those APIs take a raw `fieldNumber: UInt32` plus, for integer fields, an `encoding: ProtoIntEncoding`.

Today no generated symbol carries either value: callers copy the literals out of the `.proto` source by hand, and nothing keeps the copies in sync with codegen. A wrong field number or encoding silently decodes wrong instead of failing to compile. (GPB exposes the same information via `GPBExtensionDescriptor`; Wire has no Swift equivalent.) We hit this while migrating envelopes — a proto2 message whose payloads are all extensions — from GPB to Wire: extension field numbers had to be hardcoded and pinned with hand-written tag-agreement tests. With this change the constants come from codegen, so they are contract-checked at the source.

Scope notes:

- The `fieldNumber_` constant is generated for every extension field, including repeated fields (which get no `default_` constant).
- The `fieldEncoding_` constant is generated exactly when the generated accessor passes `encoding:` — signed, fixed, and variable integer fields, singular and repeated — so the (fieldNumber, encoding) pair travels together. Non-integer fields have no `encoding:` parameter and get no constant.
- For heap-allocated (CopyOnWrite) messages the constants are emitted on the extended type only, not again on its `Storage` type.
- Naming follows the `default_<name>` precedent; the types are `UInt32` and `ProtoIntEncoding` so the constants can be passed straight to `parseUnknownField`/`setUnknownField`.

## Testing

- Three new `SwiftGeneratorTest` cases: field-number constants for singular and repeated extension fields; encoding constants for variable/signed/fixed integer fields (and their absence on non-integer fields); and single emission (extended type only) for heap-allocated messages, with a `public struct Storage` precondition so the single-emission assertions cannot pass trivially.
- A Swift-side `ExtensibleTests` case reads the constants off real generated types (`Extensible`, `LargeExtensible`) and round-trips a value through `setUnknownField`/`parseUnknownField` using only generated constants.
- Goldens regenerated with `./gradlew generateTests` (`AllTypes`, `FooBar`, `SwiftModuleOneMessage`); `git status` is byte-clean after a fresh regeneration.
- `swift test` passes locally on macOS.

Suggested CHANGELOG entry (left out of the diff since entries appear to be authored at release time):

> * New: Swift — generated extension accessors are now accompanied by `fieldNumber_<name>` constants (and, for integer fields, `fieldEncoding_<name>` constants) exposing the extension's field number and wire encoding, mirroring the existing `default_<name>` constants.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoIntCodable.swift
+++ b/wire-runtime-swift/src/main/swift/ProtoCodable/ProtoIntCodable.swift
@@ -52,3 +52,6 @@ public enum ProtoIntEncoding {
     case signed
     case variable
 }
+
+extension ProtoIntEncoding : Sendable {
+}
--- a/wire-runtime-swift/src/test/swift/ExtensibleTests.swift
+++ b/wire-runtime-swift/src/test/swift/ExtensibleTests.swift
@@ -133,4 +133,35 @@ final class ExtensibleTests: XCTestCase {
         XCTAssertEqual(LargeExtensible.default_ext_value17, "my extension default value")
         XCTAssertEqual(LargeExtensible.default_ext_value18, "")
     }
+
+    func testExtensionFieldNumberAndEncodingConstants() {
+        XCTAssertEqual(Extensible.fieldNumber_ext_int32, 1001)
+        XCTAssertEqual(Extensible.fieldEncoding_ext_int32, .variable)
+        XCTAssertEqual(Extensible.fieldNumber_ext_sint32, 1003)
+        XCTAssertEqual(Extensible.fieldEncoding_ext_sint32, .signed)
+        XCTAssertEqual(Extensible.fieldNumber_ext_fixed32, 1004)
+        XCTAssertEqual(Extensible.fieldEncoding_ext_fixed32, .fixed)
+        XCTAssertEqual(Extensible.fieldNumber_ext_string, 1014)
+        XCTAssertEqual(LargeExtensible.fieldNumber_ext_value17, 17)
+        XCTAssertEqual(LargeExtensible.fieldNumber_rep_ext_sint32, 21)
+        XCTAssertEqual(LargeExtensible.fieldEncoding_rep_ext_sint32, .signed)
+
+        // Round-trip a value through the raw APIs using only generated constants.
+        var message = Extensible()
+        message.setUnknownField(
+            fieldNumber: Extensible.fieldNumber_ext_sint32,
+            newValue: Int32(-42),
+            encoding: Extensible.fieldEncoding_ext_sint32
+        )
+        XCTAssertEqual(
+            message.parseUnknownField(
+                fieldNumber: Extensible.fieldNumber_ext_sint32,
+                type: Int32.self,
+                encoding: Extensible.fieldEncoding_ext_sint32
+            ),
+            -42
+        )
+        // The generated accessor reads the same field the raw APIs just wrote.
+        XCTAssertEqual(message.ext_sint32, -42)
+    }
 }
--- a/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
+++ b/wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt
@@ -1438,6 +1438,7 @@ class SwiftGenerator private constructor(
               addProperty(defaultProperty)
             }
 
+            // The constants belong on the extended type; the storage pass must skip them or they'd be declared twice.
             if (!forStorageType) {
               addProperty(extensionFieldNumberProperty(field))
               extensionFieldEncodingProperty(field)?.let { addProperty(it) }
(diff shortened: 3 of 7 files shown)


loganblevins:
@oldergod GH won't let me request reviews... tagging instead

dnkoutso (approved):

oldergod (approved):
Approving with a few non-blocking comments.

`wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt:1342`, The constant covers only half of what `parseUnknownField` needs. Signed/fixed integer extensions also require the `encoding:` argument (see the generated `parseUnknownField(fieldNumber: 1003, type: Int32.self, encoding: .signed)` in `wire-tests-swift/no-manifest/src/main/swift/AllTypes.swift:898`), and callers still have to copy that from the .proto by hand, a wrong copy silently decodes wrong instead of failing to compile. Emit the encoding alongside the number so the pair travels together.

`wire-swift-generator/src/test/java/com/squareup/wire/swift/SwiftGeneratorTest.kt:93`, `BigMessage` sits exactly on the 16-field heap-allocation threshold, and the test never asserts the storage split actually happened. If the threshold moves or a field is dropped, only one extension block is generated and both assertions pass trivially while the test still reads as covering the guard. Add `assertThat(code).contains("public struct Storage")` as a precondition.

`wire-runtime-swift/src/test/swift/ExtensibleTests.swift:132`, The sibling `default_` constants have a Swift-side test that reads them off real generated types; the new constants have none. Add a case next to `testExtensionDefaultValues` that reads the constants off `LargeExtensible`/`Extensible` and round-trips one through `parseUnknownField`/`setUnknownField`.

`wire-swift-generator/src/main/java/com/squareup/wire/swift/SwiftGenerator.kt:1428`, The `!forStorageType` guard is the only thing preventing a duplicate declaration when both emission paths run over the same fields, and that invariant is documented nowhere but a test. Add a one-line comment stating the constant belongs on the extended type and the storage pass must skip it.

loganblevins:
🤖 _Comment posted by Logan's AI agent._

Thanks for the review @oldergod — all four points addressed, one commit each:

- **Encoding travels with the number** (037f138): the generator now emits `public static let fieldEncoding_<name>: ProtoIntEncoding` next to `fieldNumber_<name>`, gated on the same predicate that decides whether the accessor passes `encoding:` (signed/fixed/variable integer fields, singular and repeated), so the pair can't drift apart. The `AllTypes` golden picked up constants for its 30 integer extension fields. Since this grows the generated API surface beyond what you approved (naming, `ProtoIntEncoding` typing, integer-only emission rule), happy to get a quick re-look.
- **Heap test precondition** (16ee8f7): the single-emission test now asserts `public struct Storage` before the once-only assertions, so it fails loudly if `BigMessage` ever stops being heap-allocated. It also covers a `sint32` extension so the storage-pass guard is exercised for the encoding constants too.
- **Swift-side test** (d4acc8a): added `testExtensionFieldNumberAndEncodingConstants` next to `testExtensionDefaultValues`. It reads the constants off `Extensible`/`LargeExtensible`, round-trips a value through `setUnknownField`/`parseUnknownField` using only generated constants, and checks the generated `ext_sint32` accessor observes what the raw APIs wrote (which would catch a wrong-encoding constant).
- **Guard comment** (2a4a0f2): documented the invariant at the `!forStorageType` guard.


**Follow-up** (e68a19a): the swift CI job's strict-concurrency golden compile rejects `public static let`s of a non-Sendable type, so `ProtoIntEncoding` now declares `Sendable` explicitly (payload-free enum; public types get no implicit conformance across modules).

PR #3684, opened 2026-08-17
Runtime: implement withoutUnknownFields() without builders

### LINKED ISSUE #1200: Calling withoutUnknownFields() throws AssertionError on JVM
`withoutUnknownFields()` implementation in `jvmMain/.../Message.kt` calls into `newBuilder()`, which throws `AssertionError` in Kotlin code.

arimu1:
Fixes #1200.

`Message.withoutUnknownFields()` on the JVM called `newBuilder().clearUnknownFields().build()`. Kotlin generation with the default `javaInterop = false` implements `newBuilder()` as `throw AssertionError("Builders are deprecated...")`, so callers on JVM Kotlin hit that assertion. Closed #1201 only moved `UnknownFieldsTest` to `commonTest`, which uses `copy(unknownFields = ByteString.EMPTY)` and never exercises this JVM method.

Wire always encodes this message's unknown fields as a suffix (`writer.writeBytes(value.unknownFields)` last, including the reverse encoder). Decode that known-field prefix so builders are not required. Nested unknown fields stay until stripped on those messages, same as the old builder path.

## Test plan

- [x] Unfixed HEAD: `:wire-tests:multiplatform:jvmTest --tests com.squareup.wire.WithoutUnknownFieldsTest` (javaInterop=false) fails with `AssertionError: Builders are deprecated...` from `Message.withoutUnknownFields`
- [x] After the change, that suite passes (strip this-level unknown fields, keep nested, ParseTest-style unknown tag, identity when none)
- [x] `:wire-tests:jvm-java-kotlin:test --tests com.squareup.wire.UnknownFieldsTest --tests com.squareup.wire.ParseTest`
- [x] `:wire-tests:jvm-kotlin-interop:test --tests com.squareup.wire.UnknownFieldsTest` (javaInterop=true)
- [x] `:wire-runtime:apiCheck` and spotless on the touched modules
- JDK 21, macOS aarch64. `./gradlew -Dkjs=false -Dknative=false -Dkwasm=false -Pswift=false`

I will complete Square's Individual CLA if the bot asks.

### DIFF
--- a/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/Message.kt
+++ b/wire-runtime/src/jvmMain/kotlin/com/squareup/wire/Message.kt
@@ -60,8 +60,18 @@ protected actual constructor(
    */
   actual abstract fun newBuilder(): B
 
-  /** Returns this message with any unknown fields removed. */
-  fun withoutUnknownFields(): M = newBuilder().clearUnknownFields().build()
+  /**
+   * Returns this message with any unknown fields removed.
+   *
+   * Implemented as an adapter round-trip of the known-field prefix rather than via [newBuilder],
+   * which throws when Kotlin generation hides builders (`javaInterop = false`).
+   */
+  fun withoutUnknownFields(): M {
+    @Suppress("UNCHECKED_CAST")
+    val message = this as M
+    val encoded = adapter.encodeByteString(message)
+    return adapter.decode(encoded.substring(0, encoded.size - unknownFields.size))
+  }
 
   @Suppress("UNCHECKED_CAST")
   override fun toString(): String = adapter.toString(this as M)
--- a/wire-tests/multiplatform/src/jvmTest/kotlin/com/squareup/wire/WithoutUnknownFieldsTest.kt
+++ b/wire-tests/multiplatform/src/jvmTest/kotlin/com/squareup/wire/WithoutUnknownFieldsTest.kt
@@ -1,0 +1,107 @@
+/*
+ * Copyright (C) 2026 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.wire
+
+import assertk.assertThat
+import assertk.assertions.isEqualTo
+import assertk.assertions.isNotEqualTo
+import com.squareup.wire.protos.kotlin.edgecases.OneField
+import com.squareup.wire.protos.kotlin.unknownfields.NestedVersionOne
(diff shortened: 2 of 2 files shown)


wire-runtime/src/jvmMain/kotlin/com/squareup/wire/Message.kt
```diff
@@ -60,8 +60,18 @@ protected actual constructor(
...
 
-  /** Returns this message with any unknown fields removed. */
-  fun withoutUnknownFields(): M = newBuilder().clearUnknownFields().build()
+  /**
+   * Returns this message with any unknown fields removed.
+   *
+   * Implemented as an adapter round-trip of the known-field prefix rather than via [newBuilder],
+   * which throws when Kotlin generation hides builders (`javaInterop = false`).
+   */
+  fun withoutUnknownFields(): M {
```
oldergod, line 69: Consider an early return before the round trip:

```kotlin
if (unknownFields.size == 0) return this as M
```

Messages with no unknown fields are the common case, and each call now pays a full encode and decode of the whole message, where the old builder path did a cheap field copy. The round trip also adds a new failure mode: the decoder enforces a recursion limit of 100, so a message graph nested deeper than that now throws `IOException` where the old path succeeded. The early return makes the no-op case O(1) and keeps deep graphs working when there is nothing to strip.

wire-runtime/src/jvmMain/kotlin/com/squareup/wire/Message.kt
```diff
@@ -60,8 +60,18 @@ protected actual constructor(
...
+    @Suppress("UNCHECKED_CAST")
+    val message = this as M
+    val encoded = adapter.encodeByteString(message)
+    return adapter.decode(encoded.substring(0, encoded.size - unknownFields.size))
```
oldergod, line 73: This arithmetic depends on an invariant that is not stated anywhere: Wire adapters always write unknown fields as the last bytes of the encoding. It holds for every encoder in this repo today, but a hand-written adapter or a future encoder change could break it and silently drop known fields. Please state the invariant in the KDoc so the dependency is explicit.

oldergod (approved):
The fix is correct. All Wire encoders write unknown fields as the last bytes of the encoding, so the substring cut lands on a field boundary and the prefix decodes cleanly. Approving, with a few non-blocking comments.

On test coverage: no test in `wire-tests/jvm-kotlin-interop` calls `withoutUnknownFields()`. The PR description says that module covers `javaInterop = true`, but its `UnknownFieldsTest` only uses builders. Please add one assertion there that mirrors the Java variant in `wire-tests/jvm-java-kotlin/.../UnknownFieldsTest.java`. Also, no test covers a message that consists only of unknown fields, so the `substring(0, 0)` path never runs. A case such as `OneField.ADAPTER.decode("109506".decodeHex())` with an assertion that `withoutUnknownFields()` equals `OneField()` would close that gap.

oldergod:
Thanks for the contribution! One thing before we can merge: Square requires all outside contributors to sign our Contributor License Agreement, and you don't appear to be on the signers list yet.

You can sign it here: https://docs.google.com/forms/d/e/1FAIpQLSeRVQ35-gq2vdSxD1kdh7CJwRdjmUA0EZ9gRXaWYoUeKPZEQQ/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1

It only takes a minute. Please drop a note here once you've signed and we can proceed with the merge.

- [ ] wait for CLA and comment fixing.

oldergod:
@arimu1 Will you finish it?

